PR #4, opened 2023-06-22
[WIP] First draft of what GPT2 might look like.

Narsil:
Missing ops:

- GELU
- select (embedding)
- broadcast addition
- reshape (For attention)
- cat (for past key values cat hidden_states)


Narsil:
let's not add gpt2

PR #10, opened 2023-06-26
Add support for f16 and bf16

LaurentMazare:
This doesn't include the bf16 kernels though.


moreWax:
Implementing task 10 as per the reference implementation. Ensuring all code is modular, composable, and follows professional Rust standards. Focus on type safety, nonmutability, and avoiding unnecessary allocations. Verifying congruence with the reference repo.

PR #30, opened 2023-06-28
Reworking the kernel tree to include flash attention and layernorm

Narsil:
(faster RMS norm).


LaurentMazare:
Adding 3000 files? Maybe it's possible to isolate only the relevant bits?

Narsil:
Flash and layer norm depend on c10 and Aten and cutlass... Those files are highly interlinked unfortunately. If you know a way to shake useless files I'll take it. Otherwise we could have separate crates entirely, but right now it's convenient to have it in a single location I feel.

LaurentMazare:
But does this mean that the final binary will include the c10 and aten bits? This would be likely to bloat it a lot no? I guess the other way would be to find alternative flash attention kernels like this [triton one from their tutorials](https://github.com/openai/triton/blob/main/python/tutorials/06-fused-attention.py)?

Narsil:
No no ,this still compiles down everything to PTX, it's just that the cuda kernels themselves depend on macro magic.

Narsil:
> flash attention kernels like this [triton one from their tutorials](https://github.com/openai/triton/blob/main/python/tutorials/06-fused-attention.py)?

Still uses padding if I'm not mistaken.

Narsil:
Done

PR #38, opened 2023-06-29
Moving llama to f16.

Narsil:


LaurentMazare:
Could you revert all the `?` -> `unwrap` please, this makes the diff hard to read. Also on the first two lines it seems that you've tweaked some random characters?

PR #48, opened 2023-06-30
[not to be merged] OpenBlas experiment

LaurentMazare:
This uses openblas for the cpu matmul, on llama/f32 this only seemed to bring marginal improvements. No plans on merging this for now, maybe behind a feature flag at some point.

Benchmark with the llama 7B example, default prompt:
- Using the gemm crate, float16: 48s prompt processing then 3.3s per token.
- Using the gemm crate, float32: 163s prompt processing, then 125s per token (seems very very large somehow, the process wasn't swapping or anything).
- Using openblas float32: 160s prompt processing, then 118s per-token.
This is on a Ryzen 5 2600X, 6 threads enabled, 32GB memory.

_edit_
The poor performances on f32 were due to on the flight f16 to f32 conversions. Getting rid of this (at the expense of using more system memory), the updating timings are:
- Using the gemm crate, float32: 45s prompt processing, then 2.3s per token (seems very very large somehow, the process wasn't swapping or anything).
- Using openblas float32: 49s prompt processing, then 3.2s per-token.


LaurentMazare:
Closing now that the mkl feature has been merged.

PR #62, opened 2023-07-03
Adding saving capabilities.

Narsil:


candle-core/src/safetensors.rs
```diff
@@ -36,6 +107,21 @@ pub fn convert(view: st::TensorView<'_>, device: &Device) -> Result<Tensor> {
...
 
+pub fn convert_back(tensor: &Tensor) -> Result<Cow<[u8]>> {
+    match tensor.dtype() {
+        DType::U8 => convert_back_::<u8>(tensor.storage_data()?),
+        DType::U32 => convert_back_::<u32>(tensor.storage_data()?),
+        DType::F16 => convert_back_::<half::f16>(tensor.storage_data()?),
+        DType::BF16 => convert_back_::<half::bf16>(tensor.storage_data()?),
+        DType::F32 => convert_back_::<f32>(tensor.storage_data()?),
+        DType::F64 => convert_back_::<f64>(tensor.storage_data()?),
+        // DType::BF16 => convert_::<half::bf16>(view, device),
```
LaurentMazare, line 118: Remove these comments?

candle-core/src/safetensors.rs
```diff
@@ -1,5 +1,61 @@
...
+
+    fn data_len(&self) -> usize {
+        let n: usize = self.shape().elem_count();
+        let bytes_per_element = self.dtype().size_in_bytes();
+        n * bytes_per_element
+    }
+}
+
+impl Tensor {
+    pub fn save<P: AsRef<std::path::Path>>(&self, name: &str, filename: P) -> Result<()> {
```
LaurentMazare, line 54: Maybe mention the `safetensors` format in the function name so that it's obvious what gets used?
Also could you expose in the `Tensor` impl some save multi function (similar to what you used in your test below)?

Narsil: `Tensor::save_multi(map, ...)` ? I don't really like that (it's a method on some struct where `serialize_to_file(map,...)` already works.

the `save` is really a convenience method that I'm happy to remove.

LaurentMazare (approved):

PR #88, opened 2023-07-06
Fixing unsafe slow load (memcpy).

Narsil:
- Without the annotation, I think the rust compiler assumes it's all u8.

It did segfault trying to load `Roberta`.

### DIFF
--- a/candle-core/src/safetensors.rs
+++ b/candle-core/src/safetensors.rs
@@ -71,6 +71,8 @@ fn convert_<T: WithDType>(view: st::TensorView<'_>, device: &Device) -> Result<T
         let data: &[T] = unsafe { std::slice::from_raw_parts(v.as_ptr() as *const T, elem_count) };
         Tensor::from_slice(data, view.shape(), device)
     } else {
+        // XXX: We need to specify `T` here, otherwise the compiler will infer u8 because of the following cast
+        // Making this vector too small to fit a full f16/f32/f64 weights, resulting in out-of-bounds access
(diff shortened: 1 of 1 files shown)


candle-core/src/safetensors.rs
```diff
@@ -71,7 +71,7 @@ fn convert_<T: WithDType>(view: st::TensorView<'_>, device: &Device) -> Result<T
         let data: &[T] = unsafe { std::slice::from_raw_parts(v.as_ptr() as *const T, elem_count) };
         Tensor::from_slice(data, view.shape(), device)
     } else {
-        let mut c = Vec::with_capacity(elem_count);
+        let mut c: Vec<T> = Vec::with_capacity(elem_count);
```
LaurentMazare, line 74: Could you add a comment to state the problem when this annotation is not there? This will avoid potentially removing it in further changes.

Narsil, line 74:
```suggestion
        // XXX: We need to specify `T` here, otherwise the compiler will infer u8 because of the following cast
        // Making this vector too small to fit a full f16/f32/f64 weights, resulting in out-of-bounds access
        let mut c: Vec<T> = Vec::with_capacity(elem_count);
```

Narsil: Done. Is that explanation clear enough ?

LaurentMazare: Looks great, thanks!

LaurentMazare (approved):

PR #109, opened 2023-07-08
Sketch a fast cuda kernel for reduce-sum.

LaurentMazare:


LaurentMazare:
When running on `examples/cuda_sum_benchmark.rs`, runtime goes from 36.8ms to 656us for each call on my RTX 2080 (ryzen 2600x, 32G mem).

PR #118, opened 2023-07-10
Expanding a bit the README

Narsil:


Narsil:
> tradeoffs vs dfdx typed shapes

100% I definitely want to give a hat's off to `dfdx` and mention it, and why we though another library was worth it.

PR #127, opened 2023-07-10
`i(..)` indexing sugar (partial).

Narsil:
- Only range, and select (no tensor_select)
- No negative indexing


candle-core/src/indexer.rs
```diff
@@ -0,0 +1,244 @@
...
+                        Bound::Excluded(n) => *n + 1,
+                        Bound::Unbounded => 0,
+                    };
+                    let stop = match right_bound {
+                        Bound::Included(n) => *n + 1,
+                        Bound::Excluded(n) => *n,
+                        Bound::Unbounded => dims[i],
+                    };
+                    let len = stop - start;
+                    println!(" indexer {indexer:?} Start {start} stop{stop} - {len:?}");
```
Narsil, line 46:
```suggestion
```

candle-core/src/tensor.rs
```diff
@@ -444,6 +444,7 @@ impl Tensor {
                 len,
             })?
         }
+        println!("Narrow  {start:?} - {} - {len} - {dims:?}", dims[dim]);
```
Narsil, line 447:
```suggestion
```

PR #130, opened 2023-07-10
Move the var-builder in a central place.

LaurentMazare:
Merge the different codes together, the llama bit is to be rewritten.
This is pending some rewrite of the var-builder so as to untangle it a bit from safetensors, allow different storage and random initialization (based on hints given by the model, e.g. fan-in, fan-out, glorot, etc).

### DIFF
--- a/candle-examples/examples/bert/main.rs
+++ b/candle-examples/examples/bert/main.rs
@@ -4,72 +4,14 @@
 extern crate intel_mkl_src;
 
 use anyhow::{anyhow, Error as E, Result};
-use candle::{safetensors::SafeTensors, DType, Device, Shape, Tensor};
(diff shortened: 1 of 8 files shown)


Narsil:
Can we please have 2 different methods ?

`Model::load(vb)` 
and
`Model::init()` ? 

Both need to go through the model, but the first one is tied to safetensors (or whatever storage and actual files).
While the other isn't, and for which passing something with a Rng might be better suited. (And it doesn't make sense to pass a Rng when loading something existing). 

I feel strongly about this.

LaurentMazare:
This first PR just re-groups the implemenation so that it's easier to refactor, for the init vs load this will be in a further PR.

PR #135, opened 2023-07-11
Removing some `narrow` and `get` to use `.i(..)` (looks more torch-like).

Narsil:


Narsil:
Too old, I'll do it again later.

PR #139, opened 2023-07-11
Refactor the llama example to make it more in sync with the other ones.

LaurentMazare:
Obvious limitations after the clean-up:
- No more npy support, we may want to bring it back to life but at the `VarBuilder` level.
- Higher memory usage.
Upsides:
- Use the new practices, make things more uniform.

(note: the results when using cuda and safetensors are different from when using numpy weights or the mkl version and weirder, this was also the case prior to this PR, to be investigated)


Narsil:
Yes I know about the differences, which are indeed quite odd.

I know the layernorm have `inv_freq` in the transformers impl with numbers different from the standard implem.
I tried doing that modifications and that wasn't it :(

LaurentMazare:
Hopefully this should be fixed now, for posterity this was caused by the data preparation [in this script](https://github.com/huggingface/transformers/blob/2642d8d04b14c18199ebe7b35f976da02df61752/src/transformers/models/llama/convert_llama_weights_to_hf.py#L101), this changed the "alignment" of the matrixes but not their shapes so was not causing any issue besides the poor results.

PR #198, opened 2023-07-19
Add a small matmul benchmark.

LaurentMazare:
Nothing very fancy, this benchmark is just helpful when fiddling with gemm matmul.


LaurentMazare:
The next time we need a matmul benchmark, we should make one with the interesting sizes in `cpu_benchmarks.rs`.

PR #216, opened 2023-07-21
TP sharding v2

Narsil:
Second update version of TP sharding.

- This WAS NOT tested for accuracy (beacuse I didn't implement all_reduce)
- Currently waiting for a simple API to get the cuda storage so I can call `all_reduce` directly and recreate a tensor from that. (That way I don't just `pub` everything)

I tried to keep the modifications minimal in `var_builder` for now as this is not the purpose of this PR.

### DIFF
--- a/candle-examples/examples/llama_multiprocess/model.rs
+++ b/candle-examples/examples/llama_multiprocess/model.rs
@@ -4,6 +4,7 @@ use candle_nn::{Embedding, Linear, VarBuilder};
 use cudarc::nccl::safe::{Comm, ReduceOp};
 use half::f16;
 use std::collections::HashMap;
+use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
 use super::MAX_SEQ_LEN;
@@ -23,12 +24,19 @@ impl TensorParallelColumnLinear {
 
 struct TensorParallelRowLinear {
     linear: Linear,
-    comm: Arc<Comm>,
+    comm: Rc<Comm>,
 }
 
 struct AllReduce {
-    comm: Arc<Comm>,
+    comm: Rc<Comm>,
 }
+
+/// This is actually not safe: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/threadsafety.html
+/// But for this example purposes, this will work
+unsafe impl Sync for AllReduce {}
+/// This is actually not safe: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/threadsafety.html
+/// But for this example purposes, this will work
+unsafe impl Send for AllReduce {}
 
 impl CustomOp1 for AllReduce {
     fn name(&self) -> &'static str {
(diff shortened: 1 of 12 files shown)


candle-examples/examples/llama_multiprocess/model.rs
```diff
@@ -0,0 +1,494 @@
...
+        self.comm.all_reduce(s, &mut dst, &ReduceOp::Sum).unwrap();
+        let dst = candle::CudaStorage::wrap_cuda_slice(dst, dev);
+        Ok((dst, l.shape().clone()))
+    }
+}
+
+fn all_reduce_sum(x: &Tensor, comm: &Rc<Comm>) -> Result<Tensor> {
+    x.custom_op1(AllReduce { comm: comm.clone() })
+}
+
```
Narsil, line 74: This is the core of the new thing.

+ TP Row/Col

candle-nn/src/var_builder.rs
```diff
@@ -137,6 +135,58 @@ impl<'a> VarBuilder<'a> {
...
+                let view = safetensors[*index].tensor(&path)?;
+                let dtype = view.dtype();
+                let mut shape = view.shape().to_vec();
+                let size = shape[dim];
+                let block_size = size / world_size;
+                let start = rank * block_size;
+                let stop = (rank + 1) * block_size;
+
+                let iterator = if dim == 0 {
+                    view.slice(start..stop).unwrap()
```
LaurentMazare, line 172: Why unwrap rather than propagating the error?

candle-nn/src/var_builder.rs
```diff
@@ -137,6 +135,58 @@ impl<'a> VarBuilder<'a> {
...
+                } else if dim == 1 {
+                    view.slice((.., start..stop)).unwrap()
+                } else {
+                    unimplemented!("Get sharded on dimensions != 0 or 1");
```
LaurentMazare, line 176: You can use `Error::bail!` to have a proper error if we don't plan on supporting this anytime soon.

candle-nn/src/var_builder.rs
```diff
@@ -137,6 +135,58 @@ impl<'a> VarBuilder<'a> {
 }
 
 impl<'a> VarBuilder<'a> {
+    pub fn get_sharded(
+        &self,
+        tensor_name: &str,
+        dim: usize,
+        rank: usize,
+        world_size: usize,
```
LaurentMazare, line 143: It's a bit unclear what `world_size` means here, add a comment or some documention to the function?

candle-nn/src/var_builder.rs
```diff
@@ -137,6 +135,58 @@ impl<'a> VarBuilder<'a> {
...
+                    }
+                    .bt()
+                })?;
+
+                let view = safetensors[*index].tensor(&path)?;
+                let dtype = view.dtype();
+                let mut shape = view.shape().to_vec();
+                let size = shape[dim];
+                let block_size = size / world_size;
+                let start = rank * block_size;
```
LaurentMazare, line 168: It's a bit tricky to know what is a size in number of elements or a size in bytes here, maybe rename to `start_in_bytes` or something like this? (`size` seems to be in elements)

candle-nn/src/var_builder.rs
Narsil: Fixed.

candle-nn/src/var_builder.rs
Narsil: Fixed

candle-nn/src/var_builder.rs
Narsil: Everything is in number of elements. The size in bytes is taken care of in `safetensors`.

candle-nn/src/var_builder.rs
Narsil: I updated with a rustdoc on top.

`rank` and `world_size` are definitely names borrowed from torch distributed.
We could rename them to something like `block_id` , `n_blocks` or something along those lines.

candle-nn/src/var_builder.rs
Narsil: Done that.

LaurentMazare (approved):

PR #245, opened 2023-07-26
Again set a few extra params in flash-attn.

LaurentMazare:


candle-examples/examples/llama/model.rs
```diff
@@ -220,8 +220,12 @@ impl CausalSelfAttention {
         let v = self.repeat_kv(v)?;
 
         let y = if self.use_flash_attn {
+            // flash-attn expects (b_sz, seq_len, nheads, head_dim)
+            let q = q.transpose(1, 2)?;
+            let k = k.transpose(1, 2)?;
+            let v = v.transpose(1, 2)?;
             let softmax_scale = 1f32 / (self.head_dim as f32).sqrt();
-            flash_attn(softmax_scale, &q, &k, &v)?
+            flash_attn(softmax_scale, &q, &k, &v)?.transpose(1, 2)?
```
Narsil, line 228: It would be nice, to NOT use batching as an extra dimension, but really using the offsets tensors instead. This is a big perk of flash.

LaurentMazare: I'm currently prototyping with the non varlen version of flash-attention so the batch size is required at this point. Longer term we could have a look at the variable length version.

PR #247, opened 2023-07-26
Add number of tokens.

Narsil:


candle-wasm-examples/llama2-c/src/app.rs
```diff
@@ -147,6 +156,7 @@ impl Component for App {
             }
             Msg::WorkerInMsg(inp) => {
                 self.worker.send(inp);
+                self.n_tokens = 0;
```
LaurentMazare, line 159: Would seem nicer to have this at the same place where we update `self.generated` rather than having two different places where we reset stuff.

Narsil: Done.

LaurentMazare (approved):

PR #258, opened 2023-07-27
Starting the book.

Narsil:


Narsil:
Friendly ping for review.

candle-book/src/guide/hello_world.md
```diff
@@ -1 +1,192 @@
-# PyTorch cheatsheet
+# Hello world !
+
+We will now create the hello world of the ML world, building a model capable of solving MNIST dataset.
+
+Open `src/main.rs` and fill in with these contents:
```
LaurentMazare, line 5: these contents -> this content?
Also maybe we should have a repo with the actual files so that it's easier for people that just want to copy the files around?

candle-book/src/guide/hello_world.md
```diff
@@ -1 +1,192 @@
...
+
+Everything should now run with:
+
+```bash
+cargo run --release
+```
+
+## Using a `Linear` layer.
+
+Now that we have this, we might want to complexity a little, for instance by adding `bias` and creating
```
LaurentMazare, line 48: to complexity a little -> to complexify things a bit or something like that?

candle-book/src/guide/hello_world.md
```diff
@@ -1 +1,192 @@
...
+impl Model {
+    fn forward(&self, image: &Tensor) -> Result<Tensor> {
+        let x = self.first.forward(image)?;
+        let x = x.relu()?;
+        self.second.forward(&x)
+    }
+}
+```
+
+This will change the loading code into a new function
```
LaurentMazare, line 79: Not sure to understand what "loading code" means in this context, maybe more "model code" or "model running code"?

candle-book/src/guide/hello_world.md
```diff
@@ -1 +1,192 @@
...
+# impl Model {
+#     fn forward(&self, image: &Tensor) -> Result<Tensor> {
+#         let x = self.first.forward(image)?;
+#         let x = x.relu()?;
+#         self.second.forward(&x)
+#     }
+# }
+fn main() -> Result<()> {
+    // Use Device::new_cuda(0)?; to use the GPU.
+    let device = Device::Cpu;
```
LaurentMazare, line 109: `cuda_if_available` is nice for this :)

candle-book/src/guide/hello_world.md
```diff
@@ -1 +1,192 @@
...
+    Ok(())
+}
+```
+
+Now it works, great and is a great way to create your own layers.
+But most of the classical layers are already implemented in [candle-nn](https://github.com/LaurentMazare/candle/tree/main/candle-nn).
+
+## Using `candle_nn`.
+
+For instance [Linear](https://github.com/LaurentMazare/candle/blob/main/candle-nn/src/linear.rs) is already there.
```
LaurentMazare, line 132: Once the crates are published this should probably point at docs.rs instead of the code as it's nicer for users to look at.

candle-book/src/guide/hello_world.md
```diff
@@ -1 +1,192 @@
...
+
+    let dummy_image = Tensor::zeros((1, 784), DType::F32, &device)?;
+
+    let digit = model.forward(&dummy_image)?;
+    println!("Digit {digit:?} digit");
+    Ok(())
+}
+```
+
+Feel free to modify this example to use `Conv2d` to create a classical convnet instead.
```
LaurentMazare, line 183: Which `Conv2d`? :)

candle-book/src/guide/installation.md
```diff
@@ -1 +1,24 @@
...
+Start by creating a new app:
+
+```bash
+cargo new myapp
+cd myapp
+cargo add --git https://github.com/LaurentMazare/candle.git candle
+```
+
+At this point, candle will be built **without** CUDA support.
+To get CUDA support use the feature `cuda`
```
LaurentMazare, line 12: the feature `cuda` -> the `cuda` feature?

candle-book/src/guide/hello_world.md
Narsil: Fixed typo.

I'm hesitant for the final repo, I'm sure we're not going to keep it enough up to date and currently there's very little content.
But maybe later when the doc grows( and we could inner include too).

candle-book/src/guide/hello_world.md
Narsil: Indeed !

candle-book/src/guide/hello_world.md
Narsil: Ok I was referring to the loading take up more space.

candle-book/src/guide/hello_world.md
Narsil: Very nice indeed. I'll leave the other options in comment, seems core if you need to make sure when a model executes.

candle-book/src/guide/hello_world.md
Narsil: Fully agree but we said we'd write it first :)

candle-book/src/guide/hello_world.md
Narsil: Exercise left to the user.

candle-book/src/guide/installation.md
Narsil: Done.

PR #259, opened 2023-07-27
Book 2 (load/save)

Narsil:
Depends on https://github.com/LaurentMazare/candle/pull/258 Drafted until merge.

### DIFF
--- a/candle-core/src/safetensors.rs
+++ b/candle-core/src/safetensors.rs
@@ -2,6 +2,8 @@ use crate::{DType, Device, Error, Result, Tensor, WithDType};
 use safetensors::tensor as st;
 use safetensors::tensor::SafeTensors;
 use std::borrow::Cow;
+use std::collections::HashMap;
+use std::path::Path;
 
 impl From<DType> for st::Dtype {
     fn from(value: DType) -> Self {
@@ -238,6 +240,15 @@ fn convert_back(tensor: &Tensor) -> Result<Vec<u8>> {
     }
 }
 
+pub fn load<P: AsRef<Path>>(filename: P, device: &Device) -> Result<HashMap<String, Tensor>> {
+    let data = std::fs::read(filename.as_ref())?;
+    let st = safetensors::SafeTensors::deserialize(&data)?;
+    st.tensors()
+        .into_iter()
+        .map(|(name, view)| Ok((name, view.load(device)?)))
+        .collect()
+}
+
 pub struct MmapedFile(memmap2::Mmap);
 
 impl MmapedFile {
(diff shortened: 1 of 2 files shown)


candle-book/src/guide/hello_world.md
```diff
@@ -1 +1,53 @@
...
+We will now create the hello world of the ML world, building a model capable of solving MNIST dataset.
+
+Open `src/main.rs` and fill in with these contents:
+
+```rust
+# extern crate candle;
+use candle::{DType, Device, Result, Tensor};
+
+struct Model {
+    first: Tensor,
```
LaurentMazare, line 12: Maybe it makes more sense to use `candle_nn::Linear` here?
Or maybe we really want to go back to basics here in which case we should mention that `Linear` exists but for the sake of this tutorial we won't use it here and have explicit tensors instead.

Narsil: Great idea. I was wondering how to complexify this example without delving into too much details to keep it hello world like.

I will keep this example that simple, and expand by making your own Linear, then referring to candle_nn for layers in general.

For `Conv1d` and `Linear` for instance, wdyt ?

Narsil: BTW, this comment is more about https://github.com/LaurentMazare/candle/pull/258 (This PR I converted to draft because it's more about the `save/load` surface to get simple cheatsheet more complete on that front.

LaurentMazare: Sounds good, you could even even not have backprop/gradient descent for a tensor 101 though maybe it's too simplistic. But certainly good to only introduce `candle_nn` later in the process (so that users can understand that there is no magic behind the hood there).

Narsil: Yup, I want to keep the training loop in a different place.

IMO after having hello world + there's no magic you have several options:

- Run a real model (90% of people)
- Train a model (10% of people) 

It's an old stat, we're closer to 99% vs 1% now that AI is much more mainstream (but there's still a lot of finetuners out there, mostly using script already written by people)

candle-core/src/safetensors.rs
```diff
@@ -238,6 +240,19 @@ fn convert_back(tensor: &Tensor) -> Result<Vec<u8>> {
...
+pub fn load<P: AsRef<Path>>(filename: P, device: &Device) -> Result<HashMap<String, Tensor>> {
+    let data = std::fs::read(filename.as_ref())?;
+    let st = safetensors::SafeTensors::deserialize(&data)?;
+    st.tensors()
+        .into_iter()
+        .map(|(name, view)| Ok((name, view.load(device)?)))
+        .collect()
+}
+
+pub fn save<P: AsRef<Path>>(tensors: HashMap<&str, Tensor>, filename: P) -> Result<()> {
```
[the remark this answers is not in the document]
Narsil: We don't need it, removed it.

PR #262, opened 2023-07-28
Making multiprocess require flash-attn.

Narsil:
- Make the necessary changes for `hf-hub 0.2.0` 
- Use `flash_attn` in that example (it's intended to be a powerhouse example, it's OK to no support non flash imo).
- Remove `use_kv_cache`, same logic being slower doesn't seem really interesting to showcase here.
- Remove resulting dead code.


candle-examples/examples/llama_multiprocess/model.rs
```diff
@@ -31,6 +30,16 @@ struct AllReduce {
     comm: Rc<Comm>,
 }
 
+fn flash_attn(
```
LaurentMazare, line 33: Probably not much advantage defining this function as we don't support the feature-gate here, so maybe just inline `candle_flash_attn::flash_attn` where this is used.

Narsil: Done.

LaurentMazare (approved):

PR #319, opened 2023-08-03
[do-not-merge] Transpose the rhs in linear.

LaurentMazare:
This results in a speed up of > 6x when running llama2 inference as the `pack_generic_inner_loop` bits from `gemm-f16` can use the vectorized instructions. With this the time per token is ~1.98s vs ~1.3s for mkl.


dback1 (commented):
Remove me

PR #361, opened 2023-08-09
Make to_device and to_dtype more like Tensor.to in pytorch by adding a To<T> trait for tensor. 

Y-jiji:
A very minor modification on interface. 
Make to_device and to_dtype more like Tensor.to in pytorch by adding a To<T> trait for tensor.

### DIFF
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -142,6 +142,24 @@ fn from_storage<S: Into<Shape>>(
     Tensor(Arc::new(tensor_))
 }
 
+pub trait To<T> 
+where Self: Sized 
+{
+    fn to(self, to: T) -> Result<Self>;
+}
+
+impl To<&Device> for Tensor {
+    fn to(self, device: &Device) -> Result<Self> {
+        self.to_device(device)
+    }
+}
+
+impl To<DType> for Tensor {
+    fn to(self, dtype: DType) -> Result<Self> {
+        self.to_dtype(dtype)
+    }
+}
+
 impl Tensor {
     pub(crate) fn ones_impl<S: Into<Shape>>(
         shape: S,


Y-jiji:
First of all, I also advocate typed programming. 

However, things here are still typed, and I'm very sure there will not be any confusion when people plug a `DType::F16` or `&Device::Cpu` into `.to(x)`. 

This idea is widely accepted in the rust world. You can see many people writing `fn something(x: Impl Into<T>)` instead of `fn something(x: T)` . In fact, even some language items contain this kinds of implementations! Every time you use `?` operator in a function that returns type `Result<A, E>`, you are allowed to emit error of type T if From<T> is implemented for type `E`.

Y-jiji:
@Narsil

TieWay59:
Hello, I understand that this may depend on the project convention. However, I have done some research on the PR tab in the candle project and noticed that most PRs are authored by @LaurentMazare and @Narsil, and no external contributions are accepted in this repository.

See: https://github.com/huggingface/candle/pulls?q=is%3Apr+is%3Amerged++%21user%3A%22LaurentMazare%22

I appreciate that you are in an active and heated development stage. But if you are not ready to receive other people's PRs, I suggest you add some notices in the `README.md` file. Otherwise, it may waste other people's time and cause frustration.

I intend to share this with other platforms I can access. Perhaps letting more people know about the situation would save them time from engaging with a repository that is not very "open" for now.

ViliamVadocz:
@TieWay59 funniest PR comment I have ever seen thanks haha

TieWay59:
@ViliamVadocz Eh, you are welcome then? I'm so sorry to take the top one place in your mind. God bless you to find much funnier in other PRs. 😊😊😊

Anyway, my point is that it's more rude to abandon others' PR without a proper reply. Even if the intention is to reject the PR, a simple "No" would be beneficial in this context. However, it has been a month, and there has been no response.

LaurentMazare:
Sorry I definitely dropped the ball on this. I don't have a super strong opinion on the PR itself, but would rather not merge it as I prefer erring on the explicit side rather than overriding methods in the same way pytorch does.
We do welcome external contributions as can be seen in the "Contributors" pane (24 people) though as you mentioned things are under heavy development at the moment so it's hard to take everything into account.
Looking at how much effort has been put into this so far, I certainly felt a bit sad when reading `I intend to share this with other platforms I can access.` so I'm just going to close/lock this PR for now.

PR #371, opened 2023-08-09
add max_pool2d

LeeeSe:
add support for max_pool2d and provide a simple test for it.


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks for the PR. I'll merge once the CI has run. For the future, note that it would be great to provide cuda kernels too as we would rather not accumulate too many ops for which we only have a cpu implementation.

Issue #353, opened 2023-08-08
Cannot run examples with --features cuda option


dbrowne:
CARGO_PROFILE_RELEASE_BUILD_OVERRIDE_DEBUG=true
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
   Compiling libc v0.2.147
   Compiling autocfg v1.1.0
   Compiling crossbeam-utils v0.8.16
   Compiling proc-macro2 v1.0.66
   Compiling unicode-ident v1.0.11
   Compiling rayon-core v1.11.0
   Compiling memchr v2.5.0
   Compiling libm v0.2.7
   Compiling cfg-if v1.0.0
   Compiling pkg-config v0.3.27
   Compiling paste v1.0.14
   Compiling serde v1.0.183
   Compiling serde_derive v1.0.183
   Compiling scopeguard v1.2.0
   Compiling syn v1.0.109
   Compiling serde_json v1.0.104
   Compiling seq-macro v0.3.5
   Compiling vcpkg v0.2.15
   Compiling crc32fast v1.3.2
   Compiling ident_case v1.0.1
   Compiling strsim v0.10.0
   Compiling fnv v1.0.7
   Compiling thiserror v1.0.44
   Compiling either v1.9.0
   Compiling glob v0.3.1
   Compiling openssl v0.10.56
   Compiling rustls v0.21.6
   Compiling anyhow v1.0.72
   Compiling cudarc v0.9.13
   Compiling portable-atomic v1.4.2
   Compiling native-tls v0.2.11
   Compiling esaxx-rs v0.1.8
   Compiling adler v1.0.2
   Compiling rustix v0.38.7
   Compiling gimli v0.27.3
   Compiling macro_rules_attribute-proc_macro v0.1.3
   Compiling rustc-demangle v0.1.23
   Compiling miniz_oxide v0.7.1
   Compiling heck v0.4.1
   Compiling flate2 v1.0.26
   Compiling memoffset v0.9.0
   Compiling crossbeam-epoch v0.9.15
   Compiling num-traits v0.2.16
   Compiling zip v0.6.6
   Compiling crossbeam-channel v0.5.8
   Compiling aho-corasick v1.0.2
   Compiling object v0.31.1
[...]

PR #386, opened 2023-08-10
This is duplicated code on Cuda 12.2.

Narsil:
Without it we can compile for 52 (but I get Operation Not supported
when actually trying to use those kernels).

Fixes partially #353

### DIFF
--- a/candle-kernels/src/compatibility.cuh
+++ b/candle-kernels/src/compatibility.cuh
@@ -6,24 +6,6 @@
 
 // FIXME: the minimum compute capabilities are just guesses since the table is not specific enough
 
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Do you know if these functions are available on old versions of cuda? I would guess that I've added them because they were not there but don't have access to my gpu box to confirm, but maybe we want to predicate both on the arch and on the cuda version?

Narsil:
I *think* those were enabled in most recent Cuda versions, therefore for maximum support we would need to check cuda versions and so on.

The comment mentions it's very heuristic based to know whether those helpers should be enabled or not.

LaurentMazare (approved):

PR #395, opened 2023-08-10
Compat windows.

Narsil:
Checked on a real windows machine.

The CUDA docs says those functions were introduced in Cuda 12.0 (and absent in Cuda 11.8).
However I run a windows machine on cuda 12.0 which doesn't have these....

Cuda 12.0: https://docs.nvidia.com/cuda/archive/12.0.0/cuda-math-api/group__CUDA__MATH____HALF__COMPARISON.html#group__CUDA__MATH____HALF__COMPARISON
Cuda 11.8: https://docs.nvidia.com/cuda/archive/11.8.0/cuda-math-api/group__CUDA__MATH____HALF__COMPARISON.html#group__CUDA__MATH____HALF__COMPARISON

### DIFF
--- a/candle-kernels/src/compatibility.cuh
+++ b/candle-kernels/src/compatibility.cuh
@@ -6,6 +6,15 @@
 
 // FIXME: the minimum compute capabilities are just guesses since the table is not specific enough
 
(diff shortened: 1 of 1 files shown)


LaurentMazare (approved):

LaurentMazare:
Great, thanks for checking!

PR #397, opened 2023-08-10
download mnist dataset from storage.googleapis.com

cksac:
- add load method in mnist dataset to download the data from storage.googleapis.com/cvdf-datasets/mnist
- update mnist-training example

now can direct run `cargo run --example mnist-training --release` without user to download the data

### DIFF
--- a/candle-core/Cargo.toml
+++ b/candle-core/Cargo.toml
@@ -27,7 +27,7 @@ rand_distr = { workspace = true }
 safetensors = { workspace = true }
 thiserror = { workspace = true }
 zip = { workspace = true }
-opendal = { workspace = true }
+ureq = { workspace = true }
 
 [dev-dependencies]
 anyhow = { workspace = true }
--- a/candle-core/Cargo.toml
+++ b/candle-core/Cargo.toml
@@ -27,7 +27,6 @@ rand_distr = { workspace = true }
 safetensors = { workspace = true }
 thiserror = { workspace = true }
 zip = { workspace = true }
-ureq = { workspace = true }
 
 [dev-dependencies]
 anyhow = { workspace = true }
--- a/candle-core/Cargo.toml
+++ b/candle-core/Cargo.toml
@@ -27,6 +27,7 @@ rand_distr = { workspace = true }
 safetensors = { workspace = true }
 thiserror = { workspace = true }
 zip = { workspace = true }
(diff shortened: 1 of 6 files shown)


Narsil:
Hey thanks for that.

For datasets we want at some point to leverage all existing datasets on the hub because they use `parquet` and therefore we can deflate them to `arrow` just like `datasets` does and we could access everything already existing on the hub:
For instance: https://huggingface.co/datasets/mnist/tree/refs%2Fconvert%2Fparquet
If you're interested to work on that! The idea would be to follow the same locations and naming of files.

Other than that I really like making examples working on their own. I'm a bit less fan of pulling a sizeable dependency where a simple http request would do though.

cksac:
ya, sure. initially I am trying to use hf_hub, is it support to to fetch https://huggingface.co/datasets/mnist/tree/refs%2Fconvert%2Fparquet ?

for the opendal dependency, but can switch to other crates at this moment, i.e reqwest?

Narsil:
Here for https://github.com/huggingface/hf-hub we use `ureq` for the sync API. Seems to work  decently enough.

And yes you can download directly the parquet file: https://huggingface.github.io/candle/inference/hub.html#using-the-hub 
https://github.com/huggingface/hf-hub/blob/a13a8188ac64cc413e43d253a390cb6b80711fff/src/api/sync.rs#L584

Let's keep it as a separate PR though, your PR is good as it is.
If you can switch to `ureq` that's great, otherwise we can roll with this for now.

cksac:
changed to ureq.
download parquet file from hub will be next PR

candle-core/Cargo.toml
```diff
@@ -27,6 +27,7 @@ rand_distr = { workspace = true }
 safetensors = { workspace = true }
 thiserror = { workspace = true }
 zip = { workspace = true }
+ureq = { workspace = true }
```
Narsil, line 30: Sorry I didn't see earlier, but can we remove `ureq` from core dependency ? I failed to see it earlier. 

I can take care of it

cksac: sure, it will become candle::Error::Wrapped

Narsil:
I started the work since I started the book chapter on it: 

https://github.com/huggingface/candle/pull/439

cksac:
nice. Close this PR, as load implemented in #439

PR #400, opened 2023-08-10
General interpolation op

leifu1128:
Here is a more general implementation that supports both up/down-sampling with a test. I want to implement other interpolation methods like bilinear, lanczos, etc. Should they each be their own thing or do we put them in one struct and use ```match```? Also still need to rename UpsampleNearest2D.

### DIFF
--- a/candle-core/src/cpu_backend.rs
+++ b/candle-core/src/cpu_backend.rs
@@ -720,38 +720,31 @@ struct UpsampleNearest2D(usize, usize);
 
 impl Map1 for UpsampleNearest2D {
     fn f<T: WithDType>(&self, src: &[T], layout: &Layout) -> Result<Vec<T>> {
-        // TODO: Specialized implementation for the case 2*h, 2*w?
         let (dst_h, dst_w) = (self.0, self.1);
         let (b_sz, c, src_h, src_w) = layout.shape().dims4()?;
         let stride = layout.stride();
-        let (stride_h, stride_w) = (stride[2], stride[3]);
-        let src_index = layout.start_offset();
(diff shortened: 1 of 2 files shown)


LaurentMazare:
This tweaks the current implementation quite a bit (which actually already handles downsampling hopefully), e.g. we pre-computed the indexes to avoid going through the division again, try to avoid modulo, etc. Also I'm also currently pushing on the stable-diffusion example for which I'm in the process of paralellising these ops (conv1d/conv2d/avg-pool/upsample) so would rather delay on this kind of refactoring until this part is a bit more stable.

leifu1128:
My bad, I did not realize it can already downsample. I'll roll back that refactor, and work on other interpolations, unless those are already being worked on?

LaurentMazare:
I'm a bit worried that we would introduce lots of ops for which we don't have good support (e.g. no cuda / backprop / ...) so would rather refrain from adding such ops until there is some good use case for them. Some of these could probably be implemented in external crates through some custom ops, same as we showed with [flash-attn](https://github.com/huggingface/candle/blob/89ba005962495f2bfbda286e185e9c3c7f5300a3/candle-flash-attn/src/lib.rs#L152). Maybe a good way forward would be to use candle to implement new models - starting with custom ops for the bits that are not already available - and based on how the ops look like we can see if we want to merge them in the candle-nn crate, or have them in some domain specific `candle-...` crate, or keep them with the model if they are very specific.

LaurentMazare:
Closing this for now, the current upsample op is now supported with cuda etc and we plan on hooking optimized versions of resampling methods via cudnn.

PR #407, opened 2023-08-11
Add dequantization for ggmls `q4_0`, `q4_1`, `q5_0`, `q5_1` and `q8_0`

LLukas22:
Allows ggmls q-quants to be read into candle tensors. Useful for all non llama ggml models as they are mostly distributed as q-quants. 

Also exposes a `ggml::tensor_from_ggml` function which allows consuming applications to convert a raw ggml tensor into a candle tensor. Mainly useful for applications, which have already implemented the ggml loading logic for multiple architectures and want to try candle as a backend (e.g. rustformers).

Adds a ggml loading example to showcase loading a ggml llama model via:  `cargo run --release --example ggml -- --model [PATH]`


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks for the PR!

PR #413, opened 2023-08-11
Add more whisper variants.

LaurentMazare:
These can be used via the `--model` flag, e.g. by passing `--model small` or `--model medium`.


Narsil:
@LaurentMazare Can we add insteead `--model-id openai/whisper-large-v2 --revision refs/pr/xx` 

So users can also use fine-tuned models like https://huggingface.co/alvanlii/whisper-small-cantonese ?

PR #420, opened 2023-08-12
Change multilingual detection

viktorlott:
Use vocal size to detect if model is multi-lingual.

### DIFF
--- a/candle-examples/examples/whisper/main.rs
+++ b/candle-examples/examples/whisper/main.rs
@@ -250,12 +250,6 @@ enum WhichModel {
 }
 
 impl WhichModel {
-    fn is_multilingual(&self) -> bool {
-        match self {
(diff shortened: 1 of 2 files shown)


LaurentMazare:
I would rather keep the current logic rather than rely on the vocab size even if the python version does this. It seems less likely to create issues if at some point new tokenizer are supported.

viktorlott:
> I would rather keep the current logic rather than rely on the vocab size even if the python version does this. It seems less likely to create issues if at some point new tokenizer are supported.

Totally get where you're coming from. Go ahead and close the PR if you'd like, and continue the great work!

LaurentMazare:
Thanks for your understanding, also don't hesitate to start some discussions on the discord for that kind of thing, might be easier to get feedback/suggestions than through PR comments.

PR #421, opened 2023-08-12
fix: can directly save the loaded weights

wuyumin:
can directly save the loaded weights.
```rust
  let api = Api::new()?;

  let weights = repo.get("model.safetensors")?;

  let weights = candle_core::safetensors::load(weights, &Device::Cpu)?;
  // do something like this
  // ...
  candle_core::safetensors::save(&weights, "new-model.safetensors")?;
```


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #435, opened 2023-08-13
[WIP] Adding i32 and randint, randint_like Tensor methods only for CPU

flatcapital:
Firstly, **great** work on this library, I'm very interested to see it grow!

I was looking for a way to generate tensors with uniform random integers, including negative values.

Similar methods found in PyTorch

[torch.randint](https://pytorch.org/docs/stable/generated/torch.randint.html)
[torch.randint_like](https://pytorch.org/docs/stable/generated/torch.randint_like.html)

However I'm quite unsure if this is at all something you would like to support in your framework, since it required incorporating a new DType `i32`. (Unless there is some really basic conversion of tensors from the existing `rand` method I completely missed) 

Please note my Rust experience is minimal, so this is a work in progress. I just wanted to know if this is something that would be considered at some point to include since there are some use cases for generating these kind of Tensors. However if there is no obvious need please IGNORE & close this PR.

If you would consider this for inclusion, I will continue writing tests for all the cases and also implement the Cuda backend of this.

```
    let a = Tensor::randint(0u8, 255, (8,), &Device::Cpu)?;
    let b = Tensor::randint(0u32, 1000, (6,), &Device::Cpu)?;
    let c = Tensor::randint(-100i32, 100, (4,), &Device::Cpu)?;
    let d = Tensor::randint_like(&a, 0i32, 1000);
    let fl = Tensor::rand(0.0f32, 1., (4,), &Device::Cpu)?;
    let err = Tensor::randint(0i32, -10, (2,), &Device::Cpu);

```

Yields
```
a: CpuTensor[195, 206, 250, 167, 35, 109, 156, 66; u8]
b: CpuTensor[600, 151, 443, 678, 824, 474; u32]
c: CpuTensor[-49, -86, 84, 37; i32]
d: Ok(CpuTensor[453, 545, 42, 103, 757, 483, 198, 913; i32])
fl: CpuTensor[0.01507771, 0.58171165, 0.7638751, 0.83642375; f32]
err: Err(RandIntBoundsInvalid { lo: 0, up: -10 })
```

Error if `up` is less than `lo`

```
Error: the upper bound value -10 must be greater than the lower bound value 0
```

I am also in the process of looking into setting a manual seed for the **RNG** for CPU and Cuda. I would like to build on this for testing purposes to ensure the same "random" values are tested consistently.

I see the current Cuda backend has a fixed seed (at the speed of light) :)

```    
let curand = cudarc::curand::CudaRng::new(299792458, device.clone()).w()?;
```

Any comments / feedback would be appreciated, also no worries simply closing this if not needed.

### DIFF
--- a/candle-core/src/backend.rs
+++ b/candle-core/src/backend.rs
@@ -1,5 +1,5 @@
 use crate::op::{BinaryOpT, CmpOp, ReduceOp, UnaryOpT};
-use crate::{CpuStorage, DType, Layout, Result, Shape};
+use crate::{CpuStorage, DType, Layout, Result, Shape, IntDType, FloatDType};
 
 pub trait BackendStorage: Sized {
     type Device: BackendDevice;
@@ -97,7 +97,11 @@ pub trait BackendDevice: Sized + std::fmt::Debug + Clone {
 
     fn storage_from_cpu_storage(&self, _: &CpuStorage) -> Result<Self::Storage>;
 
-    fn rand_uniform(&self, _: &Shape, _: DType, _: f64, _: f64) -> Result<Self::Storage>;
+    fn rand_uniform_float<T: Copy + PartialOrd + FloatDType>(
+        &self, _shape: &Shape, _dtype: DType, _lo: T, _up: T) -> Result<Self::Storage>;
+
+    fn rand_uniform_int<T: Copy + PartialOrd + IntDType>(
+        &self, _shape: &Shape, _dtype: DType, _lo: T, _up: T) -> Result<Self::Storage>;
 
     fn rand_normal(&self, _: &Shape, _: DType, _: f64, _: f64) -> Result<Self::Storage>;
 }
--- a/candle-core/src/convert.rs
+++ b/candle-core/src/convert.rs
@@ -92,6 +92,7 @@ from_tensor!(f64);
 from_tensor!(f32);
 from_tensor!(f16);
 from_tensor!(bf16);
+from_tensor!(i32);
 from_tensor!(u32);
 from_tensor!(u8);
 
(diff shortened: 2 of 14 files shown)


LaurentMazare:
Thanks for submitting this. What would be your specific use cases for `i32` tensors?
It's a bit tricky to introduce new dtypes as they end up adding code for all backends (cuda, cpu, mkl, accelerate, and probably more to come) so there will be some maintenance burden down the road. E.g. the main reasons for having `u8` is representing boolean tensors and for `u32` to represent token ids, maybe we'll need more but we have to be careful before commiting to support them, to be fair we might actually remove `f64` at some point as there isn't much of a clear use case for it at the moment.

flatcapital:
Thanks for the quick reply! I fully understand not wanting to clutter with frivolous types. I am actually busy trying to port the TimesNet model from [https://github.com/thuml/Time-Series-Library/tree/main](https://github.com/thuml/Time-Series-Library/tree/main) 

...and purely for mocking FFTransforms I wanted to create random integer Tensors to mimic selected top_k periods before a real implementation of FFT. **So not really a valid use case for model building.**

I initially just adapted the current `rand` method and used only u32, which works fine (I only need >0 values) but thought I might parity match the PyTorch methods `randint` and `randint_like`. Might have gone overboard, but it gave me a good spin through the code base.

Would you consider not having the `i32` type added, but having a way to still create uniform random integer Tensors? Although then just in range 0 to max u32/usize?

Also, would you be in support for adding manual (configurable) seed config for RNG?

No worries closing this then, I'll perhaps submit a much lighter weight PR for just `randint` using existing `u32`?

Cheers! 👍

PR #439, opened 2023-08-14
[Book] Add small error management + start training (with generic dataset inclusion).

Narsil:
- Starting work to get generic Parquet files:
- Deliberately not owning a struct `parquet` are relatively hard to navigate without ownership, therefore giving `Serializedreader` is much more convienient that abstracting away for now (Parquet can only be read in sequence, without any typing, SerializedReader reads require ownership and are not cloneable.).
- Small book on error management.


candle-book/src/SUMMARY.md
```diff
@@ -12,7 +12,11 @@
 
 - [Running a model](inference/README.md)
     - [Using the hub](inference/hub.md)
-- [Error management]()
+- [Error management](error_manage.md)
+- [Training](training/README.md)
+    - [MNIST]()
```
Narsil, line 17:
```suggestion
    - [MNIST](training/mnist.md)
```

candle-datasets/Cargo.toml
```diff
@@ -18,3 +18,6 @@ intel-mkl-src = { workspace = true, optional = true }
 memmap2 = { workspace = true }
 tokenizers = { workspace = true, features = ["onig"] }
 rand = { workspace = true }
+thiserror = { workspace = true }
+parquet = { version = "45.0.0"}
```
LaurentMazare, line 22: Why not define the parquet version in the top level `Cargo.toml`? I would think it's better to centralize everything here even if there is a single crate using it for now.
Also do you have a sense of how big of a dependency this crate ends up being?

candle-datasets/src/hub.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        // First element is an image
+        col_iter.next();
+        assert_eq!(
+            col_iter.next().unwrap().1,
+            &parquet::record::Field::Str("a drawing of a green pokemon with red eyes".to_string())
+        );
+
+        // Keep for now to showcase how to use.
+        //for row in rows {
```
LaurentMazare, line 84: Rather than keeping this in a comment, maybe just put it in a `if false` block (with some clippy deactivation), this way the code will still be compiled and will be fixed when the api changes.

candle-datasets/src/vision/mnist.rs
```diff
@@ -63,3 +65,54 @@ pub fn load_dir<T: AsRef<std::path::Path>>(dir: T) -> Result<crate::vision::Data
...
+    }
+    let images = (Tensor::from_vec(buffer_images, (samples, 784), &Device::Cpu)?
+        .to_dtype(DType::F32)?
+        / 255.)?;
+    let labels = Tensor::from_vec(buffer_labels, (samples,), &Device::Cpu)?;
+    Ok((images, labels))
+}
+
+pub fn load() -> Result<crate::vision::Dataset> {
+    let api = Api::new().unwrap();
```
LaurentMazare, line 95: Why `unwrap` rather than making this an error? We have `wrap` that is convenient (very similar to anyhow) so that we don't need to register all errors in the candle crate.

candle-examples/examples/mnist-training/main.rs
```diff
@@ -140,7 +140,7 @@ struct Args {
 pub fn main() -> anyhow::Result<()> {
     let args = Args::parse();
     // Load the dataset
-    let m = candle_datasets::vision::mnist::load_dir("data")?;
+    let m = candle_datasets::vision::mnist::load()?;
```
LaurentMazare, line 143: Could we have a flag to use the old data loading optionally? That would be convenient at least to me as I'm using this example fairly often.

candle-examples/src/lib.rs
```diff
@@ -14,6 +14,10 @@ pub fn device(cpu: bool) -> Result<Device> {
 
 #[cfg(test)]
 mod tests {
+    use anyhow::Result;
```
LaurentMazare, line 17: I'm starting to be a bit concerned about the amount of testing that ends up being done here. Culd we put this somewhere under `candle-book` instead, even if it requires an additional `Cargo.toml` there?

candle-datasets/Cargo.toml
Narsil: `Cargo bloat --release --crates`:

```
6.6%  15.3%   2.2MiB parquet
 6.1%  14.1%   2.1MiB arrow_cast
 ```
 
 ```
 6.6%  15.3%   2.2MiB parquet
 6.1%  14.1%   2.1MiB arrow_cast
 5.2%  12.0%   1.7MiB candle_core
 3.1%   7.2%   1.1MiB std
 1.9%   4.4% 653.7KiB arrow_select
 1.9%   4.3% 647.1KiB arrow_array
 1.8%   4.2% 628.5KiB [Unknown]
 1.6%   3.8% 561.3KiB image
 1.4%   3.3% 491.6KiB rustls
 0.9%   2.1% 311.3KiB zstd_sys
 0.7%   1.6% 244.2KiB ureq
 0.6%   1.5% 222.2KiB arrow_ipc
 0.6%   1.5% 219.7KiB ring
 0.6%   1.4% 208.3KiB brotli_decompressor
 0.6%   1.4% 204.9KiB brotli
 0.5%   1.1% 160.6KiB arrow_data
 0.4%   0.9% 141.5KiB openssl
 0.4%   0.8% 126.9KiB num_bigint
 0.3%   0.8% 114.0KiB indicatif
 0.3%   0.7% 111.8KiB lz4_sys
 6.6%  15.4%   2.2MiB And 95 more crates. Use -n N to show more.
43.1% 100.0%  14.6MiB .text section size, the file size is 33.9MiB
```

So relatively sizeable.

candle-datasets/src/hub.rs
Narsil: I uncommented it simply it's not that big to run.

candle-examples/src/lib.rs
Narsil: I managed to overlap a crate and the actual book.
Both are in the same directory yet are 2 different sets of files.

candle-datasets/Cargo.toml
```diff
@@ -9,6 +9,9 @@ categories.workspace = true
 license.workspace = true
 readme = "README.md"
 
+[lib]
+crate-type=["dylib"]
```
LaurentMazare, line 13: Why do we want this rather than the default?

candle-examples/Cargo.toml
```diff
@@ -39,6 +39,8 @@ tracing-subscriber = { workspace = true }
 wav = { workspace = true }
 # Necessary to disambiguate with tokio in wasm examples which are 1.28.1
 tokio = "1.29.1"
+parquet = "45.0.0"
```
LaurentMazare, line 42: Workspace version?

candle-datasets/Cargo.toml
Narsil: I need to remove, but that's required to be able to detect bloat :)

LaurentMazare (approved):

Issue #385, opened 2023-08-10
Compilation failed for flash-attn kernels


khulaifi95:
Hi, amazing job done here!

I am trying the example provided for llama-2 multi-device inference using the following command:

```shell
cargo run --example llama_multiprocess --release --features "cuda nccl flash-attn"
```

which yields the following error messages during the building stage of `candle-flash-attn` crate:

```
  kernels/flash_fwd_hdim160_fp16_sm80.cu(26): here

  kernels/flash_fwd_kernel.h(325): error: argument list for class template "cute::Tensor" is missing
            detected during:
              instantiation of "void flash::compute_attn<Kernel_traits,Is_dropout,Is_causal,Is_even_N,Is_even_K,Return_softmax,Params>(const Params &) [with Kernel_traits=Flash_fwd_kernel_traits<160, 64, 64, 4, false, false, cutlass::half_t, Flash_kernel_traits<160, 64, 64, 4, cutlass::half_t>>, Is_dropout=true, Is_causal=true, Is_even_N=true, Is_even_K=true, Return_softmax=true, Params=Flash_fwd_params]"
  kernels/flash_fwd_launch_template.h(15): here
              instantiation of "void flash_fwd_kernel<Kernel_traits,Is_dropout,Is_causal,Is_even_N,Is_even_K,Return_softmax>(Flash_fwd_params) [with Kernel_traits=Flash_fwd_kernel_traits<160, 64, 64, 4, false, false, cutlass::half_t, Flash_kernel_traits<160, 64, 64, 4, cutlass::half_t>>, Is_dropout=true, Is_causal=true, Is_even_N=true, Is_even_K=true, Return_softmax=true]"
  kernels/flash_fwd_launch_template.h(34): here
              instantiation of "void run_flash_fwd<Kernel_traits,Is_dropout,Is_causal>(Flash_fwd_params &, cudaStream_t) [with Kernel_traits=Flash_fwd_kernel_traits<160, 64, 64, 4, false, false, cutlass::half_t, Flash_kernel_traits<160, 64, 64, 4, cutlass::half_t>>, Is_dropout=true, Is_causal=true]"
  kernels/flash_fwd_launch_template.h(155): here
              instantiation of "void run_mha_fwd_hdim160<T>(Flash_fwd_params &, cudaStream_t) [with T=cutlass::half_t]"
  kernels/flash_fwd_hdim160_fp16_sm80.cu(26): here

  Error limit reached.
[...]

PR #452, opened 2023-08-15
[build scripts] add c++17 flag

lh-ycx:
Hey guys, I noticed that the current flash attention building [script](https://github.com/huggingface/candle/blob/main/candle-flash-attn/build.rs#L89) could fail due to the absence of ```c++17``` flag (just as described in  #385). Adding the flag would solve the problem. Thanks.


Narsil:
LGTM !

LaurentMazare (approved):

LaurentMazare:
Merged! Thanks for the PR! (a bit unclear why this is required in some configuration and not others but anyway it's an easy fix)

PR #463, opened 2023-08-16
Move RmsNorm to candle-nn

guoqingbao:
Since RmsNorm is a commonly-used operator for LLMs, RmsNorm was moved to candle-nn for common usage in this PR. All llama models (llama, llama2-c, llama-multiprocess) can now use the same RMSNorm defined in candle-nn. Could you help review this commit @LaurentMazare

### DIFF
--- a/candle-examples/examples/llama/model.rs
+++ b/candle-examples/examples/llama/model.rs
@@ -152,35 +152,20 @@ fn embedding(cfg: &Config, vb: VarBuilder) -> Result<Embedding> {
 }
 
 struct RmsNorm {
-    scale: Tensor,
-    eps: f64,
+    inner: candle_nn::LayerNorm,
     span: tracing::Span,
 }
 
 impl RmsNorm {
     fn load(size: usize, eps: f64, vb: VarBuilder) -> Result<Self> {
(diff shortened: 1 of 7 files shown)


LaurentMazare:
Thanks for the PR and for pointing this out all this unfortunate code duplication :)
Actually `RmsNorm` is just a variant of `LayerNorm` so instead of creating a new layer for it with a separate implementation, I've modified `LayerNorm` in #482 so that it's used by all these models so hopefully this should be much better now. It also makes things more flexible if one wants to use `LayerNorm` without the bias (which is an option that is available in PyTorch).

LaurentMazare:
(also probably obvious but if you write new models and want to use it, just use `candle_nn::rms_norm(...)`)

guoqingbao:
> Thanks for the PR and for pointing this out all this unfortunate code duplication :) Actually `RmsNorm` is just a variant of `LayerNorm` so instead of creating a new layer for it with a separate implementation, I've modified `LayerNorm` in #482 so that it's used by all these models so hopefully this should be much better now. It also makes things more flexible if one wants to use `LayerNorm` without the bias (which is an option that is available in PyTorch).

That's great!

PR #492, opened 2023-08-17
Basic `qmatmul` parallelization

LLukas22:
Uses `rayon`'s parallel iterator with size constraints to create reasonable batches for the quantized `vec_dot` function. Achieves ~80% of `llama.cpp`'s speed. 

Maybe this should be replaced by `kernel::par_for_each` or `kernel::par_range` for more fine grained control in the future. 

Also changes the cargo config to always build with `native` cpu features.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,3 +3,6 @@ rustflags = ["-C", "target-cpu=native"]
 
 [target.wasm32-unknown-unknown]
 rustflags = ["-C", "target-feature=+simd128"]
+
+[target.x86_64-apple-darwin]
+rustflags = ["-C", "target-feature=-avx,-avx2"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,7 +1,4 @@
-[target.x86_64-unknown-linux-gnu]
-rustflags = ["-C", "target-cpu=native"]
-
-[target.aarch64-apple-darwin]
+[build]
 rustflags = ["-C", "target-cpu=native"]
 
 [target.wasm32-unknown-unknown]
--- a/candle-core/src/quantized/k_quants.rs
+++ b/candle-core/src/quantized/k_quants.rs
@@ -1,6 +1,7 @@
 use super::GgmlDType;
 use crate::Result;
 use half::f16;
+use rayon::prelude::*;
 
 // Default to QK_K 256 rather than 64.
 pub const QK_K: usize = 256;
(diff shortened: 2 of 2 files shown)


.cargo/config.toml
```diff
@@ -1,8 +1,5 @@
-[target.x86_64-unknown-linux-gnu]
-rustflags = ["-C", "target-cpu=native"]
-
-[target.aarch64-apple-darwin]
+[build]
```
LaurentMazare, line 1: Do you know if this works with the wasm build? I would be a bit afraid it tries to add the native bit in the build there too and ends up being an issue.

candle-core/src/quantized/k_quants.rs
```diff
@@ -821,10 +822,16 @@ pub fn matmul<T: GgmlType>(
...
-        }
+
+        dst_row
+            .into_par_iter()
+            .enumerate()
+            .with_min_len(128)
+            .with_max_len(512)
+            .for_each(|(col_idx, dst)| {
+                let rhs_col = &rhs_t[col_idx * k_in_rhs_blocks..(col_idx + 1) * k_in_rhs_blocks];
+                *dst = T::vec_dot(k, rhs_col, lhs_row).unwrap();
```
LaurentMazare, line 833: Unwrap doesn't look so good here, if there is an error, we would want to report it properly rather than crashing.

.cargo/config.toml
LLukas22: I can't build wasm currently as `onig_sys v69.8.1` seams to do some funky stuff. But theoretically the more specific definition should overwrite the more generic on in a cargo config.

LaurentMazare: Ok I just tried the `.cargo/config.toml` change locally and I can still build the wasm bit / it still works so good with me.

candle-core/src/quantized/k_quants.rs
LaurentMazare: Ok, if this has any potential performance impact, let's maybe forget about it. The plan would be to check for these conditions outside of the hot loop at some point anyway so having an unwrap in the meantime wouldn't be that bad.

LLukas22: Performance basically hasen't changed after implementing it, but i guess the entire `qmatmul` has to be optimized at some point.  We can keep it in for now.

LLukas22:
Is it normal for the macos ci to fail?

LaurentMazare:
> Is it normal for the macos ci to fail?

Nope, I guess it might be an aftermath of the `native` change as the error is "illegal instruction". Maybe this added somehow avx to maxcos x86 and this ends up not being available somehow. Might be worth reversing this part of the change to see if it helps.

LLukas22:
Well i guessed `native` would mean it activates **only** the available features, maybe there is some virtualization shenanigans going on in the actions runners. Anyhow i disabled `avx` for intel based macs, lets see if the CI passes.

LaurentMazare (approved):

PR #524, opened 2023-08-20
Add quantization support for `q2k`, `q3k`, `q4k` and `q5k`

LLukas22:
A first draft of the quantization implementation and round-trip testing. 

Currently `q3k` is still missing and i would like to cleanup the code quite a bit, as it's currently hard to read and very C++ like.

Regarding the tests, i extended the quantization test to round-trip 1024 long vectors ranging from -0.5 to 0.5 and from -128 to 128 and compare the results to the actual vectors with a defined tolerance. I also check some hardcoded values to ensure it produces the same values as the original `ggml` implementation.


LaurentMazare:
Thanks for the PR, from looking at it quickly it looks pretty good! Would you mind running rustfmt and clippy so as to get the CI to be happy?

LLukas22:
`clippy` wanted a lot of changes as there was a lot of indexing going on. But as i already said im gonna give this another refactoring pass if `q3k` is implemented and tested.

candle-core/tests/quantized_tests.rs
```diff
@@ -124,51 +124,162 @@ fn quantize_q4_0() -> Result<()> {
...
+
+/// Round a vector
+fn round_vector(values: &[f32]) -> Vec<f32> {
+    values
+        .iter()
+        .map(|x| (1000. * x).round() / 1000.)
+        .collect::<Vec<_>>()
+}
+
+fn compare_with_error(values: &[f32], expected: &[f32], precision: usize, tolerance: f32) {
```
LaurentMazare, line 153: Why do we have both a precision and a tolerance here? Wouldn't just one of them be sufficient?

candle-core/tests/quantized_tests.rs
```diff
@@ -124,51 +124,162 @@ fn quantize_q4_0() -> Result<()> {
...
+#[test]
+fn quantize_q4k() -> Result<()> {
+    use k_quants::BlockQ4K;
+
+    let (src, mut dst) = get_test_vector(0.5, Some(1024));
+    let _quant = quantize_roundtrip::<BlockQ4K>(src.as_slice(), dst.as_mut_slice())?;
+    compare_with_error(src.as_slice(), dst.as_slice(), 6, 0.1);
+
+    // Test some specific values
+    dst = round_vector(&dst);
```
LaurentMazare, line 209: No need to change `dst`, `let dst = ...` should work fine if I'm not missing something.

candle-core/src/quantized/utils.rs
```diff
@@ -0,0 +1,295 @@
...
+) -> (f32, f32) {
+    // Get min/max
+    let mut min = x
+        .iter()
+        .take(n)
+        .fold(f32::MAX, |min, &val| if val < min { val } else { min });
+    let max = x
+        .iter()
+        .take(n)
+        .fold(f32::MIN, |max, &val| if val > max { val } else { max });
```
LaurentMazare, line 172: Use `max_by` with `total_cmp`?

LLukas22:
Thanks for giving it a quick review. I finally finished `q3k` which was more tedious to implement than i'd like to admit. I'll now start my refactoring pass to get rid of most of the indexing going on and i'll include your review comments.

LaurentMazare:
Maybe you can just deactivate the test manually for now only on macos with a `#[cfg(target_os = "linux")]` like directive?

LLukas22:
Well the CI passes on windows, but fails on linux/macos. Probably some weirdness going on with these lines: 
```rust
            //Reconstruct the scales
            let mut aux = unsafe {
                std::mem::transmute::<&mut [u8; 12], &mut [u32; 4]>(&mut block.scales.clone())
            };

            let tmp = aux[2];
            aux[2] = ((aux[0] >> 4) & KMASK2) | (((tmp >> 4) & KMASK1) << 4);
            aux[3] = ((aux[1] >> 4) & KMASK2) | (((tmp >> 6) & KMASK1) << 4);
            aux[0] = (aux[0] & KMASK2) | (((tmp) & KMASK1) << 4);
            aux[1] = (aux[1] & KMASK2) | (((tmp >> 2) & KMASK1) << 4);

            //Transfer the scales into an i8 array
            let scales = unsafe { std::mem::transmute::<&mut [u32; 4], &mut [i8; 16]>(&mut aux) };
```
 
I'll take another look on my linux machine.

LLukas22:
Alright, found the error. Some `uint` instance was free-floating in memory 😅 I'm out of time for today, will do the refactoring pass in the next few days.

candle-core/tests/quantized_tests.rs
LaurentMazare: This wasn't fixed or am I missin something? It's not that important though but when reading the code, people would wonder why `dst` mutable rather than defined a second time.

candle-core/tests/quantized_tests.rs
```diff
@@ -124,51 +124,180 @@ fn quantize_q4_0() -> Result<()> {
...
+    let (src, mut dst) = get_test_vector(0.5, Some(1024));
+    let _quant = quantize_roundtrip::<BlockQ4K>(src.as_slice(), dst.as_mut_slice())?;
+    compare_with_error(dst.as_slice(), src.as_slice(), 0.017);
+
+    // Test some specific values
+    dst = round_vector(&dst);
     assert_eq!(
-        [src[0], src[128], src[256], src[512], src[800], src[1023]],
-        [-0.5, -0.375, -0.25, 0.0, 0.28125, 0.49902344]
+        [dst[0], dst[128], dst[256], dst[512], dst[800], dst[1023]],
```
LaurentMazare, line 229: Did the assertions on `src` disappear? The goal was to have both assertions on `src` and `dst` so that its easy to check tha they are close.

LaurentMazare (commented):
Just a point for next time: as this PR combines both moving lots of code around and also some new code, it's pretty hard to make a sense of what is new and need actual review and what just moved around. Next time it might be better to have separate PRs for the moving and adding new code bits :)

candle-core/tests/quantized_tests.rs
LLukas22: Yeah i removed the `src` assertions as i added the `compare_with_error` function, to assure every value in the `dst` vector matches it `src` value with a certain tolerance. I could add the `src` asserts back but for lower quants (`q2k`, `q3k`) they can diverge quite a bit from the `dst` results. 

> This wasn't fixed or am I missin something? It's not that important though but when reading the code, people would wonder why dst mutable rather than defined a second time.

I forgot to redefine some of the `dst` values, gonna fix it now.

LaurentMazare: If it's easy to add back some of the `src` asserts, that would be nice, it's only a couple lines and gives a bit more context on how much difference to expect.

LLukas22:
Alright, should be good to go. 

> Just a point for next time: as this PR combines both moving lots of code around and also some new code, it's pretty hard to make a sense of what is new and need actual review and what just moved around. Next time it might be better to have separate PRs for the moving and adding new code bits :)

Yeah sorry about that, got a bit carried away 😅. My next vec-dot PRs will be a lot more organized 👍

LaurentMazare (approved):

LaurentMazare:
Merged, thanks again!

PR #536, opened 2023-08-21
implement full method

cksac:
create tensor with fill value like `torch.full`
```rust
let a = Tensor::full(3.14f32, (2, 3), &Device::Cpu)?;

let b = Tensor::full(128u32, (2, 3), &Device::Cpu)?;
```

### DIFF
--- a/candle-core/src/backend.rs
+++ b/candle-core/src/backend.rs
@@ -1,5 +1,5 @@
 use crate::op::{BinaryOpT, CmpOp, ReduceOp, UnaryOpT};
-use crate::{CpuStorage, DType, Layout, Result, Shape};
+use crate::{CpuStorage, DType, Layout, Result, Shape, WithDType};
 
 pub trait BackendStorage: Sized {
     type Device: BackendDevice;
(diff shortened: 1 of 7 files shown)


LaurentMazare:
I think this could also be done via `Tensor.ones(...)? * 42` and this would avoid adding complexity on all the different layers. It's arguably slighly less efficient but do you have a use case where this would really matter?

cksac:
```rust
use anyhow::Result;

fn main() -> Result<()> {
    let a = Tensor::ones((2, 3), candle_core::DType::U32, &Device::Cpu)?;
    let b = (&a * 42)?;
    Ok(())
}
```
I got
```
12 |     let b = (a * 42)?;
   |                ^ the trait `Borrow<Tensor>` is not implemented for `{integer}`
```

cksac:
I think the full method added little complexity, especially cuda using existing const_impl, but provide smoother pytorch migration experience and more efficient.

there are also some expected to work but don't, e.g.
```rust
  // works
  let b = (&a * 42f64)?;
  // not works
  // let c = a.mul(42f64)?;
```

cksac:
update full to take typed fill value

LaurentMazare:
A simpler version of this has been implemented in #1455.

PR #590, opened 2023-08-24
Align tensor device print more with PyTorch

patrickvonplaten:
I have been playing around a bit with tensor displaying and I think it could make sense to align `candle-core` more with PyTorch.

PyTorch tensor printing looks as follows:
**CPU**
```py
import torch
torch.randn((2,2))
```

```bash
tensor([[-2.1989, -2.5581],
        [ 2.2192,  0.8749]])
```

**GPU**
```py
import torch
torch.randn((2,2), device="cuda")
```

```bash
tensor([[ 0.5948, -0.7603],
        [-0.2274,  0.2770]], device='cuda:0')
```

I think it's very useful to see which device the tensor is on and as stated here: https://github.com/huggingface/candle/blob/2cde0cb74be3ada6b2d845a018d4746929318afb/candle-core/src/display.rs#L3 it printing should align `candle-core` printing as much as possible to PyTorch.

This PR makes a couple of changes so that tensor printing gives access to the cuda device and align more with PyTorch.

**CPU**
```bash
[[ 0.1413,  0.4800],
 [-0.9911, -1.1071]]
Tensor[[2, 2], f32]
```

**GPU**
```bash
[[-1.8936,  0.9397],
 [ 1.0709, -0.0135]]
Tensor[[2, 2], f32, cuda:0]
```


LaurentMazare:
Ah I struggled quite a bit to align the formatting to somehow match the PyTorch side (and it's far from perfect) but yeah certainly good to add these details.
Could you look at the CI failures? Roughly you cannot go directly to the `cuda_backend` internals as it may not be available (and the dummy backend is used).

candle-core/src/display.rs
```diff
@@ -9,11 +9,20 @@ impl Tensor {
         &self,
         f: &mut std::fmt::Formatter,
     ) -> std::fmt::Result {
-        let prefix = match self.device() {
-            crate::Device::Cpu => "Cpu",
-            crate::Device::Cuda(_) => "Cuda",
+        #[cfg(feature = "cuda")]
```
LaurentMazare, line 12: What about using `self.device().location()` instead, this should avoid the need for using the `cuda` feature gate so is more likely to work well when/if we add a new device some day?

patrickvonplaten: Indeed much cleaner! Thanks

LaurentMazare:
Thanks!

PR #591, opened 2023-08-24
[WIP] Improve Yolo WASM UI example

radames:
Only missing the last UI iteration, I'll push soon


candle-wasm-examples/yolo/src/worker.rs
radames, line 1: hi @LaurentMazare added the  conf_threshold and iou_threshold, does this make sense?

LaurentMazare: Yes, you may want to look at the clippy lint that failed in the CI - and get into the habit of running clippy locally - it will tell you a lot about best practices, here around `Option<_>`.

LaurentMazare:
Looks pretty great, some small points:
- If you haven't already done so, could you test the pure rust UI building it via trunk and checking that it still looks fine?
- Not sure if we want to include the images as they add to the repo size, maybe just have a link to them?
- Either make a script for building the wasm bits for the JS UI (and download the image too?) or put some indications in a local `README.md` on how to build/deploy?

candle-wasm-examples/yolo/src/model.rs
```diff
@@ -623,16 +623,31 @@ fn iou(b1: &Bbox, b2: &Bbox) -> f32 {
...
+    pred: &Tensor,
+    img: DynamicImage,
+    w: usize,
+    h: usize,
+    conf_threshold: f32,
+    iou_threshold: f32,
+) -> Result<Vec<Vec<Bbox>>> {
     let (pred_size, npreds) = pred.dims2()?;
     let nclasses = pred_size - 4;
+    let conf_threshold = Some(conf_threshold)
```
radames, line 636: I don't get this error on my env, I'm on OSX, is there a better way to do this ````used `unwrap_or()` on `Some` value```` ?

LaurentMazare: `unwrap_or` only applies when the option is `None` so doing `Some(a).unwrap_or(b)` is actually equivalent to `a` :) Not sure how your setup is but hopefully clippy should be easy to install on macos too (or at least it was for me :) )

radames: I was trying to set a default value as well as constrain to 0-1, is there a better way to do this? thanks

candle-wasm-examples/yolo/src/worker.rs
radames: thanks going to try clippy and see if it can help me improve

candle-wasm-examples/yolo/src/model.rs
LaurentMazare: As the parameter of the function is `u32`, it will always be present so there is no notion of default value. If you want to make it optional, you can change it to `Option<u32>` in the argument list?
The min/max looks reasonable, slightly more idiomatic is `.clamp(0., 1.)`.

radames: ok that's great feedback, I'll assume there's always a preset value then, and clamp it.

LaurentMazare:
Looks pretty good to me, if you think it's ready can you remove the draft tag from the PR and we can merge?

radames:
thanks @LaurentMazare
> If you haven't already done so, could you test the pure rust UI building it via trunk and checking that it still looks fine?
I've tested with the new changes and it's as before. I had to hardcode the new params as the worker now expect `model_size`, `conf_threshold` and `iou_threshold`

LaurentMazare (approved):

LaurentMazare:
Merged, great work on this UI, that's super helpful!

PR #601, opened 2023-08-25
Repairing cast bf16/f16

Narsil:
Currently complains because there's multiple cast possible.

Just forcing the actual cast type fixes it.

### DIFF
--- a/candle-kernels/src/cast.cu
+++ b/candle-kernels/src/cast.cu
@@ -13,13 +13,13 @@ extern "C" __global__ void FN_NAME( \
     const size_t *strides = info + num_dims; \
     if (is_contiguous(num_dims, dims, strides)) { \
         for (unsigned int i = blockIdx.x * blockDim.x + threadIdx.x; i < numel; i += blockDim.x * gridDim.x) { \
-            out[i] = (DST_TYPENAME) inp[i]; \
+            out[i] = DST_TYPENAME(inp[i]); \
         } \
     } \
     else { \
         for (unsigned int i = blockIdx.x * blockDim.x + threadIdx.x; i < numel; i += blockDim.x * gridDim.x) { \
             unsigned strided_i = get_strided_index(i, num_dims, dims, strides); \
-            out[i] = (DST_TYPENAME) inp[strided_i]; \
+            out[i] = DST_TYPENAME(inp[strided_i]); \
         } \
     } \
 } \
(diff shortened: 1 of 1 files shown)


LaurentMazare (approved):

candle-kernels/src/cast.cu
```diff
@@ -13,13 +13,13 @@ extern "C" __global__ void FN_NAME( \
     const size_t *strides = info + num_dims; \
     if (is_contiguous(num_dims, dims, strides)) { \
         for (unsigned int i = blockIdx.x * blockDim.x + threadIdx.x; i < numel; i += blockDim.x * gridDim.x) { \
-            out[i] = inp[i]; \
+            out[i] = DST_TYPENAME(inp[i]); \
```
LaurentMazare, line 16: If this version works that's fine but if it doesn't create issue maybe a `static_cast` would be the most idiomatic way?

LaurentMazare: (CI actually doesn't seem very happy with this "function call" version)

Narsil: Quite odd, I can't reproduce on 11.8 on the host machine, even when forcing CUDA_11.8.

Narsil: Ok reproduced, the issue is not in the 11.8 headers, but in the 11.8 nvcc.

Looking how to fix.

Narsil: Ok the `bf16 <-> f16` cast is actually missing in cu118....

Narsil: I don't feel good at all with the proposed fix... There must be a better way..

candle-kernels/src/cast.cu
```diff
@@ -24,19 +24,42 @@ extern "C" __global__ void FN_NAME( \
...
+    const size_t num_dims, \
+    const size_t *info, \
+    const SRC_TYPENAME *inp, \
+    DST_TYPENAME *out \
+) { \
+    const size_t *dims = info; \
+    const size_t *strides = info + num_dims; \
+    if (is_contiguous(num_dims, dims, strides)) { \
+        for (unsigned int i = blockIdx.x * blockDim.x + threadIdx.x; i < numel; i += blockDim.x * gridDim.x) { \
+            out[i] = (DST_TYPENAME) (float) inp[i]; \
```
Narsil, line 39: Can't find anything better, happy to try things.

The low level cast just doesn't exist. This seems OK since f32 >> (bf16, f16) so the uplift shouldn't be an issue.
I'm wondering how much the compiler can remove the intermediary conversion though.

LaurentMazare: Ok let's go ahead with this version then.

LaurentMazare (approved):

PR #641, opened 2023-08-28
Improve the mnist-training error message when dataset is not found

shinglyu:
The user need to manually download the dataset and place it in the right location. In the current implementation, the user will only get a generic "No such file or directory" error and don't know what to do. Adding an error message to guide the user to download the dataset.


LaurentMazare:
Thanks, I think this is superseded by #439 which now gets the dataset to be automatically downloaded from the huggingface hub.

shinglyu:
Yes, I confirmed that it works without manual intervention now. Closing the PR.

PR #665, opened 2023-08-29
Adds chat mode for quantized models example

kstavro:
Added completion and chat mode to the quantized example, where we prompt the model from the CLI. The model does not need to reload each time we prompt it. In chat mode, the conversation history is taken into account.

I tried to also cache the q values of the context history when reprompting the model in chat mode, but somehow I couldn't get the attention masking working. In particular, the mask can't be rebroadcasted into the attention shape in L165 (of the new script), because now the mask is of shape `[prompt, prompt+context_history]` (in other words, `[Dim(prompt_q), Dim(prompt_k) + Dim(k_cache)]`). I don't understand all the intricacies of the implementation of the mask in the ggml models, so (for) now each time we reprompt, we feed the model with all the context history (up to the last 4096 tokens). Which works, but has a significant impact to the processing of the prompt after several back and forths with the model. If you have an idea how to make a `q_cache` work (basically getting the masking as above to work), that would be nice to add.

Let me know if that works for the example. Or of any additions you think might be nice to have.

### DIFF
--- a/candle-examples/examples/quantized/main.rs
+++ b/candle-examples/examples/quantized/main.rs
@@ -16,7 +16,6 @@ use candle_nn::{Embedding, Module};
 use candle_transformers::generation::LogitsProcessor;
 
 const MAX_SEQ_LEN: usize = 4096;
-const DEFAULT_PROMPT: &str = "My favorite theorem is ";
 
 struct RmsNorm {
     inner: candle_nn::LayerNorm,
@@ -114,7 +113,13 @@ impl LayerWeights {
         Ok(rope)
     }
 
-    fn forward_attn(&mut self, x: &Tensor, mask: &Tensor, index_pos: usize) -> Result<Tensor> {
+    fn forward_attn(
+        &mut self,
+        x: &Tensor,
+        mask: &Tensor,
+        index_pos: usize,
+        update_kv_cache: bool,
+    ) -> Result<Tensor> {
         let _enter = self.span_attn.enter();
         let (b_sz, seq_len, n_embd) = x.dims3()?;
         let q = self.attention_wq.forward(x)?;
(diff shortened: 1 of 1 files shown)


kstavro:
I just noticed that you also added llama-code in the meanwhile to the example, this is missing from my PR, need to add the missing stuff.

Narsil:
My personal take is that this is an example, not an actual product/finished code, so I wouldn't add it.

kstavro:
> My personal take is that this is an example, not an actual product/finished code, so I wouldn't add it.

That is fair, no strong arguments there 👍 

I would still like though to bring to your attention the following perspective from the user POV, which was my very first motivation to touch the example's code: I wanted to play with the quantized models and prompt them multiple times to see how they behave from within candle. This is quite cumbersome with the current state of the example, because you have to run main and load the models weights from scratch each time the model is prompted, so at least ~2 unnecessary extra secs with each prompt (for the 7b models and on my machine, could be less or more for others). So I would still consider changing the interface of the model in the example to keep it loaded in memory and prompt it from the CLI, this would improve the out-of-the box experience of playing with this example a lot imho, and practically for free, just 3 lines of code or so (and wrapping most of main into a loop as in this PR).  My feeling is that right now the example does showcase candle (and this works great), but sort of discourages you from playing with the actual models from within candle. But I do realize this might not be very important in the big scope of the examples of the repo (and if you decide this can be changed, it doesn't have to be with this PR of course).

The chat mode just came only afterwards, after converting the interface to cli and keeping the model in memory, I wanted to also play with the models by talking to them back and forth as a dialogue. No intention of adding any finished code or the likes, I just thought it might be useful to the example since I already wrote it in the first place. 

I leave it to you to close this PR if it has nothing else to offer to the example. Looking forward to the the next stuff you implement, will keep playing with the examples, very excited with the repo (and its pace 🔥 ). 

Cheers!

LaurentMazare:
Thanks for the PR and for the detailed explanation. It certainly makes sense from a user POV but on the other hand I totally agree with @Narsil that these examples are "showcase" for candle and we wouldn't want to include all the llama.cpp feature - we would like the examples to be as easy to read and understand as possible. We would certainly like to see other folks develop a feature rich llama.cpp equivalent based on candle, it would bring great feedback to candle on how usable things are and would avoid mixing everything in the same repo.
As a middle ground, I just merged #690 which tries to add the `interactive` (no history retained but multiple prompt) and `chat` (history retained, kv-cache flushing so slow) to the current model. So hopefully you can play with this to get a sense of the model without reloading the weights each time.
It tries to change as little things as possible and preserve the existing modes, that said I'm not totally sure about the added complexity so we may well strip this feature at some point (and would certainly prefer not adding more complexity on top of it for now).

LaurentMazare:
Closing this now, but again I would encourage you to create an implementation with a lot more feature in its own repo if you have some interest (and would be keen to have an "external resources" section in the candle readme that would point at it).

PR #667, opened 2023-08-29
Lamb example

leifu1128:
Self contained example of custom optimizer, in this case Lamb, which is the go to for gigabatch training. I copied the main function from the other example, hope thats ok. Also would be nice to have official ops for clamp and norm.

### DIFF
--- a/candle-nn/examples/lamb_optimizer.rs
+++ b/candle-nn/examples/lamb_optimizer.rs
@@ -1,0 +1,164 @@
+use candle::{Tensor, Device, Var, Result, Module, DType};
+use candle_nn::{Linear, linear, VarBuilder, VarMap};
+
+fn clamp(x: &Tensor, min: f64, max: f64) -> Result<Tensor> {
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Thanks for the PR, looks pretty interesting but I would be in favor of putting this in an external repo instead, I think we're lacking external resources at the moment and having more would be great for several reasons:
- It would help ensure that the public apis are properly defined and easy to use from external crates.
- Potential users would have a better sense of what can be achieved just by using the library without internal hooks.
- Less of a need for reviews, uniform code style, etc, so this will favor a diversity of approach.
- The external crates could use all the dependencies they want (not that it matters here).
- It would give you more space so that you can expand on what the example does, how it should be used, and maybe even have some forms of tutorials.
The only downside I would see is discoverability and for that we could have some links to external resources from the readme so that it's easy to find these out.

leifu1128:
Makes sense, closing.

PR #668, opened 2023-08-30
fix: use correct BLOCK_SIZE in matmul function for ggml type

xffxff:
To compute `k_in_lhs_blocks`, we should use `T::VecDottype::BLOCK_SIZE` because we obtain `lhs_b` using `T::VecDotType::from_float`. Similarly, to calculate `k_in_rhs_blocks`, we need to use `T::BLOCK_SIZE` since `rhs_t` is a slice of `T` (I'm just starting to learn quantization, so I'm not sure if it's correct, but it seems intuitive to me)

image


ivarflakstad:
`T::BLCK_SIZE` is (or at least should always be) equal to `T::VecDotType::BLCK_SIZE`.

PR #669, opened 2023-08-30
Improve Whisper WASM UI example

radames:
* Examples are from https://huggingface.co/datasets/Narsil/candle-examples
* I'm not sure why [`tiny.en`](https://huggingface.co/spaces/lmz/candle-whisper/blob/main/tiny.en.safetensors) only has 75MB on your repo and here I'm point to [openai repo](https://huggingface.co/openai/whisper-tiny.en/resolve/refs%2Fpr%2F17/) which has 151MB
* I see on the WASM worker config is always set to `let config = Config::tiny_en();` does it worth to enable loading this from the model `config.json` considering the size of the other models? 

You can see live demo [here](https://huggingface.co/spaces/radames/candle-whisper-wasm)


candle-wasm-examples/whisper/src/bin/m.rs
```diff
@@ -0,0 +1,41 @@
...
+    #[wasm_bindgen(constructor)]
+    pub fn new(
+        weights: Vec<u8>,
+        tokenizer: Vec<u8>,
+        mel_filters: Vec<u8>,
+    ) -> Result<Decoder, JsError> {
+        let decoder = D::load(ModelData {
+            tokenizer,
+            mel_filters,
+            weights,
```
radames, line 20: potentially we could have `config: JsValue` here? and fetch `config.json` from the model file, but I need to figure out how to convert it to a `struct Config `

LaurentMazare: Sounds like a great idea if we want to support multiple versions (the command line version of the whisper example does). You're probably looking for something like:
```rust
let config: Config = serde_json::from_str(&json_string)?)?
```
You can grep in the repo to find some examples of this being used, it mostly requires the `Config` struct to have the proper `#[derive(Serialize, Deserialize)]` annotation so that the serialization/deseralization codes are generated.

LaurentMazare (approved):

candle-wasm-examples/whisper/src/bin/m.rs
LaurentMazare: Certainly happy to defer this to a future PR if you want. Also note that this is still using the very old version of the whisper model so I plan on updating it to the latest version used by the command line tool, hopefully this should improve performance significantly.

radames:
Fell free to merge, when you can. Thanks

PR #699, opened 2023-09-01
Allow retrieving and setting prefix of VarBuilder

chronicl:
Since a recent change `VarBuilder::get` doesn't panic anymore when a tensor couldn't be found and instead initializes it silently. `VarBuilder::contains_tensor` allows for manually checking wether the tensor exists, but I didn't find any way to figure out what the prefix of the current `VarBuilder` is, which is why I need `VarBuilder::prefix`.
`VarBuilder::set_prefix` is useful when some tensor that isn't on the current path needs to be retrieved, for example when weights are tied.

### DIFF
--- a/candle-nn/src/var_builder.rs
+++ b/candle-nn/src/var_builder.rs
@@ -102,6 +102,20 @@ impl<'a, B: Backend> VarBuilderArgs<'a, B> {
         }
     }
 
+    /// Returns the prefix of the `VarBuilder`.
+    pub fn prefix(&self) -> String {
+        self.path.join(".")
+    }
+
+    /// Returns a new `VarBuilder` with the prefix set to `prefix`.
+    pub fn set_prefix(&self, prefix: impl ToString) -> Self {
+        Self {
+            data: self.data.clone(),
+            path: vec![prefix.to_string()],
+            _phantom: std::marker::PhantomData,
+        }
+    }
+
     /// Return a new `VarBuilder` adding `s` to the current prefix. This can be think of as `cd`
     /// into a directory.
     pub fn push_prefix<S: ToString>(&self, s: S) -> Self {


LaurentMazare:
`VarBuilder::get` should only pass the `get` query to the varbuilder backend which implementation could decide to initialise the tensor or not. Only the `VarMap` one can initialize things. Could you expand a bit more on how you want to actually use the newly exposed `prefix`? I don't really mind them being exposed but there might be some easier way around this.

chronicl:
I see, then it wasn't an update, but me switching from having a safetensors backend to a varmap backend that caused the change.

Short answer: Having the prefix exposed helps debugging when creating new models.
Long answer: I was implementing a vision encoder decoder model in candle, meaning it's a combination of two models and I used a `VarMap` to store the weights of the two models in one place. The model didn't work as expected and it took me a while to figure out that some weights weren't being loaded from the varmap but were instead initialized as zero, because I got a path wrong. So I created a helper function `assert_tensor_exists` that checked at every point where tensors are being loaded with `VarBuilder::get` wether the tensor actually exists and thus won't be silently initialized as zero. That helped me find the place where things went wrong, but I then had to go backward through all `VarBuilder::pp` calls to figure out what the actual path is that is wrong. `VarBuilder::prefix` would have allowed me to immediately include the path name in the panic call of `assert_tensor_exists`.

Now why I would like to have `VarBuilder::set_prefix`: The decoder part of the model I was implementing is `BertLMHeadModel` from the `transformers` library and it ties weights from the embedding layer with weights in the prediction head. The weights are stored as `bert.embeddings.word_embeddings.weight` in the safetensors file, which is the correct path for the embedding layer, but for the prediction head the `VarBuilder` is in a completely different path and I didn't find a way to easily access `bert.embeddings.word_embeddings.weight` from there. I just resorted to passing an extra clone of the original `VarBuilder` without prefix all the way down to the prediction head.

Are there easier ways to accomplish these things?

LaurentMazare:
Ok, I'll merge this as it definitely sounds helpful.
Re your use case, `from_varmap` is mostly for training, I'll make sure the documentation/name reflect this better. Note that `from_safetensors` can combine multiple safetensors files and if this doesn't work you can use `from_tensors` that just use a hashmap and would fail on missing tensors.
For `VarMap` we could also add the possibility to make it fail if some tensors haven't been set in the `load` call, this would avoid implicitely using tensors that have been initialized rather than loaded.
Using `set_prefix` sounds reasonable though it's a bit weird / un-idiomatic, I'll also make the doc reflect this. I would just build the path from the root var-builder using `.pp` on each part (`vb_root.pp("bert.embeddings.word_embeddings.weight")` would work too).

LaurentMazare (approved):

PR #705, opened 2023-09-01
Cleanup the pyo3 setup.

LaurentMazare:


Narsil:
I recommend maturin here ! Very nice tool for the building process.
The  GH action enable building for many platforms too

LaurentMazare:
Yeah at this stage I care more about something that is super easy to understand and audit than having something more involved but we should certainly set it up soon ish.

PR #725, opened 2023-09-03
Add `GGUFv2` support

LLukas22:
GGUF was extended by `u64`, `i64` and `f64` types as mentioned [here](https://github.com/ggerganov/ggml/pull/302#issuecomment-1694484996). All ranges were also changed from `u32` to `u64` as someone actually build a file which exceeded the `u32` limit of the tensor sizes. 

This PR allows to read V1 and V2 files, but only V2 files can be created.


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks!

PR #751, opened 2023-09-05
WASM example fix

radames:
This fix a caching issue, since HF models are not correctly cache, here using Cache API to store the models appropriately
update UI to change seq_len depending on model size


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #768, opened 2023-09-07
Add small customization to the build 

gzsombor:
I tried to build locally with flash-attention, but I run into a couple of problems.
First, Cuda 11.x doesn't work well with the gcc 11, which is the default on Ubuntu 22, but the current build script didn't provide any knob to control the used compiler, so I've added support for a new environment variable.
Second,  when I specified a relative path as CANDLE_FLASH_ATTN_BUILD_DIR, the build failed again - so I've fixed that too.

### DIFF
--- a/candle-flash-attn/build.rs
+++ b/candle-flash-attn/build.rs
@@ -60,6 +60,10 @@ fn main() -> Result<()> {
         Ok(build_dir) => PathBuf::from(build_dir),
     };
     set_cuda_include_dir()?;
+
+    let ccbin_env = std::env::var("CANDLE_NVCC_CCBIN");
+    println!("cargo:rerun-if-env-changed=CANDLE_NVCC_CCBIN");
+
     let compute_cap = compute_cap()?;
 
     let out_file = build_dir.join("libflashattention.a");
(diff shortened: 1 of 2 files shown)


gzsombor:
The original problem was this broadly recited compilation problem: https://github.com/NVlabs/instant-ngp/issues/119

candle-flash-attn/build.rs
```diff
@@ -57,9 +57,22 @@ fn main() -> Result<()> {
             #[allow(clippy::redundant_clone)]
             out_dir.clone()
         }
-        Ok(build_dir) => PathBuf::from(build_dir),
+        Ok(build_dir) =>
+        {
+            let path = PathBuf::from(build_dir);
+            let path = if path.is_relative() {
```
LaurentMazare, line 63: Shouldn't canonicalize take care of making the path absolute? The doc mentions that it's similar to `realpath`.

gzsombor: That's true. However, the error message, while true, is a bit misleading. When I first did that way, I was confused, for an embarassingly long time, why it complains, that "./target/libflash" does not exist, even though I've just created that directory. In hindsight, it was obvious, that the working directory of this script is "./candle-flash-attn" - and not the root of the repository - so if I want to store something under "./target", I need to specify as "../target/libflash". 
 I thought, about interpreting the relative path from the repository root, but I'm not sure, how much confusion that would occur.

LaurentMazare: Maybe we should just tweak the error message to mention the current directory?

gzsombor: Yes, that could work as well.

LaurentMazare (approved):

LaurentMazare:
Thanks!

gzsombor:
Welcome!

PR #788, opened 2023-09-09
Basic type-hinting for the pyo3 wrapper.

LLukas22:
This PR adds some very basic type-hints for the PyO3 wrapper. This should improve the development experience as the python language server can now infer the types and support the user a bit.

https://github.com/huggingface/candle/assets/65088241/2b1bab7a-2343-43cb-8abc-f436063423bd

Currently only a few functions (mainly everything used in the `test.py`) got added to the stub file. If we want to add more type-hints i could add them later.

### DIFF
--- a/candle-pyo3/candle/candle.pyi
+++ b/candle-pyo3/candle/candle.pyi
@@ -1,0 +1,111 @@
+from typing import Any, List, Tuple, Union
+from enum import Enum, auto
+
+class Device(Enum):
+    """
+    Backend device for a tensor.
+    """
+    Cpu = auto(),
+    Cuda = auto(),
+
+class DType:
+    """
+    The DType of a tensor.
+    """
+    ...
+
+class QTensor:
+    """
+    Represents a quantized `candle` Tensor.
+    """
+    def dequantize(self) -> Tensor:
+        ...
+
+class Tensor:
+    """
+    Represents a internal `candle` Tensor.
+    """
+    def __init__(self, data: Any) -> None: 
+        """
+        Create a Tensor from given data.
+        """
+
+    @property
+    def shape(self) -> List[int]:
+        """
+        Returns the shape of the Tensor.
+        """
+
+    @property
+    def rank(self) -> int:
+        """
+        Returns the rank of the Tensor.
+        """
+    
+    @property
+    def device(self) -> Device:
+        """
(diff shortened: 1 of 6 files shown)


candle-pyo3/src/lib.rs
```diff
@@ -101,28 +102,6 @@ impl PyDevice {
     }
 }
 
-impl<'source> FromPyObject<'source> for PyDevice {
```
LaurentMazare, line 104: The goal here was to be able to specify the device as a string. Could we maintain this functionality?

candle-pyo3/candle/candle.pyi
```diff
@@ -0,0 +1,111 @@
+from typing import Any, List, Tuple, Union
```
LaurentMazare, line 1: Did you have to type all this manually? I'm a bit worried that this will be hard to maintain when making changes to the bindings, I imagine that there is no way to generate these automatically instead at the moment? (I would have thought that it's not that hard actually as it's easy to introspect module elements and documentations in python)

LLukas22: Yeah i created those manually and you are absolutely right, this will be a nightmare to maintain. I looked into generating them automatically (or at least from the `text_signature`), but PyO3 doesn't seam to support this currently. See:

-  https://github.com/PyO3/pyo3/issues/2454
- https://github.com/PyO3/pyo3/issues/1112

This seams to be blocked by CPython directly: https://github.com/python/cpython/pull/101872 

I'm think we should close this PR as there will probably be a lot of changes in candles implementation. This is kind of unfortunate as using the python bindings without type-hints can be a bit tedious. Do you have any suggestions?

LaurentMazare: Up to you, as you're planning on making changes to the pyo3 bits as long as you keep this file in mind, I'm happy with merging it - especially if you think it makes you more productive when trying out the python bindings.

LLukas22: Yeah i don't think creating them manually is sustainable in the long run. I probably look into how [tokenizers](https://github.com/huggingface/tokenizers/tree/main) solves this with their [stub.py](https://github.com/huggingface/tokenizers/blob/main/bindings/python/stub.py) file.

Issue #680, opened 2023-08-30
MatMulNonContiguous Cuda Error when multiply tensors


Ssslakter:
Hi, the code below shows how to get the error from cuda about non-contiguous matrix multiplication, though I believe this should work since I just want to multiply two zero matrices
```rust
use candle_core::{DType, Device, Result, Tensor};

fn main() -> Result<()> {
    // Use Device::new_cuda(0)?; to use the GPU.
    let device = Device::new_cuda(0)?;

    let a = Tensor::randn(0f32, 1., (2, 2), &device)?;
    let b = Tensor::randn(0f32, 1., (2, 2), &device)?;

    let c = a.matmul(&b)?;

    println!("Digit {c:?} digit");

    let a = Tensor::zeros((1, 2), DType::F32, &device)?;
    let b = Tensor::zeros((2, 2), DType::F32, &device)?;

    let c = a.matmul(&b);
    
    // will print Digit Err(WithBacktrace { inner: Cuda(MatMulNonContiguous { lhs_stride: [0, 0], rhs_stride: [0, 0], mnk: (1, 2, 2) }), ...
    println!("Digit {c:?} digit");
    Ok(())
}
```
I tried to understand the part with gemm algorithms, but it's too hard just from reading the nvidia docs. 
The problem is with the `Tensor.layout.stride`  being just zeros in `Tensor::zeros`
The same happens with `Tensor::ones`.

_Machine info:
GPU - RTX 2060 
Windows 10
Cuda v11.8_

LaurentMazare:
The `Tensor::zeros` function broadcast the zeros rather than creating a dense matrix with them (which would be contiguous), this explains the issue you're seeing. You probably just want to call `.contiguous()?` after creating the matrixes that you want to ensure are contiguous - maybe at some point this will be done automatically for you by `matmul` but for now we prefer erring on the explicit side.

Ssslakter:
Ok, I see, I guess I was misleaded by the example from the book https://huggingface.github.io/candle/guide/hello_world.html which does not have a .`contiguous()?` before creating the model

LaurentMazare:
Right, that's indeed a bit of a bummer than the default examples don't work on cuda. @Narsil what do you think? The easiest fix would probably be to replace the `zeros` with `randn` in the examples, would that sound reasonable to you?

Narsil:
Very reasonable.

I'm torned on the `zeros` being broadcast tbh too.
[...]

PR #791, opened 2023-09-09
Little docs changes

Ssslakter:
Closes https://github.com/huggingface/candle/issues/680 

This pr basically is a little documentation fix to make example working straight from copy-pasting them and changing the device to cuda.

### DIFF
--- a/candle-book/Cargo.toml
+++ b/candle-book/Cargo.toml
@@ -11,7 +11,7 @@ readme = "README.md"
 
 [dependencies]
 accelerate-src = { workspace = true, optional = true }
-candle-core = { path = "../candle-core", version = "0.2.1", package = "candle-core" }
+candle = { path = "../candle-core", version = "0.2.1", package = "candle-core" }
 candle-datasets = { path = "../candle-datasets", version = "0.2.1" }
 candle-nn = { path = "../candle-nn", version = "0.2.1" }
 candle-transformers = { path = "../candle-transformers", version = "0.2.1" }
--- a/candle-book/Cargo.toml
+++ b/candle-book/Cargo.toml
@@ -11,7 +11,7 @@ readme = "README.md"
 
 [dependencies]
 accelerate-src = { workspace = true, optional = true }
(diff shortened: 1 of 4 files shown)


LaurentMazare (approved):

LaurentMazare:
It seems that your `Cargo.toml` change renaming `candle` to `candle-core` is actually breaking a few things in the clippy CI. Could you have a look?

Ssslakter:
Ok, this is interesting, I tried to run the `mdbook test -L ..\target\release\deps` on this chapter and it only worked after I changed the name to `candle-core`. I haven't noticed the imports in `lib.rs`. My guess is that it is breaking since docs use `extern crate candle_core;` and tests use  `use candle;` so different names break one of these. Maybe we should make the imports in the docs and `lib.rs` consistent, so we have everywhere `use candle;` or everywhere `use candle_core;` ?

LaurentMazare:
Sounds right, I would suggest `candle` everywhere (as we're doing in all examples etc) though not a very strong opinion here.

Ssslakter:
Well... I believe I would have to keep it as it was before. Probably has to do with how rustdoc uses compiled libraries

LaurentMazare (approved):

LaurentMazare:
Ah there still seems to be some CI issues, this time in the "Test candle-book" which is probably relevant :) Would you mind having another look?

LaurentMazare:
Looks good now, thanks for the PR and sorry for all the back and forth!

Ssslakter:
Whoops, I accidently forgot to replace to random in this line https://github.com/huggingface/candle/blob/main/candle-book/src/guide/hello_world.md?plain=1#L32 
Omg, how silly to make a mistake in that simple😅 
@LaurentMazare I could make a new one-line pr or it seems would be better to edit this line alongside with other future documentation changes

LaurentMazare:
1 line PR is good if you have the time for it (otherwise I'll make it on my side, pretty sure I would forget about it if not addressed now :) ).

Ssslakter:
Good! Here it is https://github.com/huggingface/candle/pull/804 🤗

PR #792, opened 2023-09-09
Group support in conv-transpose2d.

LaurentMazare:


LaurentMazare:
Too old, lots of things have changed since then...

PR #794, opened 2023-09-10
Move some models to candle-transformers so that it's easier to re-use.

LaurentMazare:


LaurentMazare:
This will also allow for models to be tested more easily.

PR #816, opened 2023-09-11
Add weight, bias, and hidden_size methods

EricLBuehler:
# Description
I added `weight` and `bias` methods to Conv1d and Conv2d, and the `hidden_size` method to Embedding

# Breaking changes
None, because this only adds functionality.

### DIFF
--- a/candle-nn/src/embedding.rs
+++ b/candle-nn/src/embedding.rs
@@ -21,7 +21,7 @@ impl Embedding {
 
     /// Get the hidden size of the embedding matrix
     pub fn hidden_size(&self) -> usize {
-        self.embeddings.dim(1).unwrap()
+        self.hidden_size
     }
 }
 


candle-nn/src/embedding.rs
```diff
@@ -18,6 +18,11 @@ impl Embedding {
     pub fn embeddings(&self) -> &Tensor {
         &self.embeddings
     }
+
+    /// Get the hidden size of the embedding matrix
+    pub fn hidden_size(&self) -> usize {
+        self.embeddings.dim(1).unwrap()
```
LaurentMazare, line 24: Expose `self.hidden_size` instead? This would avoid the `unwrap` (better having as few of these as possible)

EricLBuehler: True, I will convert that.

LaurentMazare: Thanks

LaurentMazare (approved):

PR #819, opened 2023-09-11
Implement top_p / nucleus sampling

jbochi:
This is an implementation of nucleus sampling, introduced in the paper [The Curious Case of Neural Text Degeneration](https://papers.labml.ai/paper/1904.09751).

It's supported by [llama-cpp](https://github.com/ggerganov/llama.cpp/blob/d54a4027a6ebda98ab0fef7fa0c2247d0bef132a/llama.h#L444), [llama2.c](https://github.com/karpathy/llama2.c/blob/35deb5e0fa55f0a257040bcf1624ed8386e63dc7/run.c#L631), [open ai](https://platform.openai.com/docs/api-reference/chat/create#top_p) and others.

### DIFF
--- a/candle-transformers/src/generation/mod.rs
+++ b/candle-transformers/src/generation/mod.rs
@@ -4,32 +4,77 @@ use rand::{distributions::Distribution, SeedableRng};
 pub struct LogitsProcessor {
     rng: rand::rngs::StdRng,
     temperature: Option<f64>,
+    top_p: Option<f64>,
 }
 
 impl LogitsProcessor {
-    pub fn new(seed: u64, temperature: Option<f64>) -> Self {
+    pub fn new(seed: u64, temperature: Option<f64>, top_p: Option<f64>) -> Self {
         Self {
             rng: rand::rngs::StdRng::seed_from_u64(seed),
             temperature,
+            top_p,
         }
+    }
+
+    fn sample_argmax(&mut self, logits: Tensor) -> Result<u32> {
+        let logits_v: Vec<f32> = logits.to_vec1()?;
+        let next_token = logits_v
+            .iter()
+            .enumerate()
+            .max_by(|(_, u), (_, v)| u.total_cmp(v))
+            .map(|(i, _)| i as u32)
+            .unwrap();
+        Ok(next_token)
+    }
+
+    fn sample_multi(&mut self, prs: &Vec<f32>) -> Result<u32> {
+        let distr = rand::distributions::WeightedIndex::new(prs).map_err(Error::wrap)?;
+        let next_token = distr.sample(&mut self.rng) as u32;
+        Ok(next_token)
+    }
+
+    fn sample_topp(&mut self, prs: &mut Vec<f32>, top_p: f32) -> Result<u32> {
+        // top-p sampling (or "nucleus sampling") samples from the smallest set of
+        // tokens that exceed probability top_p. This way we never sample tokens that
+        // have very low probabilities and are less likely to go "off the rails".
+        let mut argsort_indices = (0..prs.len()).collect::<Vec<_>>();
+        
+        // Sort by descending probability.
+        argsort_indices.sort_by(|&i, &j| prs[j].partial_cmp(&prs[i]).unwrap());
+        
+        // Clamp smaller probabilities to zero.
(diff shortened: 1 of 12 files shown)


LaurentMazare:
Looks great, thanks! Could you describe a bit the testing that you've done? Just want to be sure that it covers running all the modified CLIs as well as the different web UIs.
Also could you add an entry to the top level `CHANGELOG.md` file?

candle-transformers/src/generation/mod.rs
```diff
@@ -4,32 +4,77 @@ use rand::{distributions::Distribution, SeedableRng};
...
+        Ok(next_token)
+    }
+
+    fn sample_multi(&mut self, prs: &Vec<f32>) -> Result<u32> {
+        let distr = rand::distributions::WeightedIndex::new(prs).map_err(Error::wrap)?;
+        let next_token = distr.sample(&mut self.rng) as u32;
+        Ok(next_token)
+    }
+
+    fn sample_topp(&mut self, prs: &mut Vec<f32>, top_p: f32) -> Result<u32> {
```
LaurentMazare, line 36: What is the intended behavior if top_p is larger than the highest probability in the array? Weight everything equally? Is it what would happen or will we get some error because of an invalid distribution?

candle-transformers/src/generation/mod.rs
```diff
@@ -4,32 +4,77 @@ use rand::{distributions::Distribution, SeedableRng};
...
+
     pub fn sample(&mut self, logits: &Tensor) -> Result<u32> {
         let logits = logits.to_dtype(DType::F32)?;
         let temperature = self.temperature.unwrap_or(0.);
-        let next_token = if temperature > 0. {
-            let prs = candle_nn::ops::softmax(&(&logits / temperature)?, D::Minus1)?;
-            let prs: Vec<f32> = prs.to_vec1()?;
-            let distr = rand::distributions::WeightedIndex::new(prs).map_err(Error::wrap)?;
-            distr.sample(&mut self.rng) as u32
+        let top_p = self.top_p.unwrap_or(1.);
```
LaurentMazare, line 64: Is it the appropriate default? I would have though 0 would be the right thing?

candle-transformers/src/generation/mod.rs
jbochi: It will include all tokens necessary for the probability of the selected tokens to be larger than `top_p`.

For example, if `top_p` is 0.9 and probabilities are `[0.8, 0.1, 0.06, 0.04]` the first three will be selected.

As long as `top_p` is higher than 0, which we check before calling this function, at least one of the tokens will be selected, and we won't get an error.

candle-transformers/src/generation/mod.rs
jbochi: 1.0 also works because it means all tokens are selected.  That's also the default open ai uses: https://platform.openai.com/docs/api-reference/chat/create#top_p

jbochi:
> Looks great, thanks!

Thank you! And thank you for creating candle!

> Could you describe a bit the testing that you've done? Just want to be sure that it covers running all the modified CLIs as well as the different web UIs.

Here's `trunk serve`:
image

I also tested that even with high `temperature`, the output is close to greedy decoding (temperature=0) if `top_p` is low:

image

And here's `lib-example.html`:

image


> Also could you add an entry to the top level CHANGELOG.md file?

Done

candle-transformers/src/generation/mod.rs
LaurentMazare: Ah I misunderstood what it does, thought it would only select tokens with some minimum probability.

LaurentMazare:
Great thanks, will merge once the CI has run!

LaurentMazare:
rustfmt seems to fail, could you have a look please?

jbochi:
> rustfmt seems to fail, could you have a look please?

sorry about that! Should be fixed now.

I also added the new test file I forgot to commit before

LaurentMazare:
Thank for apply rustfmt, clippy seems to be having some failures now :)

jbochi:
Sorry again. Hopefully, it's all fixed now 🙏

LaurentMazare:
Thanks, merged!

PR #821, opened 2023-09-12
Add LoRA

EricLBuehler:
I implemented LoRA for `Linear`, `Embedding`, `Conv1d`, and `Conv2d`, each according to the official `peft` implementation (with the exception of `Conv1d` which was implemented according to a [Microsoft implementation](https://github.com/microsoft/LoRA) as there is no `peft` implementation).

# Details
I added a package, called `candle-lora` that implements LoRA for the above types. Merged weights are not yet implemented, although I plan on implementing them and submitting a PR which will be an addition to `candle-lora`.

To use `candle-lora`, the `Linear`, `Embedding`, `Conv1d`, or `Conv2d` types must be replaced by a `Box<dyn ...LayerLike>` type in the model struct(s). This is so that they can be swapped out with the new layers without requiring an error-prone model struct duplication.

Because Rust cannot check the types of members without macros, the developer must select layers to convert. Selected layers are passed to `candle-lora`, and new layers that have a concrete type (Lora...) are returned. To use them, they must be manually inserted into the model struct(s).

# Limitations and future plans
Because Rust can not allow dynamic attribute reassignment like Python, `candle-lora` cannot have as simple an interface as `peft` does. However, I plan on developing a library that leverages macros to make using `candle-lora` more ergonomic. This library will allow automatic selection and swapping of layers akin to `peft`.


LaurentMazare:
Thanks for all the work on this, this look very nice. Maybe we could keep this in an external repo rather than including it here? We would like to emphasize that it's easy to build on top of the candle core libraries and that it's not a monolithic design and `candle-lora` looks like a great example of this. Fwiw we should probably also move out of the repo things like `candle-datasets`, the wasm examples, the stable diffusion example if we start to make it more fully featured, the pyo3 bindings, etc.
Additional benefits of a separate repo:
- Faster compilation time for the main repo.
- It forces us into having a clean boundary between the different parts and not breaking the external facing apis.
The only downside I can see is discoverability of crates but for that I feel that we should add new sections in the main readme such as "Models using candle", "Libraries using candle".

EricLBuehler:
Ok, thank you. I agree - it would be great to have a non-monolithic canndle.

LaurentMazare:
Do you want to make a PR to mention it in the readme? (I can also do it if you prefer but would be great to mention it)

PR #829, opened 2023-09-12
Extract T5 module and add main function to use it

jbochi:
This moves the T5 module from the musicgen example to `candle-transformers/src/models/t5.rs`

I added a new `candle-examples/examples/t5/main.rs` that computes embeddings and tested it with `t5-small`, `t5-base`, and `t5-large`. They all work.

`google/flan-t5-base` doesn't work yet because they changed the activation function to a Gated Gelu.

I was planning to tackle this and also to add a decoder next, but perhaps it makes sense to stop here to not make the PR too large.

Please let me know what you think.

Thanks


LaurentMazare:
Sounds like a good idea to move this code to a more central location, thanks!

LaurentMazare:
Actually did you check the generated embeddings vs the python version or just that it produced some output? I've just added some support for sentence similarities in the example and the results are not that good so I'm a bit concerned the model might be buggy.

LaurentMazare:
Ah nevermind, this was likely caused by the way padding was handled, I removed it and now the similarities make a lot more sense. I also computed the embeddings for some samples in both python and candle and they lined up fairly well so all good!

jbochi:
Thanks for reviewing this and making sure the embeddings are correct!

PR #836, opened 2023-09-13
Add `CANDLE_NVCC_CCBIN` support for `candle-kernels`, and eliminate warnings.

crlf0710:
On windows, it seems currently `nvcc` will not auto detect `cl.exe` position, so this allows the compiler bindir be stored as an environment variables, for convenience.

This also eliminate a warning under conditional compilation.


LaurentMazare (approved):

LaurentMazare:
Merged, thanks!

PR #869, opened 2023-09-16
[Request for comment (RFC)] Allow printing out keys and values of VarBuilder

patrickvonplaten:
When translating models into PyTorch or other frameworks (JAX, TF), the crux is often to match the names and tensors of the checkpoint to the expected names and tensors of PyTorch's nn.Module

When playing with building models in candle, the `VarBuilder` is a very important object as it's expected by almost every candle-nn layer object.

It would be super helpful to be able to better debug the var builder.

A common use case (I guess) is the following. One loads a safetensors checkpoint and tries to build a candle model on top of it:

```rs
        let weights = unsafe { candle::safetensors::MmapedFile::new(filename)? };
        let weights = weights.deserialize()?;

        let vs_model = nn::VarBuilder::from_safetensors(vec![weights], dtype, device);
```

Having loading `vs_model`, I would now like to inspect the object for the structure it has **before** I pass it into a candle struct definition. In this PR I propose to add a `state_dict()` method to `VarBuilderArgs`:

```rs
println!("{:?}", vs_model.state_dict());
```

This would nicely display all the expected names along side it's tensor shape:
```
{"encoder.down.0.block.0.conv2.weight": Tensor[dims 128, 128, 3, 3; f32], "encoder.down.3.block.1.conv1.weight": Tensor[dims 512, 512, 3, 3; f32], "decoder.up.4.block.0.conv2.bias": Tensor[dims 768; f32], "encoder.down.2.block.1.conv2.weight": Tensor[dims 256, 256, 3, 3; 
f32], "decoder.up.2.block.2.conv2.bias": Tensor[dims 256; f32],  ...}
```
which would help a lot to know how one should implement the model in rust.

This PR is far from being complete, I'm mainly just seeing if there is interest for this function.

One way to implement this function is to the following:
- 1.) Make passing the shape argument to the backend's `get(...)` functions optional so that one get retrieve a tensor simply based on the name
- 2.) Add a `fn keys()` function to every backend which returns a list of all keys. Note that for now this is incomplete

cc @Narsil @LaurentMazare - curious to hear your thoughts :-)

### DIFF
--- a/candle-nn/src/var_builder.rs
+++ b/candle-nn/src/var_builder.rs
@@ -1,6 +1,8 @@
 //! A `VarBuilder` is used to retrieve variables used by a model. These variables can either come
 //! from a pre-trained checkpoint, e.g. using `VarBuilder::from_safetensors`, or initialized for
 //! training, e.g. using `VarBuilder::from_varmap`.
+//!
+#![feature(type_name_of_val)]
 use crate::VarMap;
 use candle::{safetensors::Load, DType, Device, Error, Result, Shape, Tensor};
 use safetensors::{slice::IndexOp, tensor::SafeTensors};
@@ -57,6 +59,8 @@ pub trait Backend: Send + Sync {
     ) -> Result<Tensor>;
 
     fn contains_tensor(&self, name: &str) -> bool;
+
+    fn keys(&self) -> Option<Vec<&str>>;
 }
 
 pub trait SimpleBackend: Send + Sync {
@@ -71,6 +75,8 @@ pub trait SimpleBackend: Send + Sync {
     ) -> Result<Tensor>;
 
     fn contains_tensor(&self, name: &str) -> bool;
+
+    fn keys(&self) -> Option<Vec<&str>>;
 }
 
 impl<'a> Backend for Box<dyn SimpleBackend + 'a> {
@@ -89,6 +95,10 @@ impl<'a> Backend for Box<dyn SimpleBackend + 'a> {
     fn contains_tensor(&self, name: &str) -> bool {
         self.as_ref().contains_tensor(name)
     }
+
+    fn keys(&self) -> Option<Vec<&str>> {
+        self.as_ref().keys()
+    }
 }
 
 impl<'a, B: Backend> VarBuilderArgs<'a, B> {
@@ -103,6 +113,12 @@ impl<'a, B: Backend> VarBuilderArgs<'a, B> {
             path: vec![],
             _phantom: std::marker::PhantomData,
         }
+    }
+
+    pub fn hey(&self) -> String {
+        let keys = self.data.backend.keys();
+        println!("Type: {:?}", keys);
+        "hey".to_string()
     }
 
     /// Returns the prefix of the `VarBuilder`.
@@ -199,6 +215,10 @@ impl SimpleBackend for Zeros {
     fn contains_tensor(&self, _name: &str) -> bool {
         true
     }
+
+    fn keys(&self) -> Option<Vec<&str>> {
+        None
+    }
 }
 
 impl SimpleBackend for HashMap<String, Tensor> {
@@ -233,6 +253,10 @@ impl SimpleBackend for HashMap<String, Tensor> {
     fn contains_tensor(&self, name: &str) -> bool {
         self.contains_key(name)
     }
+
+    fn keys(&self) -> Option<Vec<&str>> {
+        None
+    }
 }
 
 impl SimpleBackend for VarMap {
@@ -249,6 +273,10 @@ impl SimpleBackend for VarMap {
 
     fn contains_tensor(&self, name: &str) -> bool {
         self.data().lock().unwrap().contains_key(name)
+    }
+
+    fn keys(&self) -> Option<Vec<&str>> {
+        None
     }
 }
 
@@ -290,6 +318,10 @@ impl<'a> SimpleBackend for SafeTensorWithRouting<'a> {
     fn contains_tensor(&self, name: &str) -> bool {
         self.routing.contains_key(name)
     }
+
+    fn keys(&self) -> Option<Vec<&str>> {
+        Some(self.routing.keys().map(|s| &s[..]).collect())
+    }
 }
 
 impl SimpleBackend for candle::npy::NpzTensors {
@@ -322,6 +354,10 @@ impl SimpleBackend for candle::npy::NpzTensors {
 
     fn contains_tensor(&self, name: &str) -> bool {
         self.get(name).map_or(false, |v| v.is_some())
+    }
+
+    fn keys(&self) -> Option<Vec<&str>> {
+        None
     }
 }
 
@@ -506,5 +542,9 @@ impl<'a> Backend for ShardedSafeTensors<'a> {
 
     fn contains_tensor(&self, name: &str) -> bool {
         self.0.routing.contains_key(name)
+    }
+
+    fn keys(&self) -> Option<Vec<&str>> {
+        None
     }
 }
--- a/candle-nn/src/var_builder.rs
+++ b/candle-nn/src/var_builder.rs
@@ -49,7 +49,7 @@ pub trait Backend: Send + Sync {
     /// Retrieve a tensor with some target shape.
     fn get(
         &self,
-        s: Shape,
+        s: Option<Shape>,
         name: &str,
         h: Self::Hints,
         dtype: DType,
@@ -65,7 +65,7 @@ pub trait SimpleBackend: Send + Sync {
     /// Retrieve a tensor based on a target name and shape.
     fn get(
         &self,
-        s: Shape,
+        s: Option<Shape>,
         name: &str,
         h: crate::Init,
         dtype: DType,
@@ -81,7 +81,7 @@ impl<'a> Backend for Box<dyn SimpleBackend + 'a> {
     type Hints = crate::Init;
     fn get(
         &self,
-        s: Shape,
+        s: Option<Shape>,
         name: &str,
         h: Self::Hints,
         dtype: DType,
@@ -113,8 +113,18 @@ impl<'a, B: Backend> VarBuilderArgs<'a, B> {
         }
     }
 
-    pub fn keys(&self) -> Option<Vec<&str>> {
-        self.data.backend.keys()
+    pub fn state_dict(&self) -> HashMap<String, Tensor> {
+        let binding = self.data.backend.keys().unwrap();
+        let keys = binding.iter();
+        let state_dict: HashMap<_, _> = keys
+            .map(|&name| {
+                (
+                    name.to_string(),
+                    self.get_without_shape(&name, Default::default()).unwrap(),
+                )
+            })
+            .collect();
+        state_dict
     }
 
     /// Returns the prefix of the `VarBuilder`.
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Interesting, I certainly agree that matching the model tensor names with the names from the tensor files is a tricky thing. The way I get around this and that should be better documented is to run:
```
cargo run --example tensor-tools -- ls foo.safetensors
```
This will print the checkpoint tensors and their shape. I even have a prebuilt binary for this in my path.
Do you think this could help with your workflow?

I'm a bit reluctant about having a `model_state_dict` as if I understood properly this contains the content of the file rather than what the model asks for. A var builder may actually not have a notion of content e.g. when initializing weights for training.
If you wanted to inspect all the weights a model ask for you can use a `zero` var builder and see the tensor that it holds after the model initialization though somehow I never find myself needing this but it should work.

Sorry all this should be actually explained a bit more somewhere, maybe we should have a "porting a model from pytorch" document to give some tips on how to usually do this as it's a super common use case and with the proper approach it's actually quite easy and mechanical to do so.

Anyway certainly curious to know what you think about this.

patrickvonplaten:
> ```
> cargo run --example tensor-tools -- ls foo.safetensors
> ```
That tool is super nice! Yes that definitely helps!

> I'm a bit reluctant about having a model_state_dict as if I understood properly this contains the content of the file rather than what the model asks for. A var builder may actually not have a notion of content e.g. when initializing weights for training.

That makes sense! Yes I think the name `state_dict` is maybe badly chosen because it refers too much to a model. 
Maybe something like `tensors_dict()` or `tensors()` is a better name? Fully agree that this function should not have any notion/relation to the model.

> A var builder may actually not have a notion of content e.g. when initializing weights for training.

I see, so a var builder can both link/load tensors from a weight file for inference, but also have "no" content and instead just dynamically allocate new zero tensors (e.g. when initializing for training?).

So for the `Zeros` struct it would definitely make no sense to add a `tensors(...)` (or a better named one) function.

However, I still think it would be quite important for people to not have to rely on external tools to have an easier time looking into the parameters that are loaded into the different model components. For example, let's see I'm in the process of porting a model that looks as follows:

```rs
public struct AutoencoderKL {
    encoder: Encoder,
    decoder: Decoder,
}

impl AutoEncoderKL {
    pub fn new(
        vs: nn::VarBuilder,
        ...,
    ) -> Result<Self> {
        vs_encoder = vs.pp("encoder")
        vs_decoder = vs.pp("decoder")
        let encoder = Encoder::new(vs_encoder, in_channels, latent_channels, encoder_cfg)?;
        let decoder = Decoder::new(vs_decoder, in_channels, latent_channels, encoder_cfg)?;
        ...
    }
```

Here, I think it would be important that the user is able to effectively debug `vs_decoder`. For example let's say that:
- The encoder structure is already correct, but just the decoder is not correct. It would be nice to be able print out just the decoder var builder tensors directly in the code.
- Or all the weights shape match, but when running the forward pass the results start to differ for a very specific weight - in this case one would debug whether the problem might be that the values of this weight aren't correct and it might be nice to print out all the values of that specific weight directly in the code (before running the forward at building time)

But maybe I'm debugging also too much in PyTorch-style and this is not the most effective way here.


Some other related thoughts (heavily biased from someone that has worked heavily with PyTorch / JAX / TF in Python and very little with rust, so to be taken with a grain of salt :sweat_smile:)

- 1.) I struggle a bit with the notion of "variable builder" because:
   - For inference the variables are already built no? Coming a bit also from the JAX/Flax world where model parameters are passed around all the time, e.g. see `"params"` [here](https://flax.readthedocs.io/en/latest/guides/flax_basics.html#module-basics) something just like "variables" would be more intuitive for me
   - I understand that when randomly initializing however the "variable builder" is empty, but I guess it would overall still be more intuitive for me to have it called "variables" 
  
- 2.) 
> If you wanted to inspect all the weights a model ask for you can use a zero var builder and see the tensor that it holds after the model initialization

That's a great mechanism to see the diff between a safetensors file and what is expected! Think that's a good way for beginners to port models! 
Here how would one exactly look into the randomly initialized var builder afterwards though? Would you save the var builder as a file? Is there a save function for the var builder? I see a lot of `from_...` loading functions but no "save" functions. Here it'd also very convenient for me to have a 

- 3.) It's a bit unintuitive for me that one has to pass a `"shape"` to the `"get"` function of the different backend implementations. I see the "var builder" as a container that holds tensors with different names (just like the dict of a PyTorch file) and I just want to retrieve a certain tensor by name - why do I have to pass a "shape" here?
I understand that this is absolutely necessary when randomly initializing the model, but I don't think it's necessary when the var builder already holds tensors. In this case it seems like passing "shape" is mostly just used to verify that the expected model shape matches the tensor shape? 

- 4.) From playing with the code a bit, the Zeros backend feels a bit different than the other backends, e.g. if I understand correctly the `get(..)` method of the zeros backend allocates new tensors whereas all the other backends just return existing tensors. Wonder if it could make more sense to separate them more

- 5.) Happy to write a doc about how to port from PyTorch to Rust :-) Think it could also be a good way for me to learn more tricks!

Sorry for the long message :sweat_smile: Just writing down some thoughts in case they are helpful. 
I think it could make sense to think about potentially renaming the function to `tensors()` and thus allowing the user a more dynamic way to print tensors of the weights, but also totally fine for me to close the PR!

LaurentMazare:
> I see, so a var builder can both link/load tensors from a weight file for inference, but also have "no" content and instead just dynamically allocate new zero tensors (e.g. when initializing for training?).

Yes exactly. A var-builder is just a backend ([trait](https://github.com/huggingface/candle/blob/7cafca835a4bb9a21f3c8111e2f61b7a6b1270fd/candle-nn/src/var_builder.rs#L46)) that one can query for a tensor name, and a current path. It's totally up to the backend to fetch the tensor in the way it wants. The backend can even be implemented fully on the user side so that it's easy to hook up new file format, training initializers, backends that mix both loading fixed tensors from files and initializing some (e.g. for lora style stuff), using a hashmap from names to tensors etc.

> Here, I think it would be important that the user is able to effectively debug `vs_decoder`. For example let's say that:
> 
> * The encoder structure is already correct, but just the decoder is not correct. It would be nice to be able print out just the decoder var builder tensors directly in the code.

That's a good point. To me it's more up to the backend to provide something like this rather than the var builder but maybe we can have some helper function to tunnel things easily, and have an additional backend method to query for available tensors (we already have one to query for whether a tensor exist).

> * Or all the weights shape match, but when running the forward pass the results start to differ for a very specific weight - in this case one would debug whether the problem might be that the values of this weight aren't correct and it might be nice to print out all the values of that specific weight directly in the code (before running the forward at building time)

There is probably something I misunderstand here but why couldn't you print the tensor once you've obtained it from the var builder (so before the forward pass), just a `println!("{tensor}")` would hopefully work fine.

> 
> But maybe I'm debugging also too much in PyTorch-style and this is not the most effective way here.
> 
> Some other related thoughts (heavily biased from someone that has worked heavily with PyTorch / JAX / TF in Python and very little with rust, so to be taken with a grain of salt 😅)
> 
> * 1.) I struggle a bit with the notion of "variable builder" because:
>   
>   * For inference the variables are already built no? Coming a bit also from the JAX/Flax world where model parameters are passed around all the time, e.g. see `"params"` [here](https://flax.readthedocs.io/en/latest/guides/flax_basics.html#module-basics) something just like "variables" would be more intuitive for me
>   * I understand that when randomly initializing however the "variable builder" is empty, but I guess it would overall still be more intuitive for me to have it called "variables"

The naming is certainly poor, variables would be fine too me if folks think it's better, maybe something that also indicates that there is a notion of path but well, naming is hard...

> * 3.) It's a bit unintuitive for me that one has to pass a `"shape"` to the `"get"` function of the different backend implementations. I see the "var builder" as a container that holds tensors with different names (just like the dict of a PyTorch file) and I just want to retrieve a certain tensor by name - why do I have to pass a "shape" here?
>   I understand that this is absolutely necessary when randomly initializing the model, but I don't think it's necessary when the var builder already holds tensors. In this case it seems like passing "shape" is mostly just used to verify that the expected model shape matches the tensor shape?

Yes, and I think it's actually good to do so. The idea is to fail as quickly as possible, i.e. when loading the weights rather than with a shape mismatch in the forward pass. We could add a way to query for the shape based on a name (if it exists) but if it's possible to avoid I would prefer keeping the api as easy and uniform as possible. Certainly interested if you have some use cases in mind where that would be helpful, certainly interested.

> * 4.) From playing with the code a bit, the Zeros backend feels a bit different than the other backends, e.g. if I understand correctly the `get(..)` method of the zeros backend allocates new tensors whereas all the other backends just return existing tensors. Wonder if it could make more sense to separate them more

Actually the main other backend that is interesting is the VarMap one which is used for training - and this structure supports saving ([here](https://github.com/huggingface/candle/blob/7cafca835a4bb9a21f3c8111e2f61b7a6b1270fd/candle-nn/src/var_map.rs#L30)) for when one wants to checkpoint the model during training - it also supports loading a partially trained state. The idea there is to have a single api so that the exact same code is used for training/inference without the need of using a trait in all the `new` functions (the actual trait is sort of hidden from the user via a box-dyn).

> * 5.) Happy to write a doc about how to port from PyTorch to Rust :-) Think it could also be a good way for me to learn more tricks!

+1 to this. Also if you want to port some interesting examples to see what sticks well that would be great but maybe that's already what you're doing.

> 
> Sorry for the long message 😅 Just writing down some thoughts in case they are helpful. I think it could make sense to think about potentially renaming the function to `tensors()` and thus allowing the user a more dynamic way to print tensors of the weights, but also totally fine for me to close the PR!

Don't be, super interesting to see your thoughts on this. Overall, I would say:
- Having `tensors()` sounds good and helpful, would be interesting to know if it's only the tensors "under" the path or all of them. Maybe we have a boolean flag so that the user is explicit about which version they want.
- I would prefer keeping the shape mandatory for now, and maybe adding a way to query for it if helpful.

Also just to point out that this is certainly convoluted and part of it is because we currently accomodate for using mmaped safetensors data with no copy (though it's actually not fully implemented). An alternative design would be to have a var builder that actually doesn't contain the safetensor data and just stores the `(name, tensor)` that the users ask for (initializing them randomly even for inference), and then have a `varbuilder.load("foo.safetensors")` function to replace the weights with the stored ones.

jeromeku:
@patrickvonplaten 
Implemented simple backend for 'tracing' a Candle model as well as some python utilities for comparing the traced output with the expected weights as would be loaded when `Varbuilder::from_xxx` is called, where the 'expected' weight map is parsed from `hf hub` `safetensors` headers, `pytorch.index.json`, etc. 

See [here](https://github.com/huggingface/candle/issues/1070#issuecomment-1767917332) for additional details.  The unit tests in the [model_tracing](https://github.com/jeromeku/candle/blob/candle-transformers-model-testing-utils/candle-transformers/tests/utilities/model_tracing.rs) module demonstrates basic usage, and the [notebook](https://github.com/jeromeku/candle/blob/candle-transformers-model-testing-utils/candle-transformers/tests/python/model_conversion.ipynb) shows how to verify these output traces with their `HF` sources. 

Trying to create some utilities to ease porting and verification of `HF` --> `Candle` models.  Let me know if this is along the lines of your original request.

LaurentMazare:
Closing this as stale and the idioms for candle models have evolved a bit since then. Feel free to re-open or open a fresh PR/issue for discussion.

PR #870, opened 2023-09-16
Generate `*.pyi` stubs for PyO3 wrapper

LLukas22:
Well, let's try this again. 

This approach is heavily inspired by the way [tokenizers](https://github.com/huggingface/tokenizers/tree/main/bindings/python) solved this problem, by generating all stub files via an `stub.py` script. That way all typing and doc-strings are defined on the rust side and there is no need to keep the python files in sync. 

There are 3 ways to add typehints: 

1. Do nothing: That way, PyO3 will autogenerate the text-signature of your function. Meaning we at least have the signature but no type information in the generated `*.pyi` file.
2. Add a `signature`: That way, we know the function signature and which fields are optional. 
3. Add a `text_signature`: The preferred way of defining a function signature. Contains all parameter names and their types. Allows the generated `*.pyi` file to provide type-hinting support to the language server. 

Drawback: It is currently not possible to add a return type via a `text_signature` meaning all functions either return `None` or `Any`.

Additional Stuff: 
I added `text_signatures` to most of the static functions and sprinkled in some comments. I also moved all file loading and saving functions to the `utils` module as that felt like the right place to put them.


LaurentMazare (approved):

LaurentMazare:
Ah that's pretty amazing to have switched this to automated generation, it will make maintaining this a lot easier in the future, that's great!

PR #872, opened 2023-09-16
change to idiomatic code

chris-ha458:
actually 2~3 changes in one.
if this style does not read better I could introduce them separately


LaurentMazare:
Closing this as seems stale, I don't think such styling changes make much sense or is obviously better/more idiomatic.

PR #880, opened 2023-09-17
Add return types to `*.pyi` stubs

LLukas22:
This PR extends the automatic stub generation via `stub.py` by adding an optional way of adding a functions return type from the rust side. It is now possible to type-hint the return type of a function by adding it with the `&RETURNS&:` flag to the end of a doc-string. If a function has this flag in it's doc-string it will be parsed as the functions return type and removed from the doc-string. 
This behavior is completely optional and if no return type is provided the Pyo3 defaults of `None` or `Any` will be used.

I also added doc-strings and text-signatures to all `Tensor` and `QTensor` functions.

https://github.com/huggingface/candle/assets/65088241/56c1e662-07e3-4527-873b-69e7a355fe67

Additional Changes:
I also added a `save_gguf` function as I will mainly use the python wrapper to convert HF models to quantized GGUF models.


LaurentMazare (approved):

LaurentMazare:
Very cool, thanks!

PR #925, opened 2023-09-21
CUDA cast i64

grzuy:


LaurentMazare (approved):

LaurentMazare:
Thanks, looks like we totally overlooked these when adding i64 support.

Issue #637, opened 2023-08-28
Need train example


evgenyigumnov:
Could you add the following case to the examples?

Take the llama model (from llama example) as a basis and train it on a simple dataset from a file.
For example:
[
  {
  "input": "What is your name?", "output": "What is your name? My name is ROBOTRON"
  },
  {
  "input": "How old are you?", "output": "How old are you? I am immortal robot!"
  },
]

Then we run the trained model and submit file for the test as input and print the result of the test work on stdout.
For example (file)
[
  {
  "test": "Hello! What is your name?"
  },
  {
  "test": "Good morning! How old are you?"
  },
]
For example (stdout)
test 1:
Hello! What is your name? Hello my friend. I am ROBOTRON.
test 2:
Good morning! How old are you? Good morning. I am no age robot!

evgenyigumnov:
If is not so hard. You could explain me how to do this. And I add this example by myself to you repository.

evgenyigumnov:
As I seen in README.md

- Simple syntax, looks and feels like PyTorch.
    - Model training.

evgenyigumnov:
https://github.com/huggingface/candle/tree/main/candle-book/src/training
Thats all what I found :( 

I understand the importance of idiomatic learning examples on handwriting recognition. But somehow everyone wants to retrain language models. I beg you. Although you did not help me with servers with video cards. Give at least links to articles or contacts of people that would explain how to do this.

HoodieCollin:
Hey there, I've been looking for the same resources and sadly I haven't had much luck finding a “walk-through” on how to train a model with Candle (I assume it's too young of a project for there to be blogs and more fleshed-out tutorials).

If you are like me and new to ML in general but a seasoned programmer, I recommend the following process while we wait for improved docs.

- Find a tutorial that uses PyTorch or Tensorflow to build and train a simple model on the MNIST dataset.
- Follow that guide and experiment with it in a Google Collab notebook.
- Ask ChatGPT to clarify anything you don't understand (the MNIST dataset is classic so ChatGPT has a good understanding of the concepts involved).

Now that you have a good grasp of how that specific implementation works, go to the examples workspace in this repo. You should inspect the MNIST example and compare it to the Python version you just finished. [Here is a link to the example.](https://github.com/huggingface/candle/blob/main/candle-examples/examples/mnist-training/main.rs)

[...]

PR #926, opened 2023-09-22
Book train simlified example

evgenyigumnov:
Based on ticket https://github.com/huggingface/candle/issues/637

PS 

Unfortanly I cant check how will be look like "#include ../simplified.rs:book_training_simplified1" blocks in book :(


Narsil:
Thanks @evgenyigumnov Looks really good.

you should be able to check out the book using `mdbook serve` no ?

evgenyigumnov:
I was check by mdbook serve
And fix small mistake
At that moment all looks fine

PR #940, opened 2023-09-23
Add a general LLMDataset

EricLBuehler:
Hello everybody,

I propose adding an LLMDataset type for easy training of LLM models. It will take tokens of any `T: WithDType`, and provides several functions for applying mappings to the dataset, modifying it, and retrieving metadata.

## Rationale
Currently, Candle only provides a dataset for TinyStories. LLMDataset  is a more general dataset that does not provide functions to load from a file but allows for any type token. This can be used as a base dataset and then abstracted upon to create more datasets.

## Use
To use LLMDataset, it is converted into an LLMDatasetIter which (because it internally holds an immutable reference), it tied to the lifetime of the LLMDataset and prevents modification. This LLMDatasetIter allows for iteration through the dataset with batching and automatic token shifting.

The relavent issue is here #941

### DIFF
--- a/candle-datasets/src/nlp/llm.rs
+++ b/candle-datasets/src/nlp/llm.rs
@@ -70,15 +70,16 @@ impl<'a, T: WithDType> LLMDatasetIter<'a, T> {
     /// The input and target Tensors are automatically created from each token sequence by
     /// truncating the rightmost token to create the input, and by truncating
     /// the tokens to remove the leftmost token to create the target.
-    pub fn new(dataset: &'a LLMDataset<T>) -> Self {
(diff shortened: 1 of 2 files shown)


LaurentMazare:
Thanks for the PR. Seeing the questions in #944 I imagine that you're still working on training models based on the changes in this PR and on other changes. I think it would be more convenient and you would be able to iterate faster if you start by developing this in a separate dedicated repo. Once the api has solidified and you've been able to train some models using this we could look at what bits are general enough to be merged in the `candle-dataset` crate. Would this sound reasonable to you?

EricLBuehler:
Hi @LaurentMazare , I agree. Perhaps after it matures I can re-open this PR.

PR #945, opened 2023-09-23
[segment-anything] Support multi-point as the prompt input

GeauxEric:
Add the support for prompting more than one point; which was already supported by the original segment anything model.


LaurentMazare:
Thanks for the PR, looks pretty good!
Could you expand a bit on the testing that you did on this, ideally I think we would want:
- Test with multiple points with some images comparing the generated masks when using only one of the points vs more.
- Check that the web UI still works fine.
- Check that the command line tool still works fine.

radames:
very cool! quickly looking at the wasm module, it might work as before with a single point even thought now it would accept multiple points. If this pass other tests, I'd be happy to adapt the UI/UX to send more points later. Also it would be very cool to add the extra prompting bbox + points, labels

GeauxEric:
> Thanks for the PR, looks pretty good! Could you expand a bit on the testing that you did on this, ideally I think we would want:
> 
> * Test with multiple points with some images comparing the generated masks when using only one of the points vs more.
> * Check that the web UI still works fine.
> * Check that the command line tool still works fine.

I just updated the example `main.rs` Cli args to support multiple point prompt. Generated new segment results (one point v.s. two) and updated it in the readme. Pls take a look.

I wonder how integration tests are performed for this repo, does the CICD worfklow support download any model file and do any inference?

LaurentMazare:
The CI only covers testing the core and nn libraries, not the models at the moment, hence asking you to test it manually including the web UI (which would be hard to automate anyway).
Could you look at the clippy failures in the CI btw?

LaurentMazare:
I actually fixed the lint so as to unlock this!

PR #965, opened 2023-09-26
add eos token to phi example

radames:


LaurentMazare:
Thanks, I've made some small tweaks to apply rustfmt and fail if the token would not be there (which shouldn't happen but anyway - the previous version would use the first token in the encoding which might be odd).

PR #966, opened 2023-09-26
Phi 1.5 wasm module

radames:
Add Phi quantized model as Wasm module example
@LaurentMazare if you would like to test, I'll finish the PR tomorrow PST. It still needs a blurb description, but overall I think all the features are there. Maybe a better default prompt?

### DIFF
--- a/candle-wasm-examples/phi/phiWorker.js
+++ b/candle-wasm-examples/phi/phiWorker.js
@@ -70,7 +70,7 @@ async function generate(data) {
     );
 
     self.postMessage({ status: "loading", message: "Initializing model" });
-    model.init_with_prompt(
+    const firstToken = model.init_with_prompt(
       prompt,
       temp,
       top_p,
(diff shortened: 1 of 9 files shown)


candle-wasm-examples/phi/phiWorker.js
```diff
@@ -0,0 +1,131 @@
...
+    self.postMessage({ status: "loading", message: "Starting Phi" });
+    const model = await Phi.getInstance(
+      weightsURL,
+      modelID,
+      tokenizerURL,
+      quantized
+    );
+
+    self.postMessage({ status: "loading", message: "Initializing model" });
+    model.init_with_prompt(
```
LaurentMazare, line 73: I think this return the first generated token but somehow it gets dropped at the moment?

radames: great catch! I completely missed that it return the first token 🤦🏼 need to fix the worker! thanks

LaurentMazare (approved):

LaurentMazare:
Amazing work, thanks!

Issue #878, opened 2023-09-17
Why Gelu is zero for u8/u32/i64?


npuichigo:
https://github.com/huggingface/candle/blob/5f83c13f17a7b16955c9b649424aca276d5e930d/candle-core/src/op.rs#L573-L584

chris-ha458:
might have to be `todo!`

UnsafeOats:
How would you implement it in any reasonable way for integers?  Getting even close to the standard tanh-derived approximation for GeLU u8/u32/i64 would either involve rough approximation or converting to f32/f64 and converting back, which defeats the point.

It would probably be a "good enough" integer approximation for GeLU to just have a step function f(x) where f(x) = -1 if x < 0, f(x) = 0 if x == 0, f(x) = x if x > 0.  Outside of this, I can't think of reasonably close to GeLU for u8/u32/i64.

If this approach is amendable to the maintainers, I'd love to put a PR together for it.

chris-ha458:
Since gelu itself is not defined as an integer to integer operation (it is a Real to Real or Float to float).

This particular implementation is mapped into integers to integers, so a "correct" implementation is not possible.

That being said, many implementations still could be possible that rounds the output value.

The more important issue is that this is not well documented.
Even if somebody does not intentionally invokes this, (they shouldn't), it could be invoked inadvertantly by type inference or type coercion.

If there is a plan to support this but it hasn't been decided yet it should be `todo!`
if it is agreed that it should not be invoked, then either the traitbounds can be changed so that invoking it becomes a compile time error or a panic! should be there so it becomes a runtime error (the former is more preferred imo)

UnsafeOats:
My suggested approximation above was such a simple implementation that I just [implemented it](https://github.com/huggingface/candle/pull/970).

ivarflakstad:
Hey!
Going through old unaddressed issues/PRs and found this.
I agree that certain ops should not be implemented for certain data types, as there is no correct way to represent them.
I think that in this case I'd rather we get obviously incorrect results `0` than give credibility to what is essentially incorrect usage. And yes, `panic` is probably a step up from "silently failing" with `0`, but due to the overall design it hasn't caused any major problems that I am aware of.

In any case I'll close this as the original question has been answered long ago 👍

PR #970, opened 2023-09-27
GeLU implementation for u8/u32/i64

shaneish:
In response to [issue-878](https://github.com/huggingface/candle/issues/878), I've added an integer approximation for GeLU.


ivarflakstad:
Afaik most frameworks have some kind of runtime error in this situation. Perhaps we should do the same.
For now I'll keep the behaviour as-is.

PR #988, opened 2023-09-28
fix: fixes slice_scatter dim type

grzuy:
Hi,

Guessing this was supposed to be like this?

Thanks.


LaurentMazare (approved):

LaurentMazare:
Ah good catch, thanks for the fix!

PR #1020, opened 2023-10-02
[SAM] Add undo button and background point mode

radames:
Demo Updated: https://huggingface.co/spaces/radames/candle-segment-anything-wasm

### DIFF
--- a/candle-wasm-examples/segment-anything/lib-example.html
+++ b/candle-wasm-examples/segment-anything/lib-example.html
@@ -168,6 +168,7 @@
         if (!hasImage || isEmbedding || isSegmenting) {
           return;
         }
+        const backgroundMode = event.shiftKey ? bgPointMode^event.shiftKey : bgPointMode;
         const targetBox = event.target.getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Looks amazing!
Two suggestions though happy to defer to subsequent PR if you prefer:
- Could we use different colors for the background and "normal" points (both appear cyan to me)?
- Maybe when clicking close enough to an existing point, the point gets removed rather than a new point added?

radames:
thanks for the feedback! can send the changes on this PR. I like the idea of removing the point if it's close enough. I'll keep the same button to toggle modes and add the remove feature

radames:
very neat idea @LaurentMazare , you can test it now, the distance is about 3 pixels radius.

radames:
Before you merge, I'll add a shift click for desktop as well!

LaurentMazare:
Works very nicely, let me know when good to merge!

radames:
it looks good now @LaurentMazare thanks

LaurentMazare (approved):

PR #1028, opened 2023-10-04
Whisper quantized wasm

radames:
* add support for the quantized wasm
* remove the model.rs and pointed to transformers/whisper.rs
* tested trunk demo and move assets around 
* send o a PR here https://huggingface.co/spaces/lmz/candle-whisper

It will be nice to adapt the `convert_and_run` in the future to send segments as it decode its, something like `next_segment`


candle-wasm-examples/whisper/src/worker.rs
```diff
@@ -247,17 +300,44 @@ impl Decoder {
...
+        let task = match md.task {
+            Some(task_str) => {
+                let task_string = task_str.to_owned();
+                match task_string.as_str() {
+                    "translate" => Some(Task::Translate),
+                    _ => Some(Task::Transcribe),
+                }
+            }
+            _ => Some(Task::Transcribe),
+        };
```
radames, line 330: is there a better way to to this? I want to send a `string` or None,  also tried to use an Enum with wasm_bindgen

LaurentMazare:
Oh that's so nice that you removed this model duplication, thanks!

LaurentMazare:
Could you run rustfmt? Seems like it's picking some small diff in the CI.

candle-wasm-examples/whisper/src/worker.rs
LaurentMazare: Maybe you could have a look at [this example](https://github.com/huggingface/candle/blob/11d3687cc655f8f79d856342a5539a9274e96df4/candle-examples/examples/quantized/main.rs#L306-L311), with the same`as_deref` thing you can probably do it with a single statement.

radames:
thanks @LaurentMazare, much better with `as_deref`.
I did run `cargo fmt  --all -- --check` now does it fix the issue?

LaurentMazare (approved):

LaurentMazare:
Looks great, thanks!

PR #1035, opened 2023-10-05
feat: [SAM] able to download the result as png

lichin-lin:
@radames I've add a download button and the download function for the cutout image!

https://github.com/huggingface/candle/assets/10290616/5503f55e-623e-402a-9f51-482c9b148a27

### DIFF
--- a/candle-wasm-examples/segment-anything/lib-example.html
+++ b/candle-wasm-examples/segment-anything/lib-example.html
@@ -151,6 +151,7 @@
         if (target.nodeName === "IMG") {
           const href = target.src;
           clearImageCanvas();
+          copyImageURL = href;
           drawImageCanvas(href);
           setImageEmbeddings(href);
         }
@@ -178,6 +179,7 @@
             img.onload = () => {
               resolve(img);
             };
+            img.crossOrigin = "anonymous";
             img.src = imageURL;
           });
         };
@@ -546,7 +548,7 @@
             disabled
             class="p-1 px-2 text-xs font-medium bg-white rounded-2xl outline outline-gray-200 hover:outline-orange-200 disabled:opacity-50"
           >
-            Download the result (png file)
+            Download Cut-Out
           </button>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


candle-wasm-examples/segment-anything/lib-example.html
```diff
@@ -163,6 +169,45 @@
...
       });
+      // add event to download btn
+      downloadBtn.addEventListener("click", async () => {
+        // Function to load image blobs as Image elements asynchronously
+        const loadImageAsync = (imageURL) => {
+          return new Promise((resolve) => {
+            const img = new Image();
+            img.onload = () => {
+              resolve(img);
+            };
```
radames, line 180: can you please add this here? I was getting tainted canvas with the example imagas

```
img.crossOrigin = "anonymous";
```

radames (changes requested):
Very cool! You also need to add `copyImageURL = href;` so the examples can work

https://github.com/huggingface/candle/pull/1035/files#diff-81cb3980e2868df7ea08351b7bceda0764c3a8662e9711d857e45aa510689877R151-R156

radames:
Here is live @LaurentMazare if you want to check! https://huggingface.co/spaces/radames/candle-segment-anything-wasm

LaurentMazare:
That's super cool to play with and easily detour images! One cosmetic thing, rather than "download the result" maybe it could be "download the selected part" or something like that?

radames:
How about "Download Cut-Out" ? 
It's also interesting to play with the model size, you can see the artifacts of the smaller model compared to the larger one.

lichin-lin:
Thanks for the review, will update soon! 🙌
I have a question related to cutout result: why sometime I'll get holes for the cutout mask? (is this something to do with how model generates the embeddings?)


https://github.com/huggingface/candle/assets/10290616/f3856e25-34fb-4510-97ce-9fb82f8333b6

radames:
hi @lichin-lin, as I mentioned on my last message, this is defiantly due the model size MobileSAM the default on our list. if you try SAM base, you can see it'll perform better.

lichin-lin:
Got it, thx for the explanation!
I just pushed a commit to the PR. 🙌

LaurentMazare:
Looks all good to me, @radames any more thoughts or good to merge?

radames (approved):

PR #1037, opened 2023-10-05
fix: fix index_select cuda kernel for src target dim different than ids dim when selecting dim > 0

grzuy:
Hi,

I think this might be caused possibly after https://github.com/huggingface/candle/pull/1022/files#diff-2f1279372ccba2e7f8d23f07416dccac61ff6b838532bdf5ed06519ee1005d86R894 ?


LaurentMazare (approved):

LaurentMazare:
Yes that looks much better, thanks for the fix and for the additional test. Could you run rustfmt before I merge please?

grzuy:
Sure, done.
Thanks for the quick feedback!

LaurentMazare:
Merged, thanks again!

PR #1053, opened 2023-10-07
Add LLaMA as an architecture to the PyO3 wrapper

LLukas22:


LaurentMazare:
It would be nice if this PR could be split in a bunch of smaller independent ones so as to make the reviewing easier. E.g. one PR to expose some new functions, one PR to handle the -1 in the reshape, one to add the bit of testing etc.

LLukas22:
Yeah i agree, i basically started adding all indexing/broadcasting features i needed to basically be able to copy over most pytorch models, with the transformers llama implementation as a reference. 

I will split it into the following PRs:

-  Broadcasting `+`, `-`, `\`, `*` 
-  Adding and broadcasting equality operations e.g. `<`, `==` 
-  Torch-like `tensor`, `...` and `None` indexing
-  Adding torch-like `to` function to tensor
-  Better shape handling: accept tuples or *args and handle negative indices
-  Typehinting for magic methods, e.g. `__add__`

That way it should be way simpler to review them and decide if we want those feature in the wrapper or discard them. 
These PRs should also cover most of pytorchs tensor api, meaning after adding them the wrapper should be mostly pytorch compatible.   

Is that fine from your side?

LaurentMazare:
Sounds great, looking forward to these PRs!

LLukas22:
I'm closing this for now

PR #1066, opened 2023-10-09
feat: parse Cuda compute cap from env

OlivierDehaene:
Parse compute cap from env and fallback to nvidia-smi if not found instead of the other way around.
This is needed for docker builds.

### DIFF
--- a/candle-flash-attn/build.rs
+++ b/candle-flash-attn/build.rs
@@ -166,6 +166,18 @@ fn main() -> Result<()> {
     println!("cargo:rustc-link-lib=dylib=cudart");
     println!("cargo:rustc-link-lib=dylib=stdc++");
 
+    /* laurent: I tried using the cc cuda integration as below but this lead to ptaxs never
+       finishing to run for some reason. Calling nvcc manually worked fine.
+    cc::Build::new()
+        .cuda(true)
+        .include("cutlass/include")
+        .flag("--expt-relaxed-constexpr")
+        .flag("--default-stream")
+        .flag("per-thread")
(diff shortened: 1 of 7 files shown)


LaurentMazare:
If it's easy to do on your side, would you mind splitting this PR in the actual change vs the reformatting? I feel that this would make the diff a lot easier to read.

candle-flash-attn/build.rs
```diff
@@ -152,18 +166,6 @@ fn main() -> Result<()> {
     println!("cargo:rustc-link-lib=dylib=cudart");
     println!("cargo:rustc-link-lib=dylib=stdc++");
 
-    /* laurent: I tried using the cc cuda integration as below but this lead to ptaxs never
```
LaurentMazare, line 155: Could we keep this comment, it's probably useful to know that this was tried if someone wants to pick it up again.

candle-kernels/build.rs
```diff
@@ -141,83 +222,22 @@ mod cuda {
                 }
             }
             codes.sort();
-            if !codes.contains(&compute_cap) {
-                panic!("nvcc cannot target gpu arch {compute_cap}. Available nvcc targets are {codes:?}.");
-            }
-            *codes.last().unwrap()
+            let max_nvcc_code = *codes.last().unwrap();
```
LaurentMazare, line 225: Same as above, an explicit error might be better.

OlivierDehaene:
Thanks for the review! I pushed some modifications.

LaurentMazare:
Thanks!

PR #1100, opened 2023-10-15
PyO3: Add pytorch like `.to()` operator to `candle.Tensor`

LLukas22:
Allows casting or moving (or both at once) of a tensor via `.to()`. I get why we don't want this on the rust side of things but for the wrapper it's just way more convenient as it's heavily used in existing pytorch models.


candle-pyo3/src/lib.rs
```diff
@@ -772,6 +772,39 @@ impl PyTensor {
         Ok(PyTensor(self.0.copy().map_err(wrap_err)?))
     }
 
+    #[pyo3(signature = (*args, **kwargs), text_signature = "(self, *args, **kwargs)")]
+    /// Performs Tensor dtype and/or device conversion.
+    /// &RETURNS&: Tensor
+    fn to(&self, args: &PyTuple, kwargs: Option<&PyDict>) -> PyResult<Self> {
+        let mut t = self.0.clone();
```
LaurentMazare, line 779: I wouldn't see a need for this clone and having a mutable variable here but maybe I'm missing something?

candle-pyo3/src/lib.rs
```diff
@@ -772,6 +772,39 @@ impl PyTensor {
...
+        //handle args
+        if !args.is_empty() {
+            for arg in args.iter() {
```
LaurentMazare, line 782: I don't think we should process the arguments in order. If there are multiple device arguments for example the behavior would be really weird and I don't see a legitimate use case for this.

candle-pyo3/src/lib.rs
LLukas22: I probably made it mutable, because the tensor can either be modified by the kwargs, the args or not at all. But i will probably build some sort of parser that ensures, that only one dtype and device are provided.

candle-pyo3/src/lib.rs
LLukas22: Yup i will enforce that only one dtype or device can be provided via the args or kwargs and return an error if multiple were provided.

LLukas22:
Alright, i changed it to allow a `dtype` or `device` to only be provided once, otherwise there will be an `ValueError`. I also added the option to pass in another `other` tensor as an parameter, similar to pytorch. I also added tests for these cases. I'm not to happy with how the implementation turned out, but i currently don't know how to improve it.

LaurentMazare (approved):

LaurentMazare:
Thanks, agreed that the code doesn't look super nice with all this mutability, certainly feels error prone but let's start with this and clean this up later if it sticks.

PR #1102, opened 2023-10-15
PyO3: Add typehinting for magic/dunder methodes

LLukas22:
This PR adds the missing typehints for magic methodes e.g. `__add__`. Meaning with this PR the candle  should now be completely typehinted. 
 

https://github.com/huggingface/candle/assets/65088241/68409e07-89a6-40e2-9f95-ec109a96989d

# How it works

Since `pyo3` doesn't allow us to set the text signature or even change the doc-string of a magic method the typing information have to be stored externally. For this reason the `candle-pyo3/_additional_typing/__init__.py`  module was created to enable the definition of additional functions for candle classes. The `stub.py` script will import this module and extract all classes from it and combine the function signatures of classes with the same name. 

## Drawbacks
This means the magic/dunder signatures have to be defined manually in an additional file, if they should be typehinted. Realistically this shouldn't be a problem as it's very unlikely, that the signature of these methods will ever change, but i still would like your opinion on this approach.

### DIFF
--- a/candle-pyo3/_additional_typing/README.md
+++ b/candle-pyo3/_additional_typing/README.md
@@ -1,0 +1,3 @@
+This python module contains external typehinting for certain `candle` classes. This is only necessary for `magic` methodes e.g. `__add__` as their text signature cant be set via pyo3.
+
+The classes in this module will be parsed by the `stub.py` script and interleafed with the signatures of the actual pyo3 `candle.candle` module.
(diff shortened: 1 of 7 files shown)


LaurentMazare:
Sounds good to start with, agreed that it's not very flexible but we can just see how often this ends up being an issue.

LaurentMazare (approved):

PR #1143, opened 2023-10-21
PyO3: Better shape handling

LLukas22:
This PR adds the ability to provide shapes as `*args` meaning that `rand((10,10))` can now also be written as `rand(10,10)`. I also added a new `PyRelativeShape` struct which handles negative dims, e.g. `t.reshape(-1,1)`. I decided to add this new struct to explicitly make some functions accept "relative" shapes (e.g. `reshape`) and force other functions to only accept normal `PyShapes` (e.g. `rand`).

I also needed to upgrade `pyo3` from `0.19.0` to `0.20.0` as there was a bug with `*args` and named parameters which got fixed in `0.20.0` e.g. `rand(*shape, device=None)` wouldn't let me specify a device even if the python code was valid.


candle-pyo3/py_src/candle/__init__.pyi
```diff
@@ -67,7 +67,7 @@ class u8(DType):
     pass
 
 @staticmethod
-def zeros(shape: Sequence[int], dtype: Optional[DType] = None, device: Optional[Device] = None) -> Tensor:
+def zeros(shape: Shape, dtype: Optional[DType] = None, device: Optional[Device] = None) -> Tensor:
```
LaurentMazare, line 70: Why isn't this one a `*shape` too?

candle-pyo3/src/shape.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl From<PyShape> for ::candle::Shape {
+    fn from(val: PyShape) -> Self {
+        val.0.into()
+    }
+}
+
+#[derive(Clone, Debug)]
+/// Represents a relative shape e.g. (1, -1, 3)
+pub struct PyRelativeShape(Vec<isize>);
```
LaurentMazare, line 36: In the main candle crate, we use the name `ShapeWithHole` which is I think a bit more informative than `RelativeShape`. Could we do the same here?

LLukas22: Yup, i should probably also validate that the shape contains only one "hole".

candle-pyo3/py_src/candle/__init__.pyi
LLukas22: Yup, i provided the wrong typehint there.

LLukas22:
Alright, i renamed it to `PyShapeWithHole` and added the validation logic.

LaurentMazare:
Merged, thanks!

PR #1157, opened 2023-10-23
Add Binary Cross Entropy With Logit Loss to nn crate

ToluClassics:


candle-nn/src/loss.rs
```diff
@@ -48,3 +48,27 @@ pub fn cross_entropy(inp: &Tensor, target: &Tensor) -> Result<Tensor> {
...
+///
+/// * [inp]: The input tensor of dimensions `N, C` where `N` is the batch size and `C` the number
+///          of categories. This is expected to raw logits.
+/// * [target]: The ground truth labels as a tensor of u32 of dimension `N, C` where `N` is the batch size and `C` the number
+///          of categories.
+///
+/// The resulting tensor is a scalar containing the average value over the batch.
+pub fn binary_cross_entropy_with_logit(inp: &Tensor, target: &Tensor) -> Result<Tensor> {
+    let inp = crate::ops::sigmoid(inp)?;
+    let one_tensor = Tensor::new(1.0, &inp.device())?;
```
LaurentMazare, line 65: This would only work for `f64` tensors and not for `f16`, `bf16`, `f32` and would allocate a tensor object. Instead you could use `target.affine(1., -1.)` etc which would avoid these drawbacks.

candle-nn/tests/loss.rs
```diff
@@ -39,3 +39,38 @@ fn nll_and_cross_entropy() -> Result<()> {
     assert_eq!(to_vec0_round(&loss, 4)?, 1.1312);
     Ok(())
 }
+
```
LaurentMazare, line 42: Could you put in as a comment the equivalent PyTorch snippet that can be used to check the test final output?

LaurentMazare:
Oh also there seems to be some test failures in the CI that you want to check.

ToluClassics:
> Oh also there seems to be some test failures in the CI that you want to check.

Yeah I notices, just fixed them in the new commit

LaurentMazare:
Just to mention that the cuda CIs are expected to fail, no need to look into these!

PR #1160, opened 2023-10-23
derivative for GELU

KGrewal1:
Implementation of backprop derivative for GELU using the analytical derivative of the GELU approximation 
 $$0.5v\left(1+\tanh\left(\sqrt{\frac{2}{\pi}} v(1+0.044715v^{2})\right)\right)$$
 as 
$$0.5 + (0.398942 v + 0.0535161 v^3) (1 - \tanh^{2}(0.797885 v + 0.0356774 v^{3})) + 0.5 \tanh(0.797885 v + 0.0356774 v^{3})$$

### DIFF
--- a/candle-core/src/backprop.rs
+++ b/candle-core/src/backprop.rs
@@ -473,10 +473,10 @@ impl Tensor {
                     }
                     Op::Unary(arg, UnaryOp::Gelu) => {
                         let sum_grad = grads.or_insert(arg)?;
-                        let tanh = (0.0356774 * arg.powf(3.)? + (0.797885 * arg)?)?.tanh()?;
+                        let cube = arg.powf(3.)?;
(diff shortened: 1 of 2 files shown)


LaurentMazare:
Looks nice, could you also add a test to [unary_grad](https://github.com/huggingface/candle/blob/86e1803191e2ed44c57ad01807b29a886c0263bb/candle-core/tests/grad_tests.rs#L81) and ensure that it matches the PyTorch output?
Also probably better to compute `arg.powf(3.)` only once rather than twice.

KGrewal1:
Tests added (using same input tensor as other tests) and changed to cube tensor only once

LaurentMazare (approved):

LaurentMazare:
Great, thanks for the PR!

PR #1161, opened 2023-10-23
Add step indexing

### LINKED ISSUE #1133: How do I do step indexing on a tensor?
What's the `candle` equivalent of indexing into a tensor with `t[::2]`?

laptou:
Adds a new public type called `StepRange` which implements `Into<TensorIndexer>`, so it can be used to index tensors with a step size other than one. Supports negative step sizes, which index starting from the end instead of starting from the beginning.

Resolves #1133.

### DIFF
--- a/candle-core/src/indexer.rs
+++ b/candle-core/src/indexer.rs
@@ -46,6 +46,40 @@ impl Tensor {
                     current_dim += 1;
                     out
                 }
+                TensorIndexer::Step(left_bound, right_bound, step) => {
+                    let start = match left_bound {
+                        Bound::Included(n) => *n,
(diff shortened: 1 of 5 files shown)


laptou:
What are your thoughts on adding an impl `Into<TensorIndexer> for (R, isize) where R: RangeBounds`? It would shorten something like this:
```rust
let input_feature_0 = input_feature.i((.., StepRange::new(.., 2), StepRange::new(.., 2), ..));
```

Into this:
```rust
let input_feature_0 = input_feature.i((.., (.., 2), (.., 2), ..));
```

But that might be too obscure.

PR #1166, opened 2023-10-24
[Wasm] Add puffin phi model to wasm

radames:
Made the config load from a file

Demo
https://huggingface.co/spaces/radames/Candle-Phi-1.5-Wasm


LaurentMazare:
Thanks, that's great! I've merged the change on the HF model repo so maybe you can switch the branch to main.
Probably not in this change but it would be great to have some sample prompts to show how to use this model for code completion, chat etc (or even better but more involved, have some form of chat like UI that would generate the appropriate prompts).

radames:
great points @LaurentMazare, I'll add the examples to the UI. One note form the original configs I had to change `gelu-new` to `gelu`, following your implementation

radames:
@LaurentMazare I've added prompt examples following phi paper.

LaurentMazare:
Looks neat, thanks a lot!

PR #1169, opened 2023-10-24
Add step indexing

laptou:
Same as #1161, but GitHub wouldn't let me change the branch from `main` so I made a new PR.

### DIFF
--- a/candle-core/src/indexer.rs
+++ b/candle-core/src/indexer.rs
@@ -46,6 +46,40 @@ impl Tensor {
                     current_dim += 1;
                     out
                 }
+                TensorIndexer::Step(left_bound, right_bound, step) => {
+                    let start = match left_bound {
+                        Bound::Included(n) => *n,
+                        Bound::Excluded(n) => *n + 1,
+                        Bound::Unbounded => 0,
+                    } as i64;
+                    let stop = match right_bound {
+                        Bound::Included(n) => *n + 1,
+                        Bound::Excluded(n) => *n,
+                        Bound::Unbounded => dims[i],
+                    } as i64;
+
+                    let step = *step as i64;
+                    let device = x.device();
+                    let indices = match step {
+                        1.. => Tensor::arange_step(start as i64, stop as i64, step, device)?,
+                        ..=-1 => {
+                            let mut data = vec![];
+                            let mut current = stop - 1;
+                            while current >= start {
+                                data.push(current);
+                                current += step;
+                            }
+                            let len = data.len();
+
+                            Tensor::from_vec_impl(data, len, device, false)?
+                        }
+                        0 => panic!("step size cannot be zero"),
+                    };
+
+                    let out = x.index_select(&indices, current_dim)?;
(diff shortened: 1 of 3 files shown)


candle-core/src/indexer.rs
```diff
@@ -46,6 +46,40 @@ impl Tensor {
                     current_dim += 1;
                     out
                 }
+                TensorIndexer::Step(left_bound, right_bound, step) => {
+                    let start = match left_bound {
+                        Bound::Included(n) => *n,
+                        Bound::Excluded(n) => *n + 1,
```
LaurentMazare, line 52: I don't think this is correct if the step is negative, maybe I'm missing something?

candle-core/src/indexer.rs
```diff
@@ -46,6 +46,40 @@ impl Tensor {
...
+                        Bound::Unbounded => 0,
+                    } as i64;
+                    let stop = match right_bound {
+                        Bound::Included(n) => *n + 1,
```
LaurentMazare, line 56: Same as above.

candle-core/src/indexer.rs
```diff
@@ -46,6 +46,40 @@ impl Tensor {
...
+                        Bound::Excluded(n) => *n,
+                        Bound::Unbounded => dims[i],
+                    } as i64;
+
+                    let step = *step as i64;
+                    let device = x.device();
+                    let indices = match step {
+                        1.. => Tensor::arange_step(start as i64, stop as i64, step, device)?,
+                        ..=-1 => {
```
LaurentMazare, line 65: Hopefully this is not necessary anymore as `arange_step` now supports negative steps.

candle-core/src/indexer.rs
```diff
@@ -46,6 +46,40 @@ impl Tensor {
...
+                            let mut current = stop - 1;
+                            while current >= start {
+                                data.push(current);
+                                current += step;
+                            }
+                            let len = data.len();
+
+                            Tensor::from_vec_impl(data, len, device, false)?
+                        }
+                        0 => panic!("step size cannot be zero"),
```
LaurentMazare, line 76: Let's use `candle::bail!` rather than `panic!` here so that it's easier to catch the error.

candle-core/src/indexer.rs
```diff
@@ -61,18 +95,43 @@ impl Tensor {
     }
 }
 
+pub struct StepRange(Bound<usize>, Bound<usize>, isize);
+
+impl StepRange {
+    pub fn new<T: RangeBounds<usize>>(range: T, step: isize) -> Self {
+        if step == 0 {
+            panic!("step size cannot be zero");
```
LaurentMazare, line 103: Same as above, `candle::bail!`.

LaurentMazare (commented):
Sorry for the very long time to come back to this, seems generally good. A few comments inline.

LaurentMazare:
Closing this as now update since the review a couple months back.

PR #1171, opened 2023-10-24
Candle Optim: separate crate for optimisers

KGrewal1:
This creates a new crate called candle-optim.

The main optimizer trait still lives in candle-nn as moving it out would currently be a breaking API change.

This currently adds 3 new optimizers:

* MomentumEnhancedSGD able to implement all functionality of pytorch SGD although currently does not validate the parameters in the same manner
* AdaDelta
* AdaGrad

These have all been checked against their implementation in pytorch 2.1.0+cu118.

The main advantage of this being in the huggingface/candle repo as opposed to my own is visibility, to increase the likelihood of people able to implement other algorithms for optimisation and people able to speed up the current algorithms (such as implementing and equivalent of for_each and fused methods in candle)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "candle-examples",
     "candle-book",
     "candle-nn",
+    "candle-optim",
     "candle-pyo3",
     "candle-transformers",
     "candle-wasm-examples/llama2-c",
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 [![Documentation](https://docs.rs/candle-core/badge.svg)](https://docs.rs/candle-core)
 ![License](https://img.shields.io/crates/l/candle-core.svg)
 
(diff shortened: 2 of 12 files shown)


KGrewal1:
As an alternative, if it is preferred to have optimisers be completely separate, see https://github.com/KGrewal1/optimisers/tree/master

LaurentMazare:
I think separate repo would be great to start with and favor clean APIs and not tying things too much together. Also better for iteration on the optimizer side and when it comes to discoverability we should add it mentioned in the readme.
We can check how it evolved in a couple weeks and revisit if needed.

KGrewal1:
Great sounds like a plan, though I'm likely to have to slow down a bit due to uni work now I've been given stuff to do (also that indeterminism on mac CI is really irritating).

LaurentMazare:
Let's maybe close this PR now that the [separate crate](https://github.com/KGrewal1/optimisers) is available @KGrewal1 ?

Issue #1087, opened 2023-10-13
Convert pytorch tensors to/from candle tensors in the python api


LaurentMazare:
It would be neat to support converting pytorch tensors to/from candle tensors when using the pytorch api, e.g. do something like this:
```python
import torch
import candle

t = torch.randn((10, 8))
t2 = candle.Tensor(t)
print(t2)

t3 = t.to_torch() # returns a torch tensor.
```
Ideally there should be no dependency on `torch` when running `candle.Tensor([1, 2, 3])` after the change.

Tagging @LLukas22 in case you have some thoughts about this or wanted to take a stab at it.

LLukas22:
Just adding that the same would be great for `numpy` arrays. Implementing this without adding a dependency should be possible but we probably have to copy the tensors data 🤔

jeromeku:
`Arrow` (same in-memory format as used by huggingface `datasets`) should be able to handle zero-copy array interoperability.  

See [arrow < -- > torch](https://issues.apache.org/jira/browse/ARROW-1701), [arrow < -- > numpy](https://arrow.apache.org/docs/python/numpy.html), and [arrow Tensor](https://arrow.apache.org/docs/python/generated/pyarrow.Tensor.html).

LaurentMazare:
I don't think arrow would help sharing the data on gpu, but anyway I feel that we should just go ahead with making a copy to start with - this would already be better that the current way of doing the conversion.

jeromeku:
Arrow has [CUDA](https://arrow.apache.org/docs/python/api/cuda.html) support, not sure how mature this is.  Also, the docs on the [C Array Interface](https://arrow.apache.org/docs/format/CDeviceDataInterface.html) discusses how NVIDIA's `rapids` data science library -- GPU-accelerated dataframes and other scientific compute tools -- utilizes `Arrow` under the hood.

macroexpansion:
Hi, I would like to help. As far as I know, I have to use `libtorch`/`tch-rs` in order to create a `torch` tensor in Rust, copy data from `candle.Tensor` to it, wrap it in `PyObject` and return it to Python. But that would add dependency on `libtorch`/`tch-rs`?
Would you mind helping me get started on this?

LLukas22:
[...]

PR #1172, opened 2023-10-24
convert pytorch's tensor in Python API

macroexpansion:
regarding #1087, this PR implements the ability to convert `torch` tensor from/to `candle` tensor as follows:

- `candle` to `torch`: extract `candle` tensor value and use `PyO3` interop to create `torch` tensor.
- `torch` to `candle`: extract `numpy` value from the input `torch` tensor using `PyO3` interop and create `candle` tensor with it.

### DIFF
--- a/candle-pyo3/test.py
+++ b/candle-pyo3/test.py
@@ -1,4 +1,5 @@
 import candle
+import torch
 
 print(f"mkl:         {candle.utils.has_mkl()}")
 print(f"accelerate:  {candle.utils.has_accelerate()}")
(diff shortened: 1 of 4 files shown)


candle-pyo3/test.py
```diff
@@ -1,4 +1,5 @@
 import candle
+import torch
```
LaurentMazare, line 2: Could we make a second test file specifically for the torch interaction? This would avoid requiring torch to be installed for running the main test.

LaurentMazare (commented):
Looks nice and a very small change, thanks!

candle-pyo3/test.py
macroexpansion: sure

LaurentMazare (approved):

LaurentMazare:
Merged, thanks again for the PR!

Issue #1070, opened 2023-10-10
Support Meta Nougat model


Christof23:
Hi, it would be really fantastic if there was example code to host Nougat inference on candle! Is there any chance of this being supported? The documentation doesn't seem too far off what is provided for say running a BERT model inference session https://huggingface.co/docs/transformers/main/model_doc/nougat.

LaurentMazare:
Thanks for suggesting this, sounds like a great model to have indeed.
A slightly tricky thing is that nougat seems to rely on a couple other models: sbart, donut (for the swin part), some vision-encoder-decoder so it may take a bit of time to port all these.

LaurentMazare:
Tagging with help-wanted in case anyone wants to take a stab at it. Might be good to start with donut as it's reused in part by nougat. There are some PyTorch implementations in the transformers library, e.g. [here for donut](https://github.com/huggingface/transformers/blob/fc63914399b6f60512c720959f9182b02ae4a45c/src/transformers/models/donut/modeling_donut_swin.py#L335), these are likely good starting points as the candle api is close to PyTorch.

jeromeku:
@LaurentMazare Happy to take a stab at this (starting with `Donut`).  Are there any docs (or canonical examples) for porting `transformers` models to `candle`?

LaurentMazare:
We don't have docs for this I'm afraid. If you want to start with an even simpler model than donut and document the process through some tutorial or blog post, that would be very useful to others I guess :) I'll create a separate github issue for this.
The PyTorch and candle apis are pretty similar so that it's usually not that hard to convert models by transcribing the Python code into Rust. Here is a sample class [Python implementation](https://github.com/huggingface/transformers/blob/e1cec43415e72c9853288d4e9325b734d36dd617/src/transformers/models/t5/modeling_t5.py#L280) and the [Rust equivalent](https://github.com/huggingface/candle/blob/e7560443e4680b7655d011948d3cf178268fcfff/candle-transformers/src/models/t5.rs#L161-L182) - note that the tracing bits are not really necessary.

LaurentMazare:
Created #1075 for adding some documentation around the model conversion process, I also added a bunch of details about how I usually proceed.

Christof23:
@jeromeku it'd be great to team up on this! let me know if you're making a fork.

jeromeku:
[...]

PR #1174, opened 2023-10-25
Implemented meshgrid

wdoppenberg:
Hi,

While working on implementing the Swin model as part of the [Nougat issue](https://github.com/huggingface/candle/issues/1070) I quickly ran into missing the `meshgrid` tensor operator. Hence this PR.

Let me know what you think!

### DIFF
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -575,7 +575,6 @@ impl Tensor {
     /// # Errors
     ///
     /// * Will return `Err` if `args` contains less than 2 tensors.
-    /// * Will return `Err` if an unrecognized indexing mode is specified.
     ///
     pub fn meshgrid<A: AsRef<Tensor>>(args: &[A], xy_indexing: bool) -> Result<Vec<Self>> {
         if args.is_empty() || args.len() <= 1 {
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -545,8 +545,10 @@ impl Tensor {
     /// # Arguments
     ///
     /// * `args` - A slice of 1D tensors.
-    /// * `indexing` - An optional indexing mode. Either 'xy' or 'ij'. If not specified, 'xy' is
-    ///  used.
+    /// * `xy_indexing` - Whether to use xy indexing or ij indexing. If xy is selected, the
+    /// first dimension corresponds to the cardinality of the second input and the second
+    /// dimension corresponds to the cardinality of the first input. If ij is selected, the
+    /// dimensions are in the same order as the cardinality of the inputs.
     ///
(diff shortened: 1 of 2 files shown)


candle-core/src/tensor.rs
```diff
@@ -540,6 +540,64 @@ impl Tensor {
...
+    ///
+    pub fn meshgrid<A: AsRef<Tensor>>(args: &[A], indexing: Option<&str>) -> Result<Vec<Self>> {
+        if args.is_empty() || args.len() == 1 {
+            Err(Error::OpRequiresAtLeastOneTensor { op: "meshgrid" }.bt())?
+        }
+
+        let mut grids = Vec::with_capacity(args.len());
+        for idx in 0..args.len() {
+            // Repeat the tensor across the given dimensions to infiltrate it in the resulting grid
+            let repeats: Vec<_> = (0..args.len()).map(|i| if i == idx {1} else {args[i].as_ref().shape().dims()[0]}).collect();
```
LaurentMazare, line 582: You may want to run rustfmt to get this properly formatted, also this looks like a `args.iter().enumerate().map(|(i, arg)|...` this would likely be better to just iterating on the index. And to extract the first dimension, one can do `tensor.dim(0)?`.

candle-core/src/tensor.rs
```diff
@@ -540,6 +540,64 @@ impl Tensor {
...
+            let repeated_tensor = args[idx].as_ref().clone().repeat(repeats.clone())?;
+
+            // Reshape the tensor to match the dimensions of the grid
+            let reshaped_tensor = repeated_tensor.reshape(repeats)?;
+
+            grids.push(reshaped_tensor);
+        }
+
+        match indexing {
+            Some("xy") | None => Ok(grids),
```
LaurentMazare, line 592: Using a string to specify the indexing doesn't seem very rust idiomatic. As there are only two possibilities, I would suggest using a boolean argument, e.g. `xy_indexing: bool` (and if there were more than two args I would suggest an enum).
Also no need for the option in this case, as it's as simple for the user to have `true` in the code rather than `None` and more explicit (no need to look at the docs to understand what is happening).

LaurentMazare (commented):
Looks pretty good thanks, a few small comments attached.

LaurentMazare:
Merged, thanks!

wdoppenberg:
> Actually I think the documentation tests were not run and that the implementation was a bit buggy. Sorry I should have reviewed this more thoroughly, I've just pushed #1175 that should hopefully fix this. Please ensure that the tests are being run, typically tweaking them to fail before running a `cargo run` is a good way to do this.

I see, I'll pay better attention next time. Thanks for the fixes!

PR #1179, opened 2023-10-25
Make model_id configurable for Llama2-c example 

EricLBuehler:
As stated above, this PR resolves a TODO comment by adding another command-line parameter to the eval command.

### DIFF
--- a/candle-examples/examples/llama2-c/main.rs
+++ b/candle-examples/examples/llama2-c/main.rs
@@ -70,6 +70,9 @@ struct EvaluationCmd {
     /// https://huggingface.co/karpathy/tinyllamas/tree/main
     #[arg(long, default_value = "stories15M.bin")]
     which_model: String,
+
+    #[arg(long, default_value = "TinyStories-valid.txt")]
+    dataset_name: String,
 }
 
 #[derive(Parser, Debug, Clone)]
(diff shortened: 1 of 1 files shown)


candle-examples/examples/llama2-c/main.rs
```diff
@@ -178,10 +181,9 @@ fn run_eval(args: &EvaluationCmd, common_args: &Args) -> Result<()> {
     let tokens = match &args.pretokenized_dir {
         None => {
             let api = hf_hub::api::sync::Api::new()?;
-            let model_id = "roneneldan/TinyStories"; // TODO: Make this configurable.
```
EricLBuehler, line 181: Here was the TODO comment.

EricLBuehler:
@LaurentMazare, what do you think of this PR? Do you think it is necessary?

LaurentMazare:
> @LaurentMazare, what do you think of this PR? Do you think it is necessary?

I don't think this adds much and I would even think that it may break the existing code if I'm not mistaking. The dataset was coming from `roneneldan/TinyStories` and after that it seems to come from `karpathy/tinyllamas` which I wouldn't think has the data. Maybe I'm missing something here, not sure how much testing was done. Anyway i think it would be great if you want to transform this example into a full replication of Karpathy's awesome work on tinyllama but this likely belongs to an external repo. Happy to link to it if you put something into place that works well.

EricLBuehler:
Ok, sounds good. I suppose the default value for `model_id` should have been `roneneldan/TinyStories`. Regardless, I will look into making an example of `tinyllama`. Thanks!

PR #1207, opened 2023-10-28
Add DDPG and fix Gym wrapper

dashdeckers:


candle-examples/examples/reinforcement-learning/ddpg.rs
```diff
@@ -0,0 +1,461 @@
...
+        &self,
+        batch_size: usize,
+    ) -> Option<(Tensor, Tensor, Tensor, Tensor, Vec<bool>, Vec<bool>)> {
+        if self.size < batch_size {
+            None
+        } else {
+            let transitions: Vec<Transition> = thread_rng()
+                .sample_iter(Uniform::from(0..self.size))
+                .take(batch_size)
+                .map(|i| self.buffer.get(i).unwrap().clone())
```
LaurentMazare, line 109: Is there really a need for a clone here? Seems like having a reference should be enough as we would makee clones when creating the states etc vectors but maybe I'm missing something?

candle-examples/examples/reinforcement-learning/ddpg.rs
```diff
@@ -0,0 +1,461 @@
...
+                .collect();
+
+            let states: Vec<Tensor> = transitions
+                .iter()
+                .map(|t| t.state.clone().unsqueeze(0).unwrap())
```
LaurentMazare, line 114: Avoid the `unwrap` and return a proper error on failures.

candle-examples/examples/reinforcement-learning/ddpg.rs
```diff
@@ -0,0 +1,461 @@
...
+            "actor",
+            &self.dims,
+            tau,
+        )
+    }
+}
+
+struct Critic<'a> {
+    varmap: VarMap,
+    vb: VarBuilder<'a>,
```
LaurentMazare, line 255: Why is there a lifetime annotation here? I thought we had removed it from `VarBuilder` a long time ago.

candle-examples/examples/reinforcement-learning/ddpg.rs
```diff
@@ -0,0 +1,461 @@
...
+            network,
+            target_network,
+            size_state,
+            size_action,
+            dims,
+        })
+    }
+
+    fn forward(&self, state: &Tensor, action: &Tensor) -> Result<Tensor> {
+        let xs = Tensor::cat(&[action.clone(), state.clone()], 1)?;
```
LaurentMazare, line 317: The clone here is very likely redundant and useless.

candle-examples/examples/reinforcement-learning/ddpg.rs
```diff
@@ -0,0 +1,461 @@
...
+        self.network.forward(&xs)
+    }
+
+    fn target_forward(&self, state: &Tensor, action: &Tensor) -> Result<Tensor> {
+        let xs = Tensor::cat(&[action.clone(), state.clone()], 1)?;
```
LaurentMazare, line 322: Same as above.

candle-examples/examples/reinforcement-learning/ddpg.rs
```diff
@@ -0,0 +1,461 @@
...
+                terminateds,
+                truncateds,
+            ))
+        }
+    }
+}
+
+fn track(
+    mut varmap: VarMap,
+    vb: VarBuilder,
```
LaurentMazare, line 145: Passing this as a value rather than a ref makes it necessary to have a potentially large clone, could we just use references instead?

candle-examples/examples/reinforcement-learning/ddpg.rs
dashdeckers: The compiler is telling me I need them:

```
error[E0106]: missing lifetime specifier
   --> candle-examples/examples/reinforcement-learning/ddpg.rs:255:9
    |
255 |     vb: VarBuilder,
    |         ^^^^^^^^^^ expected named lifetime parameter
    |
help: consider introducing a named lifetime parameter
    |
253 ~ struct Critic<'a> {
254 |     varmap: VarMap,
255 ~     vb: VarBuilder<'a>,
    |
```

Maybe because I'm keeping the `VarBuilder` in the struct?

candle-examples/examples/reinforcement-learning/ddpg.rs
```diff
@@ -99,50 +99,50 @@ impl ReplayBuffer {
...
-            let transitions: Vec<Transition> = thread_rng()
+            let transitions: Vec<&Transition> = thread_rng()
                 .sample_iter(Uniform::from(0..self.size))
                 .take(batch_size)
-                .map(|i| self.buffer.get(i).unwrap().clone())
+                .map(|i| self.buffer.get(i).unwrap())
                 .collect();
 
-            let states: Vec<Tensor> = transitions
+            let states: Result<Vec<Tensor>> = transitions
```
LaurentMazare, line 112: No need for all the `Result` annotations like this, you can probably just use the `?` after `collect`.

dashdeckers: I tried that actually and I'm having trouble getting it to compile without these `Result` annotations..
First I'm told I need to specify `::<Vec<_>>` in each `collect()`, and then when I do that (which already doesn't reduce the amount of syntax) I get the following error:

```
error[E0277]: the `?` operator can only be applied to values that implement `Try`
   --> candle-examples/examples/reinforcement-learning/ddpg.rs:112:39
    |
112 |               let states: Vec<Tensor> = transitions
    |  _______________________________________^
113 | |                 .iter()
114 | |                 .map(|t| t.state.clone().unsqueeze(0))
115 | |                 .collect::<Vec<_>>()?;
    | |_____________________________________^ the `?` operator cannot be applied to type `Vec<Result<candle_core::Tensor, candle_core::Error>>`
    |
    = help: the trait `Try` is not implemented for `Vec<Result<candle_core::Tensor, candle_core::Error>>`
```

LaurentMazare: You need to replace `collect::<Vec...>()?` with `collect::<Result...>()?` I think.

dashdeckers: Got it!

candle-examples/examples/reinforcement-learning/ddpg.rs
LaurentMazare: That's odd, you're using the latest github tip? Seems a bit like a relicate from a very old version.

dashdeckers: I forked a fresh version of candle after my messed up pull request and as you said transplanted my changes into it, so I guess so? The latest commit that is not mine is #1205.

LaurentMazare: Ah ok, indeed it seems that I haven't cleaned the `VarBuilder` code as much as what I thought. I'll do that soonish but that seems good.

LaurentMazare:
Could you run rustfmt? Seems to break in the CI.

dashdeckers:
Done! I have to leave my laptop now, hopeful everything is good, otherwise tomorrow :)

LaurentMazare:
Merged after pushing some small cleanups, thanks!

PR #1209, opened 2023-10-28
implement VGG16 with example

drbh:
This PR adds the VGG16 model to `candle-transformers` and a small example to the `candle-example` folder.

Example output:
vgg16test

Please let me know if I should make any changes! Thanks for this awesome repo!!

### DIFF
--- a/candle-examples/examples/vit/main.rs
+++ b/candle-examples/examples/vit/main.rs
@@ -1,0 +1,59 @@
+#[cfg(feature = "mkl")]
+extern crate intel_mkl_src;
+
+#[cfg(feature = "accelerate")]
+extern crate accelerate_src;
+
+use clap::Parser;
+
+use candle::{DType, IndexOp, D};
+use candle_nn::VarBuilder;
+use candle_transformers::models::vit;
+
+#[derive(Parser)]
+struct Args {
+    #[arg(long)]
+    model: Option<String>,
+
+    #[arg(long)]
+    image: String,
+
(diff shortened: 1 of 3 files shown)


candle-examples/examples/vgg16/README.md
```diff
@@ -0,0 +1,11 @@
+
+
+```bash
+python3 export_models.py
```
LaurentMazare, line 4: We would rather get all the examples to be self-contained through the use of the `hf-hub` crate, see the other examples. In this case, you can find some sample model weights [here](https://huggingface.co/timm/vgg16.tv_in1k/tree/main).

candle-examples/examples/vgg16/assets/meerkat.jpg
LaurentMazare, line 1: Let's avoid adding new images to the repo if possible, it's already on the large side. Maybe you can just use one of the existing image for the example readme.

LaurentMazare (commented):
Thanks for the PR. Rather than just adding vgg16, it would be nicer to add vgg13/16/19 in one go (and so rename the files from e.g. `vgg16.rs` to `vgg.rs`).

drbh:
> Thanks for the PR. Rather than just adding vgg16, it would be nicer to add vgg13/16/19 in one go (and so rename the files from e.g. `vgg16.rs` to `vgg.rs`).

thanks for the quick feedback! 🙏 I agree it makes sense to cover those models as well; will push changes shortly

PR #1211, opened 2023-10-28
feat: implement VGG13, VGG16 and VGG19

drbh:
This PR is a follow up to https://github.com/huggingface/candle/pull/1209 and implemented VGG13/16/19 and relies on `hf-hub` for the models

vgg131619


LaurentMazare:
Thanks for the PR, I'm pushed a couple tweaks on top of it. Most of them are cosmetic and to make the code a bit more rust idiomatic and avoid intermediary `clone()` that are not required. A change that is a bit more important is that the weights were reloaded each time the final layer was applied as this happened in the closure, this should now be fixed.

PR #1215, opened 2023-10-29
PyO3: Add CI to build & upload wheels as artifacts.

LLukas22:
Default `maturin` CI to build wheels for all platforms/target-architectures. I also changed the `abi3` to `py38`, which means that the created wheels are compatible with python versions >= 3.8. 

Currently this CI only uploads  the wheels as an artifact, but in the future we probably want to auto deploy to `pypi` on a release tag, or maybe create some sort of development/nightly package releases.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1222, opened 2023-10-30
candle-transformers: minor differences between t5 and quantized_t5

michaelfeil:
Background: I am thinking about implementing quantized_bert. To better understand the changes necessary, I went through t5 vs quantized_t5 - and noticed a smaller difference.

Not sure if this is just a nit, or if there is some potential bug that can be caused from this.

`quantized_t5.rs` uses:
```rust
vb.pp(ff_i)
```
but `t5.rs` uses
```rust
vb.pp(&ff_i.to_string()
```


LaurentMazare:
These have exactly the same behavior, no reason to prefer one over the other (as long as clippy is happy with it).

PR #1244, opened 2023-11-02
Add hard-sigmoid and hard-swish activations

jamjamjon:
I am building a model which use hardsigmoid() activation, basically equals: relu6(x+3) / 6
More details at: [https://pytorch.org/docs/stable/generated/torch.nn.functional.hardsigmoid.html#torch.nn.functional.hardsigmoid](url)

And I found Candle hasn't provide it yet. So This PR is for that.

### DIFF
--- a/candle-nn/src/activation.rs
+++ b/candle-nn/src/activation.rs
@@ -13,7 +13,9 @@ pub enum Activation {
     Relu6,
     Silu,
     Sigmoid,
+    HardSigmoid,
     Swish,
+    HardSwish,
     Elu(f64),
     LeakyRelu(f64),
 }
@@ -29,7 +31,9 @@ impl super::Module for Activation {
             Self::Relu6 => xs.clamp(0f32, 6f32),
             Self::Silu => crate::ops::silu(xs),
             Self::Sigmoid => crate::ops::sigmoid(xs),
+            Self::HardSigmoid => crate::ops::hard_sigmoid(xs),
             Self::Swish => xs * crate::ops::sigmoid(xs)?,
+            Self::HardSwish => xs * crate::ops::hard_sigmoid(xs)?,
             &Self::Elu(alpha) => xs.elu(alpha),
             &Self::LeakyRelu(negative_slope) => crate::ops::leaky_relu(xs, negative_slope),
         }
--- a/candle-nn/src/ops.rs
+++ b/candle-nn/src/ops.rs
@@ -1,5 +1,6 @@
 use candle::{CpuStorage, Layout, Result, Shape, Tensor};
 use rayon::prelude::*;
+use std::ops::Div;
 
 /// Applies the softmax function to the input tensor, rescaling the element so that elements on
 /// a slice of fixed index on dimension `dim` are between 0 and 1 and sum to 1.
@@ -46,8 +47,7 @@ pub fn sigmoid(xs: &Tensor) -> Result<Tensor> {
 
 pub fn hard_sigmoid(xs: &Tensor) -> Result<Tensor> {
     // TODO: Should we have a specialized op for this?
-    let six = Tensor::new(6f32, xs.device())?;
-    ((xs + 3.0) / six)?.clamp(0f32, 1f32)
+    ((xs + 3.0)?.div(6.0))?.clamp(0f32, 1f32)
 }
 
 pub fn leaky_relu(xs: &Tensor, negative_slope: f64) -> Result<Tensor> {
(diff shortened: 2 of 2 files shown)


candle-nn/src/ops.rs
```diff
@@ -44,6 +44,12 @@ pub fn sigmoid(xs: &Tensor) -> Result<Tensor> {
     (xs.neg()?.exp()? + 1.0)?.recip()
 }
 
+pub fn hard_sigmoid(xs: &Tensor) -> Result<Tensor> {
+    // TODO: Should we have a specialized op for this?
+    let six = Tensor::new(6f32, xs.device())?;
+    ((xs + 3.0) / six)?.clamp(0f32, 1f32)
```
LaurentMazare, line 50: I think this is hard-swish and not hard-sigmoid in the pytorch nomenclature.

candle-nn/src/activation.rs
```diff
@@ -29,7 +31,9 @@ impl super::Module for Activation {
             Self::Relu6 => xs.clamp(0f32, 6f32),
             Self::Silu => crate::ops::silu(xs),
             Self::Sigmoid => crate::ops::sigmoid(xs),
+            Self::HardSigmoid => crate::ops::hard_sigmoid(xs),
             Self::Swish => xs * crate::ops::sigmoid(xs)?,
+            Self::HardSwish => xs * crate::ops::hard_sigmoid(xs)?,
```
LaurentMazare, line 36: Why does hard-swish point at hard-sigmoid? My understanding is that these should be two different functions.

jamjamjon: hard-switch = x * hard-sigmoid 
Details at: [https://pytorch.org/docs/stable/generated/torch.nn.functional.hardswish.html#torch.nn.functional.hardswish](url)

candle-nn/src/ops.rs
jamjamjon: I add two activation, hard-sigmoid and hard-swish.
hard-sigmoid = relu6(x+3) / 6
hard-swish = x * hard-sigmoid

LaurentMazare: Oops, my bad, this is hard-sigmoid, can you replace the `six` with an actual `6`?

candle-nn/src/activation.rs
LaurentMazare: Ah right, thanks for clarifying.

candle-nn/src/ops.rs
jamjamjon: There will be an error:
error[E0277]: cannot divide `std::result::Result<candle_core::Tensor, candle_core::Error>` by `{float}`
  --> candle-nn/src/ops.rs:50:17
   |
50 |     ((xs + 3.0) / 6.0)?.clamp(0f32, 1f32)
   |                 ^ no implementation for `std::result::Result<candle_core::Tensor, candle_core::Error> / {float}`
   |
   = help: the trait `std::ops::Div<{float}>` is not implemented for `std::result::Result<candle_core::Tensor, candle_core::Error>`
   = help: the following other types implement trait `std::ops::Div<Rhs>`:
             <std::result::Result<B, candle_core::Error> as std::ops::Div<candle_core::Tensor>>
             <std::result::Result<B, candle_core::Error> as std::ops::Div<&candle_core::Tensor>>

For more information about this error, try `rustc --explain E0277`.

LaurentMazare: Try `((xs + 3.0)? / 6)?` instead?

jamjamjon: `((xs + 3.0)? / 6)?.clamp(0f32, 1f32)` still got an error:     
error[E0277]: the trait bound `{integer}: std::borrow::Borrow<candle_core::Tensor>` is not satisfied
  --> candle-nn/src/ops.rs:53:18
   |
53 |     ((xs + 3.0)? / 6)?.clamp(0f32, 1f32)
   |                  ^ the trait `std::borrow::Borrow<candle_core::Tensor>` is not implemented for `{integer}`
   |
   = note: required for `candle_core::Tensor` to implement `std::ops::Div<{integer}>`

For more information about this error, try `rustc --explain E0277`.
warning: `candle-nn` (lib) generated 1 warning
error: could not compile `candle-nn` (lib) due to previous error; 1 warning emitted

But, This will be Ok

LaurentMazare: Yes, should be 6.0 instead of 6 indeed. Please use `div` instead of `/` and should be all good.

jamjamjon: Ok, fixed it.

LaurentMazare:
Merged, thanks for the PR and sorry for the confusion on the function names.

jamjamjon:
My bad! I made a typo!  I need to have a rest indeed. Thanks your patience.

PR #1258, opened 2023-11-03
feat: impl backprop for erf and gelu-erf

drbh:
This PR adds backprop support for erf anf gelu-erf.

ERF is based on the erf derivation supplied here: https://en.wikipedia.org/wiki/Error_function and here: https://www.wolframalpha.com/input?i=erf

GELU-ERF is based on the derivation of `0.5x*(1+erf(x/sqrt(2)))` or (`erf(v / 2f64.sqrt()) + 1.) * 0.5 * v`) which can be found here https://www.wolframalpha.com/input?i=0.5x*%281%2Berf%28x%2Fsqrt%282%29%29%29

### DIFF
--- a/candle-core/src/backprop.rs
+++ b/candle-core/src/backprop.rs
@@ -534,14 +534,14 @@ impl Tensor {
                     }
                     Op::Unary(arg, UnaryOp::Erf) => {
                         let sum_grad = grads.or_insert(arg)?;
-                        // d/dx = erf(x) = 2/sqrt(pi) * e^(-x^2)
+                        // d/dx erf(x) = 2/sqrt(pi) * e^(-x^2)
                         let erf_grad =
                             (2. / std::f64::consts::PI.sqrt()) * (arg.sqr()?.neg()?).exp()?;
                         *sum_grad = sum_grad.add(&(&grad * erf_grad)?)?
                     }
                     Op::Unary(arg, UnaryOp::GeluErf) => {
                         let sum_grad = grads.or_insert(arg)?;
-                        // d/dx = 0.5 + 0.398942 e^(-x^2/2) x + 0.5 erf(x/sqrt(2))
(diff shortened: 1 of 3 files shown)


LaurentMazare:
Thanks for the PR, could you add some tests ideally at the bottom of this [unary_test](https://github.com/huggingface/candle/blob/ad63f20781f9f5489b4f0aa9a3f24603304f2de0/candle-core/tests/grad_tests.rs#L81) and check that the generated outputs match the ones produced by PyTorch?

drbh:
@LaurentMazare tests added! Please let me know if I should make any changes. By the way these are the comparable Pytorch to the tests (not sure if/where these would be added in candle).

```python
# ERF
import torch
import torch.nn.functional as F
# Create a tensor and enable gradient computation
tensor = torch.tensor([3.0, 1.0, 4.0, 0.15], requires_grad=True)
# Apply ERF operation
y = tensor.erf()
# Print the output tensor
print(y)
# Assume a dummy scalar loss (e.g., sum of all elements in y)
loss = y.sum()
# Compute gradients
loss.backward()
# Print the gradients of the input tensor
print(tensor.grad)

# GELU-ERF
import torch
import torch.nn.functional as F
# Create a tensor and enable gradient computation
tensor = torch.tensor([3.0, 1.0, 4.0, 0.15], requires_grad=True)
# Apply GELU activation function
y = F.gelu(tensor, approximate='none')
# Print the output tensor
print(y)
# Assume a dummy scalar loss (e.g., sum of all elements in y)
loss = y.sum()
# Compute gradients
loss.backward()
# Print the gradients of the input tensor
print(tensor.grad)
```

LaurentMazare:
Would be great if you can put the pytorch code as comment in the test.

LaurentMazare:
Looks great, thanks for the PR!

PR #1260, opened 2023-11-03
Add some preliminary ONNX support

LaurentMazare:
- Add the ONNX proto3 file.
- Use `prost` to generate the rust types.
- Add a function to read an ONNX model, use it in an example.
- Simple evaluation function using directly the model proto.


LaurentMazare:
Looks great! Thanks for the PR!

PR #1261, opened 2023-11-03
add distil-whisper link

radames:


LaurentMazare:
Thanks, really nice that these are so easy to add!

LaurentMazare (approved):

PR #1269, opened 2023-11-04
feat: add backprop for elu

drbh:
adds backprop for elu and test to compare with pytorch output


LaurentMazare:
Great, thanks for the PR! (made some small cosmetic change to avoid computing the zeros twice + use the same style as for the rest). Will merge once the CI is good.

Issue #1263, opened 2023-11-03
Missing support for some penalty metrics, beam search, and top-k in `LogitsProcessor`.


EricLBuehler:
Hello everybody,

I have been working on adding sampling support to `candle-vllm`, but have noticed there are some features that could be added to `LogitsProcessor`. Although techniques like `top-p`, `multinomial`, and `argmax` are implemented, I this that the following features would be great to support.

Namely:
1) Penalty metrics (`presence_penalty`, `frequency_penalty`, and `repetition_penalty`)
    - These are important hyperparameters when sampling logits.
2) Beam search
    - Beam search is a popular sampling method.
3) `top-k` sampling
    - This is another popular sampling method.

What do you think about adding these sampling features?

EricLBuehler:
See candle-sampling, for future reference.

PR #1271, opened 2023-11-05
Add top-k support to LogitsProcessor

EricLBuehler:
Hello everybody,

During my recent development of `candle-vllm`, I have been working on sampling techniques and noticed that top-k, beam search and some other features are missing from `LogitsProcessor` (#1263).

In this PR, I have added the `top-k` technique (which selects the `k` most likely tokens and samples among those, ensuring that lesser likely tokens are not considered) to `LogitsProcessor`. Additionally, I restructured the `LogitsProcessor` API slightly to use enums in order to reduce the chance for subtle bugs.

### DIFF
--- a/candle-examples/examples/bigcode/main.rs
+++ b/candle-examples/examples/bigcode/main.rs
@@ -36,7 +36,7 @@ impl TextGeneration {
     ) -> Self {
         let top_p = match top_p {
             Some(p) => SamplingMethod::TopP(p),
-            None => SamplingMethod::Argmax,
+            None => SamplingMethod::Multinomial,
         };
         let logits_processor = LogitsProcessor::new(seed, temp, top_p);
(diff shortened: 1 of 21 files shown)


LaurentMazare:
I would suggest you have this enhanced `LogitsProcessor` as part of your `candle-vllm` crate for now - the immediate goal for candle-transformers is not to support all sampling method but instead it would be better if this can be deffered to other crates that can focus on such specific aspects - an alternative would be to make a crate that is only for logits processing if it could be useful without requiring all the dependencies of your crate.

EricLBuehler:
Ok, with that in mind, I think I will make a separate crate for only sampling. Do you think that eventually `candle-transformers`' goal will be to support all sampling methods - perhaps I can reopen an equivalent PR then?

PR #1275, opened 2023-11-06
Error compiling candle-onnx, added info about missing protoc binaries

figgefigge:
Candle-onnx dependency prost-build no longer comes bundled with prost binaries causing error when compiling. Added info to readme.md.


LaurentMazare (approved):

LaurentMazare:
Thanks for the PR, I'll actually deactivate protoc bits by default so that it's harder to run into this and move these bits in a specific ONNX readme.

PR #1283, opened 2023-11-06
Adds check for 7b-zephyr and uses correct template

DaveTJones:
At present both mistral and zephyr models use the same prompt template, which seems to cause unusual behaviour when querying zephyr. I've added a check for use of the zephyr model and added the correct prompt template as specified at https://huggingface.co/TheBloke/zephyr-7B-alpha-GGUF for use in the zephyr case.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -16,7 +16,6 @@ jobs:
         rust: [stable]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


LaurentMazare:
Thanks for the PR, I've tweaked it a bit so as to keep zephyr being handled as a "mistral model" as it's actually a fine-tuned mistral (hence the way it was actually handled before this PR). Otherwise looks generally good, I'll merge once the CI passes.

PR #1296, opened 2023-11-08
fix: negative axis

YangNianYi:
i think should be add with negative axis


LaurentMazare:
Thanks for spotting this, actually we've recently introduced a `normalize_axis` function to avoid such possible issues by factoring this code out, I've tweaked your PR to actually use this function.

PR #1300, opened 2023-11-08
Add support to UL2 model family

jbochi:
This is yet another variant of the T5 encoder/decoder models.

It works, but it's terribly slow on MacOS. I interrupted the following command after at least 15 minutes, and it had generated just the first 4 words.

```
$ RUST_BACKTRACE=1 cargo run --example t5 --release --features accelerate -- --model-id "google/flan-ul2" --revision "refs/pr/25"  --prompt "A step by step recipe to make bolognese pasta:" --decode
    Finished release [optimized] target(s) in 1.06s
     Running `target/release/examples/t5 --model-id google/flan-ul2 --revision refs/pr/25 --prompt 'A step by step recipe to make bolognese pasta:' --decode`
Running on CPU, to run on GPU, build this example with `--features cuda`
 - In a large skillet
```

### DIFF
--- a/candle-nn/src/activation.rs
+++ b/candle-nn/src/activation.rs
@@ -12,6 +12,8 @@ pub enum Activation {
     Relu2,
     Relu6,
     Silu,
+    #[serde(rename = "gated-silu")]
+    GatedSilu,
     Sigmoid,
     HardSigmoid,
     Swiglu,
@@ -31,6 +33,7 @@ impl super::Module for Activation {
             Self::Relu2 => xs.relu()?.sqr(),
             Self::Relu6 => xs.clamp(0f32, 6f32),
             Self::Silu => crate::ops::silu(xs),
+            Self::GatedSilu => crate::ops::silu(xs),
             Self::Sigmoid => crate::ops::sigmoid(xs),
             Self::HardSigmoid => crate::ops::hard_sigmoid(xs),
             Self::Swiglu => crate::ops::swiglu(xs),
--- a/candle-nn/src/activation.rs
+++ b/candle-nn/src/activation.rs
@@ -6,14 +6,11 @@ use serde::Deserialize;
 pub enum Activation {
     #[default]
     Gelu,
-    #[serde(rename = "gated-gelu")]
-    NewGelu,
+    ApproximateGelu,
     Relu,
     Relu2,
     Relu6,
     Silu,
-    #[serde(rename = "gated-silu")]
-    GatedSilu,
     Sigmoid,
     HardSigmoid,
     Swiglu,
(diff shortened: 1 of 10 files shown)


candle-nn/src/activation.rs
```diff
@@ -31,6 +33,7 @@ impl super::Module for Activation {
             Self::Relu2 => xs.relu()?.sqr(),
             Self::Relu6 => xs.clamp(0f32, 6f32),
             Self::Silu => crate::ops::silu(xs),
+            Self::GatedSilu => crate::ops::silu(xs),
```
LaurentMazare, line 36: It feels a bit error-prone to add a new activation named `GatedSilu` whereas the behavior is the same as `Silu` and the gating has to be enforced in another way. I think we should instead change the type used in the t5 model to a separate type that handles the `gated-variant`, ideally something like:
```rust
struct ActivationWithOptionalGating {
    gated: bool,
    activation: candle_nn::Activation
}
```
And we use a custom serde decoder that recognizes `gated-silu` and `gated-relu` and set `gated = true` for them and for the other values defer to the serde decoder of `candle-nn::Activation`.

jbochi: I think that's a great idea. I gave it a shot, but I'm new to `serde`.

Please let me know if there's a better way to implement it.

candle-nn/src/activation.rs
```diff
@@ -6,8 +6,7 @@ use serde::Deserialize;
 pub enum Activation {
     #[default]
     Gelu,
-    #[serde(rename = "gated-gelu")]
-    NewGelu,
+    ApproximateGelu,
```
LaurentMazare, line 9: Please do not change the name, it's a bad name but in sync with pytorch and changing it breaks the api compatibility.

jbochi: Got it. I reverted it. Thanks

candle-transformers/src/models/quantized_t5.rs
```diff
@@ -39,6 +39,34 @@ fn masked_fill(on_false: &Tensor, mask: &Tensor, on_true: f32) -> Result<Tensor>
     Ok(m)
 }
 
+#[derive(Debug, Deserialize, Default, Clone, PartialEq)]
```
LaurentMazare, line 42: Rather than duplicating this, use the version from the `t5` module.

LaurentMazare:
Also please run `rustfmt` and `clippy`, both seem to fail in the CI.

candle-transformers/src/models/quantized_t5.rs
jbochi: 👍 Done

jbochi:
> Also please run `rustfmt` and `clippy`, both seem to fail in the CI.

Sorry about that. It should be ok now.

LaurentMazare:
Merged, thanks for the PR!

jbochi:
Thank you for the prompt review, and amazing feedback as always!

PR #1303, opened 2023-11-09
Add support for TrOCR Model

ToluClassics:
Add Support for TrOCR(https://huggingface.co/microsoft/trocr-base-printed) Model


candle-examples/examples/trocr/readme.md
```diff
@@ -0,0 +1,16 @@
...
+
+`TrOCR` is a transformer OCR Model. In this example it is used to
+transcribe image text. See the associated [model
+card](https://huggingface.co/microsoft/trocr-base-printed) for details on
+the model itself.
+
+## Running an example
+
+```bash
+cargo run --example trocr --release --  --which base --cpu --image assets/trocr.png  --tokenizer tokenizer.json
```
LaurentMazare, line 11: Could you remove the tokenizer argument here and instead retrieve it by default from the hub if that's not already the case. Also better to reduce the number of arguments and use the default values where possible.

LaurentMazare (commented):
Looks like a great addition, thanks! Just one small nitpick.

candle-examples/examples/trocr/readme.md
ToluClassics: The tokenizer is missing a `tokenizer.json` on hub so i had to manually save to file. Is there another of loading the tokenizer without that file?

LaurentMazare: You can just create a `candle-trocr-tokenizer` model on the hf hub, upload the tokenizer config there so that future users don't have to do anything to get this file.

ToluClassics: Done!

LaurentMazare (approved):

LaurentMazare:
Merged, thanks a lot! Would be great to advertise this new model a bit if you felt like posting on twitter / reddit / discord so that potential users are aware of its existance.

katopz:
Thanks for nice addition! Anyway I have some issue for `microsoft/trocr-large-printed` when I try it with

```rust
    let vb = {
        let model = match args.model {
            Some(model) => std::path::PathBuf::from(model),
            None => match args.which {
                // Somehow this one is 404 
                Which::Base => Api::new()?
                    .repo(hf_hub::Repo::with_revision(
                        "microsoft/trocr-base-printed".to_string(),
                        hf_hub::RepoType::Model,
                        "main".to_string(),
                    ))
                    .get("model.safetensors")?,
                Which::Large => Api::new()?
                    .repo(hf_hub::Repo::with_revision(
                        "microsoft/trocr-large-printed".to_string(),
                        hf_hub::RepoType::Model,
                        "main".to_string(),
                    ))
                    .get("model.safetensors")?,
            },
        };
        println!("model: {:?}", model);
        unsafe { VarBuilder::from_mmaped_safetensors(&[model], DType::F32, &device)? }
    };
```
and
```
cargo run --example trocr --release --  --which large --cpu --image candle-examples/examples/trocr/assets/trocr.png
```
i got an error
```
Error: shape mismatch for encoder.embeddings.cls_token, expected: [1, 1, 768], got: [1, 1, 1024]
```

Not sure what I miss?

ToluClassics:
Hi @katopz , I think we might be missing the encoder configurations for the large model.
 Here's where the encoder config is for `base` https://github.com/huggingface/candle/blob/6d83d42efb1c8126c4fc34faee3f5a139b09dec6/candle-transformers/src/models/vit.rs#L38

PR #1306, opened 2023-11-09
Add weight and bias functions to LayerNorm

jwnz:
While porting some code from PyTorch to Candle, I needed to check the values of the weights and biases from LayerNorm layers, but as the values are private they can not be explicitly accessed. 

However, this functionality is present in the Linear layer with the use of two `pub fn weight(&self)` and `pub fn bias(&self)` functions.

I propose adding these two (QOL) functions to the LayerNorm `impl` to match the functionality of the Linear layer, while providing access to the weight and bias Tensors in the LayerNorm layer.

### DIFF
--- a/candle-nn/src/layer_norm.rs
+++ b/candle-nn/src/layer_norm.rs
@@ -95,6 +95,14 @@ impl LayerNorm {
             eps,
         }
     }
+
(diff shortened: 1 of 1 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1309, opened 2023-11-09
Adding the actual backend

Narsil:


candle-metal-kernels/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "candle-metal-kernels"
+version = "0.3.0"
+edition = "2021"
+
+description = "CUDA kernels for Candle"
```
grzuy, line 6: `%s/CUDA/Metal/` :-)

Narsil: Thanks.

Narsil, line 6:
```suggestion
description = "Metal kernels for Candle"
```

candle-core/src/metal_backend.rs
```diff
@@ -0,0 +1,838 @@
...
+    fn dtype(&self) -> DType {
+        self.dtype
+    }
+
+    fn device(&self) -> &Self::Device {
+        &self.device
+    }
+
+    fn to_cpu_storage(&self) -> Result<CpuStorage> {
+        match self.dtype {
```
LaurentMazare, line 108: Couldn't all this be rewritten as the following?
```rust
let len_in_elements = self.buffer.length() / self.dtype.size_in_bytes();
self.buffer.read_to_vec(len_in_elements)
```
(also maybe check that the size is divisible by `size_in_bytes` and `bail!` if not the case saying "internal error...")

candle-core/src/metal_backend.rs
```diff
@@ -0,0 +1,838 @@
...
+            &device.device,
+            &command_buffer,
+            &device.kernels,
+            el,
+            &self.buffer,
+            &mut buffer,
+            mul as f32,
+            add as f32,
+        )
+        .unwrap();
```
LaurentMazare, line 156: Avoid the unwrap, use a proper error instead.

candle-core/src/metal_backend.rs
```diff
@@ -0,0 +1,838 @@
...
+    }
+
+    fn affine(&self, layout: &Layout, mul: f64, add: f64) -> Result<Self> {
+        let device = self.device().clone();
+
+        let shape = layout.shape();
+        let el = shape.elem_count();
+        let dtype = self.dtype;
+
+        assert!(layout.is_contiguous());
```
LaurentMazare, line 140: Avoid the `assert!` use a `bail!` instead (advantages are that it will be checked even in release build + proper error handling is great for people that build servers on top of candle)

candle-core/src/metal_backend.rs
```diff
@@ -0,0 +1,838 @@
...
+        command_buffer.wait_until_completed();
+        return Ok(Self {
+            buffer,
+            device: device.clone(),
+            dtype,
+        });
+    }
+
+    fn powf(&self, _: &Layout, _: f64) -> Result<Self> {
+        todo!()
```
LaurentMazare, line 167: `bail!` instead (same comment applies below)

candle-core/src/metal_backend.rs
```diff
@@ -0,0 +1,838 @@
...
+
+        // The reduction loop requires the shared array to be properly initialized and for
+        // this we want the number of threads to be a power of two.
+        let (name, check_empty, return_index) = match (op, self.dtype) {
+            (ReduceOp::Sum, DType::F32) => ("fast_sum_float", false, false),
+            (ReduceOp::Min, DType::F32) => ("fast_min_float", true, false),
+            (ReduceOp::Max, DType::F32) => ("fast_max_float", true, false),
+            (ReduceOp::ArgMin, DType::F32) => ("fast_argmin_float", true, true),
+            (ReduceOp::ArgMax, DType::F32) => ("fast_argmax_float", true, true),
+            _ => todo!("Reduce op for non float"),
```
LaurentMazare, line 207: The `self.dtype` should be checked independently from the match on `op` no? (the same comment applies to the rest of this file)

candle-core/src/metal_backend.rs
```diff
@@ -0,0 +1,838 @@
...
+            &device.kernels,
+            name,
+            src_el,
+            dst_el,
+            &self.buffer,
+            &mut buffer,
+        )
+        .map_err(MetalError::from)?;
+        command_buffer.commit();
+        command_buffer.wait_until_completed();
```
LaurentMazare, line 227: Do we have to `wait_until_completed` for each op? This seems really inefficient compared to cuda async api but maybe there is a good reason that prevent us from just piping new commands in?

candle-metal-kernels/src/lib.rs
```diff
@@ -0,0 +1,1317 @@
+#![allow(clippy::too_many_arguments)]
```
LaurentMazare, line 1: Would be better to only have this on the functions that require it rather than at the file level.

candle-metal-kernels/src/lib.rs
```diff
@@ -0,0 +1,1317 @@
...
+}
+
+impl Kernels {
+    pub fn new() -> Self {
+        let libraries = RwLock::new(Libraries::new());
+        let funcs = RwLock::new(Functions::new());
+        Self { libraries, funcs }
+    }
+
+    // pub fn init(device: &Device) -> Result<Self, MetalKernelError> {
```
LaurentMazare, line 200: Already mentioned in another file but just in case please remove the commented code or add to the comment a reason while we should keep it (strong preference for it being removed :) ).

candle-metal-kernels/src/lib.rs
```diff
@@ -0,0 +1,1317 @@
...
+    );
+
+    let (thread_group_count, thread_group_size) = linear_split(&pipeline, dst_el);
+
+    encoder.dispatch_thread_groups(thread_group_count, thread_group_size);
+    encoder.end_encoding();
+    Ok(())
+}
+
+#[cfg(test)]
```
LaurentMazare, line 693: Could you move the test in a separate file? For a large number of tests I find it much better to have them split out as it simplifies the review process.

candle-metal-kernels/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "candle-metal-kernels"
+version = "0.3.0"
```
LaurentMazare, line 3: 0.3.1?

LaurentMazare (commented):
Thanks, that's a great addition. A couple minor comments attached but hopefully nothing major.
(also just to mention that I haven't reviewed the metal kernels themselves, I definitely trust you more than me on this)

candle-metal-kernels/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+edition = "2021"
+
+description = "Metal kernels for Candle"
+repository = "https://github.com/huggingface/candle"
+keywords = ["blas", "tensor", "machine-learning"]
+categories = ["science"]
+license = "MIT OR Apache-2.0"
+
+[dependencies]
+metal = { git = "https://github.com/ivarflakstad/metal-rs.git", features = ["mps"] }
```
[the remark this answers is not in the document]
Narsil: Upstream PR is here: https://github.com/gfx-rs/metal-rs/pull/285
Not a lot of activity so far we'll definitely fork in the meantime like we have done in the past for gemm and cudarc.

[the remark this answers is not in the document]
Narsil: Done

candle-core/src/metal_backend.rs
```diff
@@ -0,0 +1,838 @@
...
+impl std::ops::Deref for MetalDevice {
+    type Target = metal::DeviceRef;
+
+    fn deref(&self) -> &Self::Target {
+        &self.device
+    }
+}
+
+impl MetalDevice {
+    // pub fn metal_device(&self) -> &metal::DeviceRef {
```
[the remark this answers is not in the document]
Narsil: Done. (No good reason since there is the automatic Deref.

candle-core/src/metal_backend.rs
Narsil: Good suggestion, done.

candle-core/src/metal_backend.rs
Narsil: Done.

candle-core/src/metal_backend.rs
Narsil: Indeed. This is definitely scaffolding code which is helpful only while building the thing, I'll remove them now.

candle-core/src/metal_backend.rs
Narsil: Done.

candle-core/src/metal_backend.rs
Narsil: other dtypes are added in later PR, and using `&'static str` makes thing simpler to handle imo.
I plan to move to regular structs/enums later anyway which will remove the &'static str (This will wait after metal4)

candle-metal-kernels/src/lib.rs
Narsil: Since it affects pretty much all the API I put it there for conciseness.

I'd rather make those functions with less arguments tbh, but I don't like introducing too many custom structs to manipulate this API.

I'll make the change for clippy on the appropriate functions.

candle-metal-kernels/src/lib.rs
Narsil: Done.

candle-metal-kernels/src/lib.rs
Narsil: I prefer having tests close to their call site personally, makes imports easier and overall readability better.
It's also pretty much the standard in rust ecosystem: https://github.com/tokio-rs/tokio/blob/7a30504fd423aa782239ead92f3afa3474f8037c/tokio/src/util/memchr.rs#L29
https://github.com/serde-rs/serde/blob/2083f43a287cac8302009fda5bbe41518dd83209/test_suite/tests/test_ser.rs#L547
https://doc.rust-lang.org/book/ch11-01-writing-tests.html
https://github.com/BurntSushi/aho-corasick/blob/f227162f7c56b3f28e9dd69a655316af0107b29b/src/util/alphabet.rs#L300

For non `src/` tests, it also makes failure mode different (it fails fast by default).

I personnally don't have a strong preference tbh. I started moving but it start to feel less rusty, therefore I checked popular crates (above) and it seems to be definitely the norm to put the test in the same file as the logic they are testing.

Given this, would you reconsider?

candle-metal-kernels/Cargo.toml
Narsil: Done with the rebase.

candle-metal-kernels/src/lib.rs
LaurentMazare: Slightly prefer the separate files as I find it much better when reviewing code to know if it's mostly test being affected vs the actual code so if you prefer it this way I'm fine with it (but would rather keep the rest of the tests separate as they currently are and have this only in `candle-metal-kernels`).

Narsil: Done.

LaurentMazare (approved):

LaurentMazare:
Thanks, looks good to me so feel free to merge whenever you're happy with it too.

PR #1312, opened 2023-11-10
Check min to buckets in relative_position_bucket

bragovo:
Without min function buckets has wrong value in some cases.
In python version this check here - https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5/modeling_t5.py#L429-L431

After this fix encode has same result between rust and python versions.


LaurentMazare (approved):

LaurentMazare:
Thanks for spotting this!

PR #1313, opened 2023-11-10
Quantized example refactor

DaveTJones:
Attempted to make some areas of the code more readable/idiomatic. If there's any value in this I'm happy to try doing the same for the other examples.


DaveTJones:
The changes are small, pretty much just swapping in if let syntax in place of matching on Some None and moving big-ish match expressions into dedicated functions. Fully understand if this doesn't add value. 

Are there any areas of the project which would be suitable for a less experienced developer to contribute to at present?

LaurentMazare:
There are a couple issues marked with "help wanted" but I think they are all a bit involved. Besides this I would encourage you to use candle on the models you care about and give feedback on how well it does and what could be improved. Porting existing PyTorch models is useful both to learn about candle itself, help improve it, and also the models themselves might interest other folks.

Issue #1314, opened 2023-11-10
Quantized zephyr example repeats system prompt


tripokey:
The quantized example repeats the '<|system|>' portion of the prompt for zephyr when chat mode has been selected causing it to not handle follow up questions well. I believe that the '<|system|>' portion should only be present for the first iteration in the chat session.

PR #1317, opened 2023-11-10
Fix quantized zephyr chat prompt (#1314)

tripokey:


LaurentMazare:
Thanks for noticing this, I've made a small tweak to avoid the mutable variable but looks all good.

PR #1318, opened 2023-11-11
Starting to fix some tests.

Narsil:
Depends on https://github.com/huggingface/candle/pull/1323


candle-metal-kernels/src/reduce.metal
```diff
@@ -68,72 +70,74 @@ kernel void NAME( \
...
-     dst[idx] *= inv_acc; 
-     idx += blockDim;
-   }
-}
-
 
 REDUCE(x + y, fast_sum_float, float)
 REDUCE(x * y, fast_mul_float, float)
 REDUCE(max(x, y), fast_max_float, float)
+
```
LaurentMazare, line 77: Large macros like this are a bit of a pain both when having compilation errors and when editing the code. Instead on the cuda side for new kernels I lean on using a C++ templated function (here it would be templated on `T`) and have a very simple macro to call the appropriate templated function. Not sure if metal files let you write some C++ code but that may be a good thing to do.

candle-nn/src/ops.rs
```diff
@@ -201,6 +201,46 @@ impl candle::CustomOp1 for SoftmaxLastDim {
...
+        let kernels = device.kernels();
+        let name = match storage.dtype() {
+            DType::F32 => "softmax_float",
+            DType::F16 => "softmax_half",
+            DType::BF16 => "softmax_bfloat",
+            dtype => candle::bail!("softmax-last-dim is not implemented for {dtype:?}"),
+        };
+
+        let n = layout.stride().len();
+        if !(layout.stride()[n - 1] == 1 && layout.start_offset() == 0) {
```
LaurentMazare, line 223: Why is `start_offset()` being different from 0 an issue here? Couldn't we just add it to the buffer pointer?

candle-nn/src/ops.rs
```diff
@@ -201,6 +201,46 @@ impl candle::CustomOp1 for SoftmaxLastDim {
...
+        &self,
+        storage: &candle::MetalStorage,
+        layout: &Layout,
+    ) -> Result<(candle::MetalStorage, Shape)> {
+        use candle::{backend::BackendStorage, DType};
+        let device = storage.device();
+        let command_buffer = device.command_buffer();
+        let kernels = device.kernels();
+        let name = match storage.dtype() {
+            DType::F32 => "softmax_float",
```
LaurentMazare, line 216: The convention on the cuda side is to use names that are based on the rust type, e.g. `softmax_f32` etc. Using this we generate the name of the kernels automatically for most ops. Might be worth to be uniform here and rename these appropriately?

candle-core/src/metal_backend.rs
```diff
@@ -70,16 +112,107 @@ impl MetalDevice {
...
 
-    pub fn new_buffer(&self, element_count: usize, dtype: DType) -> Buffer {
+    pub fn new_buffer(&self, element_count: usize, dtype: DType) -> Arc<Buffer> {
         let size = (element_count * dtype.size_in_bytes()) as NSUInteger;
-        self.device
-            .new_buffer(size, MTLResourceOptions::StorageModeManaged)
+        self._new_buffer(size, MTLResourceOptions::StorageModePrivate)
+    }
+
+    fn _new_buffer(&self, size: NSUInteger, option: MTLResourceOptions) -> Arc<Buffer> {
```
LaurentMazare, line 120: The leading `_` typically are used to bypass the "is used" warnings. For internal functions, naming them with an underscore at the end seems usually better.

candle-core/src/metal_backend.rs
```diff
@@ -70,16 +112,107 @@ impl MetalDevice {
...
+        let mut buffers = self.buffers.try_write().unwrap();
```
LaurentMazare, line 121: When would this `unwrap` fails? Feels a bit that this function should return a `Result` and propagate the error properly just in case

candle-core/src/metal_backend.rs
```diff
@@ -70,16 +112,107 @@ impl MetalDevice {
...
 }
 
 #[derive(Debug, Clone)]
 pub struct MetalStorage {
-    buffer: metal::Buffer,
+    buffer: Arc<metal::Buffer>,
+    matrices: Arc<
+        RwLock<
+            HashMap<
+                (
```
LaurentMazare, line 203: Large tuple are very error prone and hard to understand, couldn`t we use a `struct` instead? This would document a bit via the field name what these different values are used for.

candle-core/src/metal_backend.rs
```diff
@@ -126,30 +268,48 @@ impl BackendStorage for MetalStorage {
...
+                &command_buffer,
+                &device.kernels,
+                name,
+                el,
+                &self.buffer,
+                &buffer,
+                mul as f32,
+                add as f32,
+            )
+            .map_err(MetalError::from)?;
```
LaurentMazare, line 290: Feels a bit that these error conversions could be handled by the `?` operator. Is there some difficulty in implementing the appropriate conversion traits so that we can remove these `map_err`?

candle-core/src/metal_backend.rs
```diff
@@ -163,11 +323,11 @@ impl BackendStorage for MetalStorage {
     fn reduce_op(&self, op: ReduceOp, layout: &Layout, sum_dims: &[usize]) -> Result<Self> {
         if !(sum_dims.len() == 1
             && sum_dims[0] == layout.shape().rank() - 1
-            && layout.is_contiguous()
-            && layout.start_offset() == 0)
+            && layout.stride()[sum_dims[0]] == 1)
         {
-            crate::bail!("Non contiguous reduce op not supported yet");
+            crate::bail!("Non last dim reduce op not supported yet");
```
LaurentMazare, line 328: Maybe split this error message a bit as it covers multiple cases, e.g. for `sum_dims.len() != 1` mention that sums over multiple dims are not supported etc.

candle-core/src/metal_backend.rs
```diff
@@ -202,8 +362,11 @@ impl BackendStorage for MetalStorage {
             Err(crate::Error::EmptyTensor { op: "reduce" }.bt())?
         }
         let dtype = if return_index { DType::U32 } else { self.dtype };
-        let mut buffer = device.new_buffer(dst_el, dtype);
-        let command_buffer = self.device.command_queue.new_command_buffer();
+        if dtype == DType::U32 {
+            crate::bail!("Implement return index reduce op");
```
LaurentMazare, line 366: Not sure to understand this error message, could you make it more useful for the final user to diagnose the actual issue and know what to tweak?

candle-core/src/metal_backend.rs
```diff
@@ -398,14 +635,22 @@ impl BackendStorage for MetalStorage {
...
-        let command_buffer = self.device.command_queue.new_command_buffer();
+        let buffer = self.device.new_buffer(el, dtype);
+        let command_buffer = self.device.command_buffer();
+        if t.dtype() != f.dtype() {
+            crate::bail!("Invalid ternary different dtypes for values");
+        }
+        let name = match (self.dtype, t.dtype()) {
+            (DType::U8, DType::F32) => "where_u8_f32",
+            (DType::U8, DType::F16) => "where_u8_f16",
+            (left, right) => crate::bail!("Ternary {left:?} - {right:?} not implemented"),
```
LaurentMazare, line 646: Improve the error message, "ternary" -> "where" etc.

candle-core/src/metal_backend.rs
```diff
@@ -398,14 +635,22 @@ impl BackendStorage for MetalStorage {
         let dims = shape.dims();
         let el = shape.elem_count();
         let dtype = t.dtype;
-        let mut buffer = self.device.new_buffer(el, dtype);
-        let command_buffer = self.device.command_queue.new_command_buffer();
+        let buffer = self.device.new_buffer(el, dtype);
+        let command_buffer = self.device.command_buffer();
+        if t.dtype() != f.dtype() {
+            crate::bail!("Invalid ternary different dtypes for values");
```
LaurentMazare, line 641: Improve the error message "ternary" -> "where", maybe mention the dtypes?

candle-core/src/metal_backend.rs
```diff
@@ -647,70 +879,112 @@ impl BackendStorage for MetalStorage {
...
+            let descriptor = if transpose {
+                MatrixDescriptor::init_multiple(n, m, b, m * size, m * n * size, type_id)
+            } else {
+                MatrixDescriptor::init_multiple(m, n, b, n * size, m * n * size, type_id)
+            };
+            let matrix = Matrix::init_with_buffer_descriptor(&self.buffer, offset, &descriptor)
+                .ok_or_else(|| {
+                    MetalError::from("Failed to create matrix multiplication kernel".to_string())
+                })?;
+            matrices.insert(key, matrix.clone());
```
LaurentMazare, line 984: How are these collected when not useful anymore?

candle-metal-kernels/src/lib.rs
```diff
@@ -218,22 +229,43 @@ impl Kernels {
...
     ) -> Result<Function, MetalKernelError> {
-        let mut funcs = self.funcs.write()?;
-        if let Some(func) = funcs.get(name) {
-            Ok(func.clone())
+        let func = self
+            .load_library(device, source)?
+            .get_function(name, None)
+            .map_err(|e| MetalKernelError::LoadFunctionError(e.to_string()))?;
+        Ok(func)
+        // let mut funcs = self.funcs.write()?;
```
LaurentMazare, line 243: As usual please avoid having commented out code.

candle-nn/src/ops.rs
Narsil: Done.

candle-metal-kernels/src/reduce.metal
Narsil: Agreed. I'll modify in the future (mostly a pain to have all those `\` everywhere, the errors are actually pretty nice unless you forget one of them).

There are some restrictions though on templating in metal language though I had issue in converting softmax for instance (which was bugged.)

candle-core/src/metal_backend.rs
Narsil: I renamed this to `allocate_buffer` which is actually a better name imo.
I wrote `_new_buffer` by lack of good name in the meantime.

candle-core/src/metal_backend.rs
Narsil: It never should, at least in single threaded, otherwise it's a pretty bad bug in the scheduling/allocator.

In multi threaded, we need to decide, if we want to wait on the lock (with a timeout probably since deadlocks could really happen).
I think I'm ok delaying this decision when the current implem for single threaded as proven it's worth (I don't think there's a lot of benefit of doing multithreaded encoding of the commands at the moment).

Moved to Result though.

candle-core/src/metal_backend.rs
Narsil: Everything is removed as I removed MPS.

candle-core/src/metal_backend.rs
Narsil: Removed all that logic.

candle-metal-kernels/src/lib.rs
Narsil: Done.

candle-core/src/metal_backend.rs
Narsil: done.

candle-core/src/metal_backend.rs
Narsil: Oh it's the reductions like argmax which have a different return types which are not implemented yet.

candle-core/src/metal_backend.rs
Narsil: Done.

candle-core/src/metal_backend.rs
Narsil: Done.

candle-nn/src/ops.rs
Narsil: Fixed.

candle-core/src/metal_backend.rs
Narsil: Because our big error type doesn't implement those froms.

We could implement the from `MetalKernelError` directly in our large error sum type, but I think it's better to encapsulate everything in raw `MetalError` which encapsulates everything linked to metal (Kernel and non kernel errors).

I don't like those map_err either, but I don't know of any simpler way to do this hierarchy (without modifyin the device trait).

PR #1338, opened 2023-11-16
Initial work on improving the book.

ryankopf:
The book is in need of serious work and completion, as this crate is relatively new. Some work in building the book is a challenge due to an issue with [mdbook](https://github.com/rust-lang/mdBook/issues/706).

After quite a bit of thinking about how best to approach this, I've found the changes in this PR should prove fruitful for future book updates.

**Code should be executable directly.**

For beginners, it's important that the code in the book be copy and pastable directly into their editor without changes, to help them have a smooth learning experience. This is already the case with the "Hello World" example, where the code has a main() function and everything.

Some of the later code, however, just looks like script snippets. This is odd in Rust, as Rust is not a scripting language, and these code fragments can not be executed.

I found out these code fragments exist AS fragments because the auto-testing facilities available in mdbook don't support some crates, where you need to specify the location. So the code was instead written in testable ".rs" files and imported into the book.

My PR and suggestion for future book updates - which I am likely to make several of if this is improved - is to make a very couple of simple modifications to the test anchors, splitting it out into a "top" (includes) and content, which allows for a very simple main() structure to be written inside the 'ignored' Rust code that the mdbook program can not execute.

To note, this PR should not be controversial because:
* Currently the book uses a mix of externally imported code and direct code in the MD files, but testing is inconsistent
* There is untested code already in the book, on the very second page, and this change will serve as an excellent guide to turning that code into tested code instead. I have not done that in this PR to keep it simple.

### DIFF
--- a/candle-book/src/inference/hub.md
+++ b/candle-book/src/inference/hub.md
@@ -10,17 +10,23 @@ Then let's start by downloading the [model file](https://huggingface.co/bert-bas
 
 
 ```rust
-# extern crate candle_core;
 # extern crate hf_hub;
+# extern crate candle_core;
 use hf_hub::api::sync::Api;
 use candle_core::Device;
 
-let api = Api::new().unwrap();
-let repo = api.model("bert-base-uncased".to_string());
+fn main() {
+    let api = Api::new().unwrap();
+    let repo = api.model("bert-base-uncased".to_string());
 
-let weights = repo.get("model.safetensors").unwrap();
+    let weights = repo.get("model.safetensors").unwrap();
 
-let weights = candle_core::safetensors::load(weights, &Device::Cpu);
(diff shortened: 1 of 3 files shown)


ryankopf:
I will note this has been tested with cargo test and md book build, and the output looks great, fully runnable code.

`use candle::Device;
use hf_hub::api::tokio::Api;

#[tokio::main]
async fn main() {
        let api = Api::new().unwrap();

    let weights_filename = repo.get("model.safetensors").await.unwrap();

    let weights = candle::safetensors::load(weights_filename, &Device::Cpu).unwrap();
    for (name, tensor) in weights.iter() {
        println!("Weight: {}, Dimensions: {:?}", name, tensor);
    }
}`

PR #1341, opened 2023-11-18
Metal quantized modifications proposal.

Narsil:
- Add a device param, wherever needed.
- Create new QMetal storage thing that implements QuantizedType.
- Update everywhere needed.


Narsil:
@LaurentMazare Wdyt of this proposal, is this the way we should go ?

Narsil:
Superseeded by : https://github.com/huggingface/candle/pull/1523 which actually implements everything.

PR #1346, opened 2023-11-19
Add OpenChat 3.5 to quantized examples

lucasavila00:
This MR adds the [OpenChat 3.5](https://huggingface.co/openchat/openchat_3.5/tree/main) model to the quantized example.

Specifically, the GGUF version from https://huggingface.co/TheBloke/openchat_3.5-GGUF/tree/main.

OpenChat is the best 7b model according to Arena ELO https://arena.lmsys.org/

It is a fine tune of mistral, and configuring it as mistral works well.

One prompt:

``` 
$ cargo run --example quantized --release -- --prompt "The best thing about coding in rust is " --which 7b-open-chat-3.5
   Compiling candle-examples v0.3.1 (/home/lucas/oss/candle/candle-examples)
    Finished release [optimized] target(s) in 4.12s
     Running `target/release/examples/quantized --prompt 'The best thing about coding in rust is ' --which 7b-open-chat-3.5`
loaded 291 tensors (4.37GB) in 0.08s
model built
The best thing about coding in rust is 0 cost abstractions. In other words, the compiler can inline functions if they meet certain conditions (having no side effects, being small enough etc.)

I've been working on a project that uses a lot of ffi (foreign function interface) and wanted to write some wrappers for the underlying library, since the language binding I was using lacked some utility functions. Now, writing these wrapper functions was not possible since any function that calls an external function is leftmost and thus cannot be inlined by rust's compiler. This was a problem since one of the functions is called very frequently (around 50'000 times per second) ...
```

Chat:

``` 
$ cargo run --example quantized --release -- --prompt "chat" --which 7b-open-chat-3.5
   Compiling candle-examples v0.3.1 (/home/lucas/oss/candle/candle-examples)
    Finished release [optimized] target(s) in 4.22s
     Running `target/release/examples/quantized --prompt chat --which 7b-open-chat-3.5`
loaded 291 tensors (4.37GB) in 0.08s
model built
> Tell me a joke.
User: Tell me a joke.<|end_of_turn|>Assistant: 1 Why don't scientists trust atoms? Because they make up everything!

  12 prompt tokens processed: 9.11 token/s
  15 tokens generated: 6.28 token/s
> 
```

### DIFF
--- a/candle-examples/examples/quantized/main.rs
+++ b/candle-examples/examples/quantized/main.rs
@@ -53,6 +53,8 @@ enum Which {
     Zephyr7bAlpha,
     #[value(name = "7b-zephyr-b")]
     Zephyr7bBeta,
+    #[value(name = "7b-open-chat-3.5")]
+    OpenChat35,
 }
 
 impl Which {
@@ -67,8 +69,9 @@ impl Which {
             | Self::L7bCode
             | Self::L13bCode
             | Self::L34bCode => false,
-            // Zephyr is a fine tuned version of mistral and should be treated in the same way.
-            Self::Zephyr7bAlpha
+            // Zephyr and OpenChart are fine tuned versions of mistral and should be treated in the same way.
+            Self::OpenChat35
+            | Self::Zephyr7bAlpha
             | Self::Zephyr7bBeta
             | Self::Mistral7b
             | Self::Mistral7bInstruct => true,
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Thanks for adding this, I've made some small tweaks to make it more in line with the other model and avoid model specific code.

PR #1356, opened 2023-11-21
Add candle-lora transformers to readme?

EricLBuehler:
Hi @LaurentMazare,

I have recently added LoRA support to many models in `candle-transformers`. These "LoRA transformers" are in `candle-lora`. However, I think that it would be helpful to add a link for these into the README as it would increase discoverability for developers who want to fine-tune models. Currently, one would need to go to the `candle-lora` repository to find these models, and while that may be intuitive I think making their location explicit would be better.

What do you think about this addition?

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,20 +141,8 @@ And then head over to
 ## Useful External Resources
 - [`candle-tutorial`](https://github.com/ToluClassics/candle-tutorial): A
   very detailed tutorial showing how to convert a PyTorch model to Candle.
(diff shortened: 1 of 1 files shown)


LaurentMazare:
That seems a bit long to me and the readme is already overlong. I think we should stick to roughly a line per item and interested readers can click on the link to get more details.

EricLBuehler:
I just shortened my addition to be a single line, hopefully this is better?

LaurentMazare (approved):

LaurentMazare:
Yeah looks reasonable thanks.

EricLBuehler:
Thanks!

PR #1358, opened 2023-11-22
Fix linspace implementation

MilkFather:
This pull request attempts to fix a potential issue in linspace implementation.

`steps` should be strictly greater than 1 to make it consistent with the context (namely, to make it consistent with the error message and to avoid a potential divide-by-zero/NaN).

Although this implementation of linspace is still not completely conformant to the PyTorch equivalent, for example:
```python3
torch.linspace(0.0, 1.0, steps=1)    # tensor([0.])
torch.linspace(0.0, 1.0, steps=0)    # tensor([])
torch.linspace(0.0, 1.0, steps=-1)    # RuntimeError: number of steps must be non-negative
```

### DIFF
--- a/candle-examples/examples/yolo-v8/main.rs
+++ b/candle-examples/examples/yolo-v8/main.rs
@@ -7,7 +7,7 @@ extern crate accelerate_src;
 mod model;
 use model::{Multiples, YoloV8, YoloV8Pose};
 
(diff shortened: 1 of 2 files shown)


LaurentMazare:
Thanks for pointing this out, I've tweaked this PR to handle the 0 and 1 cases properly instead of just failing.

PR #1360, opened 2023-11-22
Ensure to copy data to cpu before iterating.

xnorpx:
Noticed that report was quite slow when testing the yolo8 sample.

Doing a bulk copy to CPU reduce process time from 800ms -> 8ms.


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks for benchmarking this.

PR #1362, opened 2023-11-23
Run nvcc via ccache in flash-attn

mmoskal:
On my test box, compiling `candle-flash-attn` from scratch in debug takes about 3 minutes, but most of this time is spent running `nvcc` on the flash attention kernels. Running `nvcc` via `ccache` brings this down to around 15 seconds (on the second build). Also, the ccache is shared between debug and release builds.

Do you think it's OK to always run via ccache? It seems straightforward to install.

### DIFF
--- a/candle-flash-attn/build.rs
+++ b/candle-flash-attn/build.rs
@@ -104,8 +104,9 @@ fn main() -> Result<()> {
         cu_files
             .par_iter()
             .map(|(cu_file, obj_file)| {
-                let mut command = std::process::Command::new("nvcc");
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Actually we already have a caching mechanism for this though you need to use an environment variable to set it, see [this comment](https://github.com/huggingface/candle/blob/ca19a9af6220366bb0beaea1eb7f34a1a8a2e07b/candle-flash-attn/build.rs#L1).

LaurentMazare:
Closing this as stale.

PR #1370, opened 2023-11-24
Add methods to clear cache for Llama models

EricLBuehler:
Hello everybody,

I have added some methods to clear the KV and (if applicable) the mask caches for the Llama models. These methods take an `&mut self` to emphasize their mutation. 

This functionality is necessary for the resolution of EricLBuehler/candle-vllm#6. Do you think it would be a good idea to add these methods to the Llama models here, or should I have a separate, modified Llama implementation? I would prefer to have a dependency to `candle-transformers`, but please let me know what you think.

### DIFF
--- a/candle-examples/examples/llama_multiprocess/model.rs
+++ b/candle-examples/examples/llama_multiprocess/model.rs
@@ -306,6 +306,11 @@ impl CausalSelfAttention {
             cache: cache.clone(),
         })
     }
+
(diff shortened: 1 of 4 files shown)


EricLBuehler:
Closed as of #1464

LaurentMazare:
The llama examples have been tweaked to use an explicit cache instead so this shouldn't apply (and would have been quite error prone anyway).

Issue #1371, opened 2023-11-25
Add `empty` and `empty_like` functions to `Tensor`


EricLBuehler:
I noticed that there appears to be no [`empty`](https://pytorch.org/docs/stable/generated/torch.empty.html) or [`empty_like`](https://pytorch.org/docs/stable/generated/torch.empty_like.html) functions for `Tensor`.

Taking a quick look at the Candle backend and Tensor code, it appears that this would be as simple as adding a function similar to `zeros_impl` - perhaps `empty_impl`. This would do the same thing but call the [`alloc` method](https://github.com/coreylowman/cudarc/blob/42f8f861ca1c9e586ff695efb29bd3719a2802a7/src/driver/safe/alloc.rs#L132) (which is already called at other places) instead of `alloc_zeroes`.

Given the relatively easy implementation of these functions, and the fact that they provide a faster method to create a Tensor (especially for large or numerous tensors) because they do not need to write zeroes, do you think this would be a good addition? I would be happy to contribute these.

PR #1374, opened 2023-11-25
Add the empty, empty_like methods to Tensor

EricLBuehler:
I added the empty and empty_like methods to Tensor. These produce uninitialized tensors that do not require a (potentially costly) `memset` call. Please see #1371 for the original issue.

### DIFF
--- a/candle-core/src/backend.rs
+++ b/candle-core/src/backend.rs
@@ -110,6 +110,8 @@ pub trait BackendDevice: Sized + std::fmt::Debug + Clone {
 
     fn same_device(&self, _: &Self) -> bool;
 
+    fn empty_impl(&self, _shape: &Shape, _dtype: DType) -> Result<Self::Storage>;
+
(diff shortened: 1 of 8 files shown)


candle-core/src/backend.rs
```diff
@@ -110,6 +110,8 @@ pub trait BackendDevice: Sized + std::fmt::Debug + Clone {
 
     fn same_device(&self, _: &Self) -> bool;
 
+    fn empty_impl(&self, _shape: &Shape, _dtype: DType) -> Result<Self::Storage>;
```
EricLBuehler, line 113: Should this be an `unsafe` method because it produces uninitialized data?

EricLBuehler:
@LaurentMazare, what are your thoughts on this addition?

LaurentMazare:
No real interest for this, by design tensors are immutable in candle (except for vars) so contrary to PyTorch there isn't much of a point making an uninitialized tensor that couldn't be set later anyway. We would need a pretty compelling use case to add this.

PR #1384, opened 2023-11-28
Speedup ShardedSafeTensors to load Tensors with default hints

mokeyish:
ShardedSafeTensors load tensors with default hints very slowly, and there's no need to slice them.


mokeyish:
@LaurentMazare Could you merge this PR? 

I use ShardedSafeTensors to load My model(which may enable tensor parallel or not). when disable tensor parallel, it is really very slow.

Or you need to provide a method that can be temporarily converted to simplebackend.

LaurentMazare:
Merged, sorry for the delai.

PR #1386, opened 2023-11-29
Fix contiguous affine check for metal backend

jbochi:
I believe the logic is reversed. When `layout.is_contiguous` we don't need to bail.

I also broke the check into two so that the error message is more specific.


LaurentMazare:
Adding @Narsil and @ivarflakstad explicitely as they know far better than me on the metal bits.

PR #1387, opened 2023-11-29
Add gelu unary op for metal

jbochi:
I'm trying to run T5 on metal, but gelu and softmax-last-dim are missing. This adds gelu.

It also fixes the test for cos f16, which fails on my machine (M2 chip), by reducing the precision:

```
---- tests::cos_f16 stdout ----
thread 'tests::cos_f16' panicked at 'assertion failed: `(left == right)`
  left: `[0.54, -0.4165, -0.9902]`,
 right: `[0.5405, -0.4163, -0.9902]`', src/tests.rs:442:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

cos(1) is actually 0.5403023059, and cos(2) is -0.4161468365, so I don't think it makes sense to have expectations that are already approximate.

cc  @Narsil and @ivarflakstad . Please take a look 🙏


ivarflakstad (commented):
~~Nice!~~
~~Could add for bfloat as well?😊~~
~~I won't take the liberty of approving - leaving that to Narsil 👍~~

Edit: I thought this based on the metal4 branch. Sorry for the confusion!

Narsil:
Could you check out this PR, most of the work you are doing is already included in it.

It' s basically implementing most metal kernels, with a lot of bug fixes a bit everywhere.

jbochi:
Thanks @ivarflakstad and @Narsil . I'll close this in favor of #1318  :)

By the way, great work adding the metal backend, @Narsil !

Narsil:
Thanks for the kind words. If you want to help, I'll try to be more in the candle discord channel in the next few days.

I'm currently preparing some new shiny stuff over at TGI, but metal is still pretty high on the list of TODO.

PR #1388, opened 2023-11-30
Add `to_device()` to VarBuilder to load weights to a specific GPU

mokeyish:
This makes it easier to load transformer blocks onto different GPUs.

```rust

fn load(vb) {

  // ...

  let blocks: Vec<_> = (0..cfg.n_layers)
      .map(|i| {
          let dev_ordinal = i / per_device_layers;
          log::debug!("load block {i} into GPU{dev_ordinal}");

          let vb = vb.to_device(&Device::new_cuda(dev_ordinal).unwrap());

          Block::load(
              vb.pp(&format!("model.layers.{i}")),
              cache,
              &cfg,
              comm.clone(),
          )
          .unwrap()

      })
      .collect();

     // ...
}

```

### DIFF
--- a/candle-nn/src/var_builder.rs
+++ b/candle-nn/src/var_builder.rs
@@ -32,7 +32,7 @@ impl<'a, B: Backend> Clone for VarBuilderArgs<'a, B> {
 pub type VarBuilder<'a> = VarBuilderArgs<'a, Box<dyn SimpleBackend + 'a>>;
 
 struct TensorData<B: Backend> {
-    backend: B,
+    backend: Arc<B>,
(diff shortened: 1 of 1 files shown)


mokeyish:
@LaurentMazare Hi, there two checks failed because the credentials could not be obtained.

But what do you think of this feature? Or are there any other alternatives?

LaurentMazare:
I agree that the use case you mentioned is not well covered at the moment. What I was just saying is that the changes proposed by this PR have some drawbacks that we have to be a bit more careful about - I would have to think a bit more about it to see if there is a good way around this.

PR #1393, opened 2023-12-02
Add quantized Starling, fix open-chat prompt

lucasavila00:


candle-examples/examples/quantized/main.rs
```diff
@@ -368,8 +398,10 @@ fn main() -> anyhow::Result<()> {
                         prompt.pop();
                     }
                 }
-                if args.which.is_open_chat() {
-                    format!("User: {prompt}<|end_of_turn|>Assistant: ")
+                if args.which.is_starling() {
+                    format!("GPT4 Correct User: {prompt}<|end_of_turn|>GPT4 Correct Assistant: ")
+                } else if args.which.is_open_chat() {
+                    format!("GPT4 User: {prompt}<|end_of_turn|>GPT4 Assistant: ")
```
LaurentMazare, line 404: When looking at the huggingface page for openchat3.5, I see the example prompts using "GPT4 Correct User" rather than "GPT4 User", which one should be used here?
https://huggingface.co/openchat/openchat_3.5
Maybe you could add some examples to the rest you've run with the old and new prompts to help figure out how much of a difference this makes.

lucasavila00: I will provide some examples. I was testing it mostly in interactive chat and the differences are subtle. I haven't saved anything, too.

But regarding the prompt, I based this MR on the paper, which says:

https://arxiv.org/pdf/2309.11235.pdf
image

I can't find the "GPT 4 Correct User" part on the open-chat page or the GGUF page. GGUF says "GPT 4 User" too https://huggingface.co/TheBloke/openchat_3.5-GGUF#prompt-template-openchat


The Starling model, however, uses "Correct User" per the documentation https://huggingface.co/berkeley-nest/Starling-LM-7B-alpha#code-examples

Notice the GGUF is likely incorrect https://huggingface.co/TheBloke/Starling-LM-7B-alpha-GGUF#prompt-template-openchat

LaurentMazare: I see that being mentioned on the page I mentioned https://huggingface.co/openchat/openchat_3.5
prompt

lucasavila00: Examples:

1. Factorial in Rust & Elixir: https://gist.github.com/lucasavila00/b0d24fd6093da8b7021171ff204fcddc
2. Why is Rust the best programming language: https://gist.github.com/lucasavila00/cf6e166ca6385ba528f1e04410803930

lucasavila00: I'll also run the examples with the "Correct user" prompt, 1 minute

lucasavila00: I updated the gists, and changed both starling and open chat to use 

```
format!("GPT4 Correct User: {prompt}<|end_of_turn|>GPT4 Correct Assistant:")
``` 

Notice there is no space after "assistant:" where it used to have one. The gists show the difference the spaces make.

lucasavila00: As a side note, these models are using `Q4_K_M` because the Zephyr models also used this, and mistal also uses Q4.

In my personal experience, Q4 is too degraded and only Q6+ looks similar to F16 results.

I'd be happy to create another MR which changes the quantization from Q4 to Q6, and I would provide examples of the changes.

LaurentMazare (approved):

PR #1395, opened 2023-12-02
Stable Diffusion Turbo Support

edwin0cheng:
This PR added following features  in the stable-diffusion example:

* `--sd-version turbo` 
* `--guidance-scale`  for controlling guidance scale, which SD Turbo need to set it to 0 for the correct result.

you can try this by:

``` bash
 cargo run --release --example stable-diffusion --features "cuda,cudnn" -- --prompt "A cinematic shot of a baby racoon wearing an intricate italian priest robe." --use-f16 --sd-version turbo --n-steps 1 --width 512 --height 512 --guidance-scale 0
```


LaurentMazare:
Thanks for the PR, I made some small tweaks so that the default values for `n_steps` and `guidance_scale` are automatically inferred when using the turbo version so hopefully no need for the user to set these flags manually.

PR #1408, opened 2023-12-06
Fix NaN errors for Gelu in Metal

jbochi:
This is based on https://github.com/huggingface/candle/pull/1318, which is adding the gelu kernel. cc @Narsil 

The flan t5 embeddings are all NaN when metal is enabled. This is because the output of gelu(x) for x > 10 is NaN.

```
$ cargo run --example t5 --release --features metal -- --model-id "google/flan-t5-small" --prompt "translate to German: A beautiful candle."
[[[NaN, NaN, NaN, ..., NaN, NaN, NaN],
  [NaN, NaN, NaN, ..., NaN, NaN, NaN],
  [NaN, NaN, NaN, ..., NaN, NaN, NaN],
  ...
  [NaN, NaN, NaN, ..., NaN, NaN, NaN],
  [NaN, NaN, NaN, ..., NaN, NaN, NaN],
  [NaN, NaN, NaN, ..., NaN, NaN, NaN]]]
Tensor[[1, 9, 512], f32, metal:4294968146]
Took 26.179209ms
```

With this change, I get embeddings approximately equal to the ones I get on the CPU.

### DIFF
--- a/candle-metal-kernels/src/tests.rs
+++ b/candle-metal-kernels/src/tests.rs
@@ -295,7 +295,7 @@ fn run_affine<T: Clone>(v: &[T], mul: f64, add: f64) -> Vec<T> {
     output.read_to_vec::<T>(v.len())
 }
 
-fn _run_affine_strided<T: Clone>(
+fn run_affine_strided<T: Clone>(
     v: &[T],
     shape: &[usize],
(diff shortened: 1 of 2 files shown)


Narsil (approved):
LGTM

Issue #1359, opened 2023-11-22
Additional Tensor Operations Needed


MilkFather:
I am re-implementing the famous NeRF (https://github.com/bmild/nerf) model in candle. During my experience with candle, I found the following Tensor operations missing.
* `torch.linspace`
* `torch.cumprod`
* `torch.searchsorted`

I am able to find a workaround for `torch.linspace` and `torch.cumprod`, but `torch.searchsorted` looks formidable. Are there recommended ways to implement that efficiently? Thanks.

<details>
<summary>Current workarounds for <code>linspace</code> and <code>cumprod</code></summary>

* `linspace`: directly taken from the stable-diffusion example.
* `cumprod`: modified from https://github.com/mokeyish/candle-ext/blob/1b38cb1b9faa9694c204d46788fd70d479109bab/src/cumsum.rs#L6-L28 changing the add operation to multiply operation. Not the optimal solution, I admit.

</details>

LaurentMazare:
For `cumprod`, if all your values are positive I would recommend doing `log` then the built-in `cumsum` then `exp`, this should be more efficient and easier than going the row data.
For `searchsorted` and `sort`, I think you should do a custom-op for these. The goal of custom ops is to have candle be extensible so that operations that are not super common can be added by the user. You can see an example of implementation [here](https://github.com/huggingface/candle/blob/f83e14f68dc44b869a605a440feeacf3ef8945c5/candle-core/tests/custom_op_tests.rs#L24).
Let me know if you have further questions or you want me to sketch a basic `sort` function using this.

MilkFather:
I would like to see an example implementation of `sort`.

The `sort` function in PyTorch returns a tuple of tensors - one for the sorted tensor (preserving the original DType) and another one for indexes (a LongTensor that can be used for indexing).

From what I understand, the custom op traits have a fixed output type (`Result<Storage, Shape>`) that looks incompatible with this particular use case. Hence I would like to see an example implementation for that.

LaurentMazare:
Or a simple idea there would just be to have a `sort_index` custom op that return the sorted indexes and to use some gather function to get the sorted values if needed.

MilkFather:
@LaurentMazare I spent the last weekend digging through the documentation and found that the crate is very under-documented. The example given does not touch layout manipulation at all and does little help to understanding the required steps to implement a custom operation like sort or search.

By the way, none of the "some gathering function" really pops into my mind, which leads me back to the starting point.

[...]

PR #1418, opened 2023-12-09
Example `CustomOp2` of `torch.searchsorted` 

jeromeku:
## `torch.searchsorted` `CustomOp2` Implementation and Example

Per #1359, added `cpu` and `cuda` implementations of `torch.searchsorted` as a `CustomOp2` as a `candle-examples` example.

- provides same `API` as original `torch` implementation except for `sort` argument
> [!IMPORTANT]  
> The tensor being searched **MUST** be sorted
- includes a `kernel_test` folder for testing standalone `cuda` implementation
- also includes `jupyter notebook` for generating test data from `torch.searchsorted`

See `README.md` as well as `unittests` in `searchsorted.rs` for documentation and usage examples.


ivarflakstad:
Hey! Thanks for the effort.

There is already a custom op example in `candle-examples`, so I think what you have here would be better suited in it's own repo.

PR #1420, opened 2023-12-09
Upsample grad

KGrewal1:
This should implement backprop for interpolate_2D for a limited set of cases (integer scale factor that is the same for both h and w). However this should be sufficient for common use cases such as unpooling images for deconvolutions (ie when in pytorch Upsample nearest would have been used).


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks for adding this!

Issue #1422, opened 2023-12-10
Whisper outputs repeating sentences


WenqingZong:
A similar [issue](https://github.com/huggingface/candle/issues/472) was raised just a few months ago, but seems it's still not fully solved now.

[failed_audios.zip](https://github.com/huggingface/candle/files/13629198/failed_audios.zip)

I've attached some audio samples in this issue to help our discussion, now `candle`'s whisper outputs repeating sentences (sometimes even incomplete words) when the audio file reaches 30s, here are the outputs for the three audio samples I attached:

(Repeat `a little bit of`, and trailing `.`)
```text
# cargo run --example whisper --release --features=cuda -- --input=prodrecall_call.wav --timestamps
   Compiling candle-examples v0.3.1 (/workspaces/candle/candle-examples)
    Finished release [optimized] target(s) in 5.61s
     Running `target/release/examples/whisper --input=prodrecall_call.wav --timestamps`
loaded wav data: Header { audio_format: 1, channel_count: 1, sampling_rate: 16000, bytes_per_second: 32000, bytes_per_sample: 2, bits_per_sample: 16 }
pcm data loaded 1913152
loaded mel: [1, 80, 13500]
0.0s -- 30.0s
  0.0s-...: . I'm sorry. I can definitely help you with that. What is your first and last name? Sure. Steve Simmons. Okay. Can you spell that for me? S. P. E. V. E. Last name is
30.0s -- 60.0s
  0.0s-...: .com. Okay. Can I also have your email address? Sure. It's Steve Simmons, just like it's spelled at gmail.com. Okay. I have Steve Simmons, S-P-E-V-E-S-I-M-M-O-N-S at tmail.com. Yes. Okay. Great. I'm going to have a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little bit of a little
60.0s -- 90.0s
[...]

PR #1424, opened 2023-12-12
Add logsumexp function

WenqingZong:
A useful tensor function which `candle` do not currently have. 

This function is needed to implement some [missing logic](https://github.com/openai/whisper/blob/e58f28804528831904c3b6f2c0e473f346223433/whisper/decoding.py#L500) of [Whisper](https://github.com/huggingface/candle/issues/1422)

Reference implementation: https://pytorch.org/docs/stable/generated/torch.logsumexp.html#torch-logsumexp


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1432, opened 2023-12-14
Phi 2 wasm 

radames:
Here the WASM demo for the new Phi-2
The weights are on a different repo, I was getting network error while fetching, so a split it into 500mb chunks
 q80 with ~2.5GB didn't load successfully, maybe a memory limit? 
got the `_name_or_path == "microsoft/phi-2"` from the model config in order to the hacked `new_v2`, we can change that later.

Demo: https://huggingface.co/spaces/radames/Candle-phi1-phi2-wasm-demo


LaurentMazare (approved):

LaurentMazare:
Very cool, thanks!

PR #1433, opened 2023-12-14
Add support for phi-2

jbochi:
Adds support for [phi-2](https://www.microsoft.com/en-us/research/blog/phi-2-the-surprising-power-of-small-language-models/)

It loads the safetensors from https://huggingface.co/SkunkworksAI/phi-2


jbochi:
Oh, never mind, this was already implemented 👍

PR #1436, opened 2023-12-14
Fix phi example

niutech:


niutech:
@LaurentMazare Can't you use `--features cuda` in Phi example?

LaurentMazare:
You can but it will only work if you have an nvidia gpu and cuda installed. For our examples, we tend to give instructions that do not assume that.

PR #1444, opened 2023-12-16
Add support for `scatter`(torch.scatter)

mokeyish:
Hi there,

This PR achieves pytorch's

https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_.html#torch.Tensor.scatter_

---

Because `apply_repeat_penalty` is a bit slow here.
https://github.com/huggingface/candle/blob/main/candle-transformers/src/utils.rs#L3

This implementation should be faster.
https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L337
It needs `scatter` operation.

### DIFF
--- a/candle-core/src/backend.rs
+++ b/candle-core/src/backend.rs
@@ -69,6 +69,15 @@ pub trait BackendStorage: Sized {
     fn upsample_nearest2d(&self, _: &Layout, _: usize, _: usize) -> Result<Self>;
 
     fn gather(&self, _: &Layout, _: &Self, _: &Layout, _: usize) -> Result<Self>;
+    fn scatter(
+        &self,
+        _: &Layout,
+        _: &Self,
+        _: &Layout,
+        _: &Self,
+        _: &Layout,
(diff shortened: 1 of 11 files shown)


mokeyish:
@LaurentMazare do you have time to review this PR.

LaurentMazare:
Couldn't we use `scatter_add` with some zeros tensors to achieve this? It would be less efficient but the idea is to keep the number of basic op as limited as possible, especially as we introduce new backends having more ops result in additional work and maintenance so would prefer avoiding it unless it's really necessary.

mokeyish:
I understand your concerns. It does work by adding a negative number first using scatter_add

But when the index is non-unique, it cannot be completely equivalent to torch.scatter.
image


I tried it and found that scatter can be implemented in [candle-ext](https://github.com/mokeyish/candle-ext/blob/main/src/scatter.rs).

Just calling copy_strided_src_ (which is `pub crate`) requires a twist.

image

PR #1449, opened 2023-12-16
Expose AdamW parameters

### LINKED ISSUE #1448: Expose betas for AdamW to use with LR Scheduler
I am working on building out my training using Candle. Has been good so far but I missed some of my LR schedulers I was using. Specifically I have been building out OneCycleLR from https://arxiv.org/abs/1708.07120. PyTorch has an implementation [OneCycleLR](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.OneCycleLR.html#torch.optim.lr_scheduler.OneCycleLR)

In OneCycleLR you modify both the LR and the momentum. For AdamW we can modify the beta1 value for the momentum. Would exposing the betas be invited? I can make a PR but proposing it first.

Here is the scheduler code I have so far [rockerBOO/candle_scheduler](https://github.com/rockerBOO/candle_scheduler). 

Thank you!

rockerBOO:
Wasn't sure how involved this should be but this is the simplest way I thought of. Open for suggestions or improvements. Thank you!

Resolves #1448

### DIFF
--- a/candle-nn/src/optim.rs
+++ b/candle-nn/src/optim.rs
@@ -191,8 +191,8 @@ impl AdamW {
         Self::new(vars, params)
     }
 
-    pub fn params(&self) -> ParamsAdamW {
-        self.params.clone()
+    pub fn params(&self) -> &ParamsAdamW {
+        &self.params
     }
 
     pub fn set_params(&mut self, params: ParamsAdamW) {
--- a/candle-nn/src/optim.rs
+++ b/candle-nn/src/optim.rs
@@ -190,4 +190,12 @@ impl AdamW {
         };
         Self::new(vars, params)
     }
+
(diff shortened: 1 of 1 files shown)


candle-nn/src/optim.rs
```diff
@@ -190,4 +190,12 @@ impl AdamW {
         };
         Self::new(vars, params)
     }
+
+    pub fn params(&self) -> ParamsAdamW {
```
LaurentMazare, line 194: What about returning a reference instead so that it's up to the user to decide whether they want to clone it or not?

rockerBOO: Thank you. Updated.

LaurentMazare:
Thanks!

PR #1452, opened 2023-12-17
Bump the crate version to 0.3.2.

LaurentMazare:


mokeyish:
@LaurentMazare It would be better to add a git tag when release new version.

PR #1453, opened 2023-12-17
Derive Clone trait on LogitsProcessor

dan-ilin:


LaurentMazare:
Could you ellaborate a bit on the use case for this to be cloneable? I'm a little afraid of cloning the rng state as this may result in unexpected behavior but maybe that's something that you would need somehow?

dan-ilin:
@LaurentMazare sure. The use case is to allow it be cloneable so that multiple threads can use their own instance of the `LogitsProcessor ` and not interfere with each other's internal state.

LaurentMazare:
But that wouldn't work well with the rng then, the threads are likely to generate the exact same thing based on the same prompt. Couldn't you create a separate instance per thread? That's likely to be less error prone.

dan-ilin:
Fair enough. I'll go with that approach.

PR #1458, opened 2023-12-19
add fn config_chat_ml

danielclough:
Models which use Mistral as a base and the ChatML template format require a different `vocab_size` than using the `config_7b_v0_1` allows.

Adding a new fn should work with all Mistral models using the ChatML template format.

I have tested this in my project and will make another PR to adjust the [mistral cli example](https://github.com/huggingface/candle/blob/main/candle-examples/examples/mistral/main.rs) work with the [Mistral-7B-OpenOrca](https://huggingface.co/Open-Orca/Mistral-7B-OpenOrca) model.

Read more about template formats here:
https://huggingface.co/docs/transformers/main/chat_templating

### DIFF
--- a/candle-transformers/src/models/mistral.rs
+++ b/candle-transformers/src/models/mistral.rs
@@ -37,6 +37,22 @@ impl Config {
             use_flash_attn,
         }
     }
+    pub fn config_chat_ml(use_flash_attn: bool) -> Self {
+        Self {
+            vocab_size: 32002,
+            hidden_size: 4096,
+            intermediate_size: 14336,
+            num_hidden_layers: 32,
+            num_attention_heads: 32,
+            num_key_value_heads: 8,
+            hidden_act: Activation::Silu,
+            max_position_embeddings: 32768,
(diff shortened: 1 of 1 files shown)


candle-transformers/src/models/mistral.rs
```diff
@@ -37,6 +37,22 @@ impl Config {
             use_flash_attn,
         }
     }
+    pub fn config_chat_ml(use_flash_attn: bool) -> Self {
+        Self {
```
LaurentMazare, line 41: Could you add a pointer to the config/model on the huggingface hub? This will make it easier to be sure what this config should be used for.

danielclough: I added a link to the config on the [model README.md](https://huggingface.co/DanielClough/Candle_Mistral-7B-OpenOrca/blob/main/README.md).

LaurentMazare:
Merged, thanks! (I slightly tweaked the name and added a link so that it's easier to trace out the original config when looking at the code)

Issue #973, opened 2023-09-27
Transfering CUDA tensors between candle and tch-rs without synchronization


ViliamVadocz:
Is there a way to transfer CUDA tensors between candle and tch-rs without requiring GPU/CPU synchronisation?

I am doing reinforcement learning research and currently I have a ResNet that I am training in tch-rs. I have found that the main bottleneck of my implementation is GPU/CPU synchronization when creating observations of an environment. To alleviate this I want to try using candle to reimplement the environment to keep everything on the GPU. Before I can do that I need to figure out whether I can avoid send CUDA tensors from candle to a network implemented in tch-rs.

---

Related questions that you may also be able to answer: How would I send tensors between candle and pytorch+jax? What about tch-rs and pytorch+jax?

ViliamVadocz:
I managed to get this to work, but I don't think it is correct - I think I need to leak the cuda slice so it does not get freed when the original candle tensor exits the scope.

```rs
use std::ops::Deref;

use candle_core as cc;
use cc::{backend::BackendDevice, cuda_backend::cudarc::driver::DevicePtr};
use thiserror::Error;

fn candle_to_tch() -> Res<()> {
    let device = cc::Device::Cuda(cc::CudaDevice::new(0)?);
    let tensor = cc::Tensor::arange(0.0f32, 96.0f32, &device)?.reshape((8, 12))?;
    println!("before:{tensor}");

    let (storage, layout) = tensor.storage_and_layout();

    // Get all the required information from the candle tensor.
    let data = match storage.deref() {
        // I think I need to leak the CudaSlice here so that it does not get freed by candle.
        // something like this, except `.leak()` requires ownership: `cuda_storage.as_cuda_slice::<f32>()?.leak()`
        cc::Storage::Cuda(cuda_storage) => *cuda_storage.as_cuda_slice::<f32>()?.device_ptr(),
        _ => unreachable!("unexpected storage type"),
    } as *mut u8;
    let size = slice_of_usize_to_slice_of_i64(layout.shape().dims());
    let strides = slice_of_usize_to_slice_of_i64(layout.stride());
    let kind = match storage.dtype() {
        cc::DType::F32 => tch::Kind::Float,
        _ => unreachable!("unexpected data type"),
    };
    let device = match storage.device() {
        cc::Device::Cuda(cuda_device) => tch::Device::Cuda(cuda_device.ordinal()),
        _ => unreachable!("unexpected device type"),
    };

[...]

PR #1467, opened 2023-12-22
Add storage_mut safely

EricLBuehler:
This PR makes the `storage_mut_and_layout` method public. This is safe because of the use of RwLock and panicking, in contrast to the previous commend regarding the addition of unsafe.

### DIFF
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -2346,6 +2346,7 @@ impl Tensor {
     ///
     /// ## Panics
     /// - If the lock was poisoned
+    /// - If the lock is held by the current thread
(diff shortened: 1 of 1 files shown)


EricLBuehler:
@LaurentMazare, what do you think about the reasoning regarding the safety?

ViliamVadocz:
I am curious what motivated this change. Are you trying to transfer tensor data between different libraries? Maybe you would be interested in this issue: https://github.com/huggingface/candle/issues/973

EricLBuehler:
Hi @ViliamVadocz , yes I am trying to get a mutable way to access the Tensor data. This is to use with cudarc, which Candle and candle-vllm both work with. I am not transferring data between libraries, this is just a visibility modification at heart.

PR #1468, opened 2023-12-23
Add config_chat_ml example.

danielclough:
This code pulls from `DanielClough/Candle_Mistral-7B-OpenOrca` instead of [`Open-Orca/Mistral-7B-OpenOrca`](https://huggingface.co/Open-Orca/Mistral-7B-OpenOrca) in order to have access to `.safetensors` files and a `tokenizer.json`.


candle-examples/examples/mistral/main.rs
```diff
@@ -225,18 +225,27 @@ fn main() -> Result<()> {
             if args.quantized {
                 vec![repo.get("model-q4k.gguf")?]
             } else {
-                vec![
-                    repo.get("pytorch_model-00001-of-00002.safetensors")?,
-                    repo.get("pytorch_model-00002-of-00002.safetensors")?,
-                ]
+                match args.model_id.as_str() {
```
LaurentMazare, line 228: If it's in your repo, maybe you could modify the name of the safetensor files so that this part is not necessary?

candle-examples/examples/mistral/README.md
```diff
@@ -88,3 +88,18 @@ fn quick_sort(arr: &mut [i32]) {
 ``
 226 tokens generated (10.91 token/s)
 ```
+
+## Config
```
LaurentMazare, line 92: I don't feel these config details are really needed in the readme, it's more about how to run the examples rather than the technical differences.

candle-examples/examples/mistral/main.rs
danielclough: At first I followed your example, but a colleague convinced me that removing the `pytorch_` prefix on `.safetensors` files is the more common practice.

Examples: 
[mistralai/Mistral-7B-v0.1](https://huggingface.co/mistralai/Mistral-7B-v0.1/commit/26bca36bde8333b5d7f72e9ed20ccda6a618af24)
[microsoft/phi-2](https://huggingface.co/microsoft/phi-2/tree/main)
[HuggingFaceH4/zephyr-7b-alpha](https://huggingface.co/HuggingFaceH4/zephyr-7b-alpha/tree/main)

Also, this [convert script hosted by safetensors](https://huggingface.co/spaces/safetensors/convert/blob/cee67d632357dcb8365ccc935698c09051d4b29b/convert.py#L127) removes the prefix.

I worry that if my work is unlike what people are used to it will cause confusion, so I prefer to conform to what seems to be the standard practice.

PR #1470, opened 2023-12-23
Add attribute test macro

mokeyish:
@LaurentMazare  I noticed that the current test macro is not very convenient. I think attribute macro would be better, I changed part of it. If you also think this is better. I'd be happy to update all unit tests to use attribute macro.

https://github.com/huggingface/candle/blob/main/candle-core/src/test_utils.rs#L5-L7
// TODO: Switch to generating the two last arguments automatically once concat_idents is
// stable. https://github.com/rust-lang/rust/issues/29599

---

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "candle-book",
     "candle-nn",
     "candle-pyo3",
+    "candle-test-macro",
     "candle-transformers",
     "candle-wasm-examples/*",
     "candle-wasm-tests",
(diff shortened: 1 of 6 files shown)


LaurentMazare:
I don't feel that it's really worth making a separate crate for this as it's a pretty small number of lines of code and the number of code is already large. That said, if convenient to you feel free to have this in some of your helper crates.

mokeyish:
I don't want to make a separate crate either, mainly procedural macros have to be like this.

https://doc.rust-lang.org/reference/procedural-macros.html#the-proc_macro-crate

PR #1476, opened 2023-12-23
Add config_chat_ml example.

danielclough:
This code pulls from `DanielClough/Candle_Mistral-7B-OpenOrca` instead of `Open-Orca/Mistral-7B-OpenOrca` in order to have access to .safetensors files and a tokenizer.json.


ivarflakstad:
I would have merged in 2023, but in 2026 I don't think these specific weights are as important to support as they were. Apologies 🙇

danielclough:
Good job catching up on the backlog. :1st_place_medal:

PR #1489, opened 2023-12-27
Add one-hot/cold encoding

Ryanmtate:
One-hot/cold encoding.

 Given an input tensor of indices, this function returns a tensor of the same shape as the input
 tensor with an additional dimension of the given depth size. The values in the returned tensor are
 all set to the `off_value` except for the positions represented by the indices, which are set to the `on_value`.

 This method returns a tensor with a rank that is one rank larger than the input tensor.

 As an example, the following tensor will be encoded to a one-hot matrix:

 `[[0i64, 2], [1, -1]]`

 with a depth of 4 will be encoded to:

 `[[[1, 0, 0, 0], [0, 0, 1, 0]], [[0, 1, 0, 0], [0, 0, 0, 0]]]`

 When the input tensor index has a value of -1, the corresponding one-hot vector will be ignored,
 resulting in a vector of values set to the `off_value`.

 This method supports one-cold encoding by setting `on_value` to `0` and `off_value` to `1`.
 By default `on_value` is `1` and `off_value` is `0`.

 Other encoding values can be used by setting `on_value` and `off_value` to the desired values.

 # Examples

 ## One-hot encoding

 ```rust
 use candle::{Shape, Tensor, Device};
 use candle_nn::encoding::one_hot;

 let device = candle::Device::Cpu;

 let indices = Tensor::new(vec![vec![0i64, 2], vec![1, -1]], &device).unwrap();
 let depth = 4;

 let on_value = 1.0;
 let off_value = 0.0;

 let one_hot = one_hot::<f32>(indices, depth, on_value, off_value).unwrap();

 let expected_matrix = [
     [[1.0, 0.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0]],
     [[0.0, 1.0, 0.0, 0.0], [0.0, 0.0, 0.0, 0.0]],
 ];

 assert_eq!(one_hot.shape(), &Shape::from((2, 2, depth)));

 let matrix = one_hot.to_vec3::<f32>().unwrap();

 assert_eq!(matrix, expected_matrix);
```
 ## One-cold Encoding

 ```rust
 use candle::{Shape, Tensor, Device};
 use candle_nn::encoding::one_hot;


 let device = candle::Device::Cpu;
 let depth = 4;
 let indices = Tensor::new(vec![vec![0u8, 2], vec![1, 3]], &device).unwrap();

 let on_value: u8 = 0;
 let off_value = 1;

 // Note: the method does not require the turbofish operator, as the type is inferred from the `on_value` and `off_value`.
 let one_cold = one_hot(indices, depth, on_value, off_value).unwrap();

 let expected_matrix = [[[0, 1, 1, 1], [1, 1, 0, 1]], [[1, 0, 1, 1], [1, 1, 1, 0]]];

 assert_eq!(one_cold.shape(), &Shape::from((2, 2, depth)));

 let matrix = one_cold.to_vec3::<u8>().unwrap();

 assert_eq!(matrix, expected_matrix);
 ```

 # Bails

 This method will bail on tensors with a rank greater than 3.

 This method will bail if the index value is less than -1.

 This method will bail if the index value is greater than or equal to the depth value.

 This method will bail if the input data type is not one of `U8`, `U32` or `I64`.

 # API Design

 The api design for this method is loosely based on the [TensorFlow One-Hot](https://www.tensorflow.org/api_docs/python/tf/one_hot) method.

### DIFF
--- a/candle-nn/src/encoding.rs
+++ b/candle-nn/src/encoding.rs
@@ -1,61 +1,4 @@
 //! Encoding Utilities. (e.g., one-hot/cold encoding)
-//!
-//! # Examples
-//!
-//! ## One-hot encoding
-//!
-//! ```rust
-//! use candle::{Shape, Tensor, Device};
-//! use candle_nn::encoding::one_hot;
-//!
-//! let device = candle::Device::Cpu;
-//!
-//! let indices = Tensor::new(vec![vec![0f32, 2.], vec![1., -1.]], &device)?;
-//! let depth = 4;
-//!
-//! let allow_f64 = true;
-//! let on_value = 1.0;
-//! let off_value = 0.0;
-//!
-//! let one_hot = one_hot::<f32>(indices, depth, allow_f64, on_value, off_value)?;
-//!
-//! let expected_matrix = [
-//!     [[1.0, 0.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0]],
-//!     [[0.0, 1.0, 0.0, 0.0], [0.0, 0.0, 0.0, 0.0]],
-//! ];
-//!
-//! assert_eq!(one_hot.shape(), &Shape::from((2, 2, depth)));
-//!
-//! let matrix = one_hot.to_vec3::<f32>()?;
-//!
-//! assert_eq!(matrix, expected_matrix);
-//!```
-//! ## One-cold Encoding
-//!
-//! ```rust
-//! use candle::{Shape, Tensor, Device};
-//! use candle_nn::encoding::one_hot;
-//!
-//!
-//! let device = candle::Device::Cpu;
-//! let depth = 4;
-//! let indices = Tensor::new(vec![vec![0i64, 2], vec![1, 3]], &device)?;
-//!
-//! let allow_f64 = false;
-//! let on_value = 0;
-//! let off_value = 1;
-//!
-//! let one_cold = one_hot::<u8>(indices, depth, allow_f64, on_value, off_value)?;
-//!
-//! let expected_matrix = [[[0, 1, 1, 1], [1, 1, 0, 1]], [[1, 0, 1, 1], [1, 1, 1, 0]]];
-//!
-//! assert_eq!(one_cold.shape(), &Shape::from((2, 2, depth)));
-//!
-//! let matrix = one_cold.to_vec3::<u8>()?;
-//!
-//! assert_eq!(matrix, expected_matrix);
-//! ```
-//!
 
 use candle::{bail, DType, Result, Tensor, WithDType};
 
@@ -96,11 +39,7 @@ use candle::{bail, DType, Result, Tensor, WithDType};
 ///
 /// let indices = Tensor::new(vec![vec![0i64, 2], vec![1, -1]], &device).unwrap();
 /// let depth = 4;
-///
-/// let on_value = 1.0;
-/// let off_value = 0.0;
-///
-/// let one_hot = one_hot::<f32>(indices, depth, on_value, off_value).unwrap();
+/// let one_hot = one_hot(indices, depth, 1f32, 0f32).unwrap();
 ///
 /// let expected_matrix = [
 ///     [[1.0, 0.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0]],
@@ -123,12 +62,7 @@ use candle::{bail, DType, Result, Tensor, WithDType};
 /// let device = candle::Device::Cpu;
 /// let depth = 4;
 /// let indices = Tensor::new(vec![vec![0u8, 2], vec![1, 3]], &device).unwrap();
-///
-/// let on_value: u8 = 0;
-/// let off_value = 1;
-///
-/// // Note: the method does not require the turbofish operator, as the type is inferred from the `on_value` and `off_value`.
-/// let one_cold = one_hot(indices, depth, on_value, off_value).unwrap();
+/// let one_cold = one_hot(indices, depth, 0u8, 1u8).unwrap();
 ///
 /// let expected_matrix = [[[0, 1, 1, 1], [1, 1, 0, 1]], [[1, 0, 1, 1], [1, 1, 1, 0]]];
 ///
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: candle-nn/src/encoding.rs [hunk 1]
```diff
@@ -0,0 +1,193 @@
...
+) -> Result<Tensor> {
+    let on_value = on_value.unwrap_or(D::from_f64(1.0));
+    let off_value = off_value.unwrap_or(D::from_f64(0.0));
+
+    let rank = indices.rank();
+    let dims = indices.dims();
+    match rank {
+        0 => {
+            let mut v = vec![off_value; depth];
+            let index = indices.to_vec0::<f32>()?;
```
LaurentMazare, line 107: In order to use the same convention as we do for embeddings, it would be good to support the integer cases here (`u8`, `u32`, and `i64` at the moment) for indices but return an error for floating point types as it's pretty easy to get into numerical issues because of this.

candle-nn/src/encoding.rs
```diff
@@ -0,0 +1,193 @@
...
+            v[index as usize] = on_value;
+
+            Tensor::new(v, indices.device())
+        }
+        1 => {
+            let mut v = vec![off_value; depth * dims[0]];
+            let index = indices.to_vec1::<f32>()?;
+            for i in 0..dims[0] {
+                if index[i] < 0.0 {
```
LaurentMazare, line 116: These errors should be caught and sent back to the user via some `bail` rather than silently swallowed.

candle-nn/src/encoding.rs
```diff
@@ -0,0 +1,193 @@
...
+                    continue;
+                }
+
+                v[(i * depth + index[i] as usize) as usize] = on_value;
```
LaurentMazare, line 120: Let's also catch the out of bounds exceptions here rather than potentially panicking (or getting into undefined behavior).

candle-nn/src/encoding.rs
```diff
@@ -0,0 +1,379 @@
...
+///
+/// This method will bail if the input data type is `f64` and `allow_f64` is false.
+///
+/// # API Design
+///
+/// The api design for this method is loosely based on the [TensorFlow One-Hot](https://www.tensorflow.org/api_docs/python/tf/one_hot) method.
+pub fn one_hot<I: WithDType, O: WithDType>(
+    indices: Tensor,
+    depth: usize,
+    allow_f64: bool,
```
LaurentMazare, line 173: Let's maybe not introduce an argument for this. If people want this to work on `f64`, they should just cast the input array with `to_dtype` (slightly less efficient but usually not a big issue).

candle-nn/src/encoding.rs
```diff
@@ -0,0 +1,379 @@
...
+/// This method will bail if the index value is greater than or equal to the depth value.
+///
+/// This method will bail if the input data type does not equal the generic data type provided.
+///
+/// This method will bail if the input data type is `f64` and `allow_f64` is false.
+///
+/// # API Design
+///
+/// The api design for this method is loosely based on the [TensorFlow One-Hot](https://www.tensorflow.org/api_docs/python/tf/one_hot) method.
+pub fn one_hot<I: WithDType, O: WithDType>(
```
LaurentMazare, line 170: Avoid the template with types in the external facing code, at least for the input dtype here we would rather handle internall the 3 cases `u8`, `u32`, and `i64` and bail out on everything else. Same as what we're currently doing on embeddings.

candle-nn/src/encoding.rs
```diff
@@ -0,0 +1,379 @@
...
+            }
+
+            Tensor::new(v, indices.device())
+        }
+        1 => {
+            let dim1 = indices.dims1()?;
+            let mut v = vec![off_value; depth * dim1];
+            let index = indices.to_vec1::<I>()?;
+            for i in 0..dim1 {
+                let vi = index[i].to_f64();
```
LaurentMazare, line 212: Same comment as earlier, all this code should be tweaked to manipulate integers rather than floats for indexes so as to avoid numerical precision issues (even if they are impossible to happen for small indexes values).

candle-nn/src/encoding.rs
```diff
@@ -0,0 +1,555 @@
...
+//!
+//! let matrix = one_cold.to_vec3::<u8>()?;
+//!
+//! assert_eq!(matrix, expected_matrix);
+//! ```
+//!
+
+use candle::{bail, DType, Result, Tensor, WithDType};
+
+const INVALID_ONE_HOT_INDEX_MSG: &str =
```
LaurentMazare, line 62: This const is only used once, probably better to inline it.

candle-nn/src/encoding.rs
```diff
@@ -0,0 +1,555 @@
...
+// Bail if the data type is not supported.
+fn unsupported_dtype(dtype: DType) -> Result<()> {
+    bail!(
+        "one_hot: Unsupported data type. Expected U8, U32, or I64. Received {:?}",
+        dtype
+    );
+}
+
+// Set unsigned integer index values to the given value.
+fn set_uint_value<D: WithDType>(
```
LaurentMazare, line 386: This seems to be a special case of `set_int_value` below. Could we simplify things by just keeping the second?

candle-nn/src/encoding.rs
```diff
@@ -0,0 +1,555 @@
...
+///
+/// This method will bail if the input data type is `f64`.
+///
+/// # API Design
+///
+/// The api design for this method is loosely based on the [TensorFlow One-Hot](https://www.tensorflow.org/api_docs/python/tf/one_hot) method.
+pub fn one_hot<D: WithDType>(
+    indices: Tensor,
+    depth: usize,
+    on_value: Option<D>,
```
LaurentMazare, line 170: Let's have mandatory parameters here, `D::from_f64` is a bit flaky and typing `1` is shorter than `None`.

candle-nn/src/encoding.rs
```diff
@@ -0,0 +1,555 @@
...
+        }
+        2 => {
+            let (dim1, dim2) = indices.dims2()?;
+            let mut v = vec![off_value; depth * dim1 * dim2];
+
+            let idx = |i: usize, j: usize, depth: usize, dim2: usize| -> usize {
+                i * depth * dim2 + j * depth
+            };
+
+            let iter = (0..dim1).flat_map(|i| (0..dim2).map(move |j| (i, j)));
```
LaurentMazare, line 237: Is the `move` really necessary here? I would have thought that we don't care as `usize` is `Copy` (I think the same comment applies at other places).

LaurentMazare:
Merged, thanks.

candle-nn/src/encoding.rs
```diff
@@ -165,63 +97,57 @@ pub fn one_hot<D: WithDType>(
...
+                    let indices = indices.to_vec1::<i64>()?;
+                    for (i, &index) in indices.iter().enumerate() {
+                        set_usize_value(index as usize, i * depth, depth, &mut v, on_value)?;
                     }
                 }
                 DType::U32 => {
-                    let index = indices.to_vec1::<u32>()?;
-                    for i in 0..dim1 {
-                        set_usize_value(index[i] as usize, i * depth, depth, &mut v, on_value)?;
+                    let indices = indices.to_vec1::<i64>()?;
```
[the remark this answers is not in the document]
LaurentMazare: This has been fixed in a subsequent PR that also removes all the duplication, #1514

PR #1491, opened 2023-12-28
Improves metal's not implemented error messages

grzuy:
Hi,

Thanks for adding Metal support!

Been trying it out and updating these error messages helped debug and understand what's supported and what not yet.

Hope it help others too.


Narsil (approved):
Thanks for all of those !

PR #1492, opened 2023-12-28
Metal: Removing fences as they seem to not be needed.

ivarflakstad:
Updating/waiting for fences was initially added to help tackle GPU race condition bugs. However I think that the bug was fixed and that these fences are no longer needed.

@Narsil what do you think? Seems to work perfectly on the m1 pro (faster too)

### DIFF
--- a/candle-core/src/metal_backend.rs
+++ b/candle-core/src/metal_backend.rs
@@ -1367,7 +1367,7 @@ impl BackendDevice for MetalDevice {
         let command_buffer = Arc::new(RwLock::new(command_buffer));
         let command_buffer_index = Arc::new(RwLock::new(0));
         let fence = device.new_fence();
-        let kernels = Arc::new(Kernels::new(fence.clone()));
+        let kernels = Arc::new(Kernels::new());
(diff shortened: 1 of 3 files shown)


Narsil:
No it's keeps failing on m3. You can repro with my gist

ivarflakstad:
> No it's keeps failing on m3. You can repro with my gist https://gist.github.com/Narsil/13bfeb66959a0b7e2750a23be9ecc957.
> 
I get the following - I assume you do not?
```
First matmul is OK [2.0, 3.0, 6.0, 11.0, 46.0, 55.0, 66.0, 79.0]
This should be filled [6.0, 11.0, 22.0, 39.0, 514.0, 615.0, 738.0, 883.0]
```

> Also how did you test and what did you measure. On m3 the fence is quite negligible (as long as the scheduler is kept happy by leaving a few critical ones)

Ran some models and the difference was significant. However I just did a reboot and the difference is gone 🤷
Guess I'll close

ivarflakstad:
Reopening as it seems the fence situation has changed for M3.
Is that correct, @Narsil ?😊

PR #1496, opened 2023-12-28
Implement urecip op for metal backend

bayedieng:
the reciprocal operation is necessary to run stable diffusion xl turbo.


candle-core/src/metal_backend.rs
```diff
@@ -648,6 +648,7 @@ impl BackendStorage for MetalStorage {
                 ("uceil", DType::F32) => contiguous::ceil::FLOAT,
                 ("ufloor", DType::F32) => contiguous::floor::FLOAT,
                 ("uround", DType::F32) => contiguous::round::FLOAT,
+                ("urecip", DType::F32) => contiguous::round::FLOAT,
```
Narsil, line 651:
```suggestion
                ("urecip", DType::F32) => contiguous::recip::FLOAT,
```

candle-core/src/metal_backend.rs
```diff
@@ -662,6 +663,7 @@ impl BackendStorage for MetalStorage {
                 ("uceil", DType::F16) => contiguous::ceil::HALF,
                 ("ufloor", DType::F16) => contiguous::floor::HALF,
                 ("uround", DType::F16) => contiguous::round::HALF,
+                ("urecip", DType::F16) => contiguous::round::HALF,
```
Narsil, line 666:
```suggestion
                ("urecip", DType::F16) => contiguous::recip::HALF,
```

candle-metal-kernels/src/unary.metal
```diff
@@ -19,7 +19,9 @@ METAL_FUNC uint get_strided_index(
 }
 
 template <typename T> METAL_FUNC T sqr(T in){ return in * in; }
+template <typename T> METAL_FUNC T urecip(T in){ return T(1.0 / in); }
```
Narsil, line 22: This is the inverse op, why call it urecip ?

`inv` seems like a better name. (`u` is not necessary)

bayedieng: Candle names this op recip. I could change all instances to inv as i also agree it's a better name but I the Rust library itself names it recip so that would be more consistent. Either name works for me.

Narsil (approved):
LGTM.

Let's change in another PR (quite low prio imo)

PR #1501, opened 2023-12-29
Metal fill kernel

ivarflakstad:
Still uses blit command encoder for zeroes and all u8 values.

### DIFF
--- a/candle-core/src/metal_backend.rs
+++ b/candle-core/src/metal_backend.rs
@@ -3,7 +3,8 @@ use crate::conv::{ParamsConv1D, ParamsConv2D, ParamsConvTranspose1D, ParamsConvT
 use crate::op::{BinaryOpT, CmpOp, ReduceOp, UnaryOpT};
 use crate::{CpuStorage, DType, Layout, Result, Shape};
 use candle_metal_kernels;
-use candle_metal_kernels::Kernels;
+use candle_metal_kernels::{CallFill, Fill, Kernels};
+use half::{bf16, f16};
 use metal;
 use metal::{Buffer, CommandBuffer, CommandQueue, MTLResourceOptions, NSUInteger};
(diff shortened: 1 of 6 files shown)


Narsil:
Is there any benefit to using a CPU alloc and simply doing a memcpy ? Do you have benchmarks for it ?

Also keep the blit for zeros all the way. Remove the trait follow the rest of the crates convention.
Don't add any dependencies, they are not necessary.

ivarflakstad:
It's a tiny bit faster, yeah 😉 

```shell
fill_metal_u8/iter      time:   [101.33 µs 101.64 µs 101.98 µs]
                        thrpt:  [153.22 GiB/s 153.73 GiB/s 154.20 GiB/s]
                 change:
                        time:   [-96.183% -96.117% -96.019%] (p = 0.00 < 0.05)
                        thrpt:  [+2411.8% +2475.3% +2520.1%]

fill_metal_f32/iter     time:   [409.20 µs 437.94 µs 467.82 µs]
                        thrpt:  [133.60 GiB/s 142.71 GiB/s 152.74 GiB/s]
                 change:
                        time:   [-97.242% -97.113% -96.969%] (p = 0.00 < 0.05)
                        thrpt:  [+3199.1% +3364.0% +3525.2%]
                        Performance has improved.
```

> Also keep the blit for zeros all the way.
That's what I'm doing 👍

ivarflakstad:
> Don't add any dependencies, they are not necessary.

If we want to be able to encode `f16` and `bf16` values into the kernel we kinda need it.
Alternatively we could send in the kernel name and the bits as u16 I guess?

ivarflakstad:
Forgot to merge ages ago. Thankfully the same functionality has been implemented in https://github.com/huggingface/candle/pull/2512 👍

PR #1504, opened 2023-12-30
[Breaking] Add training to batchnorm with exponential moving average

nkoppel:
Ports the [dfdx implementation](https://github.com/coreylowman/dfdx/blob/main/dfdx/src/nn/layers/batch_norm1d.rs) of batch norm training to candle, which unfortunately requires breaking changes. 

- Changes the type of `running_mean` and `running_var` from `Tensor` to `Var`. I'm not sure if this is an intended use of `Var`, but it seemed to be the best fit for this use case.
- Removes `num_features` parameter and adds `momentum` parameter to `new` and `new_no_bias` methods. The `num_features` parameter is redundant with the size of the given tensors.
- Adds a `forward_t` method that shadows the `ModuleT` implementation. This makes batchnorm more intuitive in a `ModuleT` context, but could create confusion or have unexpected behavior for some use cases. Without this, however, `forward_t(..., true)` calls `forward` instead of `forward_learning`, so I'm not really sure of the trade-off here.

### DIFF
--- a/candle-nn/src/batch_norm.rs
+++ b/candle-nn/src/batch_norm.rs
@@ -7,8 +7,8 @@
 //! running stats.
 //!
 //! [`Batch Normalization`]: https://arxiv.org/abs/1502.03167
-use candle::{DType, Module, Result, Tensor, Var};
 use crate::Init;
+use candle::{DType, Module, Result, Tensor, Var};
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct BatchNormConfig {
@@ -69,10 +69,30 @@ impl BatchNorm {
         if eps < 0. {
             candle::bail!("batch-norm eps cannot be negative {eps}")
         }
-        if weight.dims() != bias.dims() {
+        if running_mean.dims().len() != 1 {
             candle::bail!(
-                "batch-norm weight shape {:?} does not match bias shape {:?}",
+                "batch-norm running mean must have 1 dimension, has {}",
+                running_mean.dims().len()
+            )
+        }
+        if running_mean.dims() != running_var.dims() {
+            candle::bail!(
+                "batch-norm running mean shape {:?} does not match running variance shape {:?}",
+                running_mean.shape(),
+                running_var.shape(),
+            )
+        }
+        if running_mean.dims() != weight.dims() {
+            candle::bail!(
+                "batch-norm running mean shape {:?} does not match weight shape {:?}",
+                running_mean.shape(),
                 weight.shape(),
+            )
+        }
+        if running_mean.dims() != bias.dims() {
+            candle::bail!(
+                "batch-norm running mean shape {:?} does not match bias shape {:?}",
+                running_mean.shape(),
                 bias.shape(),
             )
         }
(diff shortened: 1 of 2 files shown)


candle-nn/src/batch_norm.rs
```diff
@@ -142,26 +184,51 @@ impl BatchNorm {
         let x = x.flatten_from(1)?.contiguous()?;
         let x = if self.remove_mean {
             let mean_x = x.mean_keepdim(1)?;
+            {
+                // Update running mean
+                let new_mean = ((self.running_mean.as_tensor() * (1.0 - self.momentum))?
+                    + (mean_x.reshape(((),))? * self.momentum)?)?;
```
LaurentMazare, line 190: `flatten_all`?

candle-nn/src/batch_norm.rs
```diff
@@ -34,79 +42,108 @@ impl From<f64> for BatchNormConfig {
...
         weight: Tensor,
         bias: Tensor,
         eps: f64,
+        momentum: f64,
     ) -> Result<Self> {
         if eps < 0. {
             candle::bail!("batch-norm eps cannot be negative {eps}")
         }
-        if weight.dims() != [num_features] {
+        if running_mean.dims().len() != 1 {
```
LaurentMazare, line 72: `rank`?

candle-nn/src/batch_norm.rs
```diff
@@ -34,79 +42,108 @@ impl From<f64> for BatchNormConfig {
...
 
 impl BatchNorm {
     pub fn new(
-        num_features: usize,
         running_mean: Tensor,
         running_var: Tensor,
         weight: Tensor,
         bias: Tensor,
         eps: f64,
+        momentum: f64,
```
LaurentMazare, line 67: Let's have a separate function to create a batch-norm with momentum and use the default value here so as not to break the current api (also most use cases don't care much about the momentum value anyway).

candle-nn/src/batch_norm.rs
```diff
@@ -142,26 +184,51 @@ impl BatchNorm {
...
+            }
             x.broadcast_sub(&mean_x)?
         } else {
             x
         };
         let norm_x = x.sqr()?.mean_keepdim(1)?;
+        {
+            // Update running variance
+            let running_var_weight = 1.0 - self.momentum;
+            let norm_x_weight = self.momentum * num_features as f64 / (num_features as f64 - 1.0);
```
LaurentMazare, line 202: I'm a bit confused by this, shouldn't this be the batch-size rather than the number of features to get the unbiased estimator?
But that would also mean that the two `mean_keepdim` above should be done on the first dimension (batch dim) rather than on the second (channels)?
What testing did you run on this?

candle-nn/src/batch_norm.rs
```diff
@@ -34,79 +42,108 @@ impl From<f64> for BatchNormConfig {
             eps,
             remove_mean: true,
             affine: true,
+            momentum: 0.1,
         }
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
```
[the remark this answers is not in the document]
nkoppel: This was removed due to an earlier effort to implement this with `Rwlock<Tensor>` instead of `Var`, since I don't know candle all that well yet.

candle-nn/src/batch_norm.rs
nkoppel: Okay, I've added `with_momentum` variants of the `new` methods. What are your thoughts on my removing of the `num_features` parameter from the `new` methods? It does simplify things, but is still breaking. These changes will break some code due to the changes to `BatchNormConfig`.

candle-nn/src/batch_norm.rs
nkoppel: This should be resolved now, and I have added to the batchnorm test to verify the new behavior against pytorch.

candle-nn/src/batch_norm.rs
LaurentMazare: Would prefer not breaking the api and so adding `num_features` back.

nkoppel: I've added it back, and modified `check_validity` to use it.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1507, opened 2023-12-30
Redesign Module/ModuleT traits (Implements #1506)

### LINKED ISSUE #1506: [Proposal] [candle_nn] Redesign Module/ModuleT traits
While implementing #1504, I noticed that the the blanket implementation of ModuleT for Modules makes it awkward to create different behavior between training and learning. For a module to use the train parameter of the ModuleT trait, it cannot implement Module, because doing so will cause a "conflicting implementations" error at compile time. I believe that removing the ModuleT trait and adding the `forward_t` method to the Module trait can make this api easier to use.
```rust
trait Module {
    fn forward(&self, xs: &Tensor) -> Result<Tensor>;
    fn forward_t(&self, xs: &Tensor, _train: bool) -> Result<Tensor> {
        self.forward(xs)
    }
}
```
Note that the above implementation is nearly the same to the end user but with more flexibility. Implementations of `forward` only would look exactly the same, while implementations of `forward_t` only would look like the following:
```rust
impl Module for Mlp {
        self.forward_t(xs, false)
    }

    fn forward_t(&self, xs: &Tensor, train: bool) -> Result<Tensor> {
        // ... long, complicated implementation passing 'train' to submodules
    }
}
```
We could even make the implementation of `forward` above the default implementation, though this could lead to users doing `impl Module for Mlp {}` and creating infinite recursion.

These changes would not be difficult to implement, but I understand if they need to wait until a major/minor release.

nkoppel:
Closes #1506. Turned out to be a really simple implementation. Is breaking and still needs documentation.


nkoppel:
Closing for now, as @LaurentMazare gave pretty good justification for keeping `ModuleT` around.

PR #1517, opened 2024-01-03
fix index_pos bug when kv cache is disabled.

optman:


LaurentMazare:
Thanks, I've sligthly tweaked the fix to avoid `index_pos` having different meanings depending on whether the kv cache is active or not.

PR #1518, opened 2024-01-03
chore: update flash attention kernels

OlivierDehaene:
Update flash attention kernels to support windowing and alibi slopes.


Narsil (approved):
LGTM.

I'm not sure about the ever increasing number of functions although I do understand the ergonomics make it less likely to shoot oneself in the foot (like window_right_size=0 for causal).

How are you testing ?

candle-flash-attn/kernels/flash_api.cu
```diff
@@ -106,6 +109,19 @@ extern "C" void run_mha(
     params.cu_seqlens_q = cu_seqlens_q_ptr;
     params.cu_seqlens_k = cu_seqlens_k_ptr;
     params.p_ptr = nullptr; // used for `return_softmax`.
+    params.seqused_k = nullptr;
+
+    // Causal is the special case where window_size_right == 0 and window_size_left < 0.
```
LaurentMazare, line 114: I would rather have this `is_causal` logic live on the rust side rather than here, this way it will be easier to tweak without having to recompile the cuda bits (and also it will benefit from being in rust and this would just be an ffi layer without any smartness into it).

OlivierDehaene:
@Narsil we can do a breaking change instead.
I'm testing with Jina in TEI directly.

candle-flash-attn/kernels/flash_api.cu
OlivierDehaene: done

PR #1523, opened 2024-01-04
Quantized GGUF style

Narsil:
Working quantized state for candle.

High level overview:

- Introduce QStorage to split Metal and Cpu ops.
- Quantize/Dequantize still running on CPU even when asked to run on metal. Kernels exist but led to different quantization/dequantization. GGML removed those kernels on device too (I guess because of the differences).
  I think we should keep the surface for on metal quantize/dequantize so we can easily implement them later. They *are* part of the ggml API imho.
- Added a bunch of test, using `test_device!` in order to get similar testing behavior as regular `Tensor`.
- `quantized.metal` is a direct copy of ggml's `ggml-metal.metal`. This choice was made so further dev could be made faster and bugs mentionned after can be imported more easily. All the glue logic is in `candle_metal_kernels`.
- Introduces a new GgmlDType in `candle_metal_kernels`. Ggml uses different kernels based on size of matmul and hardware capacity. This wasn't implemented here, but could with the current API.

Worthy bug already discovered (not fixed in this PR since they do not belong here):
   - Q2K Metal -> Bugged (also present in GGML). 
   - Q4K CPU -> Bugged (present reviously, new test catch it). 
   - Q5K CPU -> Bugged (present previously).
   - Q8_1 Both -> Never really implemented it seems
   - Q8K metal -> Never implemented in metal


LaurentMazare:
Thanks for making this PR. It's really large and very hard to review and it seems to contain a bunch of orthogonal things.
Could you start by making a minimal PR that adds a device to the quantized var builder and doesn't do much besides this? This sounds like the natural first step and should be a reasonably small thing to look at before making the rest of the changes.

candle-core/src/tensor.rs
```diff
@@ -426,7 +426,9 @@ impl Tensor {
         if buffer_size != shape.elem_count() {
             return Err(Error::ShapeMismatch { buffer_size, shape }.bt());
         }
+        // println!("from vec {buffer_size}");
```
FL33TW00D, line 429: Probably want these gone

candle-examples/examples/custom-ops/cuda_kernels.rs
```diff
@@ -1 +0,0 @@
-pub const LAYERNORM_KERNELS: &str = include_str!(concat!(env!("OUT_DIR"), "/layernorm_kernels.ptx"));
```
FL33TW00D, line 1: Intended?

candle-examples/examples/phi/main.rs
```diff
@@ -299,21 +299,26 @@ fn main() -> Result<()> {
...
     };
-    let (model, device) = if args.quantized {
-        let vb = candle_transformers::quantized_var_builder::VarBuilder::from_gguf(&filenames[0])?;
+    let device = candle_examples::device(args.cpu)?;
+    let model = if args.quantized {
+        let vb = candle_transformers::quantized_var_builder::VarBuilder::from_gguf(
+            &filenames[0],
+            &device,
+        )?;
+        println!("Loaded vb");
```
FL33TW00D, line 308: Missed

candle-examples/examples/quantized/main.rs
```diff
@@ -369,23 +370,24 @@ fn main() -> anyhow::Result<()> {
...
         Some("ggml" | "bin") | Some(_) | None => {
-            let model = ggml_file::Content::read(&mut file).map_err(|e| e.with_path(model_path))?;
+            let model = ggml_file::Content::read(&mut file, &device)
+                .map_err(|e| e.with_path(model_path))?;
             let mut total_size_in_bytes = 0;
             for (_, tensor) in model.tensors.iter() {
                 let elem_count = tensor.shape().elem_count();
                 total_size_in_bytes +=
-                    elem_count * tensor.dtype().type_size() / tensor.dtype().blck_size();
+                    elem_count * tensor.dtype().type_size() / tensor.dtype().block_size();
```
FL33TW00D, line 390: Could make this a method?

candle-metal-kernels/src/cast.metal
```diff
@@ -86,7 +86,6 @@ CAST(cast_i64_f32, cast_i64_f32_strided, int64_t, float)
 #endif
 
 #if defined(__HAVE_BFLOAT__)
-#if __METAL_VERSION__ >= 310
```
FL33TW00D, line 89: I think we still need this

candle-kernels/src/lib.rs
```diff
@@ -1,9 +1 @@
-pub const AFFINE: &str = include_str!(concat!(env!("OUT_DIR"), "/affine.ptx"));
-pub const BINARY: &str = include_str!(concat!(env!("OUT_DIR"), "/binary.ptx"));
-pub const CAST: &str = include_str!(concat!(env!("OUT_DIR"), "/cast.ptx"));
-pub const CONV: &str = include_str!(concat!(env!("OUT_DIR"), "/conv.ptx"));
-pub const FILL: &str = include_str!(concat!(env!("OUT_DIR"), "/fill.ptx"));
-pub const INDEXING: &str = include_str!(concat!(env!("OUT_DIR"), "/indexing.ptx"));
-pub const REDUCE: &str = include_str!(concat!(env!("OUT_DIR"), "/reduce.ptx"));
-pub const TERNARY: &str = include_str!(concat!(env!("OUT_DIR"), "/ternary.ptx"));
-pub const UNARY: &str = include_str!(concat!(env!("OUT_DIR"), "/unary.ptx"));
```
FL33TW00D, line 9: Intended?

candle-metal-kernels/src/lib.rs
```diff
@@ -1567,7 +1522,176 @@ pub fn call_upsample_nearest_2d(
...
     encoder.dispatch_thread_groups(thread_group_count, thread_group_size);
-    encoder.update_fence(&kernels.fence);
+    encoder.end_encoding();
+
+    Ok(())
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum GgmlDType {
+    Q4_0,
```
FL33TW00D, line 1532: https://github.com/rustformers/llm use an FFI here to keep in line, something to consider for the future.

candle-metal-kernels/src/lib.rs
```diff
@@ -1567,7 +1522,176 @@ pub fn call_upsample_nearest_2d(
...
+            nb10,
+            nb11,
+            nb12,
+            ne0,
+            ne1,
+            r2,
+            r3
+        )
+    );
+    encoder.set_threadgroup_memory_length(0, 8192);
```
FL33TW00D, line 1689: 8192 is fine, could introduce a version check for this in future: https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf

candle-metal-kernels/src/quantized.metal
```diff
@@ -0,0 +1,5107 @@
...
+            break;
+        }
+        dst[i+1][k] = -INFINITY;
+        if (i00 + k > n_past + i01) {
+            dst[i][k] = -INFINITY;
+        }
+    }
+}
+
+kernel void kernel_norm(
```
FL33TW00D, line 571: We can go faster ;) 

https://fleetwood.dev/posts/layernorm-as-fast-as-possible

LaurentMazare:
Ping on the refactoring into a first smaller PR as mentioned above. #1534 was actually doing something like this but not sure why it was closed.

candle-metal-kernels/src/lib.rs
Narsil: Why would we need FFI ? This is purely for internal use to call the correct kernel with the correct warps. (Each kernel is written with specific warps in mind)

candle-metal-kernels/src/lib.rs
Narsil: Yes, this is copy-pasted from ggml for now. And the kernels expect 8192.

candle-metal-kernels/src/quantized.metal
Narsil: This is not used :)  If you're up for it, we could maybe discuss implementing something like this: https://github.com/huggingface/candle-rotary  (But changing the implementation around so we could add metal backend in it).

candle-kernels/src/lib.rs
Narsil: Oops nice catch I must have messed up the build.

Narsil:
@LaurentMazare have you looked at the PR ? Aside from the gigantic file (copy pasted from GGML without any modification)  the PR is actually quite small.

The only real logic is in candle_metal_kernels::call_ggml_gemm and it's also relatively small (it's just small logic to adapt the warps for the kernels as they are written in GGML file)

LaurentMazare:
I have looked at it indeed (and a bit in details), hence my suggestion to split the new kernels aside from a simple PR that would just add the device as a parameter for the quantized var store, it would make it a lot easier to read. Also there seems to be some other unrelated changes that I would prefer to see in isolation.

candle-core/src/tensor.rs
Narsil: Thanks

candle-examples/examples/quantized/main.rs
Narsil: It is a method, I don't get what you are implying here ?

candle-examples/examples/custom-ops/cuda_kernels.rs
Narsil: Hmm building without cuda will do that it seems.

candle-examples/examples/quantized/main.rs
FL33TW00D: In this file we do ` elem_count * tensor.dtype().type_size() / tensor.dtype().block_size();`
Could be a method: `tensor.size_in_bytes`

Narsil: Maybe, but let's no do this in this PR though I think, no ?

Narsil:
Here it is https://github.com/huggingface/candle/pull/1594/files 

I closed the previous "small" PR because I had to switch a lot of things around while actually implementing. Mostly because `GgmlType` (the blocks) made too many assumptions of CPU execution, so I had to move to `QStorage` to put the enum before the block logic, so I could keep the CPU code mostly as-is while enabling the GPU code to work.

That also means moving away for the generics calls ( I benched on my computer, it doesn't seem to make such a difference.)

Narsil:
Okay I'm going to go ahead and merge this.

It has been reviewed by @FL33TW00D. @LaurentMazare all your comments on this PR and the smaller one are just about splitting PRs, not really some conceptual problem with it in any way, therefore I'll assume everything is actually mostly ok, and we can just fix whatever is wrong in other PRs if needed.

LaurentMazare:
Please don't merge things like this without adressing my review comments first.

PR #1524, opened 2024-01-05
feat(metal/bf16): additional cast support for bf16

kyle-mccarthy:
- extended bf16 cast support and testing
- added tests for bf16 binary ops
- added `is_u32_bf16` and `is_u8_bf16`
- exposed new + some existing kernels to candle-core's metal backend

### DIFF
--- a/candle-core/Cargo.toml
+++ b/candle-core/Cargo.toml
@@ -48,4 +48,3 @@ metal = ["dep:metal", "dep:candle-metal-kernels"]
 [[bench]]
 name = "bench_main"
 harness = false
-
--- a/candle-core/src/metal_backend.rs
+++ b/candle-core/src/metal_backend.rs
@@ -590,14 +590,26 @@ impl BackendStorage for MetalStorage {
                 (DType::U32, DType::F32) => "cast_u32_f32",
                 (DType::U32, DType::U8) => "cast_u32_u8",
                 (DType::U32, DType::I64) => "cast_u32_i64",
+                (DType::U32, DType::BF16) => "cast_u32_bf16",
+
                 (DType::U8, DType::U32) => "cast_u8_u32",
                 (DType::U8, DType::F32) => "cast_u8_f32",
                 (DType::U8, DType::I64) => "cast_u8_i64",
+                (DType::U8, DType::BF16) => "cast_u8_bf16",
+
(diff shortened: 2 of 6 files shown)


candle-core/Cargo.toml
```diff
@@ -48,4 +48,4 @@ metal = ["dep:metal", "dep:candle-metal-kernels"]
 [[bench]]
 name = "matmul"
 harness = false
-
+required-features = ["metal"]
```
LaurentMazare, line 51: I don't think we want metal by default here.

kyle-mccarthy: The [benchmark calls new_metal](https://github.com/huggingface/candle/blob/fa3ea98ba92835960fdd825a5b4dda30ef2baaa4/candle-core/benches/matmul.rs#L15) which is [only available when the metal feature is enabled](https://github.com/huggingface/candle/blob/fa3ea98ba92835960fdd825a5b4dda30ef2baaa4/candle-core/src/lib.rs#L92-L94).

ivarflakstad: Yes, and we'll have to fix that.
I have a branch locally where I conditionally load the correct benchmarks from different modules via `cfg`.
I can make a PR tomorrow.

kyle-mccarthy: Okay I figured that was the original intention. FWIW, rust-analyzer treats it as an error currently.

> error[E0599]: no method named `wait_until_completed` found for reference `&MetalDevice` in the current scope
  --> candle-core/benches/matmul.rs:31:24
   |
31 |                 device.wait_until_completed().unwrap();
   |                        ^^^^^^^^^^^^^^^^^^^^ method not found in `&MetalDevice`

ivarflakstad: Something like this?
https://github.com/huggingface/candle/pull/1538

Narsil: I kind of agree with should keep this out. Also the bench will be useful for other targets other than metal.

kyle-mccarthy: okay it has been removed from the PR!

kyle-mccarthy:
Is there any interest in merging PR? If so, I can fix the existing merge conflicts.

Narsil:
ofc it's interesting.

PR #1526, opened 2024-01-05
Potential bug on q4k (also affects q5k most likely).

Narsil:
Matmul error is > 11% (should be much smaller).

Test only affects neon target (Hence passing tests here)


LaurentMazare:
There was indeed an overflow bug in the neon code that would only rarely trigger, I've merged #1544 that contains both a simple repro (testing the optimized and non-optimized versions of the vec-dot operation) as well as a fix.

Narsil:
Super nice, I'll discard this then

PR #1529, opened 2024-01-06
[Proposal] Simplifying our internal cargo dependencies.

Narsil:
-> Move all our packages inside the workspace dependencies, so all folders (except the packages not in the workspace) can just use `workspace = true`. That removes the version number from a lot of places, so there is a single number for the entire workspace in the root `Cargo.toml`.

Packages outside the workspace (flash, kernels, metal-kernels, onnx) still use path dependencies.


LaurentMazare (approved):

LaurentMazare:
Looks certainly simpler than using sed as I was doing for the updates but the versions are also not "pinned" anymore in the manifest compared to the old version. That said I'm not sure there was an upside to doing so so let's merge this and see how it goes :)

PR #1531, opened 2024-01-06
Moving to a proper build crate `bindgen_cuda`.

Narsil:
There are now multiple crates in the candle ecosystem like https://github.com/huggingface/candle-rotary which import the entire `build.rs`.

This PR aims to fix that by using a newly created lib `bindgen_cuda` which enables all the previous behavior yet sharing most of the custom logic in a single place.

Biggest notable change is that `bindgen_cuda` is not linked itself to candle, therefore all candle mentions and special environement variables have been changed to remove the CANDLE prefix.

The `CANDLE_FLASH_ATTN_BUILD_DIR` variable has been kept here for backward compatibility.
The change only affects `CANDLE_NVCC_BIN` which has been renamed `NVCC_BIN`.


LaurentMazare (approved):

LaurentMazare:
Neat, thanks!

PR #1534, opened 2024-01-07
Make sure that when loading quantized model have the option to select…

ltouati:
… device. This will help in the future implementing proper quantization support in candle


candle-core/Cargo.toml
```diff
@@ -48,4 +48,4 @@ metal = ["dep:metal", "dep:candle-metal-kernels"]
 [[bench]]
 name = "matmul"
 harness = false
-
+required-features= ["metal"]
```
ltouati, line 51: I've added that otherwise cargo build --all does not build on linux

Narsil:
This PR should fix it; https://github.com/huggingface/candle/pull/1538 (The idea is not to force the feature for the bench, but enable the bench on any feature)

Narsil:
Closing since #1538 landed and fixes that issue.

Issue #1532, opened 2024-01-07
[Examples] yi running with gpu enabled gives an error loading function: "Function 'cast_bf16_f32' does not exist"


fankaiLiu:
**Describe the bug**
yi running with gpu enabled gives an error loading function: "Function 'cast_bf16_f32' does not exist"
**To Reproduce**
1. cargo run --release --features metal  --example yi -- --prompt "hi"

**Expected behavior**
Normal Output Results

**Screenshots**
``` shell
avx: false, neon: true, simd128: false, f16c: false
temp: 0.00 repeat-penalty: 1.10 repeat-last-n: 64
retrieved the files in 583ns
Error: Metal error Error while loading function: "Function 'cast_bf16_f32' does not exist"
```
**Desktop (please complete the following information):**
 - OS: mac,m1
 - Version:latest main

**Additional context**
It works fine with just the cpu.

ltouati:
I think PR #1524 need to be merged to support Metal inference of models that use Bf16

ivarflakstad:
`cast_bf16_f32` already exists [here](https://github.com/huggingface/candle/blob/main/candle-metal-kernels/src/cast.metal#L61-L64), which begs the question - does your computer have metal version > 3.10?
You can run `xcrun metal --version` to check 😊

I'm downloading the model for the example now (slow connection). I'll report back if it runs on my machine 👍

fankaiLiu:
> `cast_bf16_f32` already exists [here](https://github.com/huggingface/candle/blob/main/candle-metal-kernels/src/cast.metal#L61-L64), which begs the question - does your computer have metal version > 3.10? You can run `xcrun metal --version` to check 😊
> 
> I'm downloading the model for the example now (slow connection). I'll report back if it runs on my machine 👍
> 
> Edit: I do not get the error reported above, so it should be related to the metal version.

Thanks for the validation!
here is the result of my execution of xcrun metal --version, it looks to be version 3.10 or higher, but unfortunately I still get the message that the function in question can't be found when I run it.
```  shell
xcrun metal --version                                                 
Apple metal version 31001.720 (metalfe-31001.720.3)
Target: air64-apple-darwin23.1.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/metal/macos/bin
```

cargo run --release --features metal  --example yi -- --prompt "hi"
``` shell
[...]

PR #1540, opened 2024-01-07
Metal: use __HAVE_BFLOAT__ to check for bfloat support instead of metal version

ivarflakstad:


Narsil:
Could you maybe mention why we need to change that ?

Metal spec specify that bfloat is supported for metal >= 3.1 (https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf look for bfloat).

Afaik, the header you shared in undocumented and the spec focus on using __METAL_VERSION__ instead

ivarflakstad:
In this issue [comment](https://github.com/huggingface/candle/issues/1532#issuecomment-1880077926) a user reports not having bfloat despite having what looks like Metal 3.1 but I believe it may be a quirk in how apple versions libs.

In essence `xcrun metal --version` doesn't match `__METAL_VERSION__`.

I think `__HAVE_BFLOAT__` is fairly self-documenting (which to me is better than being documented).

It's also easy to find look up in Xcode (as you already know - just stating for any curious readers).

Narsil:
> In this issue [comment](https://github.com/huggingface/candle/issues/1532#issuecomment-1880077926) a user reports not having bfloat despite having what looks like Metal 3.1 but I believe it may be a quirk in how apple versions libs.
> 
> In essence `xcrun metal --version` doesn't match `__METAL_VERSION__`.
> 
> I think `__HAVE_BFLOAT__` is fairly self-documenting (which to me is better than being documented).
> 
> It's also easy to find look up in Xcode (as you already know - just stating for any curious readers).

Then I think changing variable will not help that user. 

```
Error: Metal error Error while loading function: "Function 'cast_bf16_f32' does not exist"
```

Means the preprocessing code didn't enter the ifdef part, and metal IS <310 on his machine (despite what seems to say xcrun indeed).

We could on the other hand try to improve that error message. Maybe catch bf16 in the op name, and reraise specifically for those an error hinting that it might be an metal version issue. Wdyt ?

I actually didn't look in Xcode headers, if it's used by apple themselves I'm more confident in using it. The main concern for undocumented things is that they might disappear.

PR #1543, opened 2024-01-07
Metal: Activate bfloat affine and add benchmark

ivarflakstad:
~~Waiting on https://github.com/huggingface/candle/pull/1538~~


Narsil:
Could you share a bit about the speed improvements `fma` is supposed to bring to this particular op ?

Narsil:
Also let's fix the benchmark dead code first (by putting all the benchs in a single file which could be renamed `op.rs` for instance).

ivarflakstad:
> Could you share a bit about the speed improvements `fma` is supposed to bring to this particular op ?

Well, best case scenario it wouldn't bring any speed improvements at all because the compiler already understands that it should use `fma`.

The reason I added it is because it is made for this specific case, and because of the description:
> Returns the correctly rounded floating-point representation of the sum of c with the infinitely precise product of a and b. Rounding of intermediate products shall not occur. Edge case behavior is per the IEEE 754-2008 standard.

I don't know if manually writing `a * b + c` has the same properties for all values. I assume there is a reason this function is built-in.

Narsil (approved):
LGTM, thanks for the explanation of fma, definitely looks correct then.

Issue #1414, opened 2023-12-07
Difference in Whisper Mel impl


xd009642:
I've just been diving into some of the mel code for whisper as a coworker said it wasn't matching the python implementation (though that's on the dimension of the output so bigger than the values). And when checking comparing to whisper.cpp  the summing of the fft output and filters seems to not match:

https://github.com/huggingface/candle/blob/9bd94c1ffa0ccfd2bbc9526569b8b8a2a3812027/candle-transformers/src/models/whisper/audio.rs#L140C7-L140C7

```rust
            for k in 0..n_fft {
                sum += fft_out[k] * filters[j * n_fft + k];
            }
```

https://github.com/ggerganov/whisper.cpp/blob/4774d2feb01a772a15de81ffc34b34a1f294f020/whisper.cpp#L2456-L2462

```cpp
            int k = 0;
            for (k = 0; k < n_fft - 3; k += 4) {
                sum +=
                    fft_out[k + 0] * filters.data[j*n_fft + k + 0] +
                    fft_out[k + 1] * filters.data[j*n_fft + k + 1] +
                    fft_out[k + 2] * filters.data[j*n_fft + k + 2] +
                    fft_out[k + 3] * filters.data[j*n_fft + k + 3];
            }
```

Just raising this initially though I may update this issue if I spot more things that explain the differences we see.

LaurentMazare:
I might be missing something here but I feel that this should have the exact same behavior (as long as n_fft is divisible by 4), I think the `llama.cpp` unroll the loops manually for efficiency reasons but we don't do it on the candle side as `fast-math` is not available in rust stable anyway so the compiler doesn't have a chance to convert an unrolled version to simd instructions.

xd009642:
Does it? Because whisper.cpp has 2 loops and then the manually unrolled one in the inner. Without it, I think the code would look like:

```cpp
        // mel spectrogram
        for (int j = 0; j < mel.n_mel; j++) {
            double sum = 0.0;

            int k = 0;
            for (k = 0; k < n_fft - 3; k += 4) {
                // Rerolling the loop
                for(int offset = 0; offset < 4; offset++) {
                    sum += fft_out[k + offset] * filters.data[j*n_fft + k + offset];
                }
            }
```

If candle isn't unrolling the loop, surely it should have 3 nested loops instead of just 2 like the whisper code?

LaurentMazare:
[...]

PR #1549, opened 2024-01-09
Feature/all-close

FL33TW00D:


LaurentMazare:
What is your use case for this? I'm a bit reluctant to adding too much helper functions in the main crate, especially as methods, it makes the tensor objects harder to use as there are plenty of methods available, and also it adds some complexity that might be deferred to other crates.
(also typically we would name this `all_close` and not `allclose`, the behavior on `nan` would have to be documented and checked, there would be no need for testing it on anything else than cpu as it doesn't define device specific basic ops)

FL33TW00D:
@LaurentMazare I'm working on #1414 currently, and this would be handy to compare against known ground truths.

Will fix the remaining issues.

LaurentMazare:
Ah great that you're looking at #1414, would certainly be good to iron out these whisper discrepancies.

PR #1553, opened 2024-01-10
feat: add dependabot to the project

Sanix-Darker:
Hi there,
I really like the project, and i was wondering if we can add dependabot to it, to be always up to date regarding deps.
I set it to 5 max PRs and also on weekly only.
Let me know if you want to adjust those params, thanks.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
+    open-pull-requests-limit-per-dependency: 2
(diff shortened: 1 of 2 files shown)


Narsil:
Seems reasonable to me. @LaurentMazare wdyt ?

LaurentMazare:
For the dependabot we can try and we'll remove it if it's a pain (I'm more on the side on keeping the dependencies frozen for a longer time rather than changing them too often).
I'm not sure why the explicit caret have been added to the requirements but to my knowledge it's the default in cargo so I would be in favor of reverting that part of the change unless there is a good reason for it.

Sanix-Darker:
am okay to with the idea on keeping things frozen, let's remove the caret for now

LaurentMazare (approved):

PR #1560, opened 2024-01-10
fix: Idiomatic & matching Whisper Mel Spectrogram

### LINKED ISSUE #1414: Difference in Whisper Mel impl

FL33TW00D:
This PR does the following:
1. A more idiomatic implementation of `log_mel_spectrogram`.
2. `log_mel_spectrogram` now matches the specification [here].(https://github.com/openai/whisper/blob/ba3f3cd54b0e5b8ce1ab3de13e32122d0d5f98ab/whisper/audio.py#L110)

Some points of contention:
1. Introduces a dependency on `realfft`
2. Requires a ground truth `.npy` file
3. Some of the operations require mapping back to a `Vec`, instead of being tensor ops.

I appreciate it's a bit scruffy, but it fixes #1414, would love some pointers to get it mergable.


LaurentMazare:
I would rather keep the small self-contained implementation that initially was there rather than introducing a dependency. Wouldn't it be possible to find where the difference is coming from instead of making a full rewrite? (it has been on my stack for a long time to do this but sadly I have been a bit busy lately)
Another possibliity would be to make a proper external `candle-whisper` crate that would aim at matching the openai implementation closely and keep the current one as an example. Would that be something that you would be interested in?

FL33TW00D:
> I would rather keep the small self-contained implementation that initially was there rather than introducing a dependency. Wouldn't it be possible to find where the difference is coming from instead of making a full rewrite? (it has been on my stack for a long time to do this but sadly I have been a bit busy lately) Another possibliity would be to make a proper external `candle-whisper` crate that would aim at matching the openai implementation closely and keep the current one as an example. Would that be something that you would be interested in?

Yeah I thought the dependency might be an issue. Their FFT implementation is a bit dense, maybe I'll go with the external crate option.

LaurentMazare:
If you do the external repo, please implement beam search too :)

PR #1561, opened 2024-01-10
Add RepVGG model.

janimo:
RepVGG model inference implementation

https://arxiv.org/abs/2101.03697

### DIFF
--- a/candle-transformers/src/models/repvgg.rs
+++ b/candle-transformers/src/models/repvgg.rs
@@ -18,6 +18,8 @@ struct RepVGGConfig {
     stages: [usize; 4],
 }
 
+const CHANNELS_PER_STAGE: [usize; 5] = [64, 64, 128, 256, 512];
+
 #[rustfmt::skip]
 fn repvgg_config(cfg: &str) -> RepVGGConfig {
     let configs = HashMap::from([
@@ -134,12 +136,12 @@ fn repvgg_layer(
 
 // Get the number of output channels per stage taking into account the multipliers
 fn output_channels_per_stage(a: f32, b: f32, stage: usize) -> usize {
-    let chans = [64, 64, 128, 256, 512][stage] as f32;
+    let channels = CHANNELS_PER_STAGE[stage] as f32;
 
     match stage {
-        0 => std::cmp::min(64, (chans * a) as usize),
-        4 => (chans * b) as usize,
-        _ => (chans * a) as usize,
+        0 => std::cmp::min(64, (channels * a) as usize),
+        4 => (channels * b) as usize,
+        _ => (channels * a) as usize,
     }
 }
 
--- a/candle-transformers/src/models/repvgg.rs
+++ b/candle-transformers/src/models/repvgg.rs
@@ -1,0 +1,233 @@
+//! RepVGG inference implementation
+//!
+//! See "RepVGG: Making VGG-style ConvNets Great Again" Ding et al. 2021
+//! https://arxiv.org/abs/2101.03697
+
+use std::collections::HashMap;
+
+use candle::{Result, Tensor, D};
+use candle_nn::{
+    batch_norm, conv2d_no_bias, linear, BatchNorm, Conv2d, Conv2dConfig, Func, VarBuilder,
+};
+
+#[derive(Clone)]
+struct RepVGGConfig {
+    a: f32,
+    b: f32,
+    groups: usize,
+    stages: [usize; 4],
+}
+
+#[rustfmt::skip]
+fn repvgg_config(cfg: &str) -> RepVGGConfig {
+    let configs = HashMap::from([
+        ( "A0", RepVGGConfig { a: 0.75, b: 2.5, groups: 1, stages: [2, 4, 14, 1]}),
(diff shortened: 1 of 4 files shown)


candle-transformers/src/models/repvgg.rs
```diff
@@ -0,0 +1,233 @@
...
+    a: f32,
+    b: f32,
+    groups: usize,
+    stages: [usize; 4],
+}
+
+#[rustfmt::skip]
+fn repvgg_config(cfg: &str) -> RepVGGConfig {
+    let configs = HashMap::from([
+        ( "A0", RepVGGConfig { a: 0.75, b: 2.5, groups: 1, stages: [2, 4, 14, 1]}),
```
LaurentMazare, line 24: Let's use an enum instead of a string for the model variants.

candle-transformers/src/models/repvgg.rs
```diff
@@ -0,0 +1,233 @@
...
+        ( "B0", RepVGGConfig { a: 1.0, b: 2.5, groups:1, stages: [4, 6, 16, 1]}),
+        ( "B1", RepVGGConfig { a: 2.0, b: 4.0, groups:1, stages: [4, 6, 16, 1]}),
+        ( "B2", RepVGGConfig { a: 2.5, b: 5.0, groups:1, stages: [4, 6, 16, 1]}),
+        ( "B3", RepVGGConfig { a: 3.0, b: 5.0, groups:1, stages: [4, 6, 16, 1]}),
+        ( "B1G4", RepVGGConfig { a: 2.0, b: 4.0, groups:4, stages: [4, 6, 16, 1]}),
+        ( "B2G4", RepVGGConfig { a: 2.5, b: 5.0, groups:4, stages: [4, 6, 16, 1]}),
+        ( "B3G4", RepVGGConfig { a: 3.0, b: 5.0, groups:4, stages: [4, 6, 16, 1]}),
+    ]);
+
+    configs[cfg].clone()
```
LaurentMazare, line 36: Creating a hashmap and then extracting a single value out of it seems a bit odd, let's just do a normal pattern matching instead.

candle-transformers/src/models/repvgg.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    Ok(Func::new(move |xs| {
+        let xs = xs.apply(&reparam_conv)?.relu()?;
+        Ok(xs)
+    }))
+}
+
+// Get the number of output channels per stage taking into account the multipliers
+fn output_channels_per_stage(a: f32, b: f32, stage: usize) -> usize {
+    let chans = [64, 64, 128, 256, 512][stage] as f32;
```
LaurentMazare, line 137: Let's extract this array as a top level constant at the top of the file?

LaurentMazare (commented):
Thanks for making this, it's a great addition and the code looks already pretty good. A few comments inline to make it more in line with our current model codebase.

candle-transformers/src/models/repvgg.rs
janimo: I used strings here because of the way I used strings in main.rs too, where the command line --which arg is being checked by clap and then only valid values are being passed as model config. I did this to avoid the  matches for each of the variants and be able to derive the model filename via a single line of string formatting. Shall I change main.rs too?

candle-transformers/src/models/repvgg.rs
janimo: It is weird indeed, unfortunately there can't be no global literal hashmaps which would have been a bit cleaner. It was the consequence of trying to parameterize the model via a string and not have to have a separate wrapper function for each of the 10 model variants.

candle-transformers/src/models/repvgg.rs
janimo: I put it there since it is only used by that function, but I can move it. Thanks for the quick feedback!

candle-transformers/src/models/repvgg.rs
LaurentMazare: Re the model filename, you can certainly have a `as_str()` method on the enum that would return `"A0"` etc (or `model_filename()`, ...
I think ideally the variant would be defined on the example side so that the clap macro for which can be applied to it - you'll find lots of examples of this in the examples directory, and the `repvgg.rs` file would define `RepVGGConfig::a0()`, etc. Similar a bit to what is done [here for vgg](https://github.com/huggingface/candle/blob/d3bdd788cfdcf49b6ea539b77647b82a0b979db0/candle-examples/examples/vgg/main.rs#L12).

candle-transformers/src/models/repvgg.rs
janimo: done in the most recent update

candle-transformers/src/models/repvgg.rs
janimo: Replaced the hashmap with a match (on &str for now), it is indeed cleaner this way.

candle-transformers/src/models/repvgg.rs
LaurentMazare: Made the final tweaks for this so that this PR can be merged, thanks!

PR #1562, opened 2024-01-10
Use candle_nn::embedding instead of local copies in a few models.

janimo:
Remove a few copies of embedding.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1567, opened 2024-01-11
Remove redundant metal version check

bayedieng:
ifdef directive is not properly closed here which causes error.


ivarflakstad:
Ah, I see what happened here. Those two `#if`s are redundant.
Could you instead remove `#if __METAL_VERSION__ >= 310`? 😊

bayedieng (commented):
Done.

PR #1571, opened 2024-01-12
chore(doc): a small contrib to add star graph progression to the README

Sanix-Darker:
## WHAT

A small graph for stars grows...


LaurentMazare:
Thanks for the PR, I feel that the readme is already pretty long and hard to navigate to find the right information so I'm a bit unsure having the history of number of stars would add that much relevant information there.

Sanix-Darker:
No worry if we do close this PR, btw i was wondering if it could not be a good idea to move the README to WIKI section ?
To have a light version, what do you think ?

PR #1606, opened 2024-01-18
fix: larger batches

### LINKED ISSUE #1596: Metal peformance issue: Sentence Transformer using Candle gets aborted by the system
When using `sentence-transformers/all-MiniLM-L6-v2` on m1 PRO mac (32GB ram) and attempting to generate embeddings for 100 sentences. the program get's aborted by the system after GPU usage spikes to 100.

### How to reproduce:
- place the following code snipet (below) in `candle-examples/examples/bert-metal/main.rs`
- cargo run --example bert-metal  --features metal

```
#[cfg(feature = "mkl")]
extern crate intel_mkl_src;

#[cfg(feature = "accelerate")]
extern crate accelerate_src;
use candle_transformers::models::bert::{BertModel, Config, DTYPE};

use anyhow::{Error as E, Result};
use candle::Tensor;
use candle_nn::VarBuilder;
use hf_hub::{api::sync::Api, Repo, RepoType};
use tokenizers::{PaddingParams, Tokenizer};

fn build_model_and_tokenizer() -> Result<(BertModel, Tokenizer)> {
    let device = candle_examples::device(false)?;
    let repo = Repo::with_revision("sentence-transformers/all-MiniLM-L6-v2".to_string(), RepoType::Model, "refs/pr/21".to_string());
    let (config_filename, tokenizer_filename, weights_filename) = {
        let api = Api::new()?.repo(repo);
        let config = api.get("config.json")?;
        let tokenizer = api.get("tokenizer.json")?;
        let weights = api.get("model.safetensors")?;
        (config, tokenizer, weights)
    };
    let config = std::fs::read_to_string(config_filename)?;
    let config: Config = serde_json::from_str(&config)?;
    let tokenizer = Tokenizer::from_file(tokenizer_filename).map_err(E::msg)?;
    let vb = unsafe { VarBuilder::from_mmaped_safetensors(&[weights_filename], DTYPE, &device)? };
    let model = BertModel::load(vb, &config)?;
    Ok((model, tokenizer))
}

pub fn normalize_l2(v: &Tensor) -> Result<Tensor> {
    Ok(v.broadcast_div(&v.sqr()?.sum_keepdim(1)?.sqrt()?)?)
}

fn main() -> Result<()> {
    let sentence_count = 500;
    let (model, mut tokenizer) = build_model_and_tokenizer()?;
    let device = &model.device;
    let sentences: Vec<&str> = vec!["This is a sample sentence that should be short enough"; sentence_count];
    if let Some(pp) = tokenizer.get_padding_mut() {
        pp.strategy = tokenizers::PaddingStrategy::BatchLongest
    } else {
        let pp = PaddingParams {
            strategy: tokenizers::PaddingStrategy::BatchLongest,
            ..Default::default()
        };
        tokenizer.with_padding(Some(pp));
    }
    let tokens = tokenizer.encode_batch(sentences, true).map_err(E::msg)?;
    let token_ids = tokens
        .iter()
        .map(|tokens| {
            let tokens = tokens.get_ids().to_vec();
            Ok(Tensor::new(tokens.as_slice(), device)?)
        })
        .collect::<Result<Vec<_>>>()?;

    let token_ids = Tensor::stack(&token_ids, 0)?;
    let token_type_ids = token_ids.zeros_like()?;
    println!("running inference on batch {:?}", token_ids.shape());
    let embeddings = model.forward(&token_ids, &token_type_ids)?;
    println!("model forward timing {:?}", start.elapsed());
    let (_n_sentence, n_tokens, _hidden_size) = embeddings.dims3()?;
    let embeddings = (embeddings.sum(1)? / (n_tokens as f64))?;
    let embeddings = normalize_l2(&embeddings)?;
    println!("pooled embeddings {:?}", embeddings.shape());
    Ok(())
}
```

Note: you might need to increase the value of `sentence_count` to reproduce, for me the cutoff is somewhere between 80 and 100 (i.e  80 ✅  but 100 🔴 ).

### Output:
```
running inference on batch [100, 12]
[1]    52157 abort      cargo run --example bert-metal --features metal
```

For reference, when using the python sentence_transformers library, I am able to embed a much larger number of 
sentences at once without the program getting aborted (sample code below).
```
import torch
from sentence_transformers import SentenceTransformer
import time

if __name__ == "__main__":
    sentences = ["This is a sample sentence that should be short enough"] * 1000
    device = torch.device("mps")
    model = SentenceTransformer("sentence-transformers/all-MiniLM-L6-v2", device=device)
    start = time.monotonic()
    embeddings = model.encode(sentences, batch_size=len(sentences), device=device)
    done = time.monotonic()
    print(f"embedding (including tokenization) {done - start} seconds")
```
sentence-transformers = "2.2.2"

FL33TW00D:
Fixes #1596

Related: https://github.com/philipturner/metal-flash-attention/pull/16


Narsil (approved):
LGTM

pcuenca (approved):
Very cool!

I assume the corresponding kernel modification is inside the compiled metallib, are we not tracking the metal sources here?

philipturner:
https://github.com/philipturner/metal-flash-attention/pull/16#issuecomment-1898868823

philipturner:
If the current fork works for your use case, that's great. The whole point of open-sourcing it and the build process. But the main repo is going to keep the more flexible API. This is a more workable approach from a software engineering perspective.

FL33TW00D:
Good points from Phil.

I've now changed to just using a buffer, but still requires changes to MFA so we will forever be on a fork. Thoughts @Narsil?

pcuenca:
If we are going to be in a fork forever, I'd rather apply the previous solution as it's simpler and requires less memory initialization.

However, if the buffer solution can be merged upstream, then I'd totally go for it.

I don't fully understand Phil's comment about a better software engineering approach. I'd prefer to understand the actual use cases it unlocks. Batching is a clear use-case to me.

philipturner:
>  I'd prefer to understand the actual use cases it unlocks. Batching is a clear use-case to me.

Let's say you want to multiply a series of 8 RHS matrices over an over again, by several sets of 8 LHS matrices. This is not something achievable with `cublasXgemmStridedBatched`. It requires a rule-based offset pattern that wouldn't be possible with a fixed stride.

```
0 x 0
1 x 1
2 x 2
3 x 3
4 x 4
5 x 5
6 x 6
7 x 7
8 x 0
9 x 1
...
```

You can probably think of an easier solution. Make the N dimension of the RHS matrix just 8x larger. But you see how the choice of where a matrix starts, where it ends in memory, is something left to the user. One might want a non-contiguous memory layout where the 49th and 50th matrices skip 200 units of memory. One might source matrix data from another tensor, originally reshaped at a different dimension &mdash; any padding within the original tensor changes the memory spacing. Emulating an irregular sliding window (sliding window attention). Overlapping accesses to the same matrix. Writing `C` into a subrange of a KV cache. There is a combinatorial explosion of use cases.

> I don't fully understand Phil's comment about a better software engineering approach.

The better approach is to keep the main repo as it is. What happens in the divergent fork is up to you.

FL33TW00D:
This should be finalized now, you can see the single line change for batching on my fork here: https://github.com/FL33TW00D/metal-flash-attention/blob/feature/larger-batches/Sources/GEMM.metal#L269

PR #1616, opened 2024-01-23
Metal: Where cond get_strided_index conditionally based on function constants

ivarflakstad:
not bad


ivarflakstad:
Totally forgot about this one

PR #1627, opened 2024-01-28
fix quantized llama example use metal failed

hubertshelley:


jamiepine:
This solved my issue, would love this to be merged!

Issue #1508, opened 2023-12-31
[Examples] SDXL should also use Euler Ancestral


sayakpaul:
The example on Stable Diffusion uses DDIM for SDXL. It should use Euler Ancestral as that is the default scheduler for it: https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0/blob/main/scheduler/scheduler_config.json#L2.

For v1.5, it should be PNDM: https://huggingface.co/runwayml/stable-diffusion-v1-5/blob/main/scheduler/scheduler_config.json#L2.

freespirit:
Hello @sayakpaul
I'm currently experimenting with SDXL and I'd be happy to change the default scheduler.

I can only find a [hardcoded Scheduler config](https://github.com/huggingface/candle/blob/main/candle-transformers/src/models/stable_diffusion/mod.rs#L218-L221) in the `candle-core`'s `StableDiffusionConfig` and no way to configure it in the example. If that's what you meant, I can change it to `EulerAncestralDiscreteSchedulerConfig`. Luckily its default values match those in the linked json.

PR #1635, opened 2024-01-29
Use the Euler Ancestral scheduler by default with SDXL

freespirit:
This PR changes the default Scheduler for SDXL to Euler Ancestral as requested in #1508

### DIFF
--- a/candle-examples/examples/stable-diffusion/README.md
+++ b/candle-examples/examples/stable-diffusion/README.md
@@ -32,10 +32,10 @@ cargo run --example stable-diffusion --release --features=cuda,cudnn \
     -- --prompt "a cosmonaut on a horse (hd, realistic, high-def)" --sd-version turbo
 ```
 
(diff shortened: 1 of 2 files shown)


freespirit:
Hello!
I have an issue with the default Euler Configuration and I was hoping for some guidance with this PR.

The image I get after 17 steps looks like noise:
sd_final copy


The [default configuration](https://github.com/huggingface/candle/blob/main/candle-transformers/src/models/stable_diffusion/euler_ancestral_discrete.rs#L40-L52) for Euler Ancestral looks exactly the same as the [JSON](https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0/blob/main/scheduler/scheduler_config.json#L2) scheduler configuration so It seems that using it should just work.

The way I run the example is with metal support:
```shell
cargo run --example stable-diffusion --features=metal -- --prompt "A black cat by the window" --sd-version xl --n-steps 17
```

I was hoping for some guidelines before diving deeper into the Scheduler implementation and trying to debug the issue further 🤗

Issue #1632, opened 2024-01-28
Mistral::new assumption about `model.` prefix


functorism:
https://github.com/huggingface/candle/blob/fd7c8565646039e35925b8730d27ddad195d7e73/candle-transformers/src/models/mistral.rs#L384

Some Mistral models don't prefix the safetensors keys with `model.` and this prevents `Mistral::new` from being able to load them.

https://huggingface.co/intfloat/e5-mistral-7b-instruct/discussions/23

functorism:
Would an addition to the Mixtral that makes lm_head `Option` be acceptable. That way the same model config can accommodate both text generation models and the aforementioned embedding model.

PR #1636, opened 2024-01-29
Add support for intfloat/e5-mistral-7b-instruct

functorism:
This addresses this issue https://github.com/huggingface/candle/issues/1632

The [intfloat/e5-mistral-7b-instruct](https://huggingface.co/intfloat/e5-mistral-7b-instruct) has no `lm_head` and also lacks the `model.` prefix in its safetensors. This PR accommodates that.


functorism:
This change would not also support loading https://huggingface.co/Salesforce/SFR-Embedding-Mistral

PR #1638, opened 2024-01-30
onnx: add the Flatten operator.

wanglong001:
onnx: add the Flatten operator.

ref: https://github.com/onnx/onnx/blob/8a6b70f0adcff5409b8cd65883bbbea7d6c4f4e1/onnx/reference/ops/op_flatten.py#L11

### DIFF
--- a/candle-onnx/src/eval.rs
+++ b/candle-onnx/src/eval.rs
@@ -770,13 +770,9 @@ pub fn simple_eval(
             "Flatten" => {
                 let axis = get_attr_opt::<i64>(node, "axis")?.copied().unwrap_or(1) as usize;
                 let input = get(&node.input[0])?;
+                let first_part: usize = input.shape().dims().iter().take(axis).product();
                 let end_index = input.shape().dims().iter().product::<usize>();
-                let new_shape = if axis == 0 {
-                    (1, end_index)
-                } else {
-                    let first_part: usize = input.shape().dims().iter().take(axis).product();
-                    (first_part, end_index / first_part)
-                };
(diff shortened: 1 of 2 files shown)


candle-onnx/src/eval.rs
```diff
@@ -766,6 +766,20 @@ pub fn simple_eval(
...
                 values.insert(node.output[0].clone(), output);
             }
+            //  https://github.com/onnx/onnx/blob/main/docs/Operators.md#flatten
+            "Flatten" => {
+                let axis = get_attr_opt::<i64>(node, "axis")?.copied().unwrap_or(1) as usize;
+                let input = get(&node.input[0])?;
+                let end_index = input.shape().dims().iter().product::<usize>();
+                let new_shape = if axis == 0 {
+                    (1, end_index)
+                } else {
```
LaurentMazare, line 776: What about the case where axis would not be 1?

wanglong001: Hi， Referencing this link https://github.com/onnx/onnx/blob/main/docs/Operators.md#flatten, in the case where axis > 1, it is compatible with axis == 1.

LaurentMazare: Couldn't the case with axis set to 0 be merged with the other case? The product from `first_part` would be `1` on an empty iterator?

wanglong001: you are right. I made changes.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1650, opened 2024-02-02
Add StableLM-2, StableLM Code and Zephyr variants

janimo:
Add Stable-Code, StableLM-2 and Zephyr variants for V1 and V2. For completeness also added the V1 variant from the stabilityai repos (same as the existing one which I renamed to V1Orig). The GGUF quantized versions of the newer models use llama.cpp tensor names that are different from the ones used in the non-quantized model so I did not add those.


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks for the addition. For GGUF model, the idea would be to create some from the safetensors weight using `tensor-tools quantize` rather than relying on the llama.cpp ones.

PR #1660, opened 2024-02-05
make llama2 derive clone

danielclough:
Added derive Clone so that I can pass the model around my app in State.
Added derive Debug because that's what is used in other models.

### DIFF
--- a/candle-transformers/src/models/llama2_c.rs
+++ b/candle-transformers/src/models/llama2_c.rs
@@ -70,7 +70,7 @@ impl Config {
     }
 }
 
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub struct Cache {
     masks: Arc<Mutex<HashMap<usize, Tensor>>>,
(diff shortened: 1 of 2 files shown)


LaurentMazare:
I think for these models it's pretty error prone to clone them as they embed a kv cache that is accessed via inner mutability. So if you clone them, the cache would be shared between the two instances and very likely to conflict. Are you sure that this would not result in a bug in your use case too?

danielclough:
Ah.
That explains the bug when I use `kv_cache` `true`.

Running with `kv_cache` `false` works as expected.

Do you have any suggestions for passing models around in Axum `State` (`Mutex`) other than cloning them?

LaurentMazare:
I've pushed some tweaks to make the llama2 and quantized-llama2 models use an explicit cache rather than internalizing it, this removes the need for a mutex so cloning these should be safer now.

danielclough:
Fantastic!

Issue #1454, opened 2023-12-17
Pth file support is broken on Windows


Nikaidou-Shinku:
## Description

Pth file support on Windows is broken due to some issues with handling path in zip files.

## Details

When reading `tensor_infos`, `PathBuf::push` will use `\` as path separator on Windows.

https://github.com/huggingface/candle/blob/94817dac56f3c91235ad94b7b3419ab8c2e87134/candle-core/src/pickle.rs#L231

But soon `ZipArchive::by_name` may need a path with `/` as separator.

https://github.com/huggingface/candle/blob/94817dac56f3c91235ad94b7b3419ab8c2e87134/candle-core/src/pickle.rs#L713

This results in a `ZipError::FileNotFound` being raised.

---

I don't have an elegant way to solve this problem, as the separator used by path inside the zip seems to be dependent on the platform it was created on.

If anyone has a good idea, I can submit a PR for it.

antimora:
Burn project has started using Candle pickle and I can confirm CI fails on Windows (Mac and Windows work). 

I am attaching `pt` file for testing. I cannot debug it fully because my computer is Mac.

[mnist.pt.zip](https://github.com/huggingface/candle/files/14015029/mnist.pt.zip)

klapaukh:
This is fixed for me now in v0.4.0. @antimora / @Nikaidou-Shinku is it working for you too now?

laggui:
Can't speak for the specific use case previously mentioned here but our [simple example](https://burn.dev/book/import/pytorch-model.html) and all CI tests work on windows (previously disabled) with the latest release.

Thanks guys!

antimora:
> This is fixed for me now in v0.4.0. @antimora / @Nikaidou-Shinku is it working for you too now?

I will test it out and see if the bug is fixed. Thank you!

antimora:
I has been fixed, it's converted by our CI job. You may close it if you'd like, @LaurentMazare

PR #1661, opened 2024-02-05
Fix: pth files don't load on Windows

klapaukh:
Fix for bug #1454, where `pth` files don't load on Windows. 
This is because the path components inside of the uncompressed zip file are being joined with `path.push` which is OS aware, and so uses the `\\` separator. However, the separators inside of an uncompressed zip file are always `/`.
Since `path_to_string` in `zip` is private, the separator character is fixed, and there is only one place where this happens, it seems reasonable to just hard code the join with format.


LaurentMazare:
Thanks for the fix, could you confirm that it works well on the files that were posted on #1454 ? (I wouldn't have a windows box to try)

klapaukh:
I tried it with the `mnist.pt` from the issue ticket and it seems to work. I also made a minimal `pth` file and a regression test case for this issue.
No worries, I'm currently primarily on Windows.

LaurentMazare (approved):

LaurentMazare:
Great, thanks!

PR #1668, opened 2024-02-06
Fix token generation in bilingual models (non-English outputs)

guoqingbao:
There is a token generation problem for bilingual models (Yi, StableLM-v2, etc.) for non-English outputs, as shown below:

NonSmooth

This patch will fix the non-smooth token generation for non-English outputs and **does not** affect English generation:

Smooth


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1670, opened 2024-02-06
Add `VarBuilder::from_backend`

danieldk:
`candle-nn` already exposes a trait to define custom backends. However, it's not possible to actually construct a `VarBuilder` with a custom backend because the constructor is not exposed.

This change makes the constructor public and renames it from `new` to `from_backend` to avoid that it is seen as the primary constructor (which could be confusing to users).


LaurentMazare (approved):

LaurentMazare:
Ah that's a good point, thanks!

PR #1671, opened 2024-02-06
Enhanced pickle support for state_dict retrieval from nested dictionaries in .pth files

### LINKED ISSUE #1663: ` pickle::read_all` returns empty when loading open-ai's whisper `pt` file.
The following returns empty list.

```rust
use candle_core::pickle;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    // downloaded from https://openaipublic.azureedge.net/main/whisper/models/d3dd57d32accea0b295c96e26691aa14d8822fac7d9d27d5dc00b4ca2826dd03/tiny.en.pt
    let path = "./tiny.en.pt";

    let tensors = pickle::read_all(path).unwrap();

    println!("{:?}", tensors);
    Ok(())
}

```

Output:

```
[]
```

antimora:
Fixes #1663

This PR enhances pickle to retrieve the `state_dict` with a specified optional key. Sometimes the `state_dict` is stored at the top level of a dictionary, alongside other metadata within a `.pth` file, as seen with OpenAI's Whisper model. This modification allows the direct retrieval of states, rather than requiring the `.pth` file to be resaved using Python.

This solution was discussed and agreed upon here: https://github.com/huggingface/candle/issues/1663#issuecomment-1930154381

The fix has been tested with a new unit test and with the Whisper model (although the model cannot currently be read entirely because the data is not saved in a contiguous manner).


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1672, opened 2024-02-07
Add support for loading Fortran contiguous tensors

antimora:
This commit introduces the ability to handle Fortran contiguous tensors in the tensor loading process. Previously, the code only supported loading tensors that were contiguous in memory, failing with an error for non-contiguous tensors. With this update, tensors identified as Fortran contiguous (column-major order) are now correctly handled by reversing their dimensions after loading. This enhancement ensures broader compatibility with different tensor layouts, improving the robustness of tensor loading operations.

- Check if a tensor is Fortran contiguous using the `is_fortran_contiguous` flag.
- For Fortran contiguous tensors, reverse the dimensions after loading to correctly represent their layout in memory.
- Continue to bail out with an error for tensors that are neither C contiguous nor Fortran contiguous, maintaining the previous behavior for non-contiguous tensors without explicit support.

This change addresses the issue of loading Fortran contiguous tensors, which was previously unsupported, thereby extending the functionality of the tensor loading mechanism to accommodate a wider variety of tensor layouts.

I have tested with Whisper's pt files and added a new unit test just to cover this special case.


candle-core/src/pickle.rs
```diff
@@ -757,7 +758,16 @@ impl PthTensors {
             tensor_info.dtype,
             &mut reader,
         )?;
-        Ok(Some(tensor))
+
+        if is_fortran_contiguous {
```
LaurentMazare, line 762: I think this would only work for tensors that have a "symmetrical" shape. If the shape using the fortran layout is `[d1, d2, d3]`, we should first reshape the tensor that was read assuming C layout to shape `[d3, d2, d1]` and then permute the dimensions.

antimora: Good catch. I have updated it now to cover other cases. I also added a unit test to confirm it works for that case.

candle-core/src/pickle.rs
```diff
@@ -757,7 +758,20 @@ impl PthTensors {
...
```
LaurentMazare, line 762: Let's maybe add to the condition that there are multiple dimensions (or that it's not c contiguous). Otherwise for array with a single dimension we will pass through this case, the output will be correct but seems a bit unnecessary.

candle-core/src/pickle.rs
```diff
@@ -736,10 +736,12 @@ impl PthTensors {
...
         let mut zip = zip::ZipArchive::new(zip_reader)?;
         let mut reader = zip.by_name(&tensor_info.path)?;
+        let is_fortran_contiguous = tensor_info.layout.is_fortran_contiguous();
+        let rank = tensor_info.layout.shape().rank();
 
         // Reading the data is a bit tricky as it can be strided, for now only support the basic
-        // case.
-        if !tensor_info.layout.is_contiguous() {
+        // case and when the tensor is fortran contiguous.
+        if rank > 1 && !tensor_info.layout.is_contiguous() && !is_fortran_contiguous {
```
LaurentMazare, line 744: Sorry but I think we want to fail on non contiguous 1d tensors too.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1674, opened 2024-02-07
feat: support multithread spectrogram and small perf tweaks

drbh:
This PR adds support for multithreaded workers in `log_mel_spectrogram_` and significantly speeds up the `pcm_to_mel` step.

On an Apple M3 Max this adds a ~6-7x speed up `139.284375ms` -> `21.299458ms` using the following commands

```bash
cargo run --release --example whisper --features metal
```

additionally there are a few new tests that can be run

```bash
cargo test --package candle-transformers --lib -- models::whisper::audio::tests --nocapture
```

**note this speed up is particularly useful with smaller models since the `pcm_to_mel` can take as much time or more as the decoding step on some backends (cuda, metal) so in those cases `pcm_to_mel` is no longer the bottleneck

### DIFF
--- a/candle-transformers/src/models/whisper/audio.rs
+++ b/candle-transformers/src/models/whisper/audio.rs
@@ -1,7 +1,18 @@
 // Audio processing code, adapted from whisper.cpp
 // https://github.com/ggerganov/whisper.cpp
 
-pub trait Float: num_traits::Float + num_traits::FloatConst + num_traits::NumAssign {}
+use candle::utils::get_num_threads;
+use std::sync::Arc;
+use std::thread;
+
+pub trait Float:
+    num_traits::Float
+    + num_traits::FloatConst
+    + num_traits::NumAssign
+    + std::marker::Send
+    + std::marker::Sync
+{
+}
 
 impl Float for f32 {}
 impl Float for f64 {}
@@ -99,45 +110,50 @@ fn log_mel_spectrogram_w<T: Float>(
     };
 
     let zero = T::zero();
-    let half = T::from(0.5).unwrap();
     let mut fft_in = vec![zero; fft_size];
     let mut mel = vec![zero; n_len * n_mel];
+    let n_samples = samples.len();
+    let end = std::cmp::min(n_samples / fft_step + 1, n_len);
 
-    for i in (ith..n_len).step_by(n_threads) {
+    for i in (ith..end).step_by(n_threads) {
         let offset = i * fft_step;
 
         // apply Hanning window
-        for j in 0..fft_size {
-            fft_in[j] = if offset + j < samples.len() {
-                hann[j] * samples[offset + j]
-            } else {
-                zero
+        for j in 0..std::cmp::min(fft_size, n_samples - offset) {
+            fft_in[j] = hann[j] * samples[offset + j];
+        }
+
+        // fill the rest with zeros
+        if n_samples - offset < fft_size {
+            for j in n_samples - offset..fft_size {
+                fft_in[j] = zero;
             }
         }
 
-        // FFT -> mag^2
+        // FFT
         let mut fft_out: Vec<T> = fft(&fft_in);
 
+        // Calculate modulus^2 of complex numbers
         for j in 0..fft_size {
             fft_out[j] = fft_out[2 * j] * fft_out[2 * j] + fft_out[2 * j + 1] * fft_out[2 * j + 1];
-        }
-        for j in 1..fft_size / 2 {
-            let v = fft_out[fft_size - j];
-            fft_out[j] += v;
-        }
-
-        if speed_up {
-            // scale down in the frequency domain results in a speed up in the time domain
-            for j in 0..n_fft {
-                fft_out[j] = half * (fft_out[2 * j] + fft_out[2 * j + 1]);
-            }
         }
 
         // mel spectrogram
         for j in 0..n_mel {
             let mut sum = zero;
-            for k in 0..n_fft {
+            let mut k = 0;
+            // Unroll loop
+            while k < n_fft.saturating_sub(3) {
+                sum += fft_out[k] * filters[j * n_fft + k]
+                    + fft_out[k + 1] * filters[j * n_fft + k + 1]
+                    + fft_out[k + 2] * filters[j * n_fft + k + 2]
+                    + fft_out[k + 3] * filters[j * n_fft + k + 3];
+                k += 4;
+            }
+            // Handle remainder
+            while k < n_fft {
                 sum += fft_out[k] * filters[j * n_fft + k];
+                k += 1;
             }
             mel[j * n_len + i] = T::max(sum, T::from(1e-10).unwrap()).log10();
         }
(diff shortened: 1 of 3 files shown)


candle-transformers/src/models/whisper/audio.rs
```diff
@@ -102,22 +113,29 @@ fn log_mel_spectrogram_w<T: Float>(
...
+            fft_in[j] = hann[j] * samples[offset + j];
         }
 
-        // FFT -> mag^2
+        // fill the rest with zeros
+        if n_samples - offset < fft_size {
+            fft_in
+                .iter_mut()
+                .skip(n_samples - offset)
+                .for_each(|x| *x = zero);
```
LaurentMazare, line 132: Couldn't we use the `fill` method on the slice `fft_in[n_samples-offset..]` for this rather than doing a `for_each`?

candle-transformers/src/models/whisper/audio.rs
```diff
@@ -180,10 +212,55 @@ fn log_mel_spectrogram_<T: Float + std::fmt::Display>(
...
-    // Use a single thread for now.
-    let mut mel = log_mel_spectrogram_w(
-        0, &hann, &samples, filters, fft_size, fft_step, speed_up, n_len, n_mel, 1,
-    );
+    // ensure that the number of threads is even and less than 12
+    let n_threads = std::cmp::min(get_num_threads() - get_num_threads() % 2, 12);
+
+    let hann = Arc::new(hann);
+    let samples = Arc::new(samples);
+    let cloned_contents_of_filters = filters.to_vec();
```
LaurentMazare, line 220: What is the rationale for cloning the filters here? This seems a bit wasteful compare to just using the reference if we can.

candle-transformers/src/models/whisper/audio.rs
```diff
@@ -180,10 +212,55 @@ fn log_mel_spectrogram_<T: Float + std::fmt::Display>(
...
+            Err(_) => panic!("Thread failed"),
+        }
+    }
+
+    let l = all_outputs[0].len();
+    let mut mel = vec![zero; l];
+
+    for i in (0..l).step_by(n_threads) {
+        for ouputs in all_outputs.iter() {
+            for j in 0..n_threads {
```
LaurentMazare, line 258: Could you add some comment to detail a bit what is going on here, the first loop on the outputs of all threads then iterating on the number of thtreads is pretty hard to follow.

candle-transformers/src/models/whisper/audio.rs
```diff
@@ -197,7 +274,7 @@ fn log_mel_spectrogram_<T: Float + std::fmt::Display>(
     mel
 }
 
-pub fn pcm_to_mel<T: Float + std::fmt::Display>(
+pub fn pcm_to_mel<T: Float + std::fmt::Display + Send + Sync + 'static>(
```
LaurentMazare, line 277: Didn't you modify the `Float` trait to include `Send` and `Sync`?

candle-transformers/src/models/whisper/audio.rs
drbh: oh yea thats a great point, thanks for the recommendation! latest commit uses fill

candle-transformers/src/models/whisper/audio.rs
drbh: no good rational 😅 haha, I was originally having lifetime issues when passing filters `&[T]` between threads and forced it to work with the clone, I meant to remove before pushing but missed it. Thanks for pointing out 🙏

This should be resolved in the latest commit and relies on passing a reference into scoped threads. Please let me know if this should be updated!

LaurentMazare (approved):

LaurentMazare:
Thanks!

candle-transformers/src/models/whisper/audio.rs
drbh: yea forsure, I agree the logic is a bit unintuitive (followed the pattern from llama.cpp) and I've rewritten it to be slightly more readable. (copied below for reference). 

```rust
    // iterate over mel spectrogram segments, dividing work by threads.
    for segment_start in (0..l).step_by(n_threads) {
        // go through each thread's output.
        for thread_output in all_outputs.iter() {
            // add each thread's piece to our mel spectrogram.
            for offset in 0..n_threads {
                let mel_index = segment_start + offset; // find location in mel.
                if mel_index < mel.len() {
                    // Make sure we don't go out of bounds.
                    mel[mel_index] += thread_output[mel_index];
                }
            }
        }
    }
```

in general the pattern distributes the work (chunks of the input) across the threads with different offsets and the output values for each thread is offset but that amount too. printing them before joining has the following pattern
```
[-10.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -4.4549956, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
[0.0, -10.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -2.9593704, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
[0.0, 0.0, -10.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.7152524, 0.0, 0.0, 0.0, 0.0, 0.0]
[0.0, 0.0, 0.0, -9.601439, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.808778, 0.0, 0.0, 0.0, 0.0]
[0.0, 0.0, 0.0, 0.0, -7.369365, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.279217, 0.0, 0.0, 0.0]
[0.0, 0.0, 0.0, 0.0, 0.0, -3.5668745, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -4.3706155, 0.0, 0.0]
[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.502727, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.2597005, 0.0]
[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.904054, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -2.861027]
[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.7282937, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.0264673, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.0763023, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -3.3680458, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

drbh:
@LaurentMazare thank you!!

PR #1678, opened 2024-02-08
feat: support microphone whisper streaming

drbh:
This PR adds support for transcribing from a microphone with the following command

```bash
cargo run \
    --release \
    --example whisper \
    --features metal,microphone -- \
    --model medium.en \
    --microphone
```

This PR's relies on https://github.com/huggingface/candle/pull/1674 to reduce the `pcm_to_mel` time.

technical note** a buffer is filled with audio samples and is fed into whisper the output of each decode is immediately printed on scree. Currently the UI does not attempt to consolidate transcription into a single string, further UI improvements can/should be made to make the output as useful as possible

## Example use

whisper-streaming


drbh:
bumping because this is related to https://github.com/huggingface/candle/pull/1674

LaurentMazare:
Would you mind putting this in a separate example e.g. whisper-microphone rather than in the main whisper example even if it duplicates some code? The idea is for examples to be easy to understand and tweak to ones own nee, having different feature flags in them makes it quite harder to follow.

candle-examples/examples/whisper-microphone/melfilters.bytes
LaurentMazare, line 1: We probably want to avoid copying these files as they are pretty large, we could instead just use the ones from the whisper example directly.

candle-examples/examples/whisper-microphone/whisper_micro.rs
```diff
@@ -0,0 +1,674 @@
...
+        if len > max_len {
+            let mut data = audio_ring_buffer.lock().unwrap();
+            let new_data = data[data_len..].to_vec();
+            *data = new_data;
+        }
+    });
+
+    // loop to process the audio data
+    println!("Recording audio for 60 seconds");
+    for i in 0..60 {
```
LaurentMazare, line 602: Maybe we want an infinite loop instead, and the code can just be interrupted by the user?

LaurentMazare (approved):

LaurentMazare:
Thanks!

drbh:
@LaurentMazare thank you!!

PR #1680, opened 2024-02-08
Remove unnecessary memset for reshape, concat and contiguous & support inplace op

guoqingbao:
Some tensor operations (e.g., reshape, concat, and contiguous) have **redundant memset operations**. This occurs when alloc_zeros is called unnecessarily to fill the destination buffer. This patch introduces an alloc operation with an optional value for buffer filling. By default, the alloc function will call device::alloc if there is no value to fill. This change will prevent unnecessary memset calls for certain tensor operations, including reshape, concat, and contiguous, thereby **speeding up the inference process**.

To streamline the implementation, it would be more intuitive to incorporate **an alloc_values function in the external crate "cudarc"**. This function would accept the optional init_value parameter. If init_value is None, the function would allocate an uninitialized buffer; otherwise, it would allocate a buffer and memset it with init_value (see changes in cuda_backend.rs -> alloc_impl). This would eliminate the need for separate implementations such as ones_impl, zeros_impl, and so on.


guoqingbao:
This patch will also support in-place custom ops by making the return values of CustomOp1, CustomOp2 and CustomOp3 optional (for inplace operations where no return tensor(s) are required).

EricLBuehler:
@guoqingbao, have you looked into this further to verify performance improvements? I am working on [Mistral.rs](https://github.com/EricLBuehler/mistral.rs), which provides an OpenAI API compatible HTTP server and includes models such as Mistral and X-LoRA. However, I noticed a 2x fewer completion tokens per second compared to llama.cpp. 

I saw you got 200% performance increases, have you released this code? I am very interested in such developments.

guoqingbao:
> @guoqingbao, have you looked into this further to verify performance improvements? I am working on [Mistral.rs](https://github.com/EricLBuehler/mistral.rs), which provides an OpenAI API compatible HTTP server and includes models such as Mistral and X-LoRA. However, I noticed a 2x fewer completion tokens per second compared to llama.cpp.
> 
> I saw you got 200% performance increases, have you released this code? I am very interested in such developments.

I have ported this project to a custom DSA and made some fused kernels including softmax, layernorm, rotary embedding, kvconcat, etc. The fused kernels were not written in CUDA, but I thought it could produce similar results on NV GPU. You may refer to the subproject (candle-gcu) under this candle-gcu repo: https://github.com/guoqingbao/Candle-GCU . While I haven't released the kernel (ubridge) and runtime (UHHI) code,  but you may refer to the changes of making fused kernels for candle https://github.com/guoqingbao/candle/commits/c294c8befad2d82ab71f8502c4206c7e2c74fcf3/.

guoqingbao:
> @guoqingbao, have you looked into this further to verify performance improvements? I am working on [Mistral.rs](https://github.com/EricLBuehler/mistral.rs), which provides an OpenAI API compatible HTTP server and includes models such as Mistral and X-LoRA. However, I noticed a 2x fewer completion tokens per second compared to llama.cpp.
> 
> I saw you got 200% performance increases, have you released this code? I am very interested in such developments.

Do you have any plans to continue the candle-vllm project? I've been working on the batched inference for candle (the current implement is not suitable for batched computing because of their naive CUDA kernels) and vllm approach seems very attractive. I achieved 20x throughput on LLaMa2 7B using fused kernels when the batch size was larger than 64 in compared to single batch inference on my custom DSA. I thought it could achieve higher performance under vllm approach (kv cache properly managed and pageattention used).

LaurentMazare:
I'm not sure which part of vllm you're after but note that if you want to use the paged attention kernels from candle, you can try [candle-paged-attention](https://github.com/huggingface/candle-paged-attention).

guoqingbao:
> I'm not sure which part of vllm you're after but note that if you want to use the paged attention kernels from candle, you can try [candle-paged-attention](https://github.com/huggingface/candle-paged-attention).

That's awesome! Thanks for the info, I will try this on my DSA.

EricLBuehler:
@guoqingbao , Mistral.rs is a continuation of candke-vllm and will likely include some Paged Attention features. Do you have benchmarks for your library?

guoqingbao:
> @guoqingbao , Mistral.rs is a continuation of candke-vllm and will likely include some Paged Attention features. Do you have benchmarks for your library?

Thanks, I will keep up with the updates of your new project. 

We use candle examples as our bechmarks, i. e., the speed of end2end token generation.

EricLBuehler:
I would really like to benchmark your library, how can I find/run the examples?

guoqingbao:
> I would really like to benchmark your library, how can I find/run the examples?

The examples are exactly the same as candle except we make some changes to use fused kernels. See the candle-gcu project mentioned before. It is not possible to run the entire project on NV GPU because it was modified for  a custom DSA. We didn't release the another two subprojects (ubridge and UHHI) because of the open-source of our internal projects to public is still under negotiation. The two subprojects used fused kernels written in different format (other than CUDA) and it requires a custom DSA hardware to run.

PR #1684, opened 2024-02-09
Add the Qwen2 model

LaurentMazare:


julien-c:
cool model to support!

PR #1703, opened 2024-02-13
Fixing quantized llama demo on metal, by removing all hardcoded values for Cpu.

Narsil:
- Decided to add surface in order to change only internal functions, we could also pass device
to more functions instead.

### DIFF
--- a/candle-core/src/quantized/ggml_file.rs
+++ b/candle-core/src/quantized/ggml_file.rs
@@ -233,6 +233,7 @@ pub struct Content {
     pub hparams: HParams,
     pub vocab: Vocab,
     pub tensors: HashMap<String, super::QTensor>,
+    pub device: Device,
 }
 
 impl Content {
@@ -252,11 +253,13 @@ impl Content {
             let (name, tensor) = read_one_tensor(reader, magic, device)?;
             tensors.insert(name, tensor);
         }
+        let device = device.clone();
         Ok(Self {
             magic,
             hparams,
             vocab,
             tensors,
+            device,
         })
     }
 
--- a/candle-core/src/quantized/metal.rs
+++ b/candle-core/src/quantized/metal.rs
@@ -14,6 +14,10 @@ impl QMetalStorage {
         self.dtype
     }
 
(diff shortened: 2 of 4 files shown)


Vaibhavs10 (approved):
Ah! much more elegant than what I was doing, explicitly transferring to device everywhere else.

Just tested it works perfectly

```
cargo run --example quantized --features metal --release -- --which 7b-chat --prompt "What is the meaning of life according to a dog?" --sample-len 100
```
```
   Compiling candle-core v0.4.0 (/Users/vb/Documents/hf_projects/candle/candle-core)
   Compiling candle-nn v0.4.0 (/Users/vb/Documents/hf_projects/candle/candle-nn)
   Compiling candle-datasets v0.4.0 (/Users/vb/Documents/hf_projects/candle/candle-datasets)
   Compiling candle-transformers v0.4.0 (/Users/vb/Documents/hf_projects/candle/candle-transformers)
   Compiling candle-examples v0.4.0 (/Users/vb/Documents/hf_projects/candle/candle-examples)
    Finished release [optimized] target(s) in 31.69s
     Running `/Users/vb/Documents/hf_projects/candle/target/release/examples/quantized --which 7b-chat --prompt 'What is the meaning of life according to a dog?' --sample-len 100`
avx: false, neon: true, simd128: false, f16c: false
loaded 291 tensors (3.79GB) in 2.03s
params: HParams { n_vocab: 32000, n_embd: 4096, n_mult: 256, n_head: 32, n_layer: 32, n_rot: 128, ftype: 2 }
model built
What is the meaning of life according to a dog?
ова expects the meaning of life to be simple and straightforward, much like a dog’s worldview. Dogs are creatures of habit and routine, and they live in the present moment without much concern for the future or past. They are content with their lives as they are, and they do not worry about big philosophical questions like “what is the meaning of life?”
According to this perspective, the meaning of life for a dog would be something like:
1. Find

  12 prompt tokens processed: 8.18 token/s
  99 tokens generated: 7.79 token/s
```

LaurentMazare (approved):

PR #1706, opened 2024-02-13
feat: add silu activation function

OlivierDehaene:


candle-core/src/backprop.rs
```diff
@@ -589,6 +589,13 @@ impl Tensor {
                         let relu_grad = arg.ge(&arg.zeros_like()?)?.to_dtype(arg.dtype())?;
                         *sum_grad = sum_grad.add(&(&grad * relu_grad)?)?
                     }
+                    Op::Unary(arg, UnaryOp::Silu) => {
+                        let sum_grad = grads.or_insert(arg)?;
+                        // d/dx silu = sigmoid(x) * (1 + x * (1 - sigmoid(x)))
+                        let sigmoid_arg = (arg.neg()?.exp()? + 1.0)?.recip()?;
```
LaurentMazare, line 595: It would probably be more efficient to compute the sigmoid based on the silu value divided by arg rather than having to do another exponential etc.

LaurentMazare: There is no need to recompute the silu value as it's already checkpointed, I think it's named `node` at this place.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1717, opened 2024-02-15
Metavoice - first cut

LaurentMazare:
Tentative to add the [MetaVoice-1B](https://github.com/metavoiceio/metavoice-src) model


candle-transformers/src/models/metavoice.rs
```diff
@@ -0,0 +1,316 @@
...
+#[derive(Debug, Clone)]
+pub struct GPTConfig {
+    pub block_size: usize,
+    pub vocab_sizes: Vec<usize>,
+    pub n_layer: usize,
+    pub n_head: usize,
+    pub n_embd: usize,
+    pub bias: bool,
+    pub causal: bool,
+    pub spk_emb_on_text: bool,
```
vatsalaggarwal, line 121: nit: this is legacy part of our code, we don't use this anymore!

candle-transformers/src/models/metavoice.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub norm_type: NormType,
```
vatsalaggarwal, line 122: nit: this is legacy part of our code, we don't use this anymore! Everything released should be RMSNorm only

candle-transformers/src/models/metavoice.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub rmsnorm_eps: f64,
+    pub nonlinearity_type: NonLinearityType,
```
vatsalaggarwal, line 124: nit: this is legacy part of our code, we don't use this anymore! Everything released should be SwiGLU only

candle-transformers/src/models/metavoice.rs
```diff
@@ -0,0 +1,316 @@
...
+// https://github.com/metavoiceio/metavoice-src/blob/11550bb4e8a1ad032cc1556cc924f7a4e767cbfa/fam/llm/layers/attn.py#L18
+struct SelfAttention {
+    c_attn: Linear,
+    c_proj: Linear,
+    n_head: usize,
+}
+
+impl SelfAttention {
+    fn new(cfg: &GPTConfig, vb: VarBuilder) -> Result<Self> {
+        // The different attention variants are likely to be identical but still we only accept
```
vatsalaggarwal, line 192: yep, they are identical!

candle-transformers/src/models/metavoice.rs
```diff
@@ -0,0 +1,316 @@
...
+        let xs = (&xs + xs.apply(&self.ln_2)?.apply(&self.mlp))?;
+        Ok(xs)
+    }
+}
+
+// https://github.com/metavoiceio/metavoice-src/blob/11550bb4e8a1ad032cc1556cc924f7a4e767cbfa/fam/llm/model.py#L79
+#[allow(unused)]
+#[allow(clippy::upper_case_acronyms)]
+pub struct GPT {
+    wtes: Vec<candle_nn::Embedding>,
```
vatsalaggarwal, line 317: nit: this is an artefact of some legacy code on our part, we're doing single hierarchy flattened interleaved output, a single Embedding layer should be fine

vatsalaggarwal (commented):
The speaker encoder defn looks good! A couple of things to check once they're implemented: i) how audio loading is working, ii) how mel-spectrogram calculation happens, iii) loudness normalization between the two steps above

vatsalaggarwal (commented):
Nice!! Should we chat at vatsal@themetavoice.xyz - happy to answer any questions or to help this along! 

meanwhile i've left some nits that may / may not be helpful... some additional things to think about (down the line?):
- How would encodec token decoding (either via encodec RQVAE decoder / MBD + DFN/ Vocos) be handled? 
- There needs to be an adaption of the output formats between the first stage (which predicts the first 2 hierachies in a flattened interleaved manner) and second stage (which predicts 6 out of the 8 last hierarchies from the first two hierarchies)

vatsalaggarwal:
Just out of curiosity, is there any benchmarking on candle performance vs. using ANE after converting via `coremltools`?

LaurentMazare:
Thanks for looking into this, the comments on the PR are definitely helpful.
> * How would encodec token decoding (either via encodec RQVAE decoder / MBD + DFN/ Vocos) be handled?

That's indeed a large missing bits, I'm not sure which route would be the easiest: the best would be to adapt a standalone python implementation if there was one but I'm afraid we would have to extract the appropriate bits  from larger libraries, e.g. audiocraft. Do you have any plans on making a python version without the audiocraft/... dependency by any chance?

> * There needs to be an adaption of the output formats between the first stage (which predicts the first 2 hierachies in a flattened interleaved manner) and second stage (which predicts 6 out of the 8 last hierarchies from the first two hierarchies)

Would you have some pointer on where this happens on the python side? I totally overlooked this :)

vatsalaggarwal:
> Do you have any plans on making a python version without the audiocraft/... dependency by any chance? The kind of thing that would be really tricky to have are ops that are very specialised for audio and not supported in candle, e.g. fft.

We want to do this because audiocraft / xformers ops also cause headaches for folks installing our repo, but I don't think it'll be soon.

I don't understand the constrains here as I'm not familiar with `candle`, but I think the original RQVAE encodec decoder should probably be the easiest to extract if we want to get the ball rolling on this as **iirc** it's in a separate repo, and the components used are simpler (quantizer LUP table + some LSTMs + convolutions)(https://github.com/facebookresearch/encodec/blob/main/encodec/modules/seanet.py#L147)... 

Vocos requires the ISTFT operation, and MBD is a bit more complicated due to audiocraft/diffusion/etc

I think we'll need to do a FFT regardless as the speaker encoder is trained on mel-spectrogram (https://github.com/metavoiceio/metavoice-src/blob/main/fam/quantiser/audio/speaker_encoder/audio.py#L10).

I'm not fully sure what the constraints on calculating a STFT in `candle` are? If we need a rust implementation, can we use something like https://github.com/ejmahler/RustFFT (for the speaker embedding, you need to do it once per speaker, so I think this could be on CPU and it would be fine)? Alternatively I don't know if we can use the [STFT PyTorch code](https://github.com/pytorch/pytorch/blob/b381a4372b627c6179cea1d048e6426dccea9fd5/aten/src/ATen/native/SpectralOps.cpp#L826) somehow?

> Would you have some pointer on where this happens on the python side? I totally overlooked this :)

Yep:
- text token offsetting: https://github.com/metavoiceio/metavoice-src/blob/main/fam/quantiser/text/tokenise.py#L14-L20
- converting the format of the output of the first stage into the format expected as input by the second stage: https://github.com/metavoiceio/metavoice-src/blob/main/fam/llm/adapters/flattened_encodec.py#L8

vatsalaggarwal:
also @FL33TW00D shared https://github.com/FL33TW00D/ratchet/blob/master/crates/ratchet-models/src/whisper/spectrogram.rs

LaurentMazare:
Converting audio to mel should be ok as we already added this for whisper (though it's probably a very naive implementation). Is there an easy way to use run the model locally using the RQVAE encodec decoder? I'm under the impression that the default paths go through MultiBandDiffusion for the last decode step [here](https://github.com/metavoiceio/metavoice-src/blob/3758971b6501f453d79c95a78fd48d23d3e7cb7b/fam/llm/decoders.py#L84) and haven't found a way/flag/... to use one of the other decoders.

In a nutshell I would like to get the python implementation to work locally with this decoder if it's the simplest so that I can line up the values with the candle implementation. Also if it uses a separate weight file, curious where to find it.

vatsalaggarwal:
Yep, there is no code path for doing that in the released code... we know it works but causes significant degradation, so we explicitly don't allow using it... can try and send a branch with it in a bit...

> Also if it uses a separate weight file, curious where to find it

Didn't get this. It doesn't change any of the weights for MetaVoice-1B, but replaces the MBD... the weights for the rqvae decoder are available here I believe: https://huggingface.co/facebook/encodec_24khz/tree/main

LaurentMazare:
> Didn't get this. It doesn't change any of the weights for MetaVoice-1B, but replaces the MBD... the weights for the rqvae decoder are available here I believe: https://huggingface.co/facebook/encodec_24khz/tree/main

Ah sorry if this was confusing, I thought that maybe you had trained a specific version of encodec but if that's the default setting then it's all good (turns out that I'm working with the main encodec author nowadays, small world :) and it should help if I run into any issues).

vatsalaggarwal:
> Ah sorry if this was confusing, I thought that maybe you had trained a specific version of encodec but if that's the default setting then it's all good (turns out that I'm working with the main encodec author nowadays, small world :) and it should help if I run into any issues).

No, we wanted to make it easy for folks to finetune etc, so for the released model these weights were kept fixed... that's cool :)

LaurentMazare:
As a first step, I've just merged #1771 which is a very basic encodec model. It's decoder only for now, with some shortcuts taken but it can decode some encodec tokens generated by the python encoder and the resulting wav file sound reasonable.
Next step would be to generate these token with the metavoice model obviously, so if you have any kind of python code for this that would be great.

vatsalaggarwal:
Thanks a lot, and apologies for the long delay here! I've put the changes into a branch, and you can see the diff at https://github.com/metavoiceio/metavoice-src/commit/929c3f395238fef3bc8d677308a67f43cdf28c36 ... I've listened to the synthesis, and it sounds sensible... 

meanwhile, just wanted to expectations for folks on this thread: the audio sounds extremely "buzzy" and "muffled" with the RVQ decoder... so we may need to work on getting a better decoder in!

LaurentMazare:
> I've put the changes into a branch, and you can see the diff at [metavoiceio/metavoice-src@929c3f3](https://github.com/metavoiceio/metavoice-src/commit/929c3f395238fef3bc8d677308a67f43cdf28c36) ... I've listened to the synthesis, and it sounds reasonable...

Great, thanks a lot for making this branch this is super helpful.
I've got a first version that is kind of working so I'm going to merge this PR. Note that I've cut a bunch of corners (e.g. speaker embeddings are hardcoded for now), and that the generated audio is significantly worse than what the Python version does, I'll make some further PRs to line things up more between the Python and the candle version - though it's a bit tricky to do because of the randomness of sampling, I'll probably have to set the temperature to 0 where possible.

LaurentMazare:
Also just to mention that the indications to try it out are in [this example](https://github.com/huggingface/candle/tree/main/candle-examples/examples/metavoice).

vatsalaggarwal:
Nice! Awesome to see this working! Any ideas what the speed is like on Apple GPUs?

Issue #1699, opened 2024-02-12
Copy to GPU Too Slow


helenzhangyc:
I noticed that when loading a model onto the GPU, it copies to the CPU first and then onto the GPU. Is there a way that I can load the model directly onto GPU?
```
pub(crate) fn storage<A: NdArray>(&self, array: A) -> Result<Storage> {
        match self {
            Device::Cpu => Ok(Storage::Cpu(array.to_cpu_storage())),
            Device::Cuda(device) => {
                let storage = array.to_cpu_storage();
                let storage = device.storage_from_cpu_storage(&storage)?;
                Ok(Storage::Cuda(storage))
            }
            Device::Metal(device) => {
                let storage = array.to_cpu_storage();
                let storage = device.storage_from_cpu_storage(&storage)?;
                Ok(Storage::Metal(storage))
            }
        }
    }
```

helenzhangyc:
In `backend.rs`, I saw
```
// Maybe this should return a Cow instead so that no copy is done on the cpu case.
fn to_cpu_storage(&self) -> Result<CpuStorage>;
```
I'm not too familiar with the repo and the Rust language, so any ideas or instructions on how to implement this would be great!

ivarflakstad:
Going from `array: A` (which could for example be a `Vec<f32>`) to `CpuStorage::F32(Vec<f32>)` is likely not the root cause for why you're experiencing slow copy to GPU.
I've been playing around with the concept of `trait TensorSource` (conceptually similar to `NdArray`, but solves a different issue), and adding `device.storage_from_tensor_source(source)`. I did see that this reduced the CPU memory usage, but it did not have much impact on the overall performance.

PR #1736, opened 2024-02-20
feat: Return Cow on CpuStorage

junjunjd:
Return a Cow from to_cpu_storage to avoid unnecessary copy.

Address one of the issues in https://github.com/huggingface/candle/issues/1699

### DIFF
--- a/candle-core/src/backend.rs
+++ b/candle-core/src/backend.rs
@@ -1,5 +1,6 @@
 use crate::op::{BinaryOpT, CmpOp, ReduceOp, UnaryOpT};
 use crate::{CpuStorage, DType, Layout, Result, Shape};
+use std::borrow::Cow;
 
 pub trait BackendStorage: Sized {
     type Device: BackendDevice;
@@ -10,8 +11,7 @@ pub trait BackendStorage: Sized {
 
     fn device(&self) -> &Self::Device;
 
-    // Maybe this should return a Cow instead so that no copy is done on the cpu case.
-    fn to_cpu_storage(&self) -> Result<CpuStorage>;
+    fn to_cpu_storage(&self) -> Result<Cow<'_, CpuStorage>>;
 
     fn affine(&self, _: &Layout, _: f64, _: f64) -> Result<Self>;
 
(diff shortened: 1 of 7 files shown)


junjunjd:
@LaurentMazare Could I get a review on this small MR?

LaurentMazare:
Would you mind adding some benchmarks to show how much this improves on GPU model loading, ideally with the details of the config this was run on? This would be helpful to check how much of an improvement this would actually bring.

junjunjd:
@LaurentMazare Sure, I will work on that.

aorticweb:
cargo build fails for me on 

command: cargo build --features metal
```
error[E0053]: method `to_cpu_storage` has an incompatible type for trait
   --> candle-core/src/metal_backend.rs:327:33
    |
327 |     fn to_cpu_storage(&self) -> Result<CpuStorage> {
    |                                 ^^^^^^^^^^^^^^^^^^
    |                                 |
    |                                 expected `Cow<'_, CpuStorage>`, found `CpuStorage`
    |                                 help: change the output type to match the trait: `std::result::Result<Cow<'_, CpuStorage>, error::Error>`
    |
note: type in trait
   --> candle-core/src/backend.rs:14:33
    |
14  |     fn to_cpu_storage(&self) -> Result<Cow<'_, CpuStorage>>;
    |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: expected signature `fn(&metal_backend::MetalStorage) -> std::result::Result<Cow<'_, CpuStorage, >, _>`
               found signature `fn(&metal_backend::MetalStorage) -> std::result::Result<CpuStorage, _>`

For more information about this error, try `rustc --explain E0053`.
error: could not compile `candle-core` (lib) due to previous error
```

perhaps the changes need to be also implemented in metal (and cude backends)?

junjunjd:
@aorticweb Thanks. This is fixed now.

ivarflakstad:
This will have basically no effect on the performance issue mentioned in #1699 . It may have a minor improvement on CPU performance.
I don't immediately see any issues with returning `Cow<CpuStorage>`, but since it's been 2 years and there are quite a few conflicts that need resolving I'll close the PR.

PR #1742, opened 2024-02-21
Add  grads for interpolate1d

KGrewal1:
This adds gradients for interpolate1d: this is done in the same manner as for interpolate2d by using a convolution to sum the grads for integer scaling factors, and checked manually in a similar way


LaurentMazare (approved):

LaurentMazare:
Nice, thanks for adding this!

PR #1743, opened 2024-02-22
add quantized rwkv v5 model

nkypy:
this add quantized rwkv v5 model, use gguf file by tensor-tools


LaurentMazare:
Thanks for putting this up, I've made a bunch of tweaks integrating the model in the main RWKV example so as to reduce duplication.

PR #1745, opened 2024-02-22
Make the cache for the llama model explicit too.

LaurentMazare:


soupslurpr:
Does this mean we can precompute a prompt and only have to compute new tokens added to it instead of the whole prompt every time? If so, could this be done the same way for RWKV by saving the State with the prompt loaded to a file and then loading the State from the file?

PR #1799, opened 2024-03-04
Add attention mask to Bert, fix BGE batch

### LINKED ISSUE #1798: Bert: Attention mask is required for embedding calculation on BGE model
Attention mask, as already reported here https://github.com/huggingface/candle/issues/1552 is not available.

However, that issue was closed because bert is used only for computing embeddings.

But the attention mask is required for proper calculation of many sentences at once, otherwise, results will not be correct.

I created a repository that reproduces it. https://github.com/lucasavila00/candlerepro

Main Logs: https://github.com/lucasavila00/candlerepro/blob/main/log.log

Cheking the logs of the main function shows us that if we process one item at a time, we get the right results.
Also, it shows that my implementation for batching (basically copied from the examples) adds the right padding.
And it shows that the batched implementation, if using just one item, is also correct.

However, if 2 items are used, the paddings will affect results due to missing attention mask.

The results were also checked against this Python implementation: https://github.com/lucasavila00/candlerepro/blob/main/test.ipynb

It shows all values match, however the tokenizer in Python creates the required attention mask and passes it to the model call.

lucasavila00:
Closes #1798

### DIFF
--- a/candle-transformers/src/models/bert.rs
+++ b/candle-transformers/src/models/bert.rs
@@ -112,6 +112,24 @@ impl Config {
     }
 }
 
+// https://github.com/huggingface/transformers/blob/6eedfa6dd15dc1e22a55ae036f681914e5a0d9a1/src/transformers/modeling_utils.py#L877C1-L910C39
+fn get_extended_attention_mask(attention_mask: &Tensor) -> Result<Tensor> {
+    // extended_attention_mask = attention_mask[:, None, None, :]
+
+    let extended_attention_mask = attention_mask.unsqueeze(1)?.unsqueeze(2)?;
+
+    // extended_attention_mask = (1.0 - extended_attention_mask) * torch.finfo(dtype).min
+    let shape = extended_attention_mask.shape();
+
+    let on_true =
+        Tensor::new(0 as f32, extended_attention_mask.device())?.broadcast_as(shape.dims())?;
+    let on_false = Tensor::new(f32::NEG_INFINITY, extended_attention_mask.device())?
+        .broadcast_as(shape.dims())?;
+
+    let m = extended_attention_mask.where_cond(&on_true, &on_false)?;
+    Ok(m)
+}
+
 struct Dropout {
     #[allow(dead_code)]
     pr: f64,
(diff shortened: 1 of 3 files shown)


candle-transformers/src/models/bert.rs
```diff
@@ -232,8 +250,8 @@ impl BertSelfAttention {
     }
 }
 
-impl Module for BertSelfAttention {
-    fn forward(&self, hidden_states: &Tensor) -> Result<Tensor> {
+impl BertSelfAttention {
+    fn forward(&self, hidden_states: &Tensor, attention_mask: &Option<Tensor>) -> Result<Tensor> {
```
lucasavila00, line 254: What is the best approach to add this extra parameter? I fear removing the trait impl is not a good idea?

candle-transformers/src/models/bert.rs
```diff
@@ -112,6 +112,24 @@ impl Config {
...
+    let shape = extended_attention_mask.shape();
+
+    let on_true =
+        Tensor::new(0 as f32, extended_attention_mask.device())?.broadcast_as(shape.dims())?;
+    let on_false = Tensor::new(f32::NEG_INFINITY, extended_attention_mask.device())?
+        .broadcast_as(shape.dims())?;
+
+    let m = extended_attention_mask.where_cond(&on_true, &on_false)?;
+    Ok(m)
+}
```
lucasavila00, line 131: I assume this is very unoptimized? I'm not that used to candle matrix functions :/

candle-transformers/src/models/bert.rs
```diff
@@ -244,7 +262,11 @@ impl Module for BertSelfAttention {
         let value_layer = self.transpose_for_scores(&value_layer)?;
 
         let attention_scores = query_layer.matmul(&key_layer.t()?)?;
-        let attention_scores = (attention_scores / (self.attention_head_size as f64).sqrt())?;
+        let mut attention_scores = (attention_scores / (self.attention_head_size as f64).sqrt())?;
+        if let Some(attention_mask) = attention_mask {
+            let shape = attention_scores.shape().clone();
```
lucasavila00, line 267: Should shape be cloned here?

lucasavila00:
Feel free to finish the MR, request edits, or close it

I can't improve it on my own anymore and I'll leave it as a draft

PR #1807, opened 2024-03-05
Improve metal buffer usage

ivarflakstad:
* Simplify new buffer with data approach.
* Use powers of two for buffer sizes so reuse is more likely.
* Select best available buffer by size.
* Add count to MetalStorage -> facilitates using buffers of different size
* Increase default compute_per_buffer to 50 (this should ideally be calculated wrt GPU capabilities and the model being run)

Tested Phi 1.5 on M3 18gb RAM:
Initial load 4.5s -> 2.8s
Total memory usage 9.9 to 7.4 gb

Observed lower memory usage as inference progresses. Ref the flatter graph below. Worth noting that the y-axis is smaller in the front window.

image

Similar results observed on the M1 Pro 32gb. (Thanks for testing @aorticweb 🤗 )

### DIFF
--- a/candle-core/src/metal_backend.rs
+++ b/candle-core/src/metal_backend.rs
@@ -9,7 +9,7 @@ use metal::{Buffer, CommandBuffer, CommandQueue, MTLResourceOptions, NSUInteger}
 use std::collections::HashMap;
 use std::ffi::c_void;
 use std::path::Path;
-use std::sync::{Arc, Mutex, RwLock, TryLockError};
+use std::sync::{Arc, Mutex, RwLock, RwLockWriteGuard, TryLockError};
 
 /// Simple way to catch lock error without
 /// depending on T
@@ -60,7 +60,8 @@ impl From<String> for MetalError {
     }
 }
 
-type AllocatedBuffers = Arc<RwLock<HashMap<(NSUInteger, MTLResourceOptions), Vec<Arc<Buffer>>>>>;
+type BufferMap = HashMap<(NSUInteger, MTLResourceOptions), Vec<Arc<Buffer>>>;
+type AllocatedBuffers = Arc<RwLock<BufferMap>>;
 
 #[derive(Clone)]
 pub struct MetalDevice {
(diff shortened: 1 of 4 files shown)


candle-core/src/metal_backend.rs
```diff
@@ -204,33 +215,22 @@ impl MetalDevice {
     /// This method will block the computation because of the
     /// lack of lifetime management through the GPU.
     /// Internal comment for technical details.
-    pub fn new_buffer_with_data<T>(&self, data: &[T]) -> Result<Arc<Buffer>> {
-        let size = core::mem::size_of_val(data) as NSUInteger;
+    pub fn new_buffer_with_data<T>(&self, data: Vec<T>) -> Result<Arc<Buffer>> {
```
LaurentMazare, line 218: I'm not sure to understand why passing a `Vec<T>` here is better than the reference. We don't seem to be passing the ownership of the `Vec` to metal as we only extract a pointer out of it, so at the end of the function `data` will be de-allocated which seems unsafe if we don't wait until completion. I'm probably missing something here?

ivarflakstad: Ah, you're right. How about I keep `data` (within MetalDevice) until we know it has been committed/synced?

ivarflakstad: Or maybe we were fine all along 🤔 

[newBufferWithBytes](https://developer.apple.com/documentation/metal/mtldevice/1433429-newbufferwithbytes?language=objc)
> Allocates a new buffer of a given length and initializes its contents by copying existing data into it.
> ...
> Return value: A new [MTLBuffer](https://developer.apple.com/documentation/metal/mtlbuffer?language=objc) instance if the method completes successfully; otherwise nil.

ivarflakstad: For comparison `newBufferWithBytesNoCopy` would be unsafe since it only references the pointer you send in.

LaurentMazare: Cool, can you add a comment in the code with a link to the documentation, that would be great for people looking at it that would wonder why it's safe.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1811, opened 2024-03-07
Metal random-generation bug fixes

niklasha:
The Metal-engine's random-generation is a bit buggy:
* There was a buffer overrun when explicitly seeding the RNG.
* The seed buffer's usage was not properly setup, and the "Read" usage would be masked by the "Write" usage setting.
* In the shader the seed buffer's contents was not used, but its address, leading to determinism across every random generation request.
* Evenly sized tensors had the element with index N/2 deterministically left as zero.

The PR also contains a check that two consecutive random tensors does not repeat.  This is a suboptimal test in that it conceivably could return a false negative, but from my perspective, that is extremely unlikely.

With these changes, the modified test succeeds.

Edit: Removed point about non-deterministic seeding per process.  The proposed change has been removed as per discussion below.
A deterministic output was detected, added a bullet point for that as well.

### DIFF
--- a/candle-core/src/metal_backend.rs
+++ b/candle-core/src/metal_backend.rs
@@ -1685,7 +1685,7 @@ impl BackendDevice for MetalDevice {
         let seed_buffer = self.seed.try_lock().map_err(MetalError::from)?;
         let contents = seed_buffer.contents();
         unsafe {
-            std::ptr::copy([seed].as_ptr(), contents as *mut u32, 4);
+            std::ptr::copy([seed].as_ptr(), contents as *mut u32, 1);
         }
         seed_buffer.did_modify_range(metal::NSRange::new(0, 4));
 
--- a/candle-core/src/metal_backend.rs
+++ b/candle-core/src/metal_backend.rs
@@ -10,7 +10,6 @@ use std::collections::HashMap;
 use std::ffi::c_void;
 use std::path::Path;
 use std::sync::{Arc, Mutex, RwLock, RwLockWriteGuard, TryLockError};
-use std::time::{SystemTime, UNIX_EPOCH};
 
 /// Simple way to catch lock error without
 /// depending on T
@@ -1586,12 +1585,8 @@ impl BackendDevice for MetalDevice {
             Ok(val) => val.parse()?,
             _ => 50,
         };
-        let seed = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("Time went backwards")
-            .as_secs() as u32;
         let seed = Arc::new(Mutex::new(device.new_buffer_with_data(
-            [seed].as_ptr() as *const c_void,
+            [299792458].as_ptr() as *const c_void,
             4,
             MTLResourceOptions::StorageModeManaged,
         )));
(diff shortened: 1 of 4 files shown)


groovybits:
I have suspected the output on my metal gpu was deterministic vs. my llama.cpp and other stable diffusion implementations. Is that basically what is happening without this?

niklasha:
> I have suspected the output on my metal gpu was deterministic vs. my llama.cpp and other stable diffusion implementations. Is that basically what is happening without this?

Yes.  The shader would always use the same value as its seed, not only per process invocation, but for *every* command submission.  I was experimenting with learning candle just using a very simplistic matrix multiplication of two small random tensors, and noticed in my debugging output that both tensor operands were always exactly the same, although generated by two distinct `Tensor::randn` calls.  Took me quite some hours to find all the issues.

LaurentMazare:
I actually don't think we want to seed the rng via a non deterministic way as this will make it trickier to replicate issues/odd behaviors. This is the case for the cpu backend for historical reasons but actually it's not the case for cuda (where most likely this metal seeding has been adapted from) and it seems like a better way to handle things.
If you notice the same tensors being produced, you would want to call `set_seed` on your device to actually change the seed and that's up to the application code to actually do it. E.g. for stable diffusion we should add a `seed` argument to the example so that users can tweak it if they want different images and still get deterministic generation across runs.

niklasha:
> I actually don't think we want to seed the rng via a non deterministic way as this will make it trickier to replicate issues/odd behaviors. This is the case for the cpu backend for historical reasons but actually it's not the case for cuda (where most likely this metal seeding has been adapted from) and it seems like a better way to handle things. If you notice the same tensors being produced, you would want to call `set_seed` on your device to actually change the seed and that's up to the application code to actually do it. E.g. for stable diffusion we should add a `seed` argument to the example so that users can tweak it if they want different images and still get deterministic generation across runs.

That is quite OK with me.  I just noted the discrepancy and thought that the CPU engine was "correct".  That change is completely orthogonal to the rest of this PR.  I will remove it.

ivarflakstad:
> (where most likely this metal seeding has been adapted from)
Correct 🙂

niklasha:
@ivarflakstad Is there a good reason why you truncate the seed to u32?  I see that the set_seed signature accepts an u64, and also that the shader Taus generator seem to accept some more seed bits (referring to the 1, 1 constants of the ulong4 passed to init())?  I have no good knowledge of RNGs, so forgive me if I am just making nonsense here :-)

ivarflakstad:
The [paper](https://indico.cern.ch/event/93877/contributions/2118070/attachments/1104200/1575343/acat3_revised_final.pdf) I based the implementation on uses a mix of int, unsigned int, and unsigned long without specifying why.
I "standardized" the numbers to u32 in an effort to reduce potential unknown side-effects. Using u64 for state and all operations did not improve the randomness anyways.

I went with more seed bits because sometimes the `uint3 [[ thread_index_in_threadgroup ]]` pattern is useful, and you can then use those indices to generate different seeds for different threads, groups, etc.

LaurentMazare (approved):

LaurentMazare:
Merged, thanks!

PR #1812, opened 2024-03-07
Add a --seed argument to the stable-diffusion example.

niklasha:
Per suggestion in https://github.com/huggingface/candle/pull/1811#issuecomment-1983079119

### DIFF
--- a/candle-examples/examples/stable-diffusion/main.rs
+++ b/candle-examples/examples/stable-diffusion/main.rs
@@ -98,8 +98,8 @@ struct Args {
     img2img_strength: f64,
 
     /// The seed to use when generating random samples.
-    #[arg(long, default_value_t = 299792458)]
-    seed: u64,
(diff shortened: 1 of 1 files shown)


ivarflakstad (approved):
👌

LaurentMazare:
I don't think this version would work, `set_seed` fail on CPU currently so we would have to check on the device. I might be missing something though, so please test on CPU and whatever GPU you have to check that it works as expected.

niklasha:
> I don't think this version would work, `set_seed` fail on CPU currently so we would have to check on the device. I might be missing something though, so please test on CPU and whatever GPU you have to check that it works as expected.

Yes I was actually aware of that, but just forgot, let me get back with a check for it.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #1819, opened 2024-03-08
Metal: Improved reduce and softmax

ivarflakstad:
* Separates strided and contiguous reduce impls
* Online softmax impl
* Block and simdgroup reductions

Improvements in throughput on my machine (150GiB/s) using f32 ops 
|            | before    | after     |
|------- | -------- | ------- |
| softmax | 21.453 GiB/s  | 92.261 GiB/s  |
| reduce   | 14.861 GiB/s  | 132.95 GiB/s |

Numerical accuracy. Difference between new and original implementation.
| model        | tokens | max  diff        | avg diff |
|------- | -------- | ------- | ------- |
| llama v32-1b | 100    | 2.9802322e-7 | 6.4991475e-13 |
| llama v32-3b | 100    | 2.3841858e-7 | 5.892976e-13 |
| llama v32-3b | 250    | 2.3841858e-7 | 5.437812e-13 |
| llama v32-3b | 500    | 2.3841858e-7 | 5.134475e-13 |
| phi   v1.5   | 84     | 0.0          | 0.0 |
| phi   v3     | 100    | 0.0          | 0.0 |
| phi   v3     | 250    | 0.0          | 0.0 |
| phi   v3     | 500    | 0.0          | 0.0 |
| helium v1-p  | 100    | 2.3841858e-7 | 1.2197854e-12 |
| helium v1-p  | 250    | 2.3841858e-7 | 1.0073067e-12 |
| helium v1-p  | 500    | 2.3841858e-7 | 9.86934e-13 |

Flux comparison. Left is new, right is old.
flux-comparison


Vaibhavs10 (commented):
cc: @EricLBuehler for vis

FL33TW00D:
image
image


Before and after, such a big improvement speed in decode is worth a look 👀

LaurentMazare:
Looks pretty good indeed, @ivarflakstad what is required before this can be merged / any reason while it's still marked as draft ? (also you may want to include the latest changes to main as there are some conflicting files)

ivarflakstad:
Well I guess it's ready now. If anyone wants to test it on some other models etc feel free 🙇

candle-metal-kernels/src/reduce_old.metal
LaurentMazare, line 1: If this is just a copy of the old `reduce.metal` file and is not used, let's just remove the file and we can always get it from the git history.

candle-metal-kernels/build.rs
LaurentMazare, line 1: How related is this to the reduce/softmax changes? If it's somewhat orthogonal, maybe this should be in a separate PR?

EricLBuehler:
This looks amazing @ivarflakstad!

I tried Llama 3.2 3b on current `main`:
```
cargo run --release --features metal --example llama -- --which v32-3b
    Finished `release` profile [optimized] target(s) in 0.22s
     Running `target/release/examples/llama --which v32-3b`
loading the model weights from meta-llama/Llama-3.2-3B
starting the inference loop
My favorite theorem is 2+2=4 and it’s a damn good one. However, it seems like maybe I may need to change my mind now that I’ve seen this.
The theorem states: The sum of the angles in any triangle add up to 180 degrees.
This sounds great but wait! What about the ones with an odd number?
Great! So does the following:
Okay so we know that the angle sums don’t quite add up in triangles but what if you took all triangles and put them together? Then… drum roll… the angles would add up to exactly 360 degrees!
Let’s start by drawing our first triangle.
Now let’s draw another one next to it:
We’ll do the same with a third, and then a fourth…
Finally we will make one more. (now the third point is on the line created from the 2nd and 4th points)
Well, when I first saw this theorem, it was indeed my favorite until now.
As a teacher of geometry I love taking complex problems and getting students excited about them. Much easier said than done though.

220 tokens generated (20.058381492864637 token/s)
```

But on this branch (e8499c87), I'm getting an error which seems to indicate some numerical issues?

```
cargo run --release --features metal --example llama -- --which v32-3b
    Finished `release` profile [optimized] target(s) in 0.25s
     Running `target/release/examples/llama --which v32-3b`
loading the model weights from meta-llama/Llama-3.2-3B
starting the inference loop
My favorite theorem is 2+2=4 and it’s a Euclidean Division Theorem!
What kind of joking are you referring to? A mathematician might notError: A weight is invalid in distribution
```

candle-metal-kernels/build.rs
ivarflakstad: It is only related in the sense that I wanted to see if I could further improve performance.
With a previous version of this PR it was a noticeable improvement, but I just tested and with the current kernels it's ~1% difference.

candle-core/benches/benchmarks/reduce.rs
```diff
@@ -0,0 +1,239 @@
...
+
+fn run_sum(a: &Tensor) {
+    a.sum_keepdim(2).unwrap();
+}
+fn run_arg_min(a: &Tensor) {
+    a.argmin_keepdim(2).unwrap();
+}
+
+// NOTE: Should this be removed? Softmax impls live in candle-nn.
+fn softmax(a: &Tensor) -> candle_core::Result<()> {
```
ivarflakstad, line 16: @LaurentMazare 
I'll have to remove this bench since this is metal specific (softmax lives in candle-nn ops so can't call it from candle-core), but I think softmax warrants a benchmark somehow. Thoughts?

LaurentMazare: I think moving the benchmark to `candle-nn` would be good, (do it with `git mv` so as to preserve history).

ivarflakstad:
> This looks amazing @ivarflakstad!
> 
> 
> 
> I tried Llama 3.2 3b on current `main`:
> 
> ```
> 
> cargo run --release --features metal --example llama -- --which v32-3b
> 
>     Finished `release` profile [optimized] target(s) in 0.22s
> 
>      Running `target/release/examples/llama --which v32-3b`
> 
> loading the model weights from meta-llama/Llama-3.2-3B
> 
> starting the inference loop
> 
> My favorite theorem is 2+2=4 and it’s a damn good one. However, it seems like maybe I may need to change my mind now that I’ve seen this.
> 
> The theorem states: The sum of the angles in any triangle add up to 180 degrees.
> 
> This sounds great but wait! What about the ones with an odd number?
> 
> Great! So does the following:
> 
> Okay so we know that the angle sums don’t quite add up in triangles but what if you took all triangles and put them together? Then… drum roll… the angles would add up to exactly 360 degrees!
> 
> Let’s start by drawing our first triangle.
> 
> Now let’s draw another one next to it:
> 
> We’ll do the same with a third, and then a fourth…
> 
> Finally we will make one more. (now the third point is on the line created from the 2nd and 4th points)
> 
> Well, when I first saw this theorem, it was indeed my favorite until now.
> 
> As a teacher of geometry I love taking complex problems and getting students excited about them. Much easier said than done though.
> 
> 
> 
> 220 tokens generated (20.058381492864637 token/s)
> 
> ```
> 
> 
> 
> But on this branch (e8499c87), I'm getting an error which seems to indicate some numerical issues?
> 
> 
> 
> ```
> 
> cargo run --release --features metal --example llama -- --which v32-3b
> 
>     Finished `release` profile [optimized] target(s) in 0.25s
> 
>      Running `target/release/examples/llama --which v32-3b`
> 
> loading the model weights from meta-llama/Llama-3.2-3B
> 
> starting the inference loop
> 
> My favorite theorem is 2+2=4 and it’s a Euclidean Division Theorem!
> 
> What kind of joking are you referring to? A mathematician might notError: A weight is invalid in distribution
> 
> ```

Nice, thanks!
Should probably accumulate with float in softmax to preserve precision. Shouldn't affect performance at all.

candle-metal-kernels/src/reduce_old.metal
ivarflakstad: Removed

candle-metal-kernels/build.rs
ivarflakstad: Done. I'd like to bring it back at some point simply because including utils.metal etc is just better QOL.
I feel like the results are slightly better with a precompiled lib as well, but I haven't put it to the test in a scientific manner.

candle-metal-kernels/src/lib.rs
```diff
@@ -566,25 +565,47 @@ pub fn call_cast_strided(
     Ok(())
 }
 
+#[inline]
+fn calculate_reduce_threads(work_per_threadgroup: usize) -> NSUInteger {
+    let work_split = work_per_threadgroup / 2;
+    let mut w = 2;
+    while w < work_split {
```
LaurentMazare, line 572: Why not use `next_power_of_two` as we were doing before?

ivarflakstad: Ah, at one point of debugging I got paranoid hehe
I simplified it to `(work_per_threadgroup / 2).next_power_of_two()` just now

ivarflakstad: The `/ 2` is simply because each thread is reading more than one value

LaurentMazare (approved):

LaurentMazare:
Could you actually try it on a bunch of f16 and bf16 models? I just got a "weigth is invalid in distribution" which doesn't occur on main.

ivarflakstad:
> Could you actually try it on a bunch of f16 and bf16 models? I just got a "weigth is invalid in distribution" which doesn't occur on main.

I've just now tested a couple of llama and phi versions and helium preview. I did not get the error, but that doesn't necessarily prove it is gone.
I thought it was fixed in [this](https://github.com/huggingface/candle/pull/1819/commits/4c949254d1f2cbdb35badd13d118f4439c48e440) commit, can you pull the latest changes and try again? I don't want to merge if you can reproduce the bug.

LaurentMazare:
I was actually testing past this point, the issue doesn't seem to be about accumulating with the proper type but something else: some of the probabilities past the softmax are negative. Interestingly, if I save the logits in a file then just load the logits and apply the softmax in a fresh binary then I don't get the negative values.
You can reproduce the issue with the command Eric gave (using the llama 3.2 1b example with the default prompt). Fwiw I had the issue both on an M2 Pro and a M3 Max.
I think to gain confidence in the change it would be good that you tweak the `sample` function so that it computes the softmax both in the fast and slow way and double check that the values are reasonably in line even on long generation.
Also would be good to test diffusion models e.g. flux or stable diffusion before releasing.

ivarflakstad:
> You can reproduce the issue with the command Eric gave (using the llama 3.2 1b example with the default prompt). 
`cargo run --release --features metal --example llama -- --which v32-3b`
No errors. 1217 tokens generated.
`cargo run --release --features metal --example llama -- --which v32-1b`
No errors. 59 tokens generated.

Could it be that the machine(s) cached the slightly outdated compiled kernels or something?
Seems unlikely, but I'm not able to reproduce.

LaurentMazare:
> Could it be that the machine(s) cached the slightly outdated compiled kernels or something? Seems unlikely, but I'm not able to reproduce.

I just tried again after removing the `target` directory and got into the same error. If you can get access to another mac, that might be worth trying as I had the issue on the 3 different macs that I tried now (a macbook pro M2, a macbook pro M3 max, and a mac mini m4 pro).

ivarflakstad:
Finally got the bug. Had to restart my machine. Let the debugging begin!

ivarflakstad:
Figured it out. I'll push later 🫡

LaurentMazare:
Great! It would also be nice if you can also measure the differences between the optimized and non optimized softmax implementation on a couple models and check that it stays within reasonable bounds.

LaurentMazare (approved):

LaurentMazare:
Thanks, great improvement!

PR #1848, opened 2024-03-14
Improve performance for long sequence generation (kvconcat kernel).

guoqingbao:
The current implementation of key-value concatenation relies on Tensor::cat which uses ucopy kernel to compute every indice for all output tensor elements based on the input shape and output stride (see below) and it also requires double-transpose. This PR implements a kvconcat kernel to bypass the indice compute and therefore it can speed up the generation speed by around 10% - 15%, especially during long sentence generation.

``` c++
__device__ unsigned int get_strided_index(
    unsigned int idx,
    const size_t num_dims,
    const size_t *dims,
    const size_t *strides
) {
    unsigned int strided_i = 0;
    for (unsigned int d = 0; d < num_dims; d++) {
        unsigned int dim_idx = num_dims - 1 - d;
        strided_i += (idx % dims[dim_idx]) * strides[dim_idx];
        idx /= dims[dim_idx];
    }
    return strided_i;
}

for (unsigned int i = blockIdx.x * blockDim.x + threadIdx.x; i < numel; i += blockDim.x * gridDim.x) { 
       unsigned strided_i = get_strided_index(i, num_dims, dims, strides); 
        TYPENAME x = inp ? inp[strided_i] : out[i]; 
         out[i] = FUNC; 
 } 
```

The current implementation, element-wise copy.
``` c++
 //chunk_l equals to chunk_r for dim2 concat
template <typename T>
__device__ __forceinline__ void kvconcat_dim2_kernel(T *ltensor, T* rtensor, T *out,
    const size_t chunk_l, const size_t chunk_r, const size_t lstride, const size_t rstride) {
    int thread_id = GetThreadIdx();
    int out_stride = lstride + rstride;
    int idx = thread_id / out_stride;
    int j = thread_id % out_stride;
    T* pLeft = ltensor + idx * lstride;
    T* pRight = rtensor + idx * rstride;
    T* pOut = out + idx * out_stride;
    if (idx < chunk_l) {
        if (j < lstride)
            pOut[j] = pLeft[j];
        else
            pOut[j] = pRight[j - lstride];
    }
}
```

### DIFF
--- a/candle-transformers/src/models/llama.rs
+++ b/candle-transformers/src/models/llama.rs
@@ -225,8 +225,10 @@ impl CausalSelfAttention {
 
         if cache.use_kv_cache {
             if let Some((cache_k, cache_v)) = &cache.kvs[block_idx] {
-                k = Tensor::cat(&[cache_k, &k], 2)?.contiguous()?;
-                v = Tensor::cat(&[cache_v, &v], 2)?.contiguous()?;
+                // k = Tensor::cat(&[cache_k, &k], 2)?.contiguous()?;
+                // v = Tensor::cat(&[cache_v, &v], 2)?.contiguous()?;
+                k = candle_nn::kvconcat(&cache_k, &k, 2)?;
+                v = candle_nn::kvconcat(&cache_v, &v, 2)?;
                 let k_seq_len = k.dims()[1];
                 if k_seq_len > MAX_SEQ_LEN {
                     k = k
--- a/candle-core/src/cuda_backend.rs
+++ b/candle-core/src/cuda_backend.rs
@@ -2247,3 +2247,163 @@ impl BackendStorage for CudaStorage {
         Ok(())
     }
 }
+
+pub struct KVConcat {
+    pub concat_dim: usize,
+}
+impl crate::CustomOp2 for KVConcat {
+    fn name(&self) -> &'static str {
+        "kvconcat"
+    }
+
+    fn cpu_fwd(
+        &self,
+        _: &CpuStorage,
+        _: &Layout,
+        _: &CpuStorage,
+        _: &Layout,
+    ) -> Result<(CpuStorage, Shape)> {
+        crate::bail!("no cpu support for kvconcat")
+    }
+
+    fn cuda_fwd(
+        &self,
+        ltensor: &CudaStorage,
+        ltensor_l: &Layout,
+        rtensor: &CudaStorage,
+        rtensor_l: &Layout,
+    ) -> Result<(CudaStorage, Shape)> {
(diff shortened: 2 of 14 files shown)


candle-transformers/src/models/llama.rs
```diff
@@ -225,8 +225,10 @@ impl CausalSelfAttention {
 
         if cache.use_kv_cache {
             if let Some((cache_k, cache_v)) = &cache.kvs[block_idx] {
-                k = Tensor::cat(&[cache_k, &k], 2)?.contiguous()?;
-                v = Tensor::cat(&[cache_v, &v], 2)?.contiguous()?;
+                // k = Tensor::cat(&[cache_k, &k], 2)?.contiguous()?;
+                // v = Tensor::cat(&[cache_v, &v], 2)?.contiguous()?;
+                k = candle_nn::kvconcat(&cache_k, &k, 2)?;
```
EricLBuehler, line 230: Perhaps this could be implemented for the quantized_llama.rs example?

guoqingbao: Yes,  it should be workable on the quantized examples as well.

LaurentMazare:
Thanks for suggesting this. I agree that there is currently an issue with how long kv cache takes so I've made #1855 to actually remove the transpose bit when doing `Tensor::cat`, this way it will benefit all operations rather than just kv-cache stuff. It can probably still be improved a bit, e.g. avoiding the remaining modulo or using `cudaMemcpy2D` but it should already brings quite a few benefits without requiring any model change.
There is probably still room for a highly optimized kv-cache kernels, but I would suggest for this making an external crate to handle it and experiment with it, you can already find a bunch of these e.g. [candle-rotary](https://github.com/huggingface/candle-rotary) for fast rotary embeddings etc.

guoqingbao:
> Thanks for suggesting this. I agree that there is currently an issue with how long kv cache takes so I've made #1855 to actually remove the transpose bit when doing `Tensor::cat`, this way it will benefit all operations rather than just kv-cache stuff. It can probably still be improved a bit, e.g. avoiding the remaining modulo or using `cudaMemcpy2D` but it should already brings quite a few benefits without requiring any model change. There is probably still room for a highly optimized kv-cache kernels, but I would suggest for this making an external crate to handle it and experiment with it, you can already find a bunch of these e.g. [candle-rotary](https://github.com/huggingface/candle-rotary) for fast rotary embeddings etc.

Have you tested the kvconcat approach vs #1855 in terms of performance improvements? I think the kvconcat approach is simple and more straightforward, and the changes seem very heavy in #1855. I'm not sure if copy2d kernel in #1855 can benefit operations other than Tensor::cat. The proposed kvconcat approach is very similar to candle_nn::ops::softmax_last_dim which is widely used in candle models.

PR #1849, opened 2024-03-14
Add support for index u8/i64 and input f16/bf16 scatter-add on metal

tomsanbear:
As the title mentions, adds support on the metal backend for u8 and i64 indices, and f16 and bf16 input tensors for the scatter-add operation. Ran into this missing functionality while running the optimizers on my M1.

This PR also adds some coverage for all of these operations across a few scenarios.

I'm not entirely sure this is the correct approach here, nor am I sure if I am missing anything glaring.


LaurentMazare (approved):

LaurentMazare:
Thanks, looks good and great to have added some testing!

PR #1860, opened 2024-03-17
Add support for casting between all supported datatypes on Metal

tomsanbear:
This PR aims to remove the discrepancy of casting DTypes for Metal. I've gone ahead and refactored the tests as well to group them appropriately.

One area of the change that I'm uncertain of is the version support, there was a callout to metal version 220 in the code which I preserved, but I could not find any documentation from Apple on these versions and which casts are supported in what release... Guidance on this would be great!


LaurentMazare (approved):

LaurentMazare:
Thanks, I don't have a good reference for this version so let's keep this for now.

PR #1863, opened 2024-03-18
Add support for max_pool2d for Metal backend

tomsanbear:
Adds support for max_pool2d for metal for the following dtypes:
- f32
- f16
- bf16
- u8
- u32

Essentially a port of the cuda backend implementation but adjusted to the metal backend style.


LaurentMazare:
Thanks for making these changes, they look pretty great. I've made some small tweaks + enabled the toplevel maxpool tests for metal and these worked nicely.
Note that adding avgpool should be very similar if you want to give this a stab too :)

PR #1864, opened 2024-03-18
Whisper microphone example: do not repeatedly acquire and release the microphone

lavafroth:
### Changes
- The input stream from the microphone is built once in the thread that handles it.
- Inside the loop block, the stream is played and paused for the defined duration (300ms) to capture the pcm data.
- The `iter::repeat(()).enumerate()` to keep track of the inference iteration is replaced with a simple `0..` range.


lavafroth:
I have a pretty low end machine but yes it works better than the previous iteration. 😇

-------- Original Message --------

PR #1869, opened 2024-03-18
Add avg_pool2d metal implementation for the metal backend

tomsanbear:
Refactored max and avg to share a common method since the only difference is the op name.

Uncommented the metal system test.

### DIFF
--- a/candle-metal-kernels/src/conv.metal
+++ b/candle-metal-kernels/src/conv.metal
@@ -206,7 +206,7 @@ kernel void FN_NAME(  \
   upsample_nearest2d<TYPENAME>(w_out, h_out, w_scale, h_scale, dims, strides, src, dst, tid); \
 } \
 
-template <typename T>
+template <typename T, typename A>
 METAL_FUNC void avg_pool2d(
     constant size_t &w_k,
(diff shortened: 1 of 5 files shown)


candle-metal-kernels/src/conv.metal
```diff
@@ -206,6 +206,67 @@ kernel void FN_NAME(  \
...
+    return;
+  }
+
+  const size_t b_idx = tid / (w_out * h_out * c);
+  const size_t c_idx = (tid / (w_out * h_out)) % c;
+  const size_t dst_w = (tid / h_out) % w_out;
+  const size_t dst_h = tid % h_out;
+
+  const size_t src_idx0 = b_idx * src_strides[0];
+  T d = 0;
```
LaurentMazare, line 237: In the cuda version, we use a different type for the accumulator so as to get around possible precision issues, e.g. for float16, bfloat16, float32 it would be a float32, for double we would still accumulate in double.

tomsanbear: ahh that's a good idea, I'll modify with this

LaurentMazare:
Great, thanks!

PR #1874, opened 2024-03-19
Add support for conv_transpose1d for metal backend

tomsanbear:
Adds support for the conv_transpose1d function by porting the cuda backend implementation.

Added support for
- f32
- f16
- bf16
- u8
- u32

### DIFF
--- a/candle-core/src/metal_backend.rs
+++ b/candle-core/src/metal_backend.rs
@@ -948,12 +948,56 @@ impl BackendStorage for MetalStorage {
 
     fn conv_transpose1d(
         &self,
-        _l: &Layout,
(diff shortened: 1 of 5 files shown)


LaurentMazare (approved):

LaurentMazare:
Neat thanks, that's a great thing to have. I have some pending PR that significantly optimizes the cuda kernel though it increases memory usage so I'm not 100% sure it's the right call and I'll have to work on it a bit more ( #1859 if you're curious).

fletcherist:
is there any chance to add backward to this?
now it gets Error: backward is not supported for conv-transpose1d

PR #1880, opened 2024-03-19
Extract RotaryEmbedding code for reuse across models.

janimo:
Most models use identical of almost identical copies of RotaryEmbedding (cfg.rope_theta vs hardcoded 10000, rope_theta being f32 or f64). A few others (mixformer, phi, chatglm) are a bit more different implementations.

I did not change Yi to use cfg.rope_theta, it had hardcoded 10.000 while the config has 5.000.000 and I can not test this larger model.


janimo:
Would this make sense elsewhere, like transformers/utils ?

ivarflakstad:
This makes a lot of sense.
It just needs to be updated with `candle_nn::rotary_emb::rope` instead of the manual rope impls (and ofc make sure it's not supposed to be `rope_i` or `rope_thd` variants).

PR #1884, opened 2024-03-20
Avoid copying the data on squeeze and unsqueeze.

LaurentMazare:


LaurentMazare:
This change is likely to break a lot of models as it will result in non-contiguous tensors at various places. On the other end, it avoids some unnecessary copies in transformers that use the qkv layout for the self-attention weights.

PR #1903, opened 2024-03-21
Add support for conv_transpose2d on Metal backend

tomsanbear:
Adds support for the conv_transpose2d backend, opted to represent the list of parameters in a struct this time as the argument list was getting quite long.

Also added a benchmark test for this operation as I was toying with the perf of this operation while implementing.

The nice part of this one is that it seems to be the last "is metal" test flag that was required for the conv test suite!


LaurentMazare (approved):

LaurentMazare:
Amazing thanks, will merge as soon as the CI is done!

PR #1909, opened 2024-03-21
Add support for strided index-select on Metal

tomsanbear:
I'm fairly certain I got the right idea here, however I'm not sure if I've missed some guard clauses or reshaping on the output buffer.

### DIFF
--- a/candle-metal-kernels/src/indexing.metal
+++ b/candle-metal-kernels/src/indexing.metal
@@ -1,20 +1,38 @@
 #include <metal_stdlib>
 using namespace metal;
 
+METAL_FUNC uint get_strided_index(
+    uint idx,
+    constant size_t &num_dims,
+    constant size_t *dims,
+    constant size_t *strides
+) {
+    uint strided_i = 0;
+    for (uint d = 0; d < num_dims; d++) {
+        uint dim_idx = num_dims - 1 - d;
+        strided_i += (idx % dims[dim_idx]) * strides[dim_idx];
+        idx /= dims[dim_idx];
+    }
+    return strided_i;
+}
+
 template<typename TYPENAME, typename INDEX_TYPENAME>
 METAL_FUNC void index( 
     constant size_t &dst_size, 
     constant size_t &left_size, 
     constant size_t &src_dim_size, 
     constant size_t &right_size, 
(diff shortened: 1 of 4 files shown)


candle-metal-kernels/src/indexing.metal
```diff
@@ -26,6 +44,7 @@ METAL_FUNC void index(
     // No need to check for zero we're only allowing unsized. 
     */ 
     const size_t src_i = left_rank_i * src_dim_size * right_size + input_i * right_size + right_rank_i; 
+    const size_t strided_src_i = contiguous ? src_i : get_strided_index(src_i, src_dim_size, src_dims, src_strides);
```
LaurentMazare, line 47: Shouldn't `strided_src_i` be used instead of `src_i` or am I missing something? And if it's an actual issue, shouldn't a test have a failed on this this?

tomsanbear: That's a great point... My test is not doing the right thing with simulating a strided buffer perhaps. Let me take a look at some of our other strided function tests 🤔

tomsanbear: Okay latest push with the test has a case that actually fails as it should with the first commit on this branch, do you mind taking a peak at the test case and let me know if that's the right direction for this?

LaurentMazare (approved):

LaurentMazare:
Thanks. In the future would be great if you can add run some benchmarks and profiling and mention it in the feature description.
For example on cuda kernels passing parameters that are slices, e.g. strides that you added in this PR, has some substantial runtime cost because of the allocation and htod communication that it introduces. I have no intuition on the cost of this on the metal side but it would be great to check that it's negligible.

tomsanbear:
> Thanks. In the future would be great if you can add run some benchmarks and profiling and mention it in the feature description. For example on cuda kernels passing parameters that are slices, e.g. strides that you added in this PR, has some substantial runtime cost because of the allocation and htod communication that it introduces. I have no intuition on the cost of this on the metal side but it would be great to check that it's negligible.

Ah that's a good callout, let me do some research into this and run some benchmarks. I've been wanting to write up a readme for the metal kernel package that has common issues/patterns etc... about metal I've discovered during my time learning.

PR #1915, opened 2024-03-22
Support scatter/index_add with i64 indices for f16

danieldk:
Another very small PR. We need `scatter_add` with `i64` indices in Oxidized Transformers for flash attention, but it was only supported for `bfl16` and not `f16`. So this change adds `i64` indexing as well.


LaurentMazare:
Thanks!

PR #1919, opened 2024-03-23
Add cast_bf16_x/cast_x_bf16 when CUDA_ARCH<800 but CUDA_VERSION >= 11000

yinqiwen:
- it make possible to load bf16 models on T4(sm75)

### DIFF
--- a/candle-kernels/src/cast.cu
+++ b/candle-kernels/src/cast.cu
@@ -83,6 +83,18 @@ CAST_OP(double,   __nv_bfloat16, cast_f64_bf16)
 CAST_THROUGH_OP(__nv_bfloat16, uint8_t, float, cast_bf16_u8)
 CAST_THROUGH_OP(__nv_bfloat16, __half,   float, cast_bf16_f16)
 CAST_THROUGH_OP(__half,   __nv_bfloat16, float, cast_f16_bf16)
+#else
+#include <cuda.h>
+#if CUDA_VERSION >= 11000
+CAST_OP(__nv_bfloat16, float,    cast_bf16_f32)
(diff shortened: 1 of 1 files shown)


LaurentMazare:
I would be a bit afraid that this could break compilation on some architectures. Do you have some pointers with details on bfloat support for cuda version greater than 11?

yinqiwen:
Already test gemma success with cast_bf16_f16 on T4.

According to the CUDA release notes, bf16 is supported as a data type since CUDA 11.0. However, if CUDA_ARCH<800, some operations may not support bf16. Nevertheless, bf16 can still be used as a data type for casting to or from other data types, including fp32.

"CUDA C++ includes support for new data types to support new 16-bit floating point data (with 1-sign bit, 8-bit exponent and 7-bit mantissa): __nv_bfloat16 and __nv_bfloat162. See include/cuda_bf16.hpp and the CUDA Math API for more information on the datatype definition and supported arithmetic operations." -- from [cuda 11.0 release notes](https://docs.nvidia.com/cuda/archive/11.0_GA/cuda-toolkit-release-notes/index.html)

And in other projects:  
- FasterTransformer, it enable bf16 only when cuda > 11.0 (https://github.com/NVIDIA/FasterTransformer/blob/main/CMakeLists.txt#L19)
- OneFlow, it cast through float for bf16 only when cuda > 11.0 (https://github.com/Oneflow-Inc/oneflow/blob/master/oneflow/core/ep/cuda/primitive/cast.cu#L58)

LaurentMazare (approved):

LaurentMazare:
Ah that's great to know, thanks for the details!

PR #1921, opened 2024-03-23
Preliminary support for inplace ops.

LaurentMazare:
This PR adds some preliminary support for inplace ops (standard and custom). These ops are key in high performance application to avoid re-allocating buffers over and over + touch generally less memory, however they are tricky to work with. In particular running a backward pass on a forward that used an inplace op is likely to end up being UB.


zeyutt:
not very clear about current InplaceOp should use UgOp as a common safe warpper or normal kernel can also be used?

PR #1934, opened 2024-03-25
bert attention mask

lz1998:


lz1998:
Is there something like `torch.finfo(dtype).min` in candle?

unlimitedsola:
Probably a duplicate of #1799

EricLBuehler:
> Is there something like torch.finfo(dtype).min in candle?

@lz1998, you can use `f32::NEG_INFINITY` and then cast to the correct Tensor type.

kczimm:
What's the status of this PR? I've noticed that this PR is necessary to match the output of `transformers` or `sentence_transformers` when batch inputs are of different token length.

harryfei:
+1 for this. Please merge it. @LaurentMazare

LaurentMazare:
Merged, thanks and sorry for the delay.

PR #1946, opened 2024-03-27
add config for mamba 2.8b model parameter

jorgeantonio21:
Motivation: Need to have access to Mamba configuration for another repo that depends on `candle`.

Content: Add a function `config_2_8b` for `Mamba`'s `Config`, following https://huggingface.co/docs/transformers/model_doc/mamba.


LaurentMazare:
Thanks for the PR, I've tweaked it a bit so as instead of exposing a hard-coded 2.8b config, we just make the config fields public and it should be easy to build arbitrary configs on your side.
Note that the configs are also `serde::Deserialize` so if you get the config via a json file from the huggingface hub, you can just use the `serde_json` deserialization (which is what we do in `candle-examples` so as to handle all the mamba variants).

PR #1950, opened 2024-03-27
CLIP model implementation with example

Tigranchick:
I'm very impressed how Rust can work in the field of artificial intelligence. I decided to try to port the CLIP model. Mainly focused on https://huggingface.co/openai/clip-vit-base-patch32/tree/main, but in the future I will adapt it to other versions.

I encountered a problem when trying to run from a Mac with the metal feature
```
Error: Metal contiguous index_select I64 F32 not implemented
```
I would like try to implement this functionality, could you suggest me the direction to dig.

I also had some misunderstandings regarding
```python
    self.class_embedding = nn.Parameter(radnn)
```
```rust
    let class_embedding = if vs.contains_tensor("class_embedding") {
             vs.get(c.embed_dim, "class_embedding")?
         } else {
             warn!("class_embedding not found in the. Initializing a new one.");
             Tensor::randn(0.0 as f32, 1.0 as f32, &[c.embed_dim], vs.device())?
         };
```
or
```python
     self.logit_scale = nn.Parameter(torch.tensor(self.config.logit_scale_init_value))
```
```rust
     let logit_scale = if vs.contains_tensor("logit_scale") {
             vs.get(&[], "logit_scale")?
         } else {
             warn!("Creating logit_scale tensor, results may vary.");
             Tensor::new(&[c.logit_scale_init_value], vs.device())?
         };
```

Spent a lot of time debugging this process with misleading use self initialization
Since originally it is nn.Parameter which, as I understand it, is mainly used during training and when loading the model, it is just loaded under the hood to the current ones weights? In this case, is it ok to just load them as shown?

Also, the text embeddings were mainly taken and adapted from stable diffusion model implementation, I will revise it from the current version of the clip.

Unfortunately not tested on gpu, If there are any problems I will try to fix them

My main goal is to create an example for wasm because I think this is one of the main feature of candle


LaurentMazare:
That looks like a great model to add, re the error `Metal contiguous index_select I64 F32 not implemented`, this was missing from the metal backend but I just added it in #1951 so should be all good now. That said, if you can it's probably better to use `u32` for indexing rather than `i64` (indexing with negative values is not supported anyway).

Tigranchick:
Hi! I tested the implementation with metal feature and also add feature to process batch images with fixing some inaccuracies that prevented that.

LaurentMazare:
Looks pretty good, would you mind removing the images? The repo is already a bit too large so you can either use some of the existing images from other examples or add some `wget` command to the batch snippets in the readme so that images actually get fetched.

candle-transformers/src/models/clip/text_model.rs
```diff
@@ -0,0 +1,362 @@
...
+            .encoder
+            .forward(&input_ids, Some(&causal_attention_mask))?;
+        self.final_layer_norm.forward(&input_ids)
+    }
+}
+
+impl Module for ClipTextTransformer {
+    fn forward(&self, input_ids: &Tensor) -> Result<Tensor> {
+        let output = self.forward_with_mask(input_ids, usize::MAX)?;
+        let batch_indices = Tensor::arange(0i64, output.dim(0)? as i64, output.device())?;
```
LaurentMazare, line 341: Could we use u32's here rather than i64? Would be more efficient and better supported.

candle-transformers/src/models/clip/text_model.rs
```diff
@@ -0,0 +1,362 @@
...
+            token_embedding,
+            position_embedding,
+            position_ids,
+        })
+    }
+}
+
+impl Module for ClipTextEmbeddings {
+    fn forward(&self, input_ids: &Tensor) -> Result<Tensor> {
+        let seq_length = input_ids.shape().dims().last().unwrap();
```
LaurentMazare, line 93: `let seq_length = input_ids.dim(D::Minus1)?;` would be a bit simpler.

candle-transformers/src/models/clip/text_model.rs
```diff
@@ -0,0 +1,362 @@
...
+    fn forward(&self, input_ids: &Tensor) -> Result<Tensor> {
+        let output = self.forward_with_mask(input_ids, usize::MAX)?;
+        let batch_indices = Tensor::arange(0i64, output.dim(0)? as i64, output.device())?;
+
+        let sequence_max_indices = input_ids.argmax(D::Minus1)?.to_dtype(DType::I64)?;
+
+        let mut indices: Vec<Tensor> = Vec::new();
+
+        // TODO research slicing capabilities in candle
+        let batch_indices = batch_indices.to_vec1::<i64>()?;
```
LaurentMazare, line 348: Why do we go through a tensor for `batch_indices`? Seems that the loop below could be just an `enumerate` on `sequence_max_indices`.

LaurentMazare (approved):

LaurentMazare:
Merged, thanks a lot for adding this. Don't hesitate to communicate about the availability of the model on twitter/reddit/..., always good so that potential users know that it's possible to use clip now!

Tigranchick:
> Merged, thanks a lot for adding this. Don't hesitate to communicate about the availability of the model on twitter/reddit/..., always good so that potential users know that it's possible to use clip now!

thanks, it was interesting working on that

PR #1959, opened 2024-03-28
Fix: ensure that the metal backend is initialized with a random seed instead of a hardcoded one

tomsanbear:
Not quite sure why this was hardcoded this way, but this at least provides us with proper seeds between executions.


LaurentMazare:
This is actually on purpose, we want the rng to be deterministic for training reproducibility. The cpu backend doesn't have deterministic sampling for performance reasons but the issue is more on this side.

tomsanbear:
> This is actually on purpose, we want the rng to be deterministic for training reproducibility. The cpu backend doesn't have deterministic sampling for performance reasons but the issue is more on this side.

Ah very interesting, thanks for the information!

PR #1964, opened 2024-03-29
adds hugging face authentication token to be able to download llama model weights

jorgeantonio21:
Motivation: I could't run the llama example, as in order to download weights it is permission access is needed. This small PR adds an authentication token field to clap `args`, so the user can optionally pass its HuggingFace auth token and download the model weights.


LaurentMazare:
For this you can use `huggingface-cli login` or alternatively use the `HF_TOKEN` environment variable, see more details [here](https://huggingface.co/docs/huggingface_hub/quick-start#login-command).

jorgeantonio21:
Oh, I see. Appreciate the explanation, @LaurentMazare

PR #1980, opened 2024-04-01
Quantized moondream implementation and BOS token

santiagomed:
Hey @LaurentMazare, 

I managed to quantize moondream to 4 bit precision which resulted in a GGUF file of 2.5GB (from an original 3.72GB). Seems to be working as expected:
image

I also realized that not using a BOS token resulted in lower "accuracy" of image description. For example, take the following image and prompt:
moondream demo
**Prompt**: What is this?
Without BOS token:
> This image shows a computer setup with a computer tower and various computer components. The computer tower is placed on a table, and there are several wires connected to the components, indicating that the computer is ready to be turned on and used. The image also features a TV, which is likely part of the computer setup, as it is placed close to the computer tower.<

With BOS token:
> This is a computer server rack with multiple computer processors mounted on it. The rack is designed to hold and organize various computer components, such as motherboards, CPU, RAM, and other peripherals, in a neat and efficient manner. In this particular image, the rack is filled with several processors, showcasing its capacity to accommodate multiple devices at once.<

Not including the BOS token leads to hallucinations in the generated response. This PR addresses the fix.


candle-examples/examples/moondream/main.rs
```diff
@@ -64,6 +72,10 @@ impl TextGeneration {
         let mut tokens = tokens.get_ids().to_vec();
         let mut generated_tokens = 0usize;
 
+        let bos_token = match self.tokenizer.get_vocab(true).get("<|endoftext|>") {
```
LaurentMazare, line 75: This feels a bit weird for a bos token. Maybe add a comment if that's expected so as to mention that it's actually what the moondream2 tokenizer is using (maybe with a pointer to some json config or to the code?).

candle-transformers/src/models/mixformer.rs
```diff
@@ -438,10 +438,16 @@ impl MixFormerSequentialForCausalLM {
...
+        &mut self,
+        bos_token: &Tensor,
+        xs: &Tensor,
+        img_embeds: &Tensor,
+    ) -> Result<Tensor> {
         let _enter = self.span.enter();
         let xs = xs.apply(&self.embedding)?;
-        let mut xs = Tensor::cat(&[img_embeds.clone(), xs], 1)?;
+        let bos_token = bos_token.apply(&self.embedding)?;
+        let mut xs = Tensor::cat(&[bos_token, img_embeds.clone(), xs], 1)?;
```
LaurentMazare, line 450: Could you add a pointer to the python veersion for the order in this sequence (bos, then image embeds, then prompt)?

candle-transformers/src/models/quantized_mixformer.rs
```diff
@@ -337,6 +337,32 @@ impl MixFormerSequentialForCausalLM {
...
+        bos_token: &Tensor,
+        xs: &Tensor,
+        img_embeds: &Tensor,
+    ) -> Result<Tensor> {
+        let _enter = self.span.enter();
+        let xs = xs.apply(&self.embedding)?;
+        let bos_token = bos_token.apply(&self.embedding)?;
+        let mut xs = Tensor::cat(&[bos_token, img_embeds.clone(), xs], 1)?;
+        let (_b_size, seq_len, _embds) = xs.dims3()?;
+        let mask = if seq_len <= 1 {
```
LaurentMazare, line 351: I don't think `seq_len` can ever be <= 1 here, so let's remove that bit as well as the equivalent part in `mixformers.rs`.

candle-examples/examples/moondream/main.rs
santiagomed: I'm also a tad confused about this. [This](https://huggingface.co/vikhyatk/moondream2/blob/main/special_tokens_map.json) json file specified all tokens as "<|endoftext|>" which is akin to Phi which uses the same token for EOS and BOS. 

However, the Python moondream implementation uses "<END>" as EOS token. Additionally, "<END>" doesn't seem to be working in candle, but "END" does work. I looked at the tokenizer and "<END>" doesn't seem to be a token either. Maybe I'm missing something.

santiagomed:
Changed quantization to q4_0. This does seem to affect response but it is still accurate
image

LaurentMazare (approved):

LaurentMazare:
Thanks, great to have this quantized version too! @radames might be willing to try it in wasm though the size might be a bit on the large side (1.5GB).

radames:
amazing work folks! I'll work on the WASM module now! very exciting!

PR #1986, opened 2024-04-01
Modify access for conv and op to public

tomsanbear:
Currently writing a custom backend for candle and ran into module access issues as I can't access the structs defined here, this change enables this for me!


LaurentMazare (approved):

tomsanbear:
Thanks!

PR #1989, opened 2024-04-02
Added link to the Coursera ML algorithm implementations

vishpat:
Added link to the Coursera ML algorithm implementations as an external resource.


julien-c:
looks like a nice resource! did candle work well for that use case?

vishpat:
Yes, candle worked great.

PR #1995, opened 2024-04-02
Add benchmarks for the candle-nn package

tomsanbear:
Been doing some work on optimizing the layernorm and conv2d, figured I can get this in earlier than when I submit a PR for those!

Let me know if you have any feedback, it's essentially a copy of the core create's bench marks


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2001, opened 2024-04-03
moondream example - use syncronous apis instead of async/tokio

shur-complement:
removing async/await in favor of sync APIs for the moondream example


LaurentMazare:
What is the problem with using tokio here? I feel that it's nice to have at least an example that uses it so that people can see the difference with the blocking api but maybe it's breaking something for you?

shur-complement:
Yeah that's fair, I'll close this.

PR #2002, opened 2024-04-03
Refactor metal reduce kernel to be more readable/friendly to changes

tomsanbear:
There should not be a change to any of the actual logic here, this PR is just aiming to make this file friendlier to edit by reducing the amount of code that is wrapped in the CPP macro.

Note: I'm not as familiar with passing functions as references as done in the reduce change I've made here, open to recommendations.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2012, opened 2024-04-04
Add support for "sign" on tensors

tomsanbear:
Both cuda and metal have stdlib implementations for sign, as such it's handy for models (like bitnet) to have access to this operator to avoid an expensive where_cond call and other behaviour to emulate this.


candle-kernels/src/unary.cu
```diff
@@ -1,91 +1,104 @@
 #define _USE_MATH_DEFINES
-#include<math.h>
-#include<stdint.h>
+#include <math.h>
```
tomsanbear, line 2: I didn't realize my editor auto formatted this, lmk if this is undesireable and I'll undo

LaurentMazare: Yeah might be worth removing all the formatting changes if it's easy on your side, this will make the diff easier to look at.

candle-core/src/backprop.rs
```diff
@@ -707,6 +707,9 @@ impl Tensor {
                         let sum_grad = grads.or_insert(arg)?;
                         *sum_grad = sum_grad.add(&arg_grad)?
                     }
+                    Op::Unary(_arg, UnaryOp::Sign) => {
+                        Err(Error::BackwardNotSupported { op: "sign" })?
```
LaurentMazare, line 711: The backward is actually easy to implement here, we should just do nothing as there is no gradient flowing back on such a discrete operator.

tomsanbear: would that simply be the following:

```
let sum_grad = grads.or_insert(arg)?;
                        let arg_grad = grad.to_device(sum_grad.device())?;
                        *sum_grad = sum_grad.add(&arg_grad)?
```

I'm not yet quite familiar with internals of the backpropagation tracking

LaurentMazare: Nope, literally nothing should be done, and also around line 110 in the same file handle it as `Round` etc. The gradient doesn't flow through operator that are not contiguous (it's not necessary for correctness but will make it more efficient).

tomsanbear: Ah I see. Okay I've updated round and sign to be no-ops now

LaurentMazare:
I think we want the same behavior as pytorch on 0, ie return zero which is probably not what signum would do on the CPU, not sure what the cuda or metal versions are doing. Also we should cover this in the test.

tomsanbear:
> I think we want the same behavior as pytorch on 0, ie return zero which is probably not what signum would do on the CPU, not sure what the cuda or metal versions are doing. Also we should cover this in the test.

Ah good catch, I wasn't aware about that behaviour, test now handles neg zero and pos zero explicitely, and the cpu implementation does not use the signnum fn now

LaurentMazare:
It doesn't look like it's compiling on cuda for me, the sign function doesn't exist. Do you have some documentation on the headers where this function is supposed to be defined?

candle-core/src/backprop.rs
LaurentMazare: Ah that's not actually what I meant, let me actually do the tweak (it's fine to do the change you did on `Round` but a bit unrelated to this PR). Also note that it's the other way around compared to your commit message, we prevent the gradient from flowing through these ops as they are not continuous.

LaurentMazare:
I've made a couple tweaks to use a branchless version for the cpu kernels as well as cuda. It also preserves the current pytorch behavior on nan (as of 2.2) which is to return 0 (I don't feel attached to this behavior though as numpy returns nan it's just that it's more convenient and the generated assembly looks better on godbolt).

tomsanbear:
Thanks for making those changes! I'm not on my desktop right now so wasn't able to confirm the cuda side, was hoping it had existed till I could confirm 😅

LaurentMazare:
Well for me it's the other way around, I don't have access to a mac in the evenings so if you can confirm that the tests are all good on metal, then happy to merge.

tomsanbear:
Just gave it a try and no issues! Thanks!

LaurentMazare (approved):

LaurentMazare:
Merged! Thanks again!

PR #2015, opened 2024-04-04
Add flag to run Moondream in f16 precision

santiagomed:
Using f16 on M1 Mac gives significant performance improvement
image


LaurentMazare:
Thanks, I had to make a small tweak to your code as it seemed to break when using quantized on cuda.

LaurentMazare (approved):

Issue #1305, opened 2023-11-09
ONNX: Gather for tensors of rank > 0. 


xd009642:
Firstly thanks for candle, it's an impressive project and the speed at which updates are coming is really cool as a user!

I'm trying to use a speedyspeech onnx model in it right now and there's a failing op:

> not yet implemented: implement gather for Tensor[dims 78, 128; f32] Tensor[dims 1, 16; i64] axis 0 

I can provide my ONNX or assist via PRs if required (may need some pointers). Otherwise just wondering if there's any idea when Gather might fully land or if there are any difficulties in the implementation that might mean I should look for an alternative solution in the near term.

LaurentMazare:
Ah, the ONNX bits are brand new and only support a subset of the operations. It would certainly be helpful if you can share the ONNX file that you tried to run so that we can see if we can make it pass all ops.

xd009642:
Sure here's the model [speedyspeech.zip](https://github.com/huggingface/candle/files/13325372/speedyspeech.zip)

smallbraingames:
did you ever get this running?

xd009642:
Not speedyspeech but whatever candle does it won't work with that exported model. Not even onnxruntime works because the version of torch they used can generate invalid onnx graphs that nothing can run

smallbraingames:
ah got it, thank you

I wonder if there's a cheaper tts in rust than piper (which doesn't run in candle as well but does in ort)

anyways, thanks

PR #2026, opened 2024-04-07
Handle more tensor shapes in onnx "Gather" operation

gabotechs:
Stepped onto this while trying to load a MusicGen onnx model into candle (just the text encoder part). This was also reported in https://github.com/huggingface/candle/issues/1305 for another onnx model, so I decided to give it a shot.

This code replicates the problem described in https://github.com/huggingface/candle/issues/1305:
```rust
use candle_core::{DType, Device, Tensor};
use std::collections::HashMap;

    // onnx model from https://github.com/huggingface/candle/issues/1305#issuecomment-1806795045
    let model = candle_onnx::read_file("speedyspeech.onnx")?; 
    let graph = model.graph.as_ref().unwrap();

    let input_1 = graph.input[0].name.to_string();
    let input_2 = graph.input[1].name.to_string();

    let mut inputs = HashMap::new();
    inputs.insert(input_1, Tensor::zeros((1, 16), DType::I64, &Device::Cpu)?);
    inputs.insert(input_2, Tensor::zeros(1, DType::I64, &Device::Cpu)?);

    candle_onnx::simple_eval(&model, inputs)?;
    Ok(())
}
```
Which throws the following error:
```
not yet implemented: implement gather for Tensor[dims 78, 128; f32] Tensor[dims 1, 16; i64] axis 0
```

With this PR, that error is solved, but another error that seems unrelated (I think) appears:
```
Error: DTypeMismatchBinaryOp { lhs: I64, rhs: F32, op: "cat" }
```

The main challenge is that the [onnx "Gather" specification](https://onnx.ai/onnx/operators/onnx__Gather.html) is completely different from `Tensor::gather` in candle, and there does not seem to be any 1:1 operation that resembles onnx's "Gather". I came around with a solution that uses `index_select` for the cases where `inidices` has a dimension of 1 and 2, which makes the tests go green and should be suitable for most use-cases.

That's the best that I could get, some guidance about better options is much appreciated.

---

**Additional piece of info**: I was also able to solve the problem while importing MusicGen's text encoder onnx model, just to step into this other problem:
```
Error: unsupported op_type ReduceMean for op NodeProto
```


candle-onnx/src/eval.rs
```diff
@@ -508,17 +508,33 @@ pub fn simple_eval(
...
-                // differentiable way.
-                let xs = if indices.rank() == 0 {
-                    let index = indices.to_vec0::<i64>()? as usize;
-                    xs.narrow(axis, index, 1)?.squeeze(axis)?
-                } else {
-                    todo!("implement gather for {xs:?} {indices:?} axis {axis}")
+
+                // In Pytorch or Numpy this can be done by indesing the xs tensor using the indices
+                // tensor directly, but candle does not support tensor indexing at the moment, so
+                // some workarounds must be done.
```
gabotechs, line 519: As a reference, here's a piece of code for converting an onnx Gather op to Pytorch https://github.com/ENOT-AutoDL/onnx2torch/blob/main/onnx2torch/node_converters/gather.py#L41

candle-onnx/src/eval.rs
```diff
@@ -508,17 +508,33 @@ pub fn simple_eval(
...
                 let axis = xs.normalize_axis(axis)?;
-                // TODO: Provide an op to handle the ONNX generalized gather op ideally in a
-                // differentiable way.
-                let xs = if indices.rank() == 0 {
-                    let index = indices.to_vec0::<i64>()? as usize;
-                    xs.narrow(axis, index, 1)?.squeeze(axis)?
-                } else {
-                    todo!("implement gather for {xs:?} {indices:?} axis {axis}")
+
+                // In Pytorch or Numpy this can be done by indesing the xs tensor using the indices
```
gabotechs, line 517:
```suggestion
                // In Pytorch or Numpy this can be done by indexing the xs tensor using the indices
```

LaurentMazare (approved):

LaurentMazare:
Great thanks! Do you plan on also trying adding `ReduceMean`?

gabotechs:
> Great thanks! Do you plan on also trying adding ReduceMean?

Yeah, my idea right now is to add pretty much any onnx operation that's left until I get MusicGen up and running.

Issue #1939, opened 2024-03-26
Quantized example slower than llama.cpp with same model and settings...


oddpxl:
quantized compiled using --> cargo build --example quantized -r --features metal 

Unsure of...  how many layers accelerated / how many threads used / clearly different sample stages

..yet I presume the speed should be on par...   ?

CANDLE
./quantized --model mistral-7b-instruct-v0.2.Q4_K_S.gguf --which 7b-mistral-instruct-v0.2 --prompt "Blueberries cost more than strawberries. Blueberries cost less than raspberries. Raspberries cost more than strawberries and blueberries. If the first two statements are true, the third statement is?" --sample-len 2048 --temperature 0.1 --seed 1337 --top-p 0.950 --repeat-penalty 1.100 --repeat-last-n 64

--> 31.83 t/s
-------------

avx: false, neon: true, simd128: false, f16c: false
temp: 0.10 repeat-penalty: 1.10 repeat-last-n: 64
loaded 291 tensors (4.14GB) in 0.09s

LLAMA.CPP
./main -p "Blueberries cost more than strawberries. Blueberries cost less than raspberries. Raspberries cost more than strawberries and blueberries. If the first two statements are true, the third statement is?" -m mistral-7b-instruct-v0.2.Q4_K_S.gguf  -n 128 -ngl 33 --threads 8 --seed 1337

--> 51.30 t/s
-------------

sampling: 
	repeat_last_n = 64, repeat_penalty = 1.100, frequency_penalty = 0.000, presence_penalty = 0.000
	top_k = 40, tfs_z = 1.000, top_p = 0.950, min_p = 0.050, typical_p = 1.000, temp = 0.800
	mirostat = 0, mirostat_lr = 0.100, mirostat_ent = 5.000

sampling order: 
CFG -> Penalties -> top_k -> tfs_z -> typical_p -> top_p -> min_p -> temperature 
generate: n_ctx = 512, n_batch = 512, n_predict = 128, n_keep = 1

LaurentMazare:
Just to mention some more benchmarks on a RTX 2080 (cuda backend), excluding the prompt.
- llama.cpp: 6.93 token/s
- candle: 5.53 token/s
Llama.cpp is ~25% faster here, candle's performance has improved significantly over the last couple weeks as we've introduced a bunch of optimizations but there is certainly more to be done. I'll do a bit of profiling to see if there are obvious remaining bits.

```bash
[...]

PR #2037, opened 2024-04-10
Refactor the metal backend to always reuse command encoders/buffers unless a shared memory access is requested

tomsanbear:
This change aims to replace the pattern of each tensor provisioning a command buffer and encoder for each kernel operation that occurs, to a pattern where an encoder is provided to a kernel to setup it's operations.

Prior to this change, we relied on the fact that command buffers were executed sequentially to handle ordering of operations and ensuring that an output from one operation has completed before being used in a downstream operation. We now leverage Metal's resource based memory barriers to do this more effectively by ensuring we only block operations on the completion of operations on input dependencies.

What are the outcomes in terms of performance for this change? Well in summary not much... in theory I would have expected a minor gain due to the improved parallelization and lower overhead of only using a few command buffers and encoders, however in practice for the example models we do not see a change in performance.

Now although we don't see a change in performance on our models, we get a much more stable gputrace output as a result of this change. On main generating a gputrace would often cause OOM errors due to the amount of extra recorded command buffer/encoders, we now get a much more streamlined recording of the models, and the memory required to load these traces has been cut down drastically.

The major changes to note here are the following:
- command buffers are lazily initialized on calls to MetalDevice::command_buffer
- command encoders are lazily initialized on calls to MetalDevice::command_encoder
- a new RwLock is introduced for tracking the command encoder 
- the device struct is modified to make access to the command encoder and buffers private to ensure users always use the helper methods for accessing those methods
- the method for copying data to the cpu has been adjusted to handle closing the compute encoder and initializing the compute encoder
- reduced usage of the "wait_until_completed" instruction for the copy instructions, we now only rely on this when we require synchronization of resources between CPU and GPU


tomsanbear:
Hoping this branch will address some portion of https://github.com/huggingface/candle/issues/1939

tomsanbear:
Okay, I'm happy with the PR in it's current state!

@ivarflakstad it'd be great to get your eyes on this change as well!

candle-core/src/metal_backend/device.rs
```diff
@@ -189,14 +259,16 @@ impl MetalDevice {
         Ok(new_buffer)
     }
 
+    /// Allocates a new buffer with zeros.
     pub fn allocate_zeros(&self, size_in_bytes: usize) -> Result<Arc<Buffer>> {
+        self.end_compute_encoding()?;
```
LaurentMazare, line 264: Why do we need this? A comment could be helpful.

candle-core/src/metal_backend/device.rs
```diff
@@ -257,7 +323,8 @@ impl MetalDevice {
             return Ok(b.clone());
         }
 
-        let size = buf_size(size);
+        // Buffers on metal must be
```
LaurentMazare, line 326: Incomplete comment, `must be powers of two`?

candle-core/src/metal_backend/mod.rs
```diff
@@ -1282,17 +1276,17 @@ impl BackendStorage for MetalStorage {
                 dst.dtype()
             )
         }
-        let command_buffer = self.device.command_buffer()?;
         if src_s == d2 && dst_s == d2 {
-            command_buffer.set_label("copy2d_contiguous");
-            let blit = command_buffer.new_blit_command_encoder();
-            blit.set_label("copy2d_contiguous");
+            self.device.end_compute_encoding()?;
```
LaurentMazare, line 1280: Same question as in other comments, would be great for all `end_compute_encoding` to be documented.

candle-metal-kernels/src/lib.rs
```diff
@@ -270,31 +270,32 @@ impl Kernels {
...
 ) -> Result<(), MetalKernelError> {
     let pipeline = kernels.load_pipeline(device, Source::Unary, kernel_name.0)?;
-    let encoder = command_buffer.new_compute_command_encoder();
     encoder.set_compute_pipeline_state(&pipeline);
 
     set_params!(encoder, (length, &input, output));
 
     let (thread_group_count, thread_group_size) = linear_split(&pipeline, length);
+
+    encoder.memory_barrier_with_resources(&[input.buffer]);
```
LaurentMazare, line 287: Isn't this enforced by the `use_resource` just below? I'm not 100% sure of the semantics of these two so any link to some doc would be great!

candle-core/src/metal_backend/device.rs
ivarflakstad: (For context I went with power of two so that buffer reuse would be more likely)

candle-metal-kernels/src/lib.rs
tomsanbear: Good question, in this case useResource is more about granting visibility of the memory, it doesn't gate execution of the kernel on the resource. The memory barrier is used to ensure that operations involving the input buffers have completed and that we can safely read from those buffers.

tomsanbear: I think there is a lot of room for improvement in reducing these calls so that we can avoid the candle_metal_kernels code from needing to manage access requests, but that would involve knowing the compute graph prior to any synchronization event.

LaurentMazare: Not sure to understand what you mean here. My mental model based on cuda is that we inform metal of the dependencies and it will keep track of the compute graph internally. The only place where you need synchronisation is when sending or retrieving data or when doing very weird things in kernels (which we don't in candle). Is that actually different in metal?

tomsanbear:
@LaurentMazare I did some cleanup of the areas commented above, lmk if you have any other ideas on improving docs

candle-metal-kernels/src/lib.rs
tomsanbear: That's true across encoders/buffers, but when we are executing many kernels on a single encoder metal does not provide that guarantee. This is at least my understanding, I'll play around with different configurations of the useResources and memoryBarrierWithResources to conrfirm though.

tomsanbear: So it seems that the following is the behaviour under different scenarios:
- no odd behaviour when using both useResources & memoryBarrierWithResources
- no odd behaviour when only using memoryBarrierWithResources
- no odd behaviour when only using useResources
- no odd behaviour when using neither

Now this is odd, I'm not entirely convinced from a quick pass at the examples that we'd be getting this correct though... looking back at llama.cpp, they seem to only leverage the memorybarrier op and not the useresource op. I'm inclined to believe that we're goth correct here and that I suppose we just need one or the other?

tomsanbear: @ivarflakstad do you have any insight here that could help us land on the preferred option?

LaurentMazare: > That's true across encoders/buffers, but when we are executing many kernels on a single encoder metal does not provide that guarantee. This is at least my understanding

Would you have any link for this?
I certainly would rather see the specs rather than trying it out as getting it wrong would make for bugs that are very tricky to debug.
Here is the most relevant bit I found.
> Compute commands are executed in the order in which they are encoded into the command buffer. A compute command finishes execution when all threadgroups associated with the command finish execution and all results are written to memory. Because of this sequencing, the results of a compute command are available to any commands encoded after it in the command buffer.
It's coming from this [doc](https://developer.apple.com/library/archive/documentation/Miscellaneous/Conceptual/MetalProgrammingGuide/Compute-Ctx/Compute-Ctx.html#//apple_ref/doc/uid/TP40014221-CH6-SW1), and I understand that as long as you're using the same buffer you're find so you would need barries when you have multiple buffers in parallel?

tomsanbear: I've mostly been using these docs for informing myself https://developer.apple.com/documentation/metal/resource_synchronization?language=objc

Re-reading, perhaps I'm understanding wrong and these are automatically tracked for us since we are creating buffers from the device interface, and the hazard tracking is handled for us

tomsanbear: However this doc, which I believe is the more relevant one for us seems to indicate the useResource is required for the argument buffer bindings we make https://developer.apple.com/documentation/metal/buffers/tracking_the_resource_residency_of_argument_buffers?language=objc

candle-examples/examples/quantized/main.rs
```diff
@@ -545,6 +549,16 @@ fn main() -> anyhow::Result<()> {
         let eos_token = *tos.tokenizer().get_vocab(true).get(eos_token).unwrap();
         let start_post_prompt = std::time::Instant::now();
         let mut sampled = 0;
+
+        // Start metal capture
+        #[cfg(feature = "metal")]
+        if args.metal_tracing {
+            use candle::Device;
+            if let Device::Metal(metal_device) = device.clone() {
+                metal_device.capture("/tmp/candle.gputrace")?;
```
LaurentMazare, line 558: Let's make the `metal-tracing` arg an `Option<String>` so that it specifies the filename for the trace rather than hardcoding it?

ivarflakstad: Agreed. Adding to this, does it make sense to have this as an env var?
Gpu tracing not dependent on what type of model you’re running.

ivarflakstad: In which case we should log that gpu tracing is enabled

LaurentMazare: Would rather keep it as an explicit argument as we try to not rely on env variable (though there might be one remaining), the gist of it is that they are hard to discover whereas arguments are documented. Also this way it doesn't depend on whether the env variable gets changed by the user code calling candle etc.
(and on another note, candle tries actively to be log free outside of the `main.rs`)

tomsanbear:
Just an update, I've run through most examples and they work fine, the exception is stable diffusion which outputs garbage on the photo compared to main, i'll be looking into this and will tag when I find the root cause, not immediately obvious so far.

candle-examples/examples/quantized/main.rs
ivarflakstad: Right. I was thinking it would only take effect with the `debug` or `release-with-debug` profiles. I imagine no one is doing gpu tracing in a production environment.
Debug logging is compiled away in release builds.

candle-core/src/metal_backend/device.rs
```diff
@@ -99,48 +132,85 @@ impl MetalDevice {
...
-            *command_buffer_lock = command_buffer.clone();
-            *index = 0;
+        let command_buffer = command_buffer_lock.to_owned();
+
+        if let Some(command_buffer) = command_buffer {
+            Ok(command_buffer)
+        } else {
+            let new_command_buffer = self.command_queue.new_command_buffer().to_owned();
+            *command_buffer_lock = Some(new_command_buffer);
+            Ok(command_buffer_lock.to_owned().unwrap())
```
LaurentMazare, line 149: This unwrap seems a bit weird, why not cloning `new_command_buffer` before storing it in `command_buffer_lock` and returning this cloned value instead?

candle-core/src/metal_backend/device.rs
```diff
@@ -99,48 +132,85 @@ impl MetalDevice {
...
+            self.command_encoder.try_write().map_err(MetalError::from)?;
+        let command_encoder = command_encoder_lock.to_owned();
+
+        if let Some(command_encoder) = command_encoder {
+            Ok(command_encoder)
+        } else {
+            let command_buffer = self.command_buffer()?;
+            let new_command_encoder = command_buffer.new_compute_command_encoder().to_owned();
+            *command_encoder_lock = Some(ComputeCommandEncoder::from(new_command_encoder));
+            Ok(command_encoder_lock.to_owned().unwrap())
```
LaurentMazare, line 165: Same comment as above on the unwrap.

candle-core/src/metal_backend/device.rs
```diff
@@ -99,48 +132,85 @@ impl MetalDevice {
...
 
-    pub fn kernels(&self) -> &Kernels {
-        &self.kernels
+    /// If the user wants to end the current command encoder, they should call this function
+    /// In the case that there is no active command encoder, this function will not return an error,
+    /// it instead is a no-op.
+    pub fn end_compute_encoding(&self) -> Result<()> {
+        let mut command_encoder_lock =
+            self.command_encoder.try_write().map_err(MetalError::from)?;
+        let command_encoder = command_encoder_lock.to_owned();
```
LaurentMazare, line 189: Do we really need this `to_owned` here, seems like this would trigger some obj-c refcounting that may be quite wasteful whereas we only need a reference to it.

LaurentMazare:
> Just an update, I've run through most examples and they work fine, the exception is stable diffusion (wuerstchen is fine though...) which outputs garbage on the photo compared to main, i'll be looking into this and will tag when I find the root cause, not immediately obvious so far.

Gave it a quick look and it indeed looks pretty weird. When enabling `intermediary-images` (to make it use the vae earlier) and comparing with the base version via some print statements I saw the first difference on the last layer of the vae, but I also got a run where it didn't seem to happen. So it may well be some form of race condition, not sure if it's between kernels, or when fetching the data back from the gpu, also worth pointing out that my mac gets very low on memory at that point (with 16GB) - but I would expect an actual error rather than data corruption if this is an issue.
Two other things that may be worth trying if easy: disabling the buffer reuse and use freshly allocated buffers each time, and limiting the maximum number of encoded ops in the buffer.

LaurentMazare:
Any luck with this? It would be great to have to analyze performance on metal so pretty keen to have it if it still works with stable diffusion etc.

tomsanbear:
> Any luck with this? It would be great to have to analyze performance on metal so pretty keen to have it if it still works with stable diffusion etc.

Unfortunately haven't had the bandwidth the last two weeks, I should have some more time coming up where I can spend some time investigating. I was going back through the implementation I had and was trying to hunt down where the change was introduced, so far nothing seemed apparent, so now I'm running through the operations that are performed in the stable diffusion model so that I can understand what's going on.

tomsanbear:
updated the branch here with main so we don't stay too far out of sync

tomsanbear:
Closing this out. The encoder/buffer reuse this was aiming at has since landed upstream by another route — the concurrent-dispatch and inter-encoder sync work in #3511 and #3532 — so this diff is against an architecture that no longer exists and would need rewriting rather than rebasing.

Thanks for the review time on it @LaurentMazare. For the record, the stable-diffusion corruption that blocked it was never root-caused; it was most likely an artifact of this PR's own buffer reuse rather than a pre-existing bug, but I never proved that either way.

PR #2043, opened 2024-04-11
Use cat for faster MQA computation.

LaurentMazare:


tomsanbear:
Just tried this out, got the following before and after

before:
```
43 prompt tokens processed: 28.04 token/s
124 tokens generated: 20.07 token/s
```
after:
```
43 prompt tokens processed: 30.79 token/s
124 tokens generated: 24.36 token/s
```

PR #2045, opened 2024-04-11
Change for the encoder-only ProstT5 model

Victor-Mihaila:
This enables running ProstT5's (https://huggingface.co/Rostlab/ProstT5) encoder-only mode in Candle. This ProstT5 mode stores it's embed_tokens weights within the encoder, as its decoding stage was replaced with a CNN.  This alone is not sufficient to run ProstT5 within Candle examples. We will develop a ProstT5 runner outside candle for now, but would be willing to upstream it to candle-examples at a later point.


LaurentMazare (approved):

LaurentMazare:
Thanks, fwiw these conditional tensor renames are certainly very hacky.
A cleaner way to handle this is have the `VarBuilder` that you provide to the model creation do the renaming (could use regex or whatever it wants), it's already possible to do so via a custom `VarBuilder` backend though not well documented, I'm putting together #2052 that should make it a bit easier - with this change you will be able to just create a `VarBuilder` based on another `VarBuilder` and a renaming function.

PR #2047, opened 2024-04-12
Avoid crashes when running T5 models with F16 tensors on CPU

Victor-Mihaila:
This change avoids crashes when running T5 models with F16 tensors on CPU.


LaurentMazare (approved):

LaurentMazare:
Nice catch, thanks.

PR #2048, opened 2024-04-12
Add benchmarks for qmatmul operations

tomsanbear:
Adds benchmarks for the qmatmul operation across all the different dtypes, will be useful moving forwards as we tune or adjust our quantized kernels.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2049, opened 2024-04-12
Add ReduceMean onnx operation

gabotechs:
This PR adds support for the "ReduceMean" onnx operation based on https://onnx.ai/onnx/operators/onnx__ReduceMean.html#reducemean-13

The onnx model that I'm trying to load was generated with https://github.com/huggingface/optimum, and that tool generates the version 13 of the "ReduceMean" operation, so I have implemented a version compatible with 13 and below.

There do is a new version 18 with different inputs (`axes` comes as a second input rather than a named attribute), but I have not seen a clear path for switching between the different versions of an operation in order to provide different implementations.

A clean way I can imagine of doing is would be to split `candle-onnx/src/eval.rs` into one file per operation, implement all the versions of an operation in each file, and heavily reuse code across versions of the same operation. That would also allow to colocate tests and implementations. It sounds like a big change though...

---

I have also gathered a list of onnx operations that are missing and are needed for my use-case:

## Still missing
Sqrt
Range
Greater
Less
Log
Min
Where

## Incomplete implementation
ConstantOfShape -> should take "value" attribute into account for selecting the correct DType

### DIFF
--- a/candle-onnx/src/eval.rs
+++ b/candle-onnx/src/eval.rs
@@ -2,7 +2,7 @@ use crate::onnx;
 use crate::onnx::attribute_proto::AttributeType;
 use crate::onnx::tensor_proto::DataType;
 use candle::{bail, DType, Device, Result, Tensor};
-use std::collections::HashMap;
+use std::{collections::HashMap, usize};
 
 pub type Value = Tensor;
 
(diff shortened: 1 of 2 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks!

Issue #2040, opened 2024-04-11
Unsound usages of unsafe function `slice::from_raw_parts`


shinmao:
Hi, we are researchers from [Sunlab](https://sunlab-gmu.github.io/). When we tried to scan Rust-based repositories with our own implemented bug detectors, we found that there are some potentially unsound usages of `slice::from_raw_parts` in the package `candle-core` at version 0.4.1.

### `quantized::ggml_file::from_raw_data`
https://github.com/huggingface/candle/blob/a4d5a414e3ae79642ecfd6b7bb410c26a8a62a06/candle-core/src/quantized/ggml_file.rs#L120-L134
At line 128, the reference to `u8` slice was casted to raw pointers of generic types that implemented traits such as `GgmlType` e.g., `BlockQ4_0` or `f32`. However, the pointer type conversion in such way would break the alignment requirement of `slice::from_raw_parts`, and it might lead to unsound behaviors.

Even though this API is private, but it is accessible from other public API such as `qtensor_from_ggml`.

### PoC
```rust
use candle_core::quantized::{ggml_file::qtensor_from_ggml, GgmlDType};
use candle_core::Device;

fn main() {
    let raw_data: [u8; 1] = [0x1; 1];
    let dim: Vec<usize> = Vec::from([1usize, 2, 3, 4]);
    let res = qtensor_from_ggml(
        GgmlDType::F32,
        &raw_data,
        dim,
        &Device::Cpu
    );
    println!("{:?}", res.unwrap().data());
}
```
Run with miri and get the result,
```
error: Undefined Behavior: dereferencing pointer failed: alloc828 has size 1, so pointer to 96 bytes starting at offset 0 is out-of-bounds
   --> /home/rafael/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/raw.rs:101:9
    |
101 |         &*ptr::slice_from_raw_parts(data, len)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ dereferencing pointer failed: alloc828 has size 1, so pointer to 96 bytes starting at offset 0 is out-of-bounds
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
help: alloc828 was allocated here:
   --> src/main.rs:5:9
    |
[...]

PR #2053, opened 2024-04-13
Refactor usages of "from_raw_parts" to use bytemuck utilities instead

tomsanbear:
Leaving as a draft to solicit feedback on the introduction of this package.

During my research on WebGPU it seemed like this was a pretty common util being used in the webgpu project, it seems like it could be worth offloading the safety of our unsafe code to this purpose built package. If we are open to this then I'll go ahead and replace pretty much all of our usages of the from_raw_parts to use this package so that safety is handled for us. Seems that the CPU package also has a lot of reliance on this too.

When running the example from this [issue](https://github.com/huggingface/candle/issues/2040), this branch correctly throws an error in release builds `cast_slice>TargetAlignmentGreaterAndInputNotAligned` whereas on main the example runs without error.

### DIFF
--- a/candle-core/examples/safetensors.rs
+++ b/candle-core/examples/safetensors.rs
@@ -1,0 +1,9 @@
+use candle_core::quantized::{ggml_file::qtensor_from_ggml, GgmlDType};
+use candle_core::Device;
+
+fn main() {
+    let raw_data: [u8; 1] = [0x1; 1];
+    let dim: Vec<usize> = Vec::from([1usize, 2, 3, 4]);
+    let res = qtensor_from_ggml(GgmlDType::F32, &raw_data, dim, &Device::Cpu);
+    println!("{:?}", res.unwrap().data());
+}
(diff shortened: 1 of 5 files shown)


candle-core/examples/safetensors.rs
```diff
@@ -0,0 +1,9 @@
+use candle_core::quantized::{ggml_file::qtensor_from_ggml, GgmlDType};
+use candle_core::Device;
+
+fn main() {
```
tomsanbear, line 4: keeping this present on this branch to repro the issue from https://github.com/huggingface/candle/issues/2040

shinmao: @tomsanbear As I saw, most of the `GgmlType` have larger alignments than `u8`. How do we make sure we use `qtensor_from_ggml` and get the correct data value?

tomsanbear:
Going to close this as I'm not going to get around to it for a long while, will reopen a new MR if I can later 👍

PR #2056, opened 2024-04-14
Metal Unary: Add benchmarks and process kernels in a tile based fashion

tomsanbear:
Going off what @LaurentMazare found a week or so ago, this just changes the unary implementation to process kernels in tiles rather than a function call per index of the buffer. See below for benchmarks on my M3 for a variety of tile sizes, 1 being the effective current behaviour on main.

From the benchmarks below we see a falloff of improvement after a tile size of 4, but also a regression on f32 performance. The main benefits on this change appear to be on f16 and bf16 operations with f32 experiencing a minor efficiency gain at a tile size of 4.

(See latest benchmarks in the comments below)


LaurentMazare:
Could you maybe make all the benchmarks relative to the current version, it's a bit hard to read at the moment as it's not clear what gets compared to what.
Also maybe you could extract a small table that summarizes the benchmarks (though it's certainly great to have the full output in the description).

tomsanbear:
Updated with table format with the breakdown of each change and it's results

LaurentMazare:
Thanks for putting these up. It looks that most of the benefit is for f16/bf16 with two elements, maybe it would be better to have specialized kernels for these two dtypes for n = 2 and use it only in this case. One benefit is that if we fix n, the compiler should be able to unroll the loop (we could even do so manually), which might improve performance.
Also in cuda, there is a specific datatype to group two f16/bf16 together `half2`, maybe there is something similar that exists for metal?

tomsanbear:
> Thanks for putting these up. It looks that most of the benefit is for f16/bf16 with two elements, maybe it would be better to have specialized kernels for these two dtypes for n = 2 and use it only in this case. One benefit is that if we fix n, the compiler should be able to unroll the loop (we could even do so manually), which might improve performance. Also in cuda, there is a specific datatype to group two f16/bf16 together `half2`, maybe there is something similar that exists for metal?

Ah yes we do have vector types like that in metal as well, I'll play around with it and see what we get with a non dynamic tile size!

tomsanbear:
Updated benchmarks with the change to a specialized kernel for the tiled case:

This benchmark is comparing a previous run which runs the f16 and bf16 on the untiled kernels vs. running them on a tiled kernels with a size of 2. 
```
metal_sqrt_BF16/iter    time:   [11.037 µs 11.075 µs 11.110 µs]
                        thrpt:  [175.80 GiB/s 176.36 GiB/s 176.97 GiB/s]
                 change:
                        time:   [-35.443% -34.780% -34.148%] (p = 0.00 < 0.05)
                        thrpt:  [+51.856% +53.328% +54.901%]
                        Performance has improved.
metal_sqrt_F16/iter     time:   [10.862 µs 10.921 µs 10.976 µs]
                        thrpt:  [177.94 GiB/s 178.83 GiB/s 179.82 GiB/s]
                 change:
                        time:   [-36.950% -36.425% -35.888%] (p = 0.00 < 0.05)
                        thrpt:  [+55.977% +57.295% +58.604%]
                        Performance has improved.
```

LaurentMazare (approved):

LaurentMazare:
Great improvement, thanks!

PR #2061, opened 2024-04-14
Refactor candle-metal-kernels to accept an encoder instead of a command buffer

tomsanbear:
This PR is setting up the metal backend for the change proposed in this PR: https://github.com/huggingface/candle/pull/2037

The goal here is to have no actual change at runtime for this diff, but to decrease the complexity of the linked MR above.


LaurentMazare:
Thanks, certainly a good idea to break the change in smaller pieces. But let's maybe put that on hold until there is a better understanding of the issue with stable-diffusion on the other PR (if that's not already the case).

tomsanbear:
Closing this too. It was parked on #2037, and candle takes an encoder rather than a command buffer anyway now — `CommandsGuard`/`ComputeCommandEncoder` arrived with the post-#3511 concurrency work, so there is nothing left here to land.

PR #2063, opened 2024-04-14
Move image tensor to device in trocr example

boustrophedon:
Was getting `Error: device mismatch in conv2d, lhs: Cpu, rhs: Cuda { gpu_id: 0 }` when running with `--features cuda` but it seems it was just missing the `to_device` call like in https://github.com/huggingface/candle/pull/1856


LaurentMazare (approved):

LaurentMazare:
Thanks!

Issue #2074, opened 2024-04-16
Quantized Matmul: Small batches are slower than no-batch


lucasavila00:
Thanks for the work on this repo, it's amazing.

The newly added quantized matmul kernels are great, but they're slower than using the vec kernels on small batch sizes.

On Llama.cpp I see they use the vec version if batch is <= 4 https://github.com/ggerganov/llama.cpp/pull/5351 https://github.com/ggerganov/llama.cpp/pull/5370

I'm reading https://github.com/huggingface/candle/blob/main/candle-kernels/src/quantized.cu and it seems the code was extracted from Llama.cpp before this PR was merged, and candle's version misses this optimization.

For context, this issue was noticed on mistral.rs https://github.com/EricLBuehler/mistral.rs/issues/139

LaurentMazare:
Interesting, do you know if the new kernels are slower than the old approach for your use case? (the old approach was to dequantize then matmul so would not involve the mmv kernels). If that's the case, switching back to the old style when the input size is smaller than 4 should be a straightforward change.

lucasavila00:
Old kernels:

batch=1 20ms
batch=2 278ms 
batch=3 354ms

New kernels:

batch=1 20ms
batch=2 133ms
batch=3 144ms


They are faster. But not batching is even faster.

lucasavila00:
What I think is happening is that candle is not checking this condition: https://github.com/ggerganov/llama.cpp/blob/master/ggml-cuda.cu#L1898-L1900

And then not using the `vec` kernels: https://github.com/ggerganov/llama.cpp/blob/master/ggml-cuda.cu#L1955

I'm not too familiar with candle's source code but I'm profiling:

bs=1
![image](https://github.com/huggingface/candle/assets/12750442/3e9cdd74-ff79-4800-b554-f5f37c333fe0)


bs=2
![image](https://github.com/huggingface/candle/assets/12750442/d40d1086-2bf9-4ac8-bafb-40c05ec02f68)

lucasavila00:
For reference, bs=2 and llama.cpp launches the `_vec` kernels

![image](https://github.com/huggingface/candle/assets/12750442/0e56e466-607f-436e-b645-52a3985b00b4)

LaurentMazare:
Ok, thanks for the details, I'm testing it with #2075 the downside is that it adds a lot of kernels (even up to only batch size 4) so will have an impact on the binary size, but hopefully it's worth it.

EricLBuehler:
[...]

PR #2075, opened 2024-04-16
Add the mmv kernels for small batch sizes.

LaurentMazare:
Only handling the batch sizes up to 4 to avoid including too many kernels.
See #2074 and https://github.com/ggerganov/llama.cpp/pull/5351#issue-2119377523 for more context.


EricLBuehler:
@LaurentMazare, thank you for adding this!

PR #2085, opened 2024-04-18
Llama v3.

LaurentMazare:


EricLBuehler:
Note: llama3 instruct uses `<|eot_id|>` and `<|end_of_text|>` as EOS tokens.

PR #2086, opened 2024-04-18
Add support for llama3 on the quantized example

tomsanbear:
So far I've found quantized versions here: https://huggingface.co/QuantFactory/Meta-Llama-3-8B-GGUF and the tokenizer here: https://huggingface.co/unsloth/llama-3-8b/tree/main


LaurentMazare:
Thanks, I assume you tested it and it went all good? Also maybe put a TODO to switch to the TheBloke model once they are available.

LaurentMazare:
Also why not use the tokenizer from the official Meta repo?

LaurentMazare:
Thanks, just tried out the official tokenizer and it went fine on your example. Probably better to use it and not a copy that may not be an authorized one :) Hopefully you'll get access soon, for me it was super quick (maybe an hour or so).

PR #2095, opened 2024-04-20
Run metal and accelerate features on CI

tomsanbear:
This adds two new configurations that enabled testing the metal and accelerate features. The macos latest runners are run on M1s, as such we should get access to MPS.


tomsanbear:
Getting some super weird errors with the CI runners, need to double check what machines these are actually running behind the scenes...

PR #2096, opened 2024-04-20
Add missing onnx operations

gabotechs:
Follow up for https://github.com/huggingface/candle/pull/2049

Adds the following onnx operations along with tests based on the official onnx examples for the following operations:
- Sqrt
- Range
- Greater
- Less
- Log
- Min
- Where

Additionally, fixes support for the following operations, also adding tests:
- ConstantOfShape

---

With this, I was able to run MusicGen's text encoder onnx model, but there's still something wrong, even though the model now runs successfully, the output is:
```
[[[NaN, NaN, NaN, ..., NaN, NaN, NaN],
  [NaN, NaN, NaN, ..., NaN, NaN, NaN]]]
Tensor[[1, 2, 768], f32]
```
I've verified that my the onnx file is correct (I can run it successfully with https://github.com/pykeio/ort), so there must be something else in `candle-onnx` that is not working correctly.

### DIFF
--- a/candle-onnx/tests/ops.rs
+++ b/candle-onnx/tests/ops.rs
@@ -937,7 +937,10 @@ fn test_constant_of_shape() -> Result<()> {
         let eval = candle_onnx::simple_eval(&manual_graph, inputs)?;
         assert_eq!(eval.len(), 1);
 
-        let z = eval.get(OUTPUT_Z).expect("Output 'z' not found").to_dtype(DType::F64)?;
+        let z = eval
+            .get(OUTPUT_Z)
+            .expect("Output 'z' not found")
+            .to_dtype(DType::F64)?;
 
         let expected = Tensor::new(expected, &Device::Cpu)?.to_dtype(DType::F64)?;
         match expected.dims().len() {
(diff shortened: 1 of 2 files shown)


candle-onnx/tests/ops.rs
```diff
@@ -1639,3 +1773,450 @@ fn test_reduce_mean() -> Result<()> {
...
     Ok(())
 }
+
+// "Sqrt"
+#[test]
+fn test_sqrt() -> Result<()> {
+    // https://github.com/onnx/onnx/blob/main/docs/Operators.md#examples-155
+    test(&[1., 4., 9.], &[1., 2., 3.])?;
+
+    fn test(data: impl NdArray, expected: impl NdArray) -> Result<()> {
```
gabotechs, line 1783: I've followed the code pattern present in other tests, but it gets a bit repetitive, I think that a refactor using a builder pattern would be really great for all this tests, something like:
```rust
OnnxTest::build()
  .input(&[1, 2, 3, 4, ...])
  .input(1i64)
  .attribute("axis", 1)
  .attribute("keepdims", false)
  .expect(&[[1, 0, 0, 1, 0]])?
```
I've tried to be very consistent in the test code in order to streamline future refactors that you might want to do.

LaurentMazare: Agreed that it's a bit repetitive but if it's in the tests that doesn't look too bad, feel free to make a separate PR to clean this out at some point if you want though.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2099, opened 2024-04-20
Updated quantized phi model

LaurentMazare:


EricLBuehler:
Ah, that was what I was missing! I had issues in [this PR](https://github.com/EricLBuehler/mistral.rs/pull/87) when trying to break apart the qkv tensors, it seems so clear now. Thanks!

Issue #975, opened 2023-09-27
AudioGen/MusicGen


KookaburraShell:
Are there any plans to implement the AudioGen/MusicGen models from Facebook?

LaurentMazare:
Actually I sketched a musicgen implementation a while back that is still in the repo [here](https://github.com/huggingface/candle/tree/main/candle-examples/examples/musicgen). At this point we didn't have the conv1d and lstm ops so I ended up not finishing it but now that these are implemented it would probably be a good idea to resurect these (the clip part of musicgen has already been merged in `candle-transformers` as it could be used outside of musicgen).

framp:
A bit OT but is candle-transformers meant to implement all pipelines from huggingface/transformers (like an extended version of rust-bert)?

I've been trying a few models so I could help there. Are PRs welcome?

LaurentMazare:
The scope of `candle-transformers` is not totally settled yet. The workflow I would like to encourage for new models is:
- Start by developing the model in its own separate repo.
- If it works well, and the model is generic enough then we can consider moving it to `candle-transformers`.

LaurentMazare:
Tagging with help-wanted in case anyone wants to take a stab at this. There is some ongoing work in `candle-examples` but the forward passes are not implemented yet - you may also prefer starting from scratch.

LaurentMazare:
Note that the encodec model is now available in candle so musicgen should be a pretty easy add at this point if anyone wants to give it a shot.

qooba:
@LaurentMazare - I have implemented [`forward`](https://github.com/huggingface/transformers/blob/main/src/transformers/models/musicgen/modeling_musicgen.py#L2173) method for `musicgen` model. If it's ok I'd like to also implement [`generate`](https://github.com/huggingface/transformers/blob/main/src/transformers/models/musicgen/modeling_musicgen.py#L2599) method.

LaurentMazare:
Great, yeah it would be pretty neat to also have the generate method in the example, that would make it far easier to test the forward method etc so feel free to do it in the PR that you've started minting!

PR #2145, opened 2024-04-30
Musicgen forward implementation

qooba:
This PR implements forward method for musicgen model: huggingface/candle#975


akashicMarga:
@qooba did you add generate method?

PR #2179, opened 2024-05-11
Add SliceSafetensors.

LaurentMazare:


candle-nn/src/var_builder.rs
```diff
@@ -487,6 +487,12 @@ impl<'a> VarBuilder<'a> {
         Ok(Self::from_backend(Box::new(tensors), dtype, dev.clone()))
     }
 
+    /// Initializes a `VarBuilder` from a binary builder in the safetensor format.
+    pub fn from_slice_safetensors(data: Vec<u8>, dtype: DType, dev: &Device) -> Result<Self> {
```
boustrophedon, line 491: This function is the same as `from_buffered_safetensors` - probably just forgot to update it. Additionally, SimpleBackend isn't implemented for SliceSafetensors.  I made a PR.

PR #2183, opened 2024-05-13
add top_k option to `LogitsProcessor` new method

jorgeantonio21:
Adds an optional `top_k` parameter to `LogitsProcessor`, so we allow `top_k` sampling externally.


LaurentMazare:
Did you try using `from_sampling`? The idea was to use this new function when one wants to be able to set all the different options (there are a couple examples in the codebase) and not break compatibility for the existing function.

jorgeantonio21:
I see, I understand now how to deal with this, thanks for the explanation

PR #2187, opened 2024-05-15
Allow the threshold argumet to be negative in the segment-anything example

vargad:
Threshold is 0.0 by default, negative values make more points included, expanding the mask. Positive values make it more picky, making the mask smaller.

Negative numbers start with a minus sign, which normally makes clap consider it a flag.

Without this change:
```
$ cargo run --example segment-anything -- --image assets/single_pt_prompt.jpg --point 0.5,0.5 --threshold -0.5
error: unexpected argument '-0' found

Usage: segment-anything [OPTIONS] --image <IMAGE>

For more information, try '--help'.
```


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2189, opened 2024-05-16
chore: add fix for windows cudarc into the readme

Cifko:
I could not run the example on windows, because of the new cudarc version. It's looking for the wrong names. So just by copy/renaming these 3 files you can run the examples on windows. I thought it's worth to have it in the FAQ for now. Once it's fixed you can removed it.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2200, opened 2024-05-21
Add RandomNormal ONNX operator

mokulus:
The implementation is very similar to the RandomUniform operator I've wrote earlier.
I wasn't sure how to test the result, I didn't want to use a statistical test as they might fail if we are very unlucky, so I'm just testing for the uniqueness of the values. If it's not ok please let me know.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2223, opened 2024-05-27
Improve versatility of `slice_assign`

EricLBuehler:
This PR enables `slice_assign` to take multiple types of ranges as an argument, instead of the current behavior where only one type may be used. For example, open and closed ranges or even a single index may be used.

To represent this effectively in the type system, this PR introduces a new trait and represents each range for a dimension as a trait object. However, this causes a breaking change (which will be trivially easy to fix) because the argument becomes a reference rather than an owned value.

## Example
```rust
let dev = Device::Cpu;

let tensor = Tensor::arange(0u32, 4 * 5, &dev)?.reshape((4, 5))?;
let src = Tensor::arange(100u32, (2 * 3) + 100, &dev)?.reshape((3, 2))?;
let out = tensor.slice_assign(&[&(..=2), &(1..3)], &src)?;
assert_eq!(
    out.to_vec2::<u32>()?,
    &[
        [0, 100, 101, 3, 4],
        [5, 102, 103, 8, 9],
        [10, 104, 105, 13, 14],
        [15, 16, 17, 18, 19]
    ]
);
Ok(())
```


candle-core/src/tensor_indexing.rs
```diff
@@ -0,0 +1,379 @@
...
+use crate::{
+    bail,
+    op::{BackpropOp, Op},
+    shape::Dim,
+    tensor::from_storage,
+    DType, Error, Result, Tensor,
+};
+
+/// Specialization of `std::ops::RangeBounds` for `usize` to allow trait objects.
+pub trait RangeBound {
```
EricLBuehler, line 12: This is the core of the change.

candle-core/src/tensor_indexing.rs
```diff
@@ -0,0 +1,379 @@
...
+    fn start_bound(&self) -> std::ops::Bound<usize>;
+    fn end_bound(&self) -> std::ops::Bound<usize>;
+}
+
+macro_rules! range_bound {
```
EricLBuehler, line 17: Macro to reduce code repetition

candle-core/src/tensor_indexing.rs
```diff
@@ -0,0 +1,379 @@
...
+
+range_bound!(Range);
+range_bound!(RangeFrom);
+range_bound!(RangeFull, ());
+range_bound!(RangeInclusive);
+range_bound!(RangeTo);
+range_bound!(RangeToInclusive);
+range_bound!((std::ops::Bound<usize>, std::ops::Bound<usize>));
+
+impl RangeBound for usize {
```
EricLBuehler, line 60: Implement for usize so that users can avoid `x..x+1` and instead do this: `x.slice_assign(&[&0, &(0..2)])`.

PR #2231, opened 2024-05-31
Adding Gemm and ArgMax operators to candle-onnx

socathie:
Gemm operator is very common in ONNX files generated by Pytorch, such as when using torch.nn.Linear. ArgMax is also a common operator for classification tasks.

I'm not super proficient in Rust, so appreciate any modification to the current code for better practice/performance.


LaurentMazare:
Thanks, I've removed the `ArgMax` bit as it was added through another PR inbetween.

PR #2238, opened 2024-06-03
Fix dataset id for MNIST

tninesling:
Attempting to load the MNIST dataset currently results in a somewhat confusing error `Api error: Header etag is missing`, which seems to be a symptom of the dataset simply not existing. 

This can be reproduced with
```
use candle_datasets::vision::mnist;

fn main() {
    mnist::load().unwrap();
}
```


LaurentMazare (approved):

LaurentMazare:
Thanks!

LaurentMazare (approved):

PR #2240, opened 2024-06-03
fix to_scalar issue of very high latency

### LINKED ISSUE #2239: Tensor::to_scalar very high latency
I identify this phenomenon by optimizing the performance of training tinyllama, which I want to collect loss value from tensor in cuda. A loss value is a scalar value. but it takes around 300ms to retrieve, which is unusual.

Minimal reproducible example:
```rust
use candle_core::{Device, IndexOp, Tensor};
use std::time::Instant;

fn time_it<T>(func: impl FnOnce() -> T) -> T {
    let start = Instant::now();
    let result = func();
    let duration = start.elapsed();
    println!("{:?}", duration);
    result
}

fn main() {
    let cuda_dev = Device::new_cuda(0).unwrap();
    let cpu_dev = Device::Cpu;

    let shape = [10000, 10000];
    let tensor = Tensor::rand(0f32, 1f32, &shape, &cpu_dev).unwrap();
    println!("Tensor on CPU: {:?}", tensor);

    let tensor = time_it(|| tensor.to_device(&cuda_dev).unwrap());
    println!("Tensor on GPU: {:?}", tensor.layout());

    let cell = time_it(|| tensor.i((42, 42)).unwrap());

    let scalar = time_it(|| cell.to_scalar::<f32>().unwrap());
    dbg!(scalar);
}
```

In the most recent crate version v0.5.1, the line `let scalar = time_it(|| cell.to_scalar::<f32>().unwrap());` will have the identical time consumption of line `let tensor = time_it(|| tensor.to_device(&cuda_dev).unwrap());`.

In my investigation, the problem is because candle's tensor uses a storage + layout approach to represent slice, this approach in most of the time is effective, but it's inefficient when copying a small slice from cuda mem to CPU mem because it will still copy the full storage and use layout to offset to the slice you want. this creates massive latency(Before my attempt to fix ~300ms, after my attempt to fix ~17us).

RoggeOhta:
I'm hoping to resolve this issue.
Since the dimension of the tensor you wish to call to_scalar is 0. 
So I changed the match arm of `CudaStorage` in `to_scalar` by directly calling `dtoh_sync_copy` on a `CudaView<T>` created designated for copy to CPU mem.
Calling `dtoh_sync_copy` directly with `CudaSlice<T>` will cause a full copy on the underlying storage. By calling CudaView<T> will resolve this problem.

Code sample.

```rust
...
CudaStorageSlice::F32(slice) => {
    let sub_slice = slice.slice(offset..offset + 1);
    let device = slice.device();
    let cpu_storage = device.dtoh_sync_copy(&sub_slice).w()?;
    return from_cpu_storage_single(&CpuStorage::F32(cpu_storage));
}
...
```

Fix #2239

### DIFF
--- a/candle-core/src/cuda_backend/mod.rs
+++ b/candle-core/src/cuda_backend/mod.rs
@@ -12,7 +12,7 @@ use half::{bf16, f16};
 #[cfg(feature = "cudnn")]
 pub mod cudnn;
 mod device;
-mod error;
+pub mod error;
 mod utils;
 pub use device::{CudaDevice, DeviceId};
 pub use error::{CudaError, WrapErr};
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -1,12 +1,17 @@
 //! Tensors are N-dimensional matrixes of elements using a single data type.
 #![allow(clippy::redundant_closure_call)]
 use crate::backend::{BackendDevice, BackendStorage};
+#[cfg(feature = "cuda")]
+use crate::cuda_backend::error::WrapErr;
+#[cfg(feature = "cuda")]
+use crate::cuda_backend::CudaStorageSlice;
 use crate::op::{BackpropOp, BinaryOp, CmpOp, Op, ReduceOp, UnaryOp};
 use crate::scalar::TensorOrScalar;
 use crate::shape::{Dim, Dims};
+use crate::CpuStorage;
 use crate::{bail, storage::Storage, DType, Device, Error, Layout, Result, Shape};
-use std::sync::{Arc, RwLock};
 
+use std::sync::{Arc, RwLock};
 /// Unique identifier for tensors.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub struct TensorId(usize);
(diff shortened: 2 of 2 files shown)


RoggeOhta:
Passed all tests with & without cuda feature under my local environment.

RoggeOhta:
> @RoggeOhta do you measure a performance increase when you run with `CUDA_LAUNCH_BLOCKING=1`?

Output of Minimal reproducible example in the linked issue:
All build with release profile.
It's a simple test without serious benchmarking, it can only prove the change in the order of magnitudes.

Before fix with CUDA_LAUNCH_BLOCKING=1
```
Tensor on CPU: Tensor[dims 10000, 10000; f32]
117.986679ms
Tensor on GPU: Layout { shape: [10000, 10000], stride: [10000, 1], start_offset: 0 }
4.219µs
259.142736ms
[candle-test/src/main.rs:26:5] scalar = 0.09058428
```

Before fix without CUDA_LAUNCH_BLOCKING=1
```
Tensor on CPU: Tensor[dims 10000, 10000; f32]
92.308598ms
Tensor on GPU: Layout { shape: [10000, 10000], stride: [10000, 1], start_offset: 0 }
7.003µs
292.98987ms
[candle-test/src/main.rs:26:5] scalar = 0.630551
```

After fix with CUDA_LAUNCH_BLOCKING=1
```
Tensor on CPU: Tensor[dims 10000, 10000; f32]
92.458806ms
Tensor on GPU: Layout { shape: [10000, 10000], stride: [10000, 1], start_offset: 0 }
2.835µs
37.574µs # to_scalar
[candle-test/src/main.rs:26:5] scalar = 0.23715067
```

After fix without CUDA_LAUNCH_BLOCKING=1
```
Tensor on CPU: Tensor[dims 10000, 10000; f32]
159.794325ms
Tensor on GPU: Layout { shape: [10000, 10000], stride: [10000, 1], start_offset: 0 }
10.94µs # to_scalar
24.328µs 
[candle-test/src/main.rs:26:5] scalar = 0.076687455
```

Before fix with CUDA_LAUNCH_BLOCKING=1

PR #2251, opened 2024-06-06
implement if, and pad reflect mode

shua:
The intent of this change is to allow eval of the current silero_vad.onnx (v4). This onnx file uses 'If' and 'Pad' nodes, which had not been supported by onnx::simple_eval until now.

### DIFF
--- a/candle-onnx/src/eval.rs
+++ b/candle-onnx/src/eval.rs
@@ -977,7 +977,11 @@ fn simple_eval_(
             "If" => {
                 // protobuf encodes boolean false as 0 and true as 1
                 let cond = get(&node.input[0])?.get(0)?.to_scalar::<u8>()?;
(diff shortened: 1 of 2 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks, could you confirm whether you got silero_vad working in the end? If it's in a public repo somewhere, certainly happy to put a link to it in the main readme as I don't think we have any example of vad model yet.

shua:
Ah, sadly I wasn't able to run it successfully yet. With Device::Cpu, I keep getting integer overflow panics evaluating some conv1 nodes (not sure if that is expected behaviour for integer overflow or not? I haven't found any onnx documentation defining what to do with over/underflows).

The file can be downloaded from the source at https://github.com/snakers4/silero-vad/tree/master/files

PR #2260, opened 2024-06-09
onnx eval: implement Slice op

shua:
As part of my effort to get silero_vad v4 running, I have also implemented the onnx "Slice" operation.

This change also contains a small fix for the "Cast" operator, specifically to support onnx "bool" datatypes which I've chosen in the "If" op to be represented as u8. This needed a small change to the `dtype` helper function to pass through "Cast" correctly.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2262, opened 2024-06-10
feat(ci): add trufflehog secrets detection

McPatate:
### What does this PR do?

Adding a GH action to scan for leaked secrets on each commit.

### Context

`trufflehog` will scan the commit that triggered the CI for any token leak. `trufflehog` works with a large number of what they call "detectors", each of which will read the text from the commit to see if there is match for a token. For example, the hugging face detector will check for hf tokens and then query our `/api/whoami{-v2}` endpoint to check if the token is valid. If it detects a valid token, the CI will fail, informing you that you need to rotate the token given it leaked.

### References

- https://github.com/trufflesecurity/trufflehog
- https://github.com/marketplace/actions/trufflehog-oss

### DIFF
--- a/.github/workflows/trufflehog.yml
+++ b/.github/workflows/trufflehog.yml
@@ -1,0 +1,21 @@
+on:
+  push:
+
+name: Secret Leaks
+
+permissions:
+  contents: read
+  id-token: write
+  issues: write
+  pull-requests: write
+
+jobs:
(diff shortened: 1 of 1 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2268, opened 2024-06-15
onnx: implement LSTM op

shua:
I'm not sure if, at this point, the simple_eval is the right place to put all this code. It's the thing I implemented, but I could use some direction from code maintainer where this makes sense to add to the codebase.

It was a bit trickier to test this as the docs don't really give inputs/outputs, their tests simply match against an alternative implementation of LSTM. So I've done basically the same thing, where I checked that the implementation here matches the output of pytorch's LSTM node.

### DIFF
--- a/candle-nn/src/rnn.rs
+++ b/candle-nn/src/rnn.rs
@@ -55,6 +55,10 @@ pub struct LSTMState {
 }
 
 impl LSTMState {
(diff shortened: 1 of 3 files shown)


shua:
ah, I just found the LSTM implementation in candle-nn, don't know how I missed it before, I think I'll try to use that instead of the one here.

LaurentMazare:
Thanks and sorry for the delay.

PR #2279, opened 2024-06-20
Depth Anything v2

jeroenvlek:
This PR adds [Depth Anything V2](https://depth-anything-v2.github.io/), which is based on the already present DinoVisionTransformer.  I'm still adding the [Spectral-R coloring](https://huggingface.co/spaces/depth-anything/Depth-Anything-V2/blob/main/run.py#L51), but this led me to another question on the list, so I thought it wise to open a (preliminary) PR to get some early answers and feedback. So here goes:

1. Currently the port is following the Python code very closely. However, the separation of concerns in the original code is not necessarily the level of encapsulation I would have chosen (e.g. the scratch layer is accessed from the head). What is your standpoint here: Stick to the original code as closely as makes sense, or improve it where you can?
2. Can I add palette as a dep to candle-examples for the spectral coloring (or use an alternative)?
3. Currently the module is named depth_anything_v2, is that ok or do you prefer a different scheme (no/less underscores)?
4. Developing this model it was very useful for me to print the list of safetensor names, shall I add that functionality to candle-examples/src/lib.rs, or maybe even the safetensors related code? (It's only a few lines, but I'm guessing more people could benefit)
5. I've added an Identity() op, analog to the one in [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.Identity.html), there's other ways to deal with a no-op though, so is this ok?

### Example:

```rust
    let out_channel_sizes = vec![48, 96, 192, 384];
    const IN_CHANNEL_SIZE: usize = 384;
    const NUM_FEATURES: usize = 64;
    let depth_anything = DepthAnythingV2::new(
        &dinov2,
        IN_CHANNEL_SIZE,
        out_channel_sizes,
        NUM_FEATURES,
        vb,
    )?;
    let (original_height, original_width, image) = load_and_prep_image(&args.image, &device)?;

    let depth = depth_anything.forward(&image)?;

    let depth = post_process_image(&depth, original_height, original_width)?;

    save_image(&depth, output_path)?;

### DIFF
--- a/candle-examples/Cargo.toml
+++ b/candle-examples/Cargo.toml
@@ -26,7 +26,7 @@ image = { workspace = true }
 intel-mkl-src = { workspace = true, optional = true }
 num-traits = { workspace = true }
 palette = { version = "0.7.6", optional = true }
-enterpolation = "0.2.1"
+enterpolation = { version = "0.2.1", optional = true}
 pyo3 = { version = "0.21.0", features = ["auto-initialize"], optional = true }
 rayon = { workspace = true }
 rubato = { version = "0.15.0", optional = true }
@@ -67,7 +67,7 @@ onnx = ["candle-onnx"]
 metal = ["candle/metal", "candle-nn/metal"]
 microphone = ["cpal"]
 encodec = ["cpal", "symphonia", "rubato"]
-depth_anything_v2 = ["palette"]
+depth_anything_v2 = ["palette", "enterpolation"]
 
 [[example]]
 name = "llama_multiprocess"
--- a/candle-examples/Cargo.toml
+++ b/candle-examples/Cargo.toml
@@ -25,6 +25,8 @@ hf-hub = { workspace = true, features = ["tokio"] }
 image = { workspace = true }
 intel-mkl-src = { workspace = true, optional = true }
 num-traits = { workspace = true }
+palette = { version = "0.7.6", optional = true }
+enterpolation = "0.2.1"
 pyo3 = { version = "0.21.0", features = ["auto-initialize"], optional = true }
 rayon = { workspace = true }
 rubato = { version = "0.15.0", optional = true }
@@ -65,6 +67,7 @@ onnx = ["candle-onnx"]
 metal = ["candle/metal", "candle-nn/metal"]
 microphone = ["cpal"]
 encodec = ["cpal", "symphonia", "rubato"]
+depth_anything_v2 = ["palette"]
 
 [[example]]
 name = "llama_multiprocess"
@@ -101,3 +104,7 @@ required-features = ["candle-datasets"]
 [[example]]
 name = "encodec"
 required-features = ["encodec"]
+
+[[example]]
+name = "depth_anything_v2"
+required-features = ["depth_anything_v2"]
--- a/candle-examples/examples/depth_anything_v2/color_map.rs
+++ b/candle-examples/examples/depth_anything_v2/color_map.rs
@@ -1,5 +1,5 @@
-use enterpolation::Generator;
 use enterpolation::linear::ConstEquidistantLinear;
+use enterpolation::Generator;
 use palette::LinSrgb;
 
 use candle::Tensor;
--- a/candle-examples/examples/depth_anything_v2/color_map.rs
+++ b/candle-examples/examples/depth_anything_v2/color_map.rs
@@ -1,7 +1,8 @@
-use candle::{Shape, Tensor};
-use enterpolation::linear::ConstEquidistantLinear;
 use enterpolation::Generator;
+use enterpolation::linear::ConstEquidistantLinear;
 use palette::LinSrgb;
+
+use candle::Tensor;
 
 pub struct SpectralRColormap {
     gradient: ConstEquidistantLinear<f32, LinSrgb, 9>,
@@ -42,7 +43,7 @@ impl SpectralRColormap {
             candle::bail!("Not enough dims!")
         };
 
-        let color = Tensor::from_vec(rgb_values, Shape::from((*height, *width, 3)), gray.device())?;
+        let color = Tensor::from_vec(rgb_values, (*height, *width, 3), gray.device())?;
 
         color.permute((2, 0, 1))
     }
--- a/candle-examples/examples/depth_anything_v2/color_map.rs
+++ b/candle-examples/examples/depth_anything_v2/color_map.rs
@@ -1,0 +1,49 @@
+use candle::{Shape, Tensor};
+use enterpolation::linear::ConstEquidistantLinear;
+use enterpolation::Generator;
+use palette::LinSrgb;
+
+pub struct SpectralRColormap {
+    gradient: ConstEquidistantLinear<f32, LinSrgb, 9>,
+}
+
+impl SpectralRColormap {
+    pub(crate) fn new() -> Self {
+        // Define a colormap similar to 'Spectral_r' by specifying key colors.
+        // got the colors from ChatGPT-4o
+        let gradient = ConstEquidistantLinear::<f32, _, 9>::equidistant_unchecked([
+            LinSrgb::new(0.3686, 0.3098, 0.6353), // Dark blue
+            LinSrgb::new(0.1961, 0.5333, 0.7412), // Blue
+            LinSrgb::new(0.4000, 0.7608, 0.6471), // Cyan
+            LinSrgb::new(0.6706, 0.8667, 0.6431), // Green
+            LinSrgb::new(0.9020, 0.9608, 0.5961), // Yellow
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by LaurentMazare** [comment]
Thanks for putting this up, this looks great and looking forward to merging this!

> 1. Currently the port is following the Python code very closely. However, the separation of concerns in the original code is not necessarily the level of encapsulation I would have chosen (e.g. the scratch layer is accessed from the head). What is your standpoint here: Stick to the original code as closely as makes sense, or improve it where you can?

I would tend to stay close to the python version, this makes it easier for other people to understand the conversion process and also is better if some further changes are made to the python side and we want to import them.

> 2. Can I add palette as a dep to candle-examples for the spectral coloring (or use an alternative)?

Let's maybe make it a cargo feature that is not enabled by default? Something similar to [this example](https://github.com/huggingface/candle/blob/36cf54525d93660f62c3601ba0988653f3567e0e/candle-examples/Cargo.toml#L91).

> 3. Currently the module is named depth_anything_v2, is that ok or do you prefer a different scheme (no/less underscores)?

Looks good to me.

> 4. Developing this model it was very useful for me to print the list of safetensor names, shall I add that functionality to candle-examples/src/lib.rs, or maybe even the safetensors related code? (It's only a few lines, but I'm guessing more people could benefit)

Sure, let's maybe have a separate PR for adding this? If it's general enough and small, `candle-core` might be a good idea.

> 5. I've added an Identity() op, analog to the one in [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.Identity.html), there's other ways to deal with a no-op though, so is this ok?

Yes that seems like a good thing to have.

File: candle-examples/Cargo.toml [hunk 1]
```diff
@@ -25,6 +25,8 @@ hf-hub = { workspace = true, features = ["tokio"] }
 image = { workspace = true }
 intel-mkl-src = { workspace = true, optional = true }
 num-traits = { workspace = true }
+palette = { version = "0.7.6", optional = true }
+enterpolation = "0.2.1"
```
LaurentMazare, line 29: Could we also make enterpolation optional?

candle-examples/examples/depth_anything_v2/color_map.rs
```diff
@@ -0,0 +1,49 @@
...
+        println!("Gray: {:?}", gray.dims());
+        let gray_values: Vec<f32> = gray.flatten_all()?.to_vec1()?;
+        let rgb_values: Vec<f32> = gray_values
+            .iter()
+            .map(|g| self.get_color(*g))
+            .flat_map(|rgb| [rgb.red, rgb.green, rgb.blue])
+            .collect();
+
+        let [.., height, width] = gray.dims() else {
+            panic!("Not enough dims!")
```
LaurentMazare, line 42: Use `candle::bail!` rather than `panic!` to notify an error.

candle-examples/examples/depth_anything_v2/color_map.rs
```diff
@@ -0,0 +1,49 @@
...
+        };
+
+        let color = Tensor::from_vec(rgb_values, Shape::from((*height, *width, 3)), gray.device())?;
```
LaurentMazare, line 45: No need to have the `Shape::from` here, the tuple only should work.

candle-nn/src/ops.rs
```diff
@@ -926,3 +926,19 @@ pub fn replication_pad2d(xs: &Tensor, pad: usize) -> Result<Tensor> {
         n => candle::bail!("replication-pad with a size of {n} is not supported"),
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct Identity {}
```
LaurentMazare, line 931: If we don't plan on providing any argument, `pub struct Identity;` seems very slightly better.

candle-transformers/src/models/dinov2.rs
```diff
@@ -258,6 +258,69 @@ impl DinoVisionTransformer {
...
+    fn get_intermediate_layers_not_chunked(&self, xs: &Tensor, blocks_to_take: Vec<usize>) -> Result<Vec<Tensor>> {
+        let mut xs = self.prepare_tokens_with_mask(xs)?;
+        let mut output = Vec::new();
+        for (i, blk) in self.blocks.iter().enumerate() {
+            xs = blk.forward(&xs)?;
+            if blocks_to_take.contains(&i) {
+                output.push(xs.clone());
+            }
+        }
+        assert_eq!(output.len(), blocks_to_take.len(), "only {} / {} blocks found", output.len(), blocks_to_take.len());
```
LaurentMazare, line 271: Make this an error rather than a `assert` that would result in a panic.

candle-transformers/src/models/depth_anything_v2.rs
```diff
@@ -0,0 +1,535 @@
...
+    batch_norm1: Option<BatchNorm>,
+    batch_norm2: Option<BatchNorm>,
+}
+
+impl ResidualConvUnit {
+    pub fn new(
+        num_features: usize,
+        activation: Activation,
+        use_batch_norm: bool,
+        var_builder: VarBuilder,
```
LaurentMazare, line 24: The usual "convention" in all other models is that the `VarBuilder` is named `vb` for shortness and we call `vb.pp` for `push_prefix`.

candle-transformers/src/models/depth_anything_v2.rs
```diff
@@ -0,0 +1,535 @@
...
+impl FeatureFusionBlock {
+    pub fn new(
+        num_features: usize,
+        activation: Activation,
+        use_batch_norm: bool,
+        target_patch_size: usize,
+        var_builder: VarBuilder,
+    ) -> Result<Self> {
+        const KERNEL_SIZE: usize = 1;
+        let conv_cfg = Conv2dConfig {
```
LaurentMazare, line 120: You can use `Default::default` for the default convolution parameters rather than repeating them.

candle-transformers/src/models/depth_anything_v2.rs
```diff
@@ -0,0 +1,535 @@
...
+}
+pub const DINO_IMG_SIZE: usize = 518;
+const PATCH_SIZE: usize = 37; //  518 // 14 TODO see how to solve this dynamically
+
+impl Module for DPTHead {
+    fn forward(&self, xs: &Tensor) -> Result<Tensor> {
+        let out = (0..NUM_CHANNELS)
+            .map(|i| {
+                let x = if self.use_class_token {
+                    let x = xs.get(i).unwrap().get(0).unwrap();
```
LaurentMazare, line 428: Convert all the `unwrap` to actual errors using the `?` operator.

candle-transformers/src/models/depth_anything_v2.rs
```diff
@@ -0,0 +1,535 @@
...
+            pretrained,
+            depth_head,
+        })
+    }
+}
+
+impl<'a> Module for DepthAnythingV2<'a> {
+    fn forward(&self, xs: &Tensor) -> Result<Tensor> {
+        let dims = xs.shape().dims();
+        println!("DepthAnythingV2 got dims: {:?}", dims);
```
LaurentMazare, line 521: We should not have any print statement in the model code as it can be used as a library, we also prefer not to have any form of logging. Better if you can do this on the `candle-examples` side.

candle-transformers/src/models/depth_anything_v2.rs
jeroenvlek: But this is a closure, so I have to use unwrap, or how could I do that?

LaurentMazare: The closure can return a `Vec<Result<_>>` and then `collect` can convert that to a result of vec on which you can use the question mark operator (or alternatively you can just make this a for loop).

jeroenvlek: True,  maybe an iterator was too fancy, was focusing on immutabilty too much. Will fix.

candle-transformers/src/models/depth_anything_v2.rs
jeroenvlek: Thanks for the tip, but padding is 1 here, not 0 like in the default.

LaurentMazare (approved):

LaurentMazare:
Thanks!

candle-transformers/src/models/depth_anything_v2.rs
asukaminato0721: > Thanks for the tip, but padding is 1 here, not 0 like in the default.

```rs
        let conv_cfg = Conv2dConfig {
            padding: 1,
            ..Default::default()
        };
```

jeroenvlek: Awesome, thanks. Will put this in the second PR

PR #2288, opened 2024-06-25
Depth Anything V2: Bigger Models

jeroenvlek:
Sorry to open a draft again, but while adding more models I found out why the DinoV2 example loads its safetensors from a different location than Facebook's own HF space: Facebook's safetensors have the query, key and value layers separate, whereas the original Python code, and thus the Candle implementation, have a single QKV layer. Furthermore, the original model file relies on separate head weights (guessing the head is not even necessary for DepthAnything, so might remove from example)

**So what I've done here:**

* Load the weights from Facebook's HF space and pack them together into a linear QKV layer
* Make the head optional with an optional VarBuilder
* Downloaded all head files from Facebook's GitHub and converted them to safetensors in my [own HF space](https://huggingface.co/jeroenvlek/dinov2-linear-heads-safetensors/tree/main)

**Question:** Do you agree with this approach? 

Happy to adapt to other ideas, but if you agree, I will continue adding the base, large and giant models.

PS: I saw you also did some formatting on my previous PR and fixed some Clippy warnings. Maybe it's an idea to have a Contributor page that details the formatting requirements and other conventions?

### DIFF
--- a/candle-examples/examples/depth_anything_v2/main.rs
+++ b/candle-examples/examples/depth_anything_v2/main.rs
@@ -12,7 +12,7 @@ use std::path::PathBuf;
 use clap::Parser;
 
 use candle::DType::{F32, U8};
-use candle::{DType, Device, Module, Result, Tensor};
+use candle::{Device, Module, Result, Tensor};
 use candle_examples::{load_image, load_image_and_resize, save_image};
 use candle_nn::VarBuilder;
 use candle_transformers::models::depth_anything_v2::{DepthAnythingV2, DepthAnythingV2Config};
@@ -34,6 +34,9 @@ struct Args {
     dinov2_model: Option<PathBuf>,
 
     #[arg(long)]
+    dinov2_head: Option<PathBuf>,
+
+    #[arg(long)]
     depth_anything_v2_model: Option<PathBuf>,
 
     #[arg(long)]
@@ -53,21 +56,34 @@ pub fn main() -> anyhow::Result<()> {
     let args = Args::parse();
     let device = candle_examples::device(args.cpu)?;
 
-    let dinov2_model_file = match args.dinov2_model {
+    let model_path = match args.dinov2_model {
         None => {
             let api = hf_hub::api::sync::Api::new()?;
-            let api = api.model("lmz/candle-dino-v2".into());
-            api.get("dinov2_vits14.safetensors")?
+            let api = api.model("facebook/dinov2-small".into());
+            api.get("model.safetensors")?
         }
-        Some(dinov2_model) => dinov2_model,
+        Some(path) => path,
     };
-    println!("Using file {:?}", dinov2_model_file);
+    println!("Using dinov2 file {:?}", model_path);
 
-    let vb = unsafe { VarBuilder::from_mmaped_safetensors(&[dinov2_model_file], F32, &device)? };
-    let dinov2 = dinov2::vit_small(vb)?;
+    let vb = unsafe { VarBuilder::from_mmaped_safetensors(&[model_path], F32, &device)? };
+
+    let head_path = match args.dinov2_head {
+        None => {
+            let api = hf_hub::api::sync::Api::new()?;
+            let api = api.model("jeroenvlek/dinov2-linear-heads-safetensors".into());
+            api.get("dinov2_vits14_linear_head.safetensors")?
+        }
+        Some(path) => path,
+    };
+    println!("Using dinov2 head file {:?}", head_path);
+
+    let vb_head = unsafe { VarBuilder::from_mmaped_safetensors(&[head_path], F32, &device)? };
+
+    let dinov2 = dinov2::vit_small(vb, vb_head)?;
     println!("DinoV2 model built");
 
-    let depth_anything_model_file = match args.depth_anything_v2_model {
+    let depth_anything_path = match args.depth_anything_v2_model {
         None => {
             let api = hf_hub::api::sync::Api::new()?;
             let api = api.model("jeroenvlek/depth-anything-v2-safetensors".into());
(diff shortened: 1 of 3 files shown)


jeroenvlek:
I've added base, large and giant. The latter only for Dino as the DepthAnything is still pending review.

Also removed the awkard config object handling.

jeroenvlek:
There still seems to be a bug in there, comparing this large implementation with the web based model output. This output has jitter and less pronounced depth. Will investigate.

jeroenvlek:
Just noticed that _load_image_ returns (channels, height, width) but _load_image_and_resize_ returns (channels, width, height)

jeroenvlek:
Small update: Originally I thought the difference in output quality was due to model size, but there were still some bugs in there. Most seem to be fixed. It's still quite hard  to get the input conditions the same. For example, they convert to RGB from BGR, but then they resize with OpenCV and they're back to BGR (I verified this in the Python code by setting channels to zero).

jeroenvlek:
@LaurentMazare I uploaded a new depth image of the bike picture. In your experience, do you think this could be caused by the difference between _UpsampleNearest2D_ and _F.interpolate(out, (int(patch_h * 14), int(patch_w * 14)), mode="bilinear", align_corners=True)_?

LaurentMazare:
It's hard to say but I wouldn't expect it to make that much of a difference - I'm assuming that the image that you uploaded is the python output which looks great vs the rust output which is not that good. Usually for the bits that are not supported in candle, I just modify the python code to have the candle behavior e.g. I would try without the bilinear on the python side here to check that the python output is still meaningful.

jeroenvlek:
What I uploaded is the Rust output of the large model . The Python output is smoother. That's why I'm suspicious of the interpolation (used also deep inside the feature fusion blocks).

Let me investigate and apologizes for all the noise :)

jeroenvlek:
@LaurentMazare I need a bit of guidance here: The pushed code follows the flow of the Python implementation fairly correct: The logic is the same. However, the results are still not as good as they should.

Python small:

small_20240224_072209

Rust small (this branch):

3_depth_20240224_072209

This is still not great. So I went further and opened another branch (not pushed) where I added: 

* image loading and resizing using opencv (I understand you might not want that as dep, this is just for testing)
* Bilinear interpolation (Only in CPU storage, there doesn't seem to be a default Cuda kernel and it seems Candle doesn't have custom kernels)
* Lots of debug statements and in between statistics (min, max and mean)
* Load the Dino weights from the DepthAnythingV2 safetensors file and not from Facebook

The OpenCV loading and prep brings the inputs fairly close together, but then already in the DinoV2 model, things start to diverge rather badly:

[model_layer_analysis.ods](https://github.com/user-attachments/files/16080553/model_layer_analysis.ods)

Now, I'm not expecting you to follow the analysis here, but there's differences over 1400% between the means there.

My main question would be: How well do other ports of PyTorch models to Candle follow the original numerically? How much tolerance is there for numerical deviation? I would argue that in principle there shouldn't be any deviations at all and these statistics should match rather closely?

jeroenvlek:
Submitting this PR for now, as it is definitely an improvement over #2279  

There's still unexplained numerical deviations in there, that already start in the applications of the DinoV2 blocks. Whether, this is something fundamental or obvious I was not able to discover, but I have stared at this problem long enough :)

LaurentMazare:
Thanks for looking into this, the blocky results with your latest version certainly feel like some form of interpolation is off (or some upsampling of the depth results?). Anyway the PR looks good but would it be possible not to make changes on the dinov2 side? This model code might already be used so better not to break compatibility and try to reduce the diff as much as possible.

jeroenvlek:
I wouldn't mind doing that, but that was a point of the first post: The weights Facebook put on HF have the query, key and value tensors separate.

However, I believe the depth anything weights are packed in the old/different format, so let me try with that.

Regarding interpolation: I tried bilinear on the candle side (like mentioned above Cpu only), and I tried nearest neighbor interpolation inside the Pytorch implementation's interpolate_pos_encosing() [like your comment here](https://github.com/huggingface/candle/blob/ecff05d72b9bbbdf1d0d07f06ebf5b7b01bfe3db/candle-transformers/src/models/dinov2.rs#L245l)

According to my rudimentary analysis the mean divergence already starts while getting the intermediate blocks (so the forward method of the Dino blocks producers different results)

Anyway, happy to reduce the diff. Will work on that tomorrow  :)

jeroenvlek:
> Thanks for looking into this, the blocky results with your latest version certainly feel like some form of interpolation is off (or some upsampling of the depth results?). Anyway the PR looks good but would it be possible not to make changes on the dinov2 side? This model code might already be used so better not to break compatibility and try to reduce the diff as much as possible.

@LaurentMazare Just to be clear: These new prefixes follow [the Facebook naming](https://huggingface.co/facebook/dinov2-small/tree/main?show_file_info=model.safetensors) and they also have the bigger models (base, large, giant). If we keep the old prefixes, then we need to compact the Q, K, V layers into a single layer and export the 3 models to safetensors. 

Is that really what you want? (I mean Candle uses versioning, although I do understand these are breaking changes)

metalmatze:
Where are we with this? 
It would be great to at least get the support for loading facebook models directly. I think that should be opened as a separate PR again and merged. The rest can probably live in this PR.

jeroenvlek:
> Where are we with this? It would be great to at least get the support for loading facebook models directly. I think that should be opened as a separate PR again and merged. The rest can probably live in this PR.

Where are we with this? I haven't touched this since over a year, since I felt I was commenting into the void. Unfortunately, I don't have time atm to put the Facebook model loading into a separate PR, but feel free to cherry-pick/copy what you need.

Or alternatively, we can merge this, but we accept that the performance of DepthAnythingV2 is visibly worse than the Python implementation due to numerical deviations that I was unable to pin down.

pcuenca:
Small update: [this PR](https://github.com/huggingface/candle/pull/3172) fixes, I think, the first bad example from [this comment](https://github.com/huggingface/candle/pull/2288#issuecomment-2219909391).

I'm using [this repo](https://huggingface.co/pcuenq/candle-depth-anything-v2-small) for my tests, with weights copied from @jeroenvlek's and Laurent's repos.

jeroenvlek:
One line? Lol. Thank you!

jeroenvlek:
@pcuenca I see your PR is merged. Shall I close this one? Or is it still needed?

pcuenca:
Hi @jeroenvlek! Your analysis is very interesting, as well as your experiments with bilinear scaling! The PR that was merged works better than what was there before, but there's certainly room for improvement. I'm not sure if any of the pieces you contributed in this PR would further improve results, what do you think?

jeroenvlek:
I'm totally fine with closing it. I learned a lot about Rust and Candle and seeing that my mistake was a single line and not something structural also puts my mind at ease ;)

I'll close it for now, as I unfortunately don't have much time for this project.

PR #2296, opened 2024-06-28
Add AVG Pooling cpu implementation

WenheLI:


EricLBuehler:
@WenheLI I think this is correct.

WenheLI:
@EricLBuehler - Thanks! I guess another question is that in the CPU backend implementation, we should be able to speed this up by using vectorization. Not sure in candle's codebase, do we already have some infrastructure that can help us?

EricLBuehler:
Hi @WenheLI, I think you could use something like Rayon, just replace the for loops (probably just choose one to replace as rayon uses the number of CPU cores as the number of threads by default) and replace `.iter()` with `.par_iter()`.

WenheLI:
Thanks! Added vectorization. Wondering if someone can I take a look and review this>

PR #2298, opened 2024-06-29
Add method to access a sub-layer in Sequential.

iwanders:
Trivial addition to [Sequential](https://docs.rs/candle-nn/latest/candle_nn/sequential/struct.Sequential.html) that allows obtaining a reference to a sublayer.

I used an index type of `i64` because that's what `len()` currently returns, I'm not sure why it is using an `i64` and not an `usize`?


iwanders:
Closing this: I think sequential is intended as a trivial forward-pass only container.

Issue #2304, opened 2024-07-01
Indexing the first dim of a 2D tensor with a 1D tensor


v-espitalier:
Hi,

I am translating Python code to Rust and can't figure out how to index a 2D tensor with a 1D one.
Can you please advise ? Thanks for any help.

Here is the code snippet.
```
println!("table.dims: {:?}", table.dims());
> table.dims: [2212, 12]

println!("index_list.dims: {:?}", index_list.dims());
> index_list.dims: [332929]       // (with tensor values within 0..2212)

//let bias = table[index_list];   // From Python, not working in Rust
let bias = table.gather(&index_list, 0)?;
> Error: shape mismatch in gather, lhs: [2212, 12], rhs: [332929]
Expected:
println!("bias.shape: {:?}", relative_position_bias.dims());
> bias.shape: torch.Size([332929, 12])
```

I think I need to use another function than **gather()**,
maybe get inspired by : https://github.com/huggingface/candle/issues/855
but I get lost with argmax_keepdim() and broadcast_as() functions..

EricLBuehler:
Hi @v-espitalier, I think the function you want is [`index_select`](https://docs.rs/candle-core/latest/candle_core/struct.Tensor.html#method.index_select). All you need to do it pass the indices (a 1 dim tensor), and the dimension to index on, and it will do the equivalent of Pytorch slice indexing with a tensor in that dim (as long as the tensor contains indices).

v-espitalier:
Thank you very much. It worked :-)

Issue #2303, opened 2024-06-30
Change single value of tensor


v-espitalier:
Hi,

I guess I missed sth, as I can't figure out how to update a single value of a tensor.
(I know it's not good practice to write code with loops on indexes; it's for testing / debugging purpose)
How to change a single value of a tensor ?
Thanks for any answer

```
let mut t = Tensor::zeros((10, 10), DType::F32, &Device::Cpu)?;

t[5][5] = 3.14;
t[5][5].to_scalar() = 3.14;
t.i((5, 5))? = 3.14;
t.i((5, 5))?.to_scalar()? = 3.14;
t.i((5, 5))? = 3.14 * Tensor::ones((1), DType::F32, &Device::Cpu)?;
```

EricLBuehler:
You can do this with the `slice_assign` method:
```rust
let a = Tensor::zeros((10, 10), DType::F32, &Device::Cpu)?;
let to_insert = (Tensor::ones((1), DType::F32, &Device::Cpu)? * 3.14f64)?;
// 0..1 means at index 0.
let b = a.slice_assign(&[0..1, 0..1], &to_insert)?;
```

v-espitalier:
Thank you for your quick answer. Could not have found it by myself.

EricLBuehler:
Happy to help!

PR #2305, opened 2024-07-01
Add Beit model ( https://arxiv.org/abs/2106.08254 )

v-espitalier:
Hi,
I implemented the Beit model ( https://arxiv.org/abs/2106.08254 ) from MS.
The provided weights are for ImageNet classification, and come from timm.
( We use this model as well as DINOv2Reg4 for our research in CIRAD/Pl@ntNet )
The code works fine without warning. I used clippy and fmt.

Thanks for your time,
Regards,
Vincent.

PS: There will be another commit / pull request later on, on 'beit.rs' only, related to the indexes for the relative positional encoding. For now, I included these indexes as a single additional tensor, to the safetensor weights (+3MB on top of 330 MB), instead of implementing the associated function. I am working on that point, but it may take a bit of time, as the missing function is not 'trivial'.

### DIFF
--- a/candle-examples/examples/beit/README.md
+++ b/candle-examples/examples/beit/README.md
@@ -1,0 +1,20 @@
+# candle-beit
+
+[Beit](https://arxiv.org/abs/2106.08254) is a computer vision model.
+In this example, it is used as an ImageNet classifier: the model returns the
+probability for the image to belong to each of the 1000 ImageNet categories.
+
+## Running some example
+
+```bash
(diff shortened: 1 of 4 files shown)


v-espitalier:
By the way, thank you @EricLBuehler for your help on #2303 and #2304 , and the index_select() solution , it helped me with the implementation of the relative positional bias, which is specific to Beit model.

LaurentMazare (approved):

v-espitalier:
Thank you Laurent for approving and launching these tests.

PR #2318, opened 2024-07-07
fix cpu backend: pow, broadcast_pow

shua:
pow implemented as f32/f64 ln and exp works only for positive base. Then intent of this change is to support negative bases as well, and to that end, the current ln+exp code has been removed in favour of stdlib pow functions or accelerated equivalents.


shua:
closing this in favour of #2439 , as that change is sufficient for my needs.

PR #2320, opened 2024-07-07
fix cpu backend: index_select fails with negative index

shua:
currently the cpu backend incorrectly converts negative i64 to usize by casting, but it should check if the index is negative, and in that case, count back from the end of the thing being indexed.

This change fixes the behaviour by adding a private trait and implementing it for all valid index types. If any more types are added as index types, the compiler should direct people to implement this same trait to calculate the index properly.


shua:
#2440 was sufficient

PR #2321, opened 2024-07-07
silero-vad v5 example

shua:
This change adds an example of how to run silero-vad v5


LaurentMazare:
Looks pretty good, would you mind renaming the example from `vad` to something like `silero-vad`? We try to name the examples after the models that they run rather than the general feature that they provide.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2325, opened 2024-07-09
Add Mobilenet v4

janimo:
Since these models can use images of different resolutions as input I've made the load_image() helper more generic.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2334, opened 2024-07-15
feat(candle-transformers/models/codegeex4-9b): add codegeex4-9

donjuanplatinum:
add codegeex4-9b transformers models support


LaurentMazare:
Looks pretty good thanks, I've push some small cosmetic tweaks on top of it.
Would you mind adding a readme on the example? Ideally this would include an example command to run the model as well as the output that was generated.

donjuanplatinum:
> Looks pretty good thanks, I've push some small cosmetic tweaks on top of it. Would you mind adding a readme on the example? Ideally this would include an example command to run the model as well as the output that was generated.

hello, Done

LaurentMazare:
Thanks!

Issue #2322, opened 2024-07-08
bug on aarch64-apple-ios: Buffer Validation Illegal MTLStorageMode 0x10


evilsocket:
I'm running candle with metal acceleration on iOS via uniffi, specifically:

```
candle-core = { version = "0.6.0", features = ["metal"] }
candle-nn = { version = "0.6.0", features = ["metal"] }
candle-transformers = { version = "0.6.0", features = ["metal"] }
```

But the call to `Device::new_metal` panics with:

```
-[MTLDebugDevice newBufferWithBytes:length:options:]:670: failed assertion `Buffer Validation
Illegal MTLStorageMode 0x10'
```

After some debugging, I figured this happens on [this line](https://github.com/huggingface/candle/blob/9cd54aa5d4fb6cf30e0df2d198c8a387db2d9144/candle-core/src/metal_backend/mod.rs#L1802) due to iOS rejecting the `metal::MTLResourceOptions::StorageModeManaged` option.

I could replicate with this simple test:

```rust
#[uniffi::export]
pub fn test_metal() {
    let device = metal::Device::all().swap_remove(0);

    println!("device: {:?}", &device);

    println!(
        "MTLResourceOptions::StorageModeManaged = 0x{:x}",
        metal::MTLResourceOptions::StorageModeManaged
    );

    // it panics here
    let seed = device.new_buffer_with_data(
        [299792458].as_ptr() as *const std::ffi::c_void,
        4,
        metal::MTLResourceOptions::StorageModeManaged, // <-- 0x10
    );

    println!("seed: {:?}", &seed);
}
```

Which prints and then panics:

```
device: <CaptureMTLDevice: 0x105207820> -> <MTLDebugDevice: 0x105207580> -> <AGXG16Device: 0x10800f400>
    name = Apple A17 Pro GPU
MTLResourceOptions::StorageModeManaged = 0x10
-[MTLDebugDevice newBufferWithBytes:length:options:]:670: failed assertion `Buffer Validation
Illegal MTLStorageMode 0x10
'
```

Other values such as StorageModeShared do not panic, but i'm not sure of the implications also because StorageModeManaged is used in several other places in the metal implementation.

evilsocket:
also mentioned in https://github.com/huggingface/candle/issues/1841

evilsocket:
[...]

PR #2335, opened 2024-07-15
WIP: Precompile metal kernels into `.metallib` files

zackangelo:
I'm running into an issue where the first time I call `apply_repeat_penalty`, it takes a very long time (in excess of 6 seconds). It seems to be coming from the `Tensor::to_vec1d` call to move the logits into a `Vec<f32>`. It seems like a simple copy like this would be very fast. 

It was suggested on Discord that this slowness might be due to some other async stuff happening in Metal, maybe the compilation of the kernels on first load. 

This PR precompiles the kernels at build time instead of on every run. Unfortunately, it doesn't seem to solve my problem but it might be useful for other reasons.

### DIFF
--- a/candle-metal-kernels/Cargo.toml
+++ b/candle-metal-kernels/Cargo.toml
@@ -23,3 +23,7 @@ half = { version = "2.3.1", features = [
   "rand_distr",
 ] }
 rand = "0.8.5"
+
+[build-dependencies]
+anyhow = "1.0.44"
+convert_case = "0.6.0"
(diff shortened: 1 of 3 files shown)


zackangelo:
Looking at #2322, Will likely need to reconfigure the build script to optionally produce iOS `.metallibs`

zackangelo:
@LaurentMazare is this something you think you would want to potentially merge? if so I can clean it up. Otherwise, we can close it.

zackangelo:
Going to close this due to inactivity, if it's something we feel like we need in the future this should serve as a good starting point.

PR #1910, opened 2024-03-21
Backwards for ConvTranspose2D

KGrewal1:
* Add backwards for ConvTranspose2D
* Add test for grad of ConvTranspose2D (chose to use somewhat of an edge case to show correctness, can add a more 'normal' test if wanted)
* Add documentation to backprop module

PR #2337, opened 2024-07-16
Fix for backprop in ConvTranspose2D with stride of 2

iwanders:
Hi, I ran into a shape mismatch error inside the backpropagation of the deconvolution when training an implementation of FCN. The stride of 2 combined with a dilation of 1 seemed to be the problem.

After a bit of reading I think the stride and dilation are just swapped around? I can't claim I actually wrote out the math myself. However, with this change, all the unit tests in `candle-core` still pass. The added unit test for stride of 2 and dilation of 1 shows no shape mismatch error occurs anymore, and the values are very close to the ones as calculated by torch.

Commits are split out, in case that helps review / reproduction:
- da243e42ac0e919263aaceb21a219a3518c09b62 merely aids the failing test with the stride of 2, this results in the shape mismatch from [this line](https://github.com/iwanders/huggingface-candle/blob/669182b896c75c91b6ba1418529214bcdad376cc/candle-core/src/backprop.rs#L325).
- 0eb4006c76e56d5c06d36548de7df0ee4c291df9 swaps the dilation and stride in `backprop.rs`, after this the added unit test no longer fails.
- da243e42ac0e919263aaceb21a219a3518c09b62 adds the full gradient test to confirm numbers are identical.

Numbers are created using [this python snippet](https://github.com/iwanders/huggingface-candle/blob/669182b896c75c91b6ba1418529214bcdad376cc/candle-core/tests/conv_tests.rs#L588-L614) from the unit test above the newly added one, with just those four parameters modified (calculated using torch 2.3.1, python 3.11.9).

The backwards method for ConvTranspose2D was added in #1910, by @KGrewal1, hopefully they are in a better position to comment on the correctness of this fix.


LaurentMazare:
Thanks for the fix and the details + tests, looks pretty good to me. There are a few clippy failures though in the CI, would you mind looking at them ? (as some occur in the tests, you might want to pass the `--tests` flag to the clippy command when running locally)

iwanders:
> you might want to pass the --tests flag to the clippy command when running locally

Yes, that's the problem thanks! I've amended the last commit as all failures occurred in that commit. Passes locally now.

LaurentMazare:
Thanks!

KGrewal1:
Thanks for this fix, sorry have been a lot busier than I was before with work!

PR #2346, opened 2024-07-20
Add masked_fill under Tensor

fgsch:
This unifies the `masked_fill` implementations under Tensor.


LaurentMazare:
I'm not sure we want to have this under `Tensor`, the goal there is not to replicate all the functions in pytorch but rather have a smaller subset of basic functions. Maybe put it in `candle-transformers/src/utils.rs` instead.
Also note that the tensor pre-allocation for some of the models was done on purpose as it avoids a sync point on cuda which was a problem for performance so we would want to keep the current implementation for these.

fgsch:
@LaurentMazare thanks for the comments. I will update the PR accordingly.

Issue #2356, opened 2024-07-25
Tensor::log_sum_exp() loses precision for small numbers


leejuyuu:
The current implementation of the `Tensor::log_sum_exp()` function does not handle small numbers correctly. For example, the following code 

```rust
let arr = Tensor::full(-1000_f64, (10,), &Device::Cpu)?;
let logsumexp = arr.log_sum_exp(0)?;
println!("{:?}", logsumexp);
```

prints 

```
Tensor[-inf; f64]
```

whereas on torch 2.3.0  

```python
>>> arr = torch.tensor([-1000]*10)
>>> torch.logsumexp(arr, 0)
tensor(-997.6974)
```

### Version info
candle-core = "0.6.0"
Tested on Arch Linux

LaurentMazare:
Closing based on the fix in #2367, feel free to reopen if it's still an issue.

PR #2367, opened 2024-07-30
Fix log_sum_exp to handle large positive/negative inputs

yunjhongwu:
Fixing #2356

For the updated unit test case, the current `log_sum_exp(-1)` yields `[[3.4076, 6.4076], [-inf, inf]]`.  With this fix, the result will be `[[3.4076, 6.4076], [-998.5924, 1001.4076]]`, which matches the output from pytorch.


LaurentMazare (approved):

LaurentMazare:
Nice, thanks!

PR #2368, opened 2024-07-30
Add/lstm state from tensors

singjc:
I added a method to retrieve an LSTM state based on a given tensor of the hidden and cell state vectors. This is useful when loading a pretrained PyTorch models weights, to initialize the state using the hidden and cell state vectors from the pretrained model.

I initially tried other ways of initializing an LSTM state when given a hidden and cell state vector using `assign_elem` on the `state.h()` method (i.e. `state_forward.h().assign_elem(&h0_forward);`, but that did not seem to work. Which is why I'm adding this method. If there is an actual existing way to do this without the added method, please let me know.

Best,

Justin

### DIFF
--- a/candle-nn/src/rnn.rs
+++ b/candle-nn/src/rnn.rs
@@ -199,6 +199,13 @@ impl RNN for LSTM {
         let states = states.iter().map(|s| s.h.clone()).collect::<Vec<_>>();
         Tensor::stack(&states, 1)
     }
+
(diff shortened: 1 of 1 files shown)


LaurentMazare:
I don't think we would want to assume that RNN states are made of two tensors. Instead we could make the state fields `pub` for `LSTMState` and `GRUState`, hopefully this would cover your use case?

PR #2379, opened 2024-08-01
Add get_ids to GradStore (Discussion #2377)

spaghetti-source:
Discussion in https://github.com/huggingface/candle/discussions/2377 

This PR adds `get_ids` function to `GradStore` that returns the list of `TensorId` in the store. It can be used to implement optimizers that only update relevant variables stored in the store. The typical use case is the word2vec model, which has many embeddings for the words but updates only a few in each iteration.


LaurentMazare:
Thanks!

PR #2384, opened 2024-08-01
update: LSTMState and GRUState fields to be public

singjc:
I added the `pub`  keyword to the fields in `LSTMState` and `GRUState` to make the states public, so that a state can be created using a predefined initial state tensor from a pretrained model.

See previous related closed [PR #2368 ](https://github.com/huggingface/candle/pull/2368).


LaurentMazare:
Thanks!

PR #2387, opened 2024-08-02
Add GGUF BF16 dtype support

EricLBuehler:
Currently, the `GgmlDType` only supports F16 and not BF16. This PR introduces support for the BF16 type. 

I would appreciate a check if this looks good! I have tested with success on my machine which has `avx` and `f16c`, and the CUDA tests also pass even though no changes were necessary.

I also noted that there will be a confusing situation in this case, though, if the tensor is part of a QMatMul. In this case (and for all other types not supported for quantized matmul in QStorage), we should perhaps dequantize and then perform the matmul using cublas? This modification could be made in `QStorage::fwd`, perhaps.


candle-core/src/quantized/cuda.rs
```diff
@@ -409,6 +409,7 @@ impl QCudaStorage {
         match self.dtype {
             GgmlDType::F32 => deq::<f32>(&buffer, block_len, &mut out)?,
             GgmlDType::F16 => deq::<half::f16>(&buffer, block_len, &mut out)?,
+            GgmlDType::BF16 => deq::<half::bf16>(&buffer, block_len, &mut out)?,
```
EricLBuehler, line 412: With #2424 this can be optimized!

PR #2393, opened 2024-08-03
Optimize gradient for silu a bit

MilkFather:
This pull request exploits some mathematical properties of the `silu` operation. By reusing the computed forward pass results as much as possible, the optimized code reduces the running time for a sliver of a millisecond.

Ideally, since the forward pass also calculates the sigmoid of the input, if we could cache that intermediate result somehow, we can avoid computing that again, further cutting down some time.

By the way, the type of `silu_grad` is `Result<Tensor, Error>`, yet the compiler still accepts the code. Perhaps we don't need so many `?` operators after all.

### DIFF
--- a/candle-core/src/backprop.rs
+++ b/candle-core/src/backprop.rs
@@ -623,9 +623,9 @@ impl Tensor {
                     }
                     Op::Unary(arg, UnaryOp::Silu) => {
                         let sum_grad = grads.or_insert(arg)?;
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Thanks. Re `?` there are some hacks that make it possible not too use them when using binary operators, it's still a bit unclear to me whether it's a good or a bad thing as it's a bit less explicit which bit can fail but it's also a lot less verbose, so the statu quo is to let whoever writes the code decides which style they prefer.

PR #2397, opened 2024-08-05
Add the MMDiT model of Stable Diffusion 3

Czxck001:
This follows [the implementation in ComfyUI](https://github.com/comfyanonymous/ComfyUI/blob/78e133d0415784924cd2674e2ee48f3eeca8a2aa/comfy/ldm/modules/diffusionmodules/mmdit.py#L1). 

Smoke test is passed with dumped input and output tensors from a running ComfyUI SD 3 workflow.

Some extra works need to be done to complete a running SD3 example with candle, including a slight modification of the two CLIP encoders used for SDXL. The comfy implemetation of SD3 takes penultimate layer of CLIP L/G as embedding, instead of the final layer output (see [sd3_clip.py](https://github.com/comfyanonymous/ComfyUI/blob/a178e25912b01abf436eba1cfaab316ba02d272d/comfy/text_encoders/sd3_clip.py#L45) and [sdxl_clip.py](https://github.com/comfyanonymous/ComfyUI/blob/a178e25912b01abf436eba1cfaab316ba02d272d/comfy/sdxl_clip.py#L7)). Also, SD3 VAE might need some effort to make/tweak.


candle-transformers/src/models/mmdit/projections.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    pub fn post_attention(&self, x: &Tensor) -> Result<Tensor> {
+        self.proj.forward(x)
+    }
+}
+
+fn split_qkv(qkv: &Tensor, head_dim: usize) -> Result<Qkv> {
+    let (batch_size, seq_len, _) = qkv.dims3()?;
+    let qkv = qkv.reshape((batch_size, seq_len, 3, (), head_dim))?;
+    let qkv = qkv.permute((2, 0, 1, 3, 4))?;
```
LaurentMazare, line 88: In this case I think you could use `get_on_dim` rather than `permute` + `get`.

candle-transformers/src/models/mmdit/embedding.rs
```diff
@@ -0,0 +1,195 @@
...
+            )?)?
+            .exp()?;
+
+        let args = t
+            .unsqueeze(1)?
+            .to_dtype(candle::DType::F32)?
+            .matmul(&freqs.unsqueeze(0)?)?;
+        let embedding = Tensor::cat(&[args.cos()?, args.sin()?], 1)?;
+
+        assert!(dim % 2 == 0, "Embedding dimension must be even");
```
LaurentMazare, line 159: Use `bail!` rather than a plain `assert!` when possible so that a proper error is returned.

candle-transformers/src/models/mmdit/projections.rs
Czxck001: Done. That's a great idea. I wasn't aware of that.

candle-transformers/src/models/mmdit/embedding.rs
Czxck001: Done. I noticed some places in the codebase uses `bail!` as a statement without the trailing semicolon (like [here](https://github.com/huggingface/candle/blob/500c9f288214cf817c8c52c7c9a6187cb279c563/candle-core/src/conv.rs#L229)). But in somewhere else the trailing semicolon is adopted (such as [here](https://github.com/huggingface/candle/blob/500c9f288214cf817c8c52c7c9a6187cb279c563/candle-core/src/quantized/mod.rs#L126)). It appears using `bail!` without trailing semicolon is more popular and I adopted that as the convention. Please advise if you prefer otherwise.

LaurentMazare (approved):

LaurentMazare:
Thanks!

candle-transformers/src/models/mmdit/embedding.rs
LaurentMazare: I wouldn't mind too much here if the codebase is not fully coherent, thanks for checking.

PR #2424, opened 2024-08-17
Avoid dtoh copy for dequantization of f16/f32

EricLBuehler:
Currently, we execute a dtoh copy when dequantizing f16/f32 on CUDA when this is not necessary. We can just add a simple cast kernel to ensure that we keep the data on the device.

This PR only updates CUDA dequantization, Metal dequantization executes on the CPU. Generall, dequantization performance shouldn't matter too much, as we should avoid it in a hot loop.


EricLBuehler:

Issue #2361, opened 2024-07-27
Tensor::arg_sort_last_dim on CUDA does not work with large vectors


EricLBuehler:
If the vector length is high, the error `CUDA_INVALID_VALUE` is returned:

```rust
use candle_core::{DType, Device, Tensor};

fn main() {
    let a = Tensor::zeros(
        32000,
        DType::F32,
        &Device::cuda_if_available(0).unwrap(),
    )
    .unwrap();
    dbg!(&a.arg_sort_last_dim(true));
}
```

```
[src/main.rs:10:5] &a.arg_sort_last_dim(true) = Err(
    Cuda(
        Cuda(
            DriverError(
                CUDA_ERROR_INVALID_VALUE,
                "invalid argument",
            ),
        ),
    ),
)
```

I think this is caused by the kernel launch params being directly controlled by the dimensions. So, if we use `u32::min` to determine the launch params, this should be fixed.

gabrielmbmb:
In addition to @EricLBuehler comment, I did some testing and I think (not a CUDA expert) the `shared_mem_bytes` needs also to be less or equal than 48KB which is the max allowed for dynamic shared memory, otherwise we also get the CUDA_ERROR_INVALID_VALUE. In `ggml` they have an assertion for this (max dynamic shared memory per block depends on architecture):

https://github.com/ggerganov/ggml/blob/9ad0906ffca1758f29270f59938d78740808ba93/src/ggml-cuda/argsort.cu#L76-L77

Not sure how to fix this because for large tensors we need more than 48KB.

gabrielmbmb:
Using `int` (4 bytes) as type for `dst_row` is inefficient/overkill as it requires reserving more shared memory bytes for storing the same indices. Changing `dst_row` to `uint8_t` which uses 1 byte and it still can represent up to 2e8 - 1 has allowed me to use `arg_sort_last_dim` method with bigger tensors, but sadly this is not enough for a llama 3.1 8b output tensor with shape (1, 128256).

I think in `ggml` they are not using this operation for the sampling part, otherwise they would have hit this error too:

https://github.com/ggerganov/ggml/blob/9ad0906ffca1758f29270f59938d78740808ba93/examples/common.cpp#L450-L456

I think the only solution that will enable using `arg_sort_last_dim` is to write a new cuda kernel?

EricLBuehler:
[...]

PR #2430, opened 2024-08-18
Add support for argsort with no shared memory usage

EricLBuehler:
As discussed in #2361, our current argsort implementation does not work on CUDA for large vectors because we use a bitonic sort implementation, which requires shared memory. For some n x m matrix (or really anything x m), shared memory scales with `m` in our bitonic sort implementation. In fact, it is calculated as `sizeof(u32) * m`. However, CUDA imposes a limit on shared memory size (depends on the architecture, but is clearly a limiting factor).

This PR sketches an argsort which runs only in global memory. However, I haven't found a way to implement something more parallel because of the limitation of no shared memory. As a result, we use a simple bubble sort, which gives horrible performance, especially when using argsort in sampling where `m` is large.

@gabrielmbmb, I was wondering if you could review and may know of a fasterway to do this? I considered implementing a merge sort so that each thread will merge sort one row, but that still gives $n \log{n}$ VS the bitonic sort's  ${\log}^{2}{n}$.

The PR can be tested with:
```
cargo test --features cuda --test tensor_tests -- asort_very_big
```
There is still some cleanup to do before merge (for example, I modified the settings.json for testing and some other things).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,14 @@
         "candle-pyo3"
     ],
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
+    "python.testing.pytestEnabled": true,
+    "rust-analyzer.cargo.features": [
+        "cuda"
+    ],
+    "files.associations": {
+        "array": "cpp",
+        "string": "cpp",
+        "string_view": "cpp",
+        "ranges": "cpp"
+    }
 }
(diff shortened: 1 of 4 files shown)


LaurentMazare:
I'm not sure we want to replace the implementation, the goal of the current bitonic implementation is to allow for things like mixture of experts where the number of elements to sort is small and we want to focus on speed (it's actually the same as in llama.cpp), we should certainly document the current limitation though.

EricLBuehler:
Yeah, we probably should not replace the bitonic implementation, this PR adds a separate kernel which will be invoked only when the shared memory would exceed some threshold.

But, given that this change causes it to be so slow, perhaps it would be better to just document/return an error if the shared memory allocation would be excessive for now?

EricLBuehler:
Closing to avoid excessive stagnant PRs and due to discussion.

PR #2432, opened 2024-08-18
Add the i32 dtype

EricLBuehler:
This PR adds `DType::I32`. Besides being a useful and more memory-efficient alternative to our I64 dtype, it is also commonly used in GPTQ and AWQ. If we were to implement loading those formats from safetensors (I have the code already for GPTQ, please let me know if that would be of interest!), they would be [upcast to I64](https://github.com/huggingface/candle/blob/736d8eb7521dd48e777827848f2b9ed8a7473571/candle-core/src/safetensors.rs#L207-L210), and take up twice the memory.


LaurentMazare:
The idea is to reduce the number of supported dtypes to keep the complexity low, we've added up i64 later than the others as we didn't have any signed int to start with but for now I don't think we want more.

EricLBuehler:
Sounds good! I'll close this.

Qubitium:
@LaurentMazare  int32 is required for efficient awq/gptq loading and inference. Why load in int64 if int32 is optimal? If no other code uses it than perhaps gptq/awq, is a reason to exclude it?

@EricLBuehler I am the maintainer for GPTQModel and would be very interested in helping as much as I can with gptq integration into candle. 

> (I have the code already for GPTQ, please let me know if that would be of interest!),

I would be very interested in helping to get this gptq code yours merged in a separate PR if possible. Have no background in rs but I can contribute in testing and validation.

PR #2436, opened 2024-08-21
Utilize the fused rmsnorm kernel for gemma/gemma2

EricLBuehler:
This PR utilizes the fused and therefore faster fused RmsNorm kernels for the Gemma and Gemma 2 models. We also expose an API for loading the weights, although perhaps renaming it to a `plus_one...` API would be best?


EricLBuehler:

PR #2439, opened 2024-08-22
onnx: workaround pow with negative base

shua:
rather than fully defining pow in the cpu backend (as in #2318), this implements a much smaller change which is sufficient to evaluate silero-vad onnx models. Specifically, checking if we can use `Tensor::powf` which is defined for negative base.

### DIFF
--- a/candle-onnx/src/eval.rs
+++ b/candle-onnx/src/eval.rs
@@ -352,8 +352,16 @@ fn simple_eval_(
             "Pow" => {
                 let input0 = get(&node.input[0])?;
                 let input1 = get(&node.input[1])?;
-                let output = input0.broadcast_pow(input1)?;
-                values.insert(node.output[0].clone(), output);
+                // HACK: current implementation of (broadcast_)pow cannot handle negative base;
+                // this is rarely an issue, but blocked support of silero-vad which uses x^2.
+                // Rather than fix pow for all negative base, we choose to handle just x^2.
+                if let Ok(2.0) = (|| input1.flatten_all()?.to_scalar::<f32>())() {
+                    let output = input0.mul(input0)?;
+                    values.insert(node.output[0].clone(), output);
+                } else {
+                    let output = input0.broadcast_pow(input1)?;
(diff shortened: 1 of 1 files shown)


candle-onnx/src/eval.rs
```diff
@@ -352,8 +352,16 @@ fn simple_eval_(
             "Pow" => {
                 let input0 = get(&node.input[0])?;
                 let input1 = get(&node.input[1])?;
-                let output = input0.broadcast_pow(input1)?;
-                values.insert(node.output[0].clone(), output);
+                // HACK: current implementation of (broadcast_)pow cannot handle negative base;
+                // this is rarely an issue, but blocked support of silero-vad which uses x^2.
+                // Rather than fix pow for all negative base, we choose to handle just x^2.
+                if let Ok(2.0) = (|| input1.flatten_all()?.to_scalar::<f32>())() {
```
LaurentMazare, line 358: Rather than handling the case 2. specifically (which indeed seems very hacky) why not handle any "constant" power with powf, this would feel less like a hack and would handle a lot more cases, e.g. square root etc.

shua: That would require looping through all the tensor values (sort of inline implementation of broadcast_pow). I guess still kind of a hack but if that's preferable I can try that.

shua: ah, it also requires switching on the `DType` of the tensor, as `powf`, as `i32`/`u32` don't have `powf(self, exp: f32)` defined, but rather `pow(self, pow: i32)`.

LaurentMazare: Yep exactly, it would use `Tensor::powf` and so work for exponents that are not 2 (and not feel anymore like a hack but rather an optimization for this case).

shua: I still left a note in case anyone in the future looks at this thinking "why not just use `broadcast_pow` for both cases?"

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2440, opened 2024-08-22
onnx: support negative index in Gather

shua:
index_select does not support negative indexing, but this change adds just enough workarounds in onnx to allow evaluating silero-vad models (which make use of negative indices).


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2452, opened 2024-08-28
FastViT fixes.

janimo:
* correct optional SE layer dimensions.
 * head_dim instead of num_heads is 32.
 * update test example output.


LaurentMazare (approved):

LaurentMazare:
Thanks!

Issue #2351, opened 2024-07-23
Incorrect Loading of LSTM Weights from PyTorch Model


singjc:
Hello,

When loading weights from a pretrained PyTorch model using the `VarBuilder::get` method, I've noticed that most of the loaded weights for different layers match correctly, except for a 2 layer birectional LSTM layer (`rt_encoder.hidden_nn.rnn`).

### Description

I'm attempting to load a PyTorch model's weights (using the `from_pth` method) into a Candle implementation. While most layers' weights are loaded correctly, the LSTM layer weights show significant discrepancies between the PyTorch and Candle versions.

### Observed Behavior

1. Non-LSTM layer weights (e.g., CNN layers, attention layers) load correctly and match the PyTorch values.
2. The first LSTM layer weight (`weight_ih_l0`) loads correctly.
3. All other LSTM weights (e.g., `weight_hh_l0`, `weight_ih_l0_reverse`, etc.) have different values in Candle compared to PyTorch.

I have double checked to make sure both the PyTorch and Candle comparisons are using the exact same pretrained PyTorch (.pth) file. I am not entirely sure why this is happening, any help would be appreciated! If there is any other information required to help debug, please let me know.

Best,

Justin

### Code Snippet

Here's how I'm loading the LSTM weights with Candle for comparing the weight loading with PyTorch:

```rust
let weight_ih_l0 = vb.get((512, 140), "rt_encoder.hidden_nn.rnn.weight_ih_l0")?;
let weight_hh_l0 = vb.get((512, 128), "rt_encoder.hidden_nn.rnn.weight_hh_l0")?;
let weight_ih_l0_reverse = vb.get((512, 140), "rt_encoder.hidden_nn.rnn.weight_ih_l0_reverse")?;
let weight_hh_l0_reverse = vb.get((512, 128), "rt_encoder.hidden_nn.rnn.weight_hh_l0_reverse")?;
```

where `vb` is a passed value of `var_store = VarBuilder::from_pth(model_path,candle_core::DType::F32,&device)?;`

This is a brief overview of the main model's implementation in Candle

```rust
impl<'a> RTModel<'a> {
    pub fn new<P: AsRef<Path>>(model_path: P, constants_path: P, device: Device) -> Result<Self> {

    var_store = VarBuilder::from_pth(model_path,candle_core::DType::F32,&device)?;`

     // Other layers ...

      let cnn_long = Conv1d::new(
[...]

PR #2455, opened 2024-08-29
Add/lstm direction

singjc:
Hi,

I added a direction field to the `LSTMConfig` using a enum for forward and reverse direction. I added this so that it's possible to retrieve the weights from reversed layers of a pretrained pytorch model. This also addresses issue #2351.


candle-nn/src/rnn.rs
```diff
@@ -70,6 +70,12 @@ impl LSTMState {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub enum Direction {
+    Forward = 1,
```
LaurentMazare, line 75: What is the need for associating integer values to this enum?

candle-nn/src/rnn.rs
```diff
@@ -128,25 +137,29 @@ pub fn lstm(
     vb: crate::VarBuilder,
 ) -> Result<LSTM> {
     let layer_idx = config.layer_idx;
+    let direction_str = match config.direction {
+        Direction::Forward => String::from(""),
```
LaurentMazare, line 141: Why use `String` rather than `&'static str` here?

candle-nn/src/rnn.rs
singjc: There actually isn't a need to assign an integer value. I removed it now

candle-nn/src/rnn.rs
singjc: Thanks for pointing that out. I'm still new to Rust, so wasn't quite sure about using `String` vs just having `&'static str'. But I see its probably more efficient to use `&'static str`. I updated the code per your suggestion.

LaurentMazare (approved):

LaurentMazare:
Thanks!

LaurentMazare:
Btw if you already have some models that use a bidirectional LSTM, would be great to have them as potential examples in candle-examples.

PR #2464, opened 2024-09-10
Add more capability to slice_assign

EricLBuehler:
Currently, slice_assign takes only homogeneous ranges, which is not as ergonomic as the IndexOp trait. This PR aims to make slice_assign a bit easier to work with heterogeneous range types, and specifically, uses trait objects.


EricLBuehler:

PR #2467, opened 2024-09-10
Improve `candle_core::Error` to make it more ergonomic

EricLBuehler:
Currently, our Error type is already useful but could benefit from some of the QoL features which the anyhow crate has.

This PR adds the following:
- Make `Error::msg` more in line with `anyhow::Error::msg`. Although the type bound changes, this *is not* a breaking change because `std::error::Error: Display` already.
- Add the `Context` trait. Inspired by anyhow's Context trait which allows for `.with_context` and `.context` on Options and Results. Very useful!

### DIFF
--- a/candle-core/src/error.rs
+++ b/candle-core/src/error.rs
@@ -1,4 +1,7 @@
-use std::fmt::{Debug, Display};
+use std::{
+    convert::Infallible,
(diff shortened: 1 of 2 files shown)


zackangelo (commented):
LGTM

PR #2481, opened 2024-09-17
Clamping t5 hidden states to avoid F16 NaNs

### LINKED ISSUE #2480: FLUX T5 XXL model produces NaN when on CUDA and using F16
Perhaps we can use clamping, as per:
> https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5/modeling_t5.py#L748-L755

Using BF16 works on CUDA.

EricLBuehler:
Fixes #2480.


abrowne2:
@EricLBuehler Was using Flux, & applied this patch to my fork as I was getting black image results. It fixed it as I was using a quantized t5 model @ F16; I assume I was getting silent NaN’s. It may fix others.

PR #2536, opened 2024-10-03
[AArch64] Quantized MatMul performance improvement

bgergely0:
This change implements an new matmul based on Armv8.6 i8mm instructions. As the change requires a nightly compiler, using the more performant version is optional, under the new `arm-nightly-feat` feature flag.

I have also added the Armv8.4 dotprod instructions under this flag. 

Performance improvement:
- +10% with only dotprod enabled (measured on LLaMa2)
- +40-60% for i8mm based matmul (measured on quantized Whisper, different cores produced different results)

### DIFF
--- a/candle-core/Cargo.toml
+++ b/candle-core/Cargo.toml
@@ -19,6 +19,7 @@ cudarc = { workspace = true, optional = true }
 gemm = { workspace = true }
 half = { workspace = true }
 intel-mkl-src = { workspace = true, optional = true }
+itertools = "0.12.1"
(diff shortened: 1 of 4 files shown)


bgergely0:
Hi @LaurentMazare!

What's the project's stance on using features that require nightly?

I would appreciate some feedback on this change. 
I do understand that it is a big diff, but the matmul unittests show that the new neon i8mm functions are correct.
I put the nightly code under a feature flag, to not interfere with the default toolchain/default environments. What is your opinion on this?

Thanks,
Gergely

PR #2542, opened 2024-10-05
Implements bidirectional RNN

kigichang:
1. Fix incorrect result dimension of GRU states_to_tensor. It should be (Batch_size, Seq_Len, Output_Size) instead of (Batch_size, Seq_Len * Output_Size).
2. Implements bidirectional LSTM and GRU.


Baiyuetribe:
Thank you for your work, I tested it correctly

PR #2555, opened 2024-10-10
feat: intergrate chinese clip  and add example

SethWen:
This PR translate [chinese clip of transformers](https://github.com/huggingface/transformers/blob/5af7d41e49bbfc8319f462eb45253dcb3863dfb7/src/transformers/models/chinese_clip/modeling_chinese_clip.py) to candle and add example of chinese clip.

Expected to merge the code.


LaurentMazare:
Thanks!

PR #2556, opened 2024-10-11
Allow marian configs to deserialize from json.

Mikarific:
Currently, there is no way to use a Marian model that isn't `opus_mt_tc_big_fr_en` or `opus_mt_tc_big_fr_en`, because the only way to get a Marian config is by calling the functions for those two models configs which are hard coded into the library. This change simply allows you to use serde to deserialize a Marian model's `config.json`, just like how other model types in candle already do (see the bert.rs file for example).


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2558, opened 2024-10-13
Add Stable Diffusion 3 Example

Czxck001:
It appears the focus of the community has been largely shifted to Flux.dev1. So the main purpose of this PR is to demonstrate the capability of Candle and serves a smoke-test to the MMDiT (#2397).

As such, I intend to minimize the intrusive change to the existing stable-diffusion codebase, such as using renaming function to adapt the VAE var-builder to the official safetensor weights of SD3 VAE. Still, there are some changes I have to make to `candle_nn::stable_diffusion` to support the CLIP and VAE of SD3, including:
- Add a `forward_until_encoder_layer` to `ClipTextTransformer`. The Comfy implementation for SD3 uses the penultimate hidden layer of CLIP-l and CLIP-g instead of the final layer  (see [sd3_clip.py](https://github.com/comfyanonymous/ComfyUI/blob/a178e25912b01abf436eba1cfaab316ba02d272d/comfy/text_encoders/sd3_clip.py#L45) and [sdxl_clip.py](https://github.com/comfyanonymous/ComfyUI/blob/a178e25912b01abf436eba1cfaab316ba02d272d/comfy/sdxl_clip.py#L7)). This practice, although not mentioned in [the SD3 tech report](https://arxiv.org/pdf/2403.03206), *is* referred and specified in Chapter 2.1 of [the SDXL tech report](https://arxiv.org/pdf/2307.01952).
- Add the `use_quant_conv` and `use_post_quant_conv` options to the `AutoEncoderKL`, as SD3's VAE does not have those layers. These changes might be considered unspecific to SD3, as `diffusers` has [these options supported](https://huggingface.co/docs/diffusers/v0.30.0/en/api/models/autoencoderkl#diffusers.AutoencoderKL).
- Uses `get_qkv_linear` to load the attention block in `candle_nn::stable-diffusion::attention`, as some weight of linear layer of VAE in official SD3 Medium safetensors follow the dimension convention of `(channel, channel, 1, 1)` instead of the regular `(channel, channel)` that is natually supported by `nn::linear` constructor.

These changes allows reusing existing CLIP and VAE implementations, but inevitably add complexity to existing codebase. @LaurentMazare Let me know if these intrusive changes are justified. We may consider alternatives like re-implementing VAE and CLIP from scratch.

On top of these changes, I added the support to flash-attention for MMDiT based on whether the feature `flash-attn` is enabled. Also done a simple performance benchmark on GPUs like 3090 Ti and 4090.

A side note is the T5 implementation on current main branch hasn't supported for FP16. I attempted to insert simple clampings within the FP16 dynamic range but it didn't work well on my GPUs. Looks like I need to wait for a more sophiscated implementation such as #2481. So for now, I use two different VarBuilders, one maps weights in safetensor into FP32 specifically for T5, the other for the rest compoents.


LaurentMazare (commented):
Looks pretty good, thanks for adding this. Would you mind replacing the sample image with a jpg version? (the png version you attached takes almost 1MB which is not great for the repo size)

Czxck001:
@LaurentMazare Thank you for reminding me this. The sample image has been replaced by a JPG. The original PNG should be excluded from Git objects after squash-merging.

candle-examples/examples/stable-diffusion-3/vae.rs
```diff
@@ -0,0 +1,95 @@
...
+            "upsamplers" => {
+                result.push("upsample");
+                i += 1; // Skip the 0 after upsamplers.
+            }
+            part => result.push(part),
+        }
+        i += 1;
+    }
+    let outcome = result.join(".");
+    outcome
```
LaurentMazare, line 94: nit:
```suggestion
    result.join(".")
```

candle-transformers/src/models/mmdit/blocks.rs
```diff
@@ -250,6 +250,7 @@ impl ContextQkvOnlyJointBlock {
 
 // A QKV-attention that is compatible with the interface of candle_flash_attn::flash_attn
 // Flash attention regards q, k, v dimensions as (batch_size, seqlen, nheads, headdim)
+#[cfg(not(feature = "flash-attn"))]
```
LaurentMazare, line 253: Would it be possible to keep flash attention optional in the same way it was before?

candle-examples/Cargo.toml
```diff
@@ -122,3 +122,7 @@ required-features = ["onnx"]
 [[example]]
 name = "colpali"
 required-features = ["pdf2image"]
+
```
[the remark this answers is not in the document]
Czxck001: Removed. Thanks.

candle-examples/examples/stable-diffusion-3/vae.rs
Czxck001: Done.

candle-examples/examples/stable-diffusion-3/main.rs
```diff
@@ -0,0 +1,186 @@
...
+    };
+
+    let x = {
+        let mmdit = MMDiT::new(
+            &MMDiTConfig::sd3_medium(),
+            vb_fp16.pp("model.diffusion_model"),
+        )?;
+
+        let num_inference_steps = num_inference_steps.unwrap_or(28);
+        let height = height.unwrap_or(1024);
```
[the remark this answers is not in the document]
Czxck001: Totally agree. I was following the handling in flux and stable-diffusion example. Thought it was a convention or so:

https://github.com/huggingface/candle/blob/0d96ec31e8be03f844ed0aed636d6217dee9c7bc/candle-examples/examples/flux/main.rs#L34-L40

I have moved the default value to Args and they can be displayed in command line documentation now.

candle-transformers/src/models/mmdit/blocks.rs
Czxck001: Sure. I have passed down a `use_flash_attn` flag to the attention block, following the convention in UNet and other models elsewhere.

LaurentMazare (approved):

LaurentMazare:
Merged, thanks a lot!

super-fun-surf:
its so great to have this. thanks for the work. stable diffusion 3.5 Large is out now and it looks amazing. as its a full base model that we can train on. looks like its not working with candle yet though.

Czxck001:
@super-fun-surf working on it. SD3.5 changes the MMDiT archetecture a little bit (namely MMDiT-X). Needs to get that done first before implementing a working example.

LaurentMazare:
@Czxck001 I actually already started adding 3.5 and just opened a PR for it #2578 , it seems to be working well with at least the turbo model so far and would be great if you can give a look at the PR.

Czxck001:
@LaurentMazare That's awesome! I'll take a look.

Issue #2561, opened 2024-10-14
binary_cross_entropy_with_logit numerically unstable for high logits


BeneSim:
Hi,

I stumbled across an issue where my losses all went to `NaN` using `binary_cross_entropy_with_logit`. I quickly found out that this happens for high logit values. When sigmoid gets applied to them the result will be a 1.0, applying an affine transformation of  
`-1 * y + 1` then results in `0.0` and finally taking the log of `0.0` yields `NaN`.

https://github.com/huggingface/candle/blob/3d1dc06cdb44e2e012559aadd8da7342da9c2ed5/candle-nn/src/loss.rs#L66

I looked up in the pytorch source how they did it and assembled a more or less identical way in candle (see PR).

I haven't actually looked up the mathematical definition of the operation or dug deep into examples to see if the proposed implementation is flawed as well, so take the PR with a grain of salt ;)

PR #2562, opened 2024-10-14
Make Binary cross entropy with logit numerically stable for high logit values

BeneSim:
See #2561 for details. I added detailed commentary to the commits.


BeneSim:
Just a quick update, I also encountered the case where the logit was so small that the sigmoid returned 0. So I guess we need a better way. Maybe the method tensorflow uses might make sense, they basically calculate the log_sigmoid by using softplus:

See [log_sigmoid](https://www.tensorflow.org/api_docs/python/tf/math/log_sigmoid) and [softplus](https://www.tensorflow.org/api_docs/python/tf/math/softplus). This could be implemented like
```rust
let log_sigmoid_input = (inp.neg()?.exp()? + 1.)?.log()?.neg()?;
```

ivarflakstad:
Noting down the pytorch [ref](https://github.com/pytorch/pytorch/blob/b5784b9e2d6cd0a7f4764e89502179245c9955b1/aten/src/ATen/native/Loss.cpp#L349) for later 👍

PR #2563, opened 2024-10-14
onnx: ReduceMin/Max Ops

AnubhabB:
Adds support for [ReduceMin](https://onnx.ai/onnx/operators/onnx__ReduceMin.html) and [ReduceMax](https://onnx.ai/onnx/operators/onnx__ReduceMax.html) Ops. Found these in use in a Detectron2 exported model.

**Implementation**
~Supports Ops version 18+~ Backward compatible as far as I can tell

**Open Questions**
- Backward compatibility: Op version `13 -> 18` has moved the `axes` from `attributes` to `input`. Though it's trivial to handle this with `get_attr_opt::<&[i64]>(node, "axes")` fallback, not sure if that's the right way of going about this!
- The Op definition allows for `empty set` as input and prescribes `-INF/ +INF` or `Datatype::Min/ Datatype::Max` for the ops. [Reference](https://onnx.ai/onnx/operators/onnx__ReduceMin.html#summary). I'm not sure if the return is supposed to be a `scalar`?

@LaurentMazare, any thoughts around these open questions?


LaurentMazare:
> * Backward compatibility: Op version `13 -> 18` has moved the `axes` from `attributes` to `input`. Though it's trivial to handle this with `get_attr_opt::<&[i64]>(node, "axes")` fallback, not sure if that's the right way of going about this!

Falling back to taking the axis from the input if they are not present in the attributes sounds pretty reasonable to me.

> * The Op definition allows for `empty set` as input and prescribes `-INF/ +INF` or `Datatype::Min/ Datatype::Max` for the ops. [Reference](https://onnx.ai/onnx/operators/onnx__ReduceMin.html#summary). I'm not sure if the return is supposed to be a `scalar`?

I guess we can just error out in this case for the moment, and if anyone runs into this issue then we can have a look at what the failing onnx model actually expects.

AnubhabB:
Thanks @LaurentMazare.

I've amended the PR with the following:
1. Fetching `axes` defaults to the current standard (via `input`) and falls back to `attribute` for backward compatibility.
2. Errors out on empty Tensors
3. Tests for backward compatibility and empty set error.

Guess this is good to go!

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2568, opened 2024-10-17
ONNX: Adds Ops GatherElements & Xor

### LINKED ISSUE #2465: ONNX: unsupported GatherElements ops
Currently `GatherElements` operator is not supported in candle-onnx. This would be convenient since a number of converted models using [HummingBird ML](https://github.com/microsoft/hummingbird) use this operator (including Tree Classifiers).

`GatherElements` slightly differs from `Gather` which is already supported. `GatherElements` is already supported by burn. Defined here: https://onnx.ai/onnx/operators/onnx__GatherElements.html

AnubhabB:
This PR adds support for the following ONNX Ops:
- [GatherElements](https://onnx.ai/onnx/operators/onnx__GatherElements.html) - closes #2465 
- [Xor](https://onnx.ai/onnx/operators/onnx__Xor.html)

@LaurentMazare please take a look at the `Xor` implementation, Candle is currently missing `Tensor::logical_<and | or | not | xor>` ops?

Unsure if my current workaround is reasonable.


LaurentMazare:
The xor implementation looks reasonable to me in the current setup. I would be curious if you have any practical model that require this op?

AnubhabB:
Yup, [this one](https://huggingface.co/unstructuredio/detectron2_mask_rcnn_X_101_32x8d_FPN_3x/tree/main) - this is a Document layout analysis model I crossed path with by [Unstructured.io](https://github.com/Unstructured-IO), originally I had intended to create a quick example for `candle-onnx` but refrained because I can't find the license information associated with the model.

AnubhabB:
In fact, these are the ops used by this model:

```
["Add", "And", "Cast", "Clip", "Concat", "Constant", "ConstantOfShape", "Conv", "ConvTranspose", "Div", "Equal", "Exp", "Expand", "Flatten", "Floor", "Gather", "GatherND", "Gemm", "Greater", "If", "Less", "Log", "MaxPool", "Mod", "Mul", "NonZero", "Not", "Pad", "Range", "ReduceMax", "ReduceProd", "Relu", "Reshape", "Resize", "RoiAlign", "ScatterND", "Shape", "Sigmoid", "Slice", "Softmax", "Split", "Sqrt", "Squeeze", "Sub", "TopK", "Transpose", "Unsqueeze", "Where", "Xor"]
```

I'll try and figure out if these ops are common across some of the more popular models in Detectron2 Model Zoo. If they are common I'll try and bridge the gaps with Candle implementation over time, otherwise I think I'm better off with something like `Sonos/tract` or `ort` based onnx runtimes for my current project!

PR #2576, opened 2024-10-26
Fix the repo name for llama 3.1.

LaurentMazare:


julien-c (approved):
yep, Meta renamed some of their recent repos recently

PR #2579, opened 2024-10-27
Support for UG kernels.

LaurentMazare:
This adds some preliminary support for kernels implemented using UG, this will allow for easy fusion of kernels. Only cuda is supported in this PR, the metal backend hasn't been activated yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,8 @@ tokenizers = { version = "0.19.1", default-features = false }
 tracing = "0.1.37"
 tracing-chrome = "0.7.1"
 tracing-subscriber = "0.3.7"
+ug = "0.0.2"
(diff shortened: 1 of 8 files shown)


akashicMarga:
@LaurentMazare I just came across ug today and that led me to this pr in Candle. it would be great if you could share some details about ug. From ug readme we can see it's similar to tinygrad, so this will enable dynamic graph optimisations for candle as I know candle works in eager mode.

LaurentMazare:
It's pretty early days at the moment but the goal is indeed to provide a backend to candle that would allow for improved fusion. ug is somewhere between triton and tinygrad and to start with candle will use it more in a triton way to make it easier to write new kernels that work on all of cpu/metal/cuda.

PR #2584, opened 2024-10-29
Add some fast Metal MLX SDPA kernels 

EricLBuehler:
This PR adds some MLX SDPA kernels on Metal.

I can observe about a 26% performance improvement with Llama 3.1 8b @ q4k and @ q8_0 when testing through mistral.rs on my Candle fork. I updated the quantized_llama.rs file here to use the new function.

This PR adds a function `candle_nn::ops::sdpa`. The MLX attention kernels don't support masking yet, so the performance gains are only for decoding on Metal. Once/if they do, I'll update them - otherwise we can explore using Flash Attention kernels for Metal from llama.cpp.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,4 +8,5 @@
     ],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
+    "rust-analyzer.cargo.features": ["metal"]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,5 +8,4 @@
     ],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
-    "rust-analyzer.cargo.features": ["metal"]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,5 @@
         "candle-pyo3"
     ],
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true,
+    "python.testing.pytestEnabled": true
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,6 @@
         "candle-pyo3"
     ],
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
+    "python.testing.pytestEnabled": true,
+    "rust-analyzer.cargo.features": ["metal"]
 }
--- a/candle-metal-kernels/src/lib.rs
+++ b/candle-metal-kernels/src/lib.rs
@@ -1860,6 +1860,7 @@ pub fn call_sdpa_vector(
     k_stride: &[usize],
     k_buffer: &Buffer,
     v_offset: usize,
+    v_stride: &[usize],
     v_buffer: &Buffer,
     output: &Buffer,
     alpha: f32,
@@ -1871,7 +1872,8 @@ pub fn call_sdpa_vector(
     let gqa_factor = (q_shape[1] / k_shape[1]) as i32;
     let n = k_shape[2] as i32;
     let b = (q_shape[0] * q_shape[1]) as i32;
-    let stride = k_stride[1];
+    let kstride = k_stride[1];
+    let vstride = v_stride[1];
 
     let name = match (bk, itype) {
         (32, SdpaDType::F16) => "sdpa_vector_float16_t_32",
@@ -1921,7 +1923,8 @@ pub fn call_sdpa_vector(
             output,
             gqa_factor,
             n,
-            stride,
+            kstride,
+            vstride,
             alpha,
             softcapping
         )
--- a/candle-metal-kernels/src/lib.rs
+++ b/candle-metal-kernels/src/lib.rs
@@ -25,6 +25,7 @@ const REDUCE: &str = include_str!("reduce.metal");
 const SORT: &str = include_str!("sort.metal");
 const TERNARY: &str = include_str!("ternary.metal");
 const UNARY: &str = include_str!("unary.metal");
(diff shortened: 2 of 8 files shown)


.vscode/settings.json
```diff
@@ -7,5 +7,6 @@
         "candle-pyo3"
     ],
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
+    "python.testing.pytestEnabled": true,
+    "rust-analyzer.cargo.features": ["metal"]
```
LaurentMazare, line 11: Wouldn't that break the editor for non metal users?

candle-metal-kernels/src/lib.rs
```diff
@@ -1627,6 +1641,331 @@ pub fn call_gemm(
...
+
+    let pipeline = kernels.load_pipeline(device, Source::Sdpa, &name)?;
+    let encoder = ep.encoder();
+    let encoder: &ComputeCommandEncoderRef = encoder.as_ref();
+    encoder.set_compute_pipeline_state(&pipeline);
+
+    // q = (bs, qhead, seq, hidden)
+    // k/v = (bs, kv_head, seq, hidden)
+
+    let _hidden = q_shape[q_shape.len() - 1];
```
LaurentMazare, line 1747: If variables such as `_hidden` are not planned to be used, maybe remove them?

candle-metal-kernels/src/lib.rs
```diff
@@ -1627,6 +1641,331 @@ pub fn call_gemm(
...
+    encoder.set_bytes(
+        6,
+        (std::mem::size_of::<i32>() * batch_shape.len()) as u64,
+        batch_shape.as_ptr() as *const i32 as *const c_void,
+    );
+    encoder.set_bytes(
+        7,
+        (std::mem::size_of::<usize>() * batch_strides.len()) as u64,
+        batch_strides.as_ptr() as *const c_void,
+    );
```
LaurentMazare, line 1831: Couldn't the `EncoderParam` trait or better the `set_params!` macro be used to simplify this?

candle-metal-kernels/src/lib.rs
```diff
@@ -1627,6 +1641,331 @@ pub fn call_gemm(
...
+    encoder.set_bytes(
+        7,
+        std::mem::size_of::<f32>() as u64,
+        &alpha as *const f32 as *const c_void,
+    );
+    encoder.set_bytes(
+        8,
+        std::mem::size_of::<f32>() as u64,
+        &softcapping as *const f32 as *const c_void,
+    );
```
LaurentMazare, line 1949: Same comment as above, `set_params!` should be an easy win here.

candle-nn/src/ops.rs
```diff
@@ -947,3 +947,193 @@ impl Module for Identity {
...
+        if v_l.dims()[v_l.dims().len() - 3] != k_l.dims()[k_l.dims().len() - 3] {
+            candle::bail!("`k` and `v` head dims must match");
+        }
+
+        // n_heads % n_kv_heads == 0; n_heads >= 1, n_kv_heads >= 1.
+        if q_l.dims()[q_l.dims().len() - 3] % k_l.dims()[k_l.dims().len() - 3] != 0 {
+            candle::bail!("query `n_heads` must be a multiple of `n_kv_heads`");
+        }
+
+        let k_head = k_l.dims()[k_l.dims().len() - 1];
```
LaurentMazare, line 1009: `k_l.dim(D::Minus1)?` would be simpler and make for a better error message than a panic (the same applies to a bunch of places in this function)

candle-nn/src/ops.rs
```diff
@@ -947,3 +947,193 @@ impl Module for Identity {
...
+
+#[allow(dead_code)]
+struct Sdpa {
+    scale: f32,
+    softcapping: f32,
+}
+
+impl candle::CustomOp3 for Sdpa {
+    fn name(&self) -> &'static str {
+        "sdpa"
```
LaurentMazare, line 959: Let's call this metal-sdpa instead.

EricLBuehler: Done!

candle-metal-kernels/src/lib.rs
EricLBuehler: Thanks, used that. Should be much cleaner now!

candle-metal-kernels/src/lib.rs
EricLBuehler: Same, used it again here!

.vscode/settings.json
LaurentMazare: I'm still seeing the line that enables metal. Maybe I'm missing something?

candle-nn/src/ops.rs
LaurentMazare: You can get the shape which should have the `dim` methods.

.vscode/settings.json
EricLBuehler: Fixed, sorry!

candle-nn/src/ops.rs
EricLBuehler: I think the Shape also doesn't have the dim methods?

https://github.com/search?q=repo%3Ahuggingface%2Fcandle+%22fn+dim%22&type=code

EricLBuehler:
Latest commit fixes a bug reported in mlx, see ml-explore/mlx#1558.

LaurentMazare (approved):

candle-nn/src/ops.rs
LaurentMazare: Ah right, I've added it quickly as part of this PR, at least it should give better error messages rather than just out of bounds.

EricLBuehler:
Thank you!

PR #2589, opened 2024-10-30
Lazy upcasting for t5.

LaurentMazare:


super-fun-surf:
wow. way cleaner. nice work.

Issue #2598, opened 2024-11-06
Cannot compile candle-transformers


sneilan:
Using the latest commit (https://github.com/huggingface/candle/pull/2584) I get the following when running

```shell
cargo run --example bert --release -- --prompt "Here is a test sentence"
```

```
   Compiling candle-transformers v0.7.2 (/Users/sneilan/gp/scratch/candle/candle-transformers)
error[E0716]: temporary value dropped while borrowed
   --> candle-transformers/src/models/chinese_clip/text_model.rs:157:22
    |
155 |         let token_type_ids = match token_type_ids {
    |             -------------- borrow later stored here
156 |             Some(token_type_ids) => token_type_ids,
157 |             None => &self.token_type_ids.i((.., 0..seq_length))?,
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^-
    |                      |                                         |
    |                      |                                         temporary value is freed at the end of this statement
    |                      creates a temporary value which is freed while still in use
    |
help: consider using a `let` binding to create a longer lived value
    |
155 ~         let binding = self.token_type_ids.i((.., 0..seq_length))?;
156 ~         let token_type_ids = match token_type_ids {
157 |             Some(token_type_ids) => token_type_ids,
158 ~             None => &binding,
    |

error[E0716]: temporary value dropped while borrowed
   --> candle-transformers/src/models/stable_diffusion/vae.rs:399:39
    |
397 |         let xs = match &self.post_quant_conv {
    |             -- borrow later stored here
398 |             None => xs,
399 |             Some(post_quant_conv) => &post_quant_conv.forward(xs)?,
    |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^-
    |                                       |                          |
    |                                       |                          temporary value is freed at the end of this statement
    |                                       creates a temporary value which is freed while still in use
    |
help: consider using a `let` binding to create a longer lived value
    |
397 ~         let binding = post_quant_conv.forward(xs)?;
398 ~         let xs = match &self.post_quant_conv {
399 |             None => xs,
400 ~             Some(post_quant_conv) => &binding,
    |

For more information about this error, try `rustc --explain E0716`.
error: could not compile `candle-transformers` (lib) due to 2 previous errors
```

LaurentMazare:
Which version of the rust toolchain are you using? It might not be recent enough so if it's not 1.81 or above I would suggest updating and seeing if it helps.

[...]

PR #2625, opened 2024-11-18
Fix example build for bigcode

synapse-overload:
- Fix temporary values returned from index and forward


LaurentMazare:
This should be valid with recent version of rustc, could you try updating past 1.81 if you're not using it already? See https://github.com/huggingface/candle/issues/2598

synapse-overload:
You're right, thanks. My compiler ver. was the default in Alma Linux on WSL2, which is 1.75, so old.

PR #2639, opened 2024-11-26
Provide a method to allow PTH files with state maps to be loaded.

zachcp:
- discussed in a Discord Thread
- added a method to allow loading of pth files that have a state_map.

### DIFF
--- a/candle-nn/src/var_builder.rs
+++ b/candle-nn/src/var_builder.rs
@@ -544,7 +544,16 @@ impl<'a> VarBuilder<'a> {
         let pth = candle::pickle::PthTensors::new(p, None)?;
         Ok(Self::from_backend(Box::new(pth), dtype, dev.clone()))
     }
-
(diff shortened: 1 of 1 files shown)


candle-nn/src/var_builder.rs
```diff
@@ -544,7 +544,17 @@ impl<'a> VarBuilder<'a> {
         let pth = candle::pickle::PthTensors::new(p, None)?;
         Ok(Self::from_backend(Box::new(pth), dtype, dev.clone()))
     }
-
+    /// Initializes a `VarBuilder` that retrieves tensors stored in a pytorch pth file.
+    /// similar to [`from_pth`] but requires a `state_key`.
+    pub fn from_pth_with_state<P: AsRef<std::path::Path>>(
+        p: P,
+        dtype: DType,
+        state_key: String,
```
LaurentMazare, line 552: Shouldn't this be a `&str` rather than a `String`?

LaurentMazare (approved):

PR #2645, opened 2024-11-27
Fix for whisper-microphone example failure if audio isn't chunk aligned

anelson:
At least on my macOS Sequoia system (MBP 14" 2021, M1 Pro), when I run the `whisper-microphone` example after it has gathered 10 seconds of audio, it fails before the transcription:

```
Error: Insufficient buffer size 384 for input channel 0, expected 1024
```

At least for the audio device I'm using (Airpods Pro Max), there is no guarantee that each audio buffer is a multiple of 1024 samples.  Thus at the end of the 10 seconds, `buffered_pcm` can have some samples at the end that do not form a complete 1024 sample chunk.

This fixes that by tracking when there is a partial chunk at the end of the buffer, and leaving it in `buffered_pcm` to be processed on the next loop iteration.

Note that, in the interest of keeping this PR as small as possible, I didn't make any other changes to this example.  That said, I think a good enhancement would be to introduce a `const` for the hard-coded 1024 sample chunk size.

### DIFF
--- a/candle-examples/examples/whisper-microphone/main.rs
+++ b/candle-examples/examples/whisper-microphone/main.rs
@@ -624,13 +624,27 @@ pub fn main() -> Result<()> {
             continue;
         }
         let mut resampled_pcm = vec![];
-        for buffered_pcm in buffered_pcm.chunks(1024) {
+        // resample the audio, one chunk of 1024 samples at a time.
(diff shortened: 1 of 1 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2649, opened 2024-11-29
Add Nvembed v2 model

cdoko:
[NV-Embed-v2](https://huggingface.co/nvidia/NV-Embed-v2) is a text embedding model that ranks No. 1 (as of Nov 25 2024) on the [MTEB](https://huggingface.co/spaces/mteb/leaderboard) benchmark.

This implementation reproduces the example in the [model card](https://huggingface.co/nvidia/NV-Embed-v2).

License: CC-BY-NC-4.0

Please let me know if any changes are required!

### DIFF
--- a/candle-examples/examples/nvembed_v2/README.md
+++ b/candle-examples/examples/nvembed_v2/README.md
@@ -1,0 +1,43 @@
+# NV-Embed-v2
+
+Candle implementation (inference only) of [NV-Embed-v2](https://huggingface.co/nvidia/NV-Embed-v2), a text embedding model that ranks No. 1 (as of Nov 25 2024) on the [MTEB](https://huggingface.co/spaces/mteb/leaderboard) benchmark with a score of 72.31 across 56 text embedding tasks.
+
+## Running an example: Retrieval
+```bash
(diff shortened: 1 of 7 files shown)


LaurentMazare:
Thanks for adding this model, one tricky bit is that the `candle-transformers` crate has an apache/mit license so we cannot really include code that is CC-BY-NC whithin it - having models for which the code is apache/mit and the weights is CC-BY--NC would be fine though.
So if you can base your implementation on code that doesn't have the NC restriction that would make it possible to have this in `candle-transformers`.

LaurentMazare:
(disclaimer: not a lawyer neither a specialist of license)
The repo on huggingface has a license of "cc-by-nc-4.0" and unless it's explicitely stated otherwise in the files in the repo, my understanding is that it applies to them, so in particular it applies to the `modeling_nvembed.py` file and this cannot be used as a base for the implementation itself.
If you can re-implement the model based on the paper or an alternative open-source implementation with a more open source license that would be great.

LaurentMazare (approved):

LaurentMazare:
Looks all good to me, I've merged it, thanks!

Issue #2654, opened 2024-12-01
Batcher iterator return None if return_last_incomplete_batch && items.is_empty()


hhllhhyyds:
In Batcher

```
impl<I: Iterator<Item = Tensor>> Iterator for Batcher<Iter1<I>> {
    type Item = Result<Tensor>;

    fn next(&mut self) -> Option<Self::Item> {
        let mut items = Vec::with_capacity(self.batch_size);
        for _i in 0..self.batch_size {
            // We have two levels of inner here so that we can have two implementations of the
            // Iterator trait that are different for Iter1 and Iter2. If rust gets better
            // specialization at some point we can get rid of this.
            match self.inner.inner.next() {
                Some(item) => items.push(item),
                None => {
                    if self.return_last_incomplete_batch {
                        break;
                    }
                    return None;
                }
            }
        }
        Some(Tensor::stack(&items, 0))
    }
}
```
should be changed to 

```
impl<I: Iterator<Item = Tensor>> Iterator for Batcher<Iter1<I>> {
    type Item = Result<Tensor>;

    fn next(&mut self) -> Option<Self::Item> {
        let mut items = Vec::with_capacity(self.batch_size);
        for _i in 0..self.batch_size {
            // We have two levels of inner here so that we can have two implementations of the
            // Iterator trait that are different for Iter1 and Iter2. If rust gets better
            // specialization at some point we can get rid of this.
            match self.inner.inner.next() {
                Some(item) => items.push(item),
                None => {
                    if self.return_last_incomplete_batch {
                        if items.is_empty() {
                            return None;
                        } else {
                            break;
                        {
                    }
                    return None;
                }
            }
        }
        Some(Tensor::stack(&items, 0))
    }
}
```

PR #2655, opened 2024-12-01
Fix Batcher iterator break condition

hhllhhyyds:
Fix Batcher iterator break when return_last_incomplete_batch and items.is_empty (#2654)


nerdai (commented):
I just encountered this bug as well. The fix looks good to me.

Do we have any unit tests for `Batcher` yet?

LaurentMazare (approved):

PR #2658, opened 2024-12-04
Change/bert encoder public

singjc:
Hello,

I've slightly modified some of the bert components, specifically making the BertEncoder public, so that it can be used on it's own as part of other models. I have implemented a pytorch model in candle, which the pytorch model imports the BertEncoder from the equivalent transformers module. So to implement the candle version, I needed to make the BertEncoder public.


LaurentMazare:
Thanks!

Issue #2680, opened 2024-12-24
Whisper example panic on single threaded


mert-kurttutan:
Set `export RAYON_NUM_THREADS=1`, then run the following
```bash
cargo run --example whisper --release --features="symphonia"
   Compiling candle-transformers v0.8.1 (/workspaces/candle/candle-transformers)
   Compiling candle-examples v0.8.1 (/workspaces/candle/candle-examples)
    Finished `release` profile [optimized] target(s) in 1m 10s
     Running `target/release/examples/whisper`
Running on CPU, to run on GPU, build this example with `--features cuda`
No audio file submitted: Downloading https://huggingface.co/datasets/Narsil/candle_demo/blob/main/samples_jfk.wav
pcm data loaded 176000
thread 'main' panicked at /workspaces/candle/candle-transformers/src/models/whisper/audio.rs:236:24:
index out of bounds: the len is 0 but the index is 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

mert-kurttutan:
It is because of this line, https://github.com/huggingface/candle/blob/11aa30be10ebf42d10799a0726a874c74e30ad3e/candle-transformers/src/models/whisper/audio.rs#L206

When number of threads is 1, which is odd, to get to even, it reduces to zero, hence collected vector has zero length.

mert-kurttutan:
Opening PR right now.

PR #2681, opened 2024-12-24
Fix bug in whisper transformer

mert-kurttutan:
- due to num_threads going to zero in single threaded case


mert-kurttutan:
Resolves #2680

LaurentMazare:
Thanks.

PR #2684, opened 2024-12-27
UniPC for diffusion sampling

nicksenger:
Hi, thanks for the awesome library! It's great having these tools available in the Rust ecosystem.

I was interested in low step-count inference for some experiments, so ported over the [UniPC](https://arxiv.org/abs/2302.04867) scheduler. I figured I'd open a PR here in case this work is useful to others. Here is a comparison with Euler A and DDIM at 5 steps which demonstrates UniPC's benefits for quick convergence:

comparison

### DIFF
--- a/candle-transformers/src/models/stable_diffusion/uni_pc.rs
+++ b/candle-transformers/src/models/stable_diffusion/uni_pc.rs
@@ -1,18 +1,19 @@
 //! # UniPC Scheduler
 //!
-//! UniPCMultistepScheduler is a training-free framework designed for
-//! fast sampling of diffusion models.
+//! UniPC is a training-free framework designed for the fast sampling of diffusion models, which consists of a
+//! corrector (UniC) and a predictor (UniP) that share a unified analytical form and support arbitrary orders.
 //!
-//! It consists of a corrector (UniC) and a predictor (UniP) that share a unified analytical form and
-//! support arbitrary orders. UniPC is by design model-agnostic, supporting pixel-space/latent-space
-//! DPMs on unconditional/conditional sampling. It can also be applied to both noise prediction and
-//! data prediction models. The corrector UniC can be also applied after any off-the-shelf solvers
-//! to increase the order of accuracy.
+//! UniPC is by design model-agnostic, supporting pixel-space/latent-space DPMs on unconditional/conditional
+//! sampling. It can also be applied to both noise prediction and data prediction models. Compared with prior
+//! methods, UniPC converges faster thanks to the increased order of accuracy. Both quantitative and qualitative
+//! results show UniPC can improve sampling quality, especially at very low step counts (5~10).
 //!
+//! For more information, see the original publication:
 //! UniPC: A Unified Predictor-Corrector Framework for Fast Sampling of Diffusion Models, W. Zhao et al, 2023.
 //! https://arxiv.org/abs/2302.04867
 //!
-//! This work is based largely on the diffusers implementation: https://github.com/huggingface/diffusers
+//! This work is based largely on UniPC implementation from the diffusers python package:
+//! https://raw.githubusercontent.com/huggingface/diffusers/refs/heads/main/src/diffusers/schedulers/scheduling_unipc_multistep.py
 use std::ops::Neg;
 use std::{cell::RefCell, collections::HashSet};
 
@@ -876,6 +877,8 @@ mod stats {
         }
     }
 
+    /// Provides total ordering for floats through this trick: http://stereopsis.com/radix.html
+    /// NaN | -Infinity | x < 0 | -0 | +0 | x > 0 | +Infinity | NaN
     #[derive(Clone, Copy)]
     struct FloatOrd(f64);
 
--- a/candle-transformers/src/models/stable_diffusion/uni_pc.rs
+++ b/candle-transformers/src/models/stable_diffusion/uni_pc.rs
@@ -14,8 +14,8 @@
 //!
 //! This work is based largely on UniPC implementation from the diffusers python package:
 //! https://raw.githubusercontent.com/huggingface/diffusers/refs/heads/main/src/diffusers/schedulers/scheduling_unipc_multistep.py
+use std::collections::HashSet;
 use std::ops::Neg;
-use std::{cell::RefCell, collections::HashSet};
 
 use super::schedulers::PredictionType;
 use super::{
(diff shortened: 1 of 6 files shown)


LaurentMazare:
Would you have some examples with good quality where this makes a difference?

nicksenger:
Sure, here is a comparison of DDIM (default cfg) and UniPC (corrector enabled past step 2, `Bh2` solver type, otherwise defaults) on sd1.5 across 50 steps with the following prompt with seed `1984` and guidance scale 15:

> a rusty robot holding a fire torch in its hand,android scrapyard,oxidized,intricate exposed wires,arcing,neon lights,sci-fi,dystopia,futuristic city background,silhouettes,strange illuminated mannequins,geisha billboard,flying cars,4k,cinematic lighting,photo-realistic,extremely detailed,high quality,epic,lfg

negative prompt:

> reduction,reducing agents,electron rich,drawing,illustrated,low quality,out of focus,blurry,cut off,simple,clean,organized,daytime,utopia,peaceful,low contrast,lowres,worst quality

steps

6 steps is about where the output starts becoming clear from both schedulers, but the composition is quite different:

1984-15-6-ddim-512
1984-15-6-unipc-512

At 50 steps both have converged to a similar output with some differences. Which one looks better is a bit subjective, but I think most would agree the composition is closer to the 6-step output from UniPC than that from DDIM:

1984-15-50-ddim-512
1984-15-50-unipc-512

Just for fun here's the 50-step output from UniPC with the same settings as above but using `3` for the solver order:

1984-15-50-order3-512

Somewhat different output again, but overall the composition pretty much resembles what it produces at 5 steps:

1984-15-5-order3-512

candle-transformers/src/models/stable_diffusion/uni_pc.rs
```diff
@@ -0,0 +1,1048 @@
...
+}
+
+struct Inner {
+    model_outputs: Vec<Option<Tensor>>,
+    lower_order_nums: usize,
+    order: usize,
+    last_sample: Option<Tensor>,
+}
+
+struct State(RefCell<Inner>);
```
LaurentMazare, line 254: Could we avoid the `RefCell` here and expose the proper mutability requirements in the various methods instead?

candle-transformers/src/models/stable_diffusion/uni_pc.rs
```diff
@@ -0,0 +1,1048 @@
...
+            for sample in samples {
+                self.add(sample);
+            }
+
+            self
+        }
+    }
+
+    #[derive(Clone, Copy)]
+    struct FloatOrd(f64);
```
LaurentMazare, line 888: Could you give some insights on the ordering that this provides?

candle-transformers/src/models/stable_diffusion/uni_pc.rs
```diff
@@ -0,0 +1,1048 @@
...
+
+    impl std::cmp::Eq for FloatOrd {}
+    impl std::cmp::PartialEq for FloatOrd {
+        fn eq(&self, other: &Self) -> bool {
+            self.convert().eq(&other.convert())
+        }
+    }
+
+    fn sort_floats(v: &mut [f64]) {
+        let v_: &mut [FloatOrd] = unsafe { std::mem::transmute(v) };
```
LaurentMazare, line 922: Unless this is performance critical, we should avoid the `transmute` here and go through something like `sort_by_key` instead.

candle-transformers/src/models/stable_diffusion/uni_pc.rs
```diff
@@ -0,0 +1,1048 @@
...
+//! It consists of a corrector (UniC) and a predictor (UniP) that share a unified analytical form and
+//! support arbitrary orders. UniPC is by design model-agnostic, supporting pixel-space/latent-space
+//! DPMs on unconditional/conditional sampling. It can also be applied to both noise prediction and
+//! data prediction models. The corrector UniC can be also applied after any off-the-shelf solvers
+//! to increase the order of accuracy.
+//!
+//! UniPC: A Unified Predictor-Corrector Framework for Fast Sampling of Diffusion Models, W. Zhao et al, 2023.
+//! https://arxiv.org/abs/2302.04867
+//!
+//! This work is based largely on the diffusers implementation: https://github.com/huggingface/diffusers
```
LaurentMazare, line 15: If there is an equivalent file in `diffusers` could we have a direct permalink to it so that it's easier to compare with the version that was ported?

LaurentMazare:
Thanks for the detailed analysis, I've put a bunch of comments inline, please also look at the clippy and rustfmt failures in the CI.

candle-transformers/src/models/stable_diffusion/uni_pc.rs
nicksenger: Yes, though it requires also changing the mutability requirements of `Scheduler::step`. I wanted to avoid changing any existing signatures since I'm not that familiar with the project, but this should be all that's needed: 73e9bc9c9e25ba789ee72a51b734179f4e4bc05f

Fwiw adding other multi-step schedulers like ipndm or dpm++ would most likely require this change as well.

candle-transformers/src/models/stable_diffusion/uni_pc.rs
nicksenger: I added a comment in `fedce6287405fe04826344c3b5537375b317760a`, the ordering is:

`NaN | -Infinity | x < 0 | -0 | +0 | x > 0 | +Infinity | NaN`

It's the same strategy used by [float-ord](https://crates.io/crates/float-ord), which is in turn a dependency of the [average](https://crates.io/crates/average) crate where this quantile computation comes from. These are only used for the dynamic thresholding, so I thought it'd be better to hide the logic within this module versus introducing additional dependencies crate-wide.

candle-transformers/src/models/stable_diffusion/uni_pc.rs
nicksenger: Sounds good, updated in b95f2f5ba705ab799281d460febea5c7ca76eda8

I didn't notice any significant performance impact with the dynamic thresholding enabled, and in any case the author doesn't recommend the thresholding for stable diffusion - I just included it for the sake of completeness.

candle-transformers/src/models/stable_diffusion/uni_pc.rs
nicksenger: Sure, updated in 2204ef8d1ddcbb588d5c764a94d91da4883e47d4

This deviates from that implementation a bit in terms of how the schedule is specified since I was mainly interested in the Karras schedule. Extending it to support all the options available there shouldn't be too difficult though if that's preferred.

nicksenger:
> Thanks for the detailed analysis, I've put a bunch of comments inline, please also look at the clippy and rustfmt failures in the CI.

Thanks, these should be addressed now.

candle-transformers/src/models/stable_diffusion/uni_pc.rs
LaurentMazare: Agreed that it's better not to introduce a dependency but why not use the total ordering `f64::total_cmp`? Do you expect some differences with it that would be helpful here?

nicksenger: Ah, nope. This was just an oversight on my part, I thought the msrv for the project was lower for some reason, but now I'm seeing `total_cmp` is used in quite a few other places. It looks like the standard lib uses the same method: https://doc.rust-lang.org/src/core/num/f64.rs.html#1350

Updated in f0384fa62c4942983f0dccffffa61c79d3d4b996

LaurentMazare:
Thanks!

super-fun-surf:
stoked!

PR #2686, opened 2024-12-29
Added XLMRobertaModel for Reranking

akshayballal95:
Some of the good reranker models are based on this architecture and I think its a good addition to the current repository of models. 

The architecture is not very different than Bert but the position embedding technique is a bit different.

### DIFF
--- a/candle-examples/examples/xlm-roberta/main.rs
+++ b/candle-examples/examples/xlm-roberta/main.rs
@@ -1,0 +1,283 @@
+use std::path::PathBuf;
+
+use anyhow::{Error as E, Result};
+use candle::{Device, Tensor};
+use candle_nn::VarBuilder;
+use candle_transformers::models::xlm_roberta::{
+    Config, XLMRobertaForMaskedLM, XLMRobertaForSequenceClassification,
+};
+use clap::{Parser, ValueEnum};
+use hf_hub::{api::sync::Api, Repo, RepoType};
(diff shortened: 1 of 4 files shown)


candle-examples/examples/xlm-roberta/main.rs
```diff
@@ -0,0 +1,283 @@
...
+    model_id: Option<String>,
+
+    #[arg(long, default_value = "main")]
+    revision: String,
+
+    #[arg(long, default_value = "bge-reranker-base")]
+    model: Model,
+
+    #[arg(long, default_value = "reranker")]
+    task: String,
```
LaurentMazare, line 44: Let's have an enum rather than a String for the task as there seems to only be two possibilities.

candle-transformers/src/models/xlm_roberta.rs
```diff
@@ -0,0 +1,545 @@
...
+        self.classifier.forward(&hidden_states)
+    }
+}
+
+fn prepare_4d_attention_mask(
+    mask: &Tensor,
+    dtype: DType,
+    tgt_len: Option<usize>,
+) -> Result<Tensor> {
+    let bsz = mask.dims()[0];
```
LaurentMazare, line 524: Use `let bsz = mask.dim(0)?` instead, this will make a proper error rather than a panic if the shape is not correct. Same for the line below.

LaurentMazare:
Thanks!

PR #2713, opened 2025-01-13
ModernBERT model

janimo:
Model and basic fill-mask example as in the Huggingface Transformers model card.


LaurentMazare (approved):

LaurentMazare:
Thanks, great addition!

PR #2718, opened 2025-01-16
Sync upstream MLX sdpa vector kernels with mask

EricLBuehler:
This PR:
- Adds the 2-pass vector SDPA inference kernel (https://github.com/ml-explore/mlx/pull/1597) - useful for long-context generation
- Syncs the recent addition of masking to the SDPA vector kernels (https://github.com/ml-explore/mlx/pull/1757)
  - I'm not sure how useful this change is as we do not use a mask during the decoding phase when these kernels are used (q_seq = 1, kv_seq > 1). However, I think keeping the kernels in-sync with upstream will be helpful.

### DIFF
--- a/candle-metal-kernels/src/lib.rs
+++ b/candle-metal-kernels/src/lib.rs
@@ -1906,10 +1906,11 @@ pub fn call_sdpa_vector(
         alpha
     };
 
-    let constants = Some(ConstantValues::new(vec![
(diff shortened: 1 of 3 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2722, opened 2025-01-16
[Draft]: Add flash attention to whisper (unsure how correct this is)

Murad-Awad:
Whisper should be able to support flash attention. I attempted to add this in this PR; I don't know much about flash attention and a lot of this was pattern matching with other examples in the repo. I need to take a more detailed look later.

### DIFF
--- a/candle-transformers/src/models/whisper/model.rs
+++ b/candle-transformers/src/models/whisper/model.rs
@@ -33,10 +33,27 @@ struct MultiHeadAttention {
     softmax_span: tracing::Span,
     matmul_span: tracing::Span,
     kv_cache: Option<(Tensor, Tensor)>,
+    use_flash_attn: bool,
+}
+
+#[cfg(feature = "flash-attn")]
+fn flash_attn(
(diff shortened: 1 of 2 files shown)


candle-transformers/src/models/whisper/model.rs
```diff
@@ -109,25 +127,33 @@ impl MultiHeadAttention {
...
-        };
-        let wv = {
-            let _enter = self.matmul_span.enter();
-            w.matmul(&v)?
+        if self.use_flash_attn {
+            let q = q.transpose(1, 2)?;
+            let k = k.transpose(1, 2)?;
+            let v = v.transpose(1, 2)?;
+            let softmax_scale = 1f32 / (self.n_head as f32).sqrt();
+            flash_attn(&q, &k, &v, softmax_scale, true)?.transpose(1, 2)
```
Murad-Awad, line 135: I think all attention is technically causal in whisper given that it predicts tokens sequentially. I could be wrong on this though and I have a very cursory understanding of flash attention.

PR #2723, opened 2025-01-17
Fix GLM4 alignment issue

guoqingbao:
This PR primarily addresses the generation issue in GLM4, specifically the missing `rope_ratio` in the construction of the cosine-sine cache. Additionally, it enables the configuration to be loaded from a JSON file and allows weights to be loaded from a local path, facilitated by a new utility function named `hub_load_local_safetensors`. Furthermore, it resolves a compilation issue caused by the updated `hf-hub` crate in the previous #2691, which now depends on a newer version of the `tokio` package.

Tested case: 

```shell
cargo run --release --example glm4 --features cuda -- --weight-path /home/weights/glm-4-9b-chat/ --prompt "Please talk about deep learning."
```

It now generates better answers (align with the official results).

### DIFF
--- a/candle-book/Cargo.toml
+++ b/candle-book/Cargo.toml
@@ -25,7 +25,7 @@ cudarc = { workspace = true, optional = true }
 half = { workspace = true, optional = true }
 image = { workspace = true, optional = true }
 anyhow = { workspace = true }
-tokio = "1.29.1"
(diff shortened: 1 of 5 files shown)


LaurentMazare:
Thanks, nice to have that fixed and in line with the official implementation!

PR #2729, opened 2025-01-19
Various improvements to low-rank + quantized backpropagation,

computer-whisperer:
Added some changes to the back-propagation system for reducing VRAM usage when not all tensors are variables (loras mostly). 

Also added support for qmatmul in back-propagation -- let me know if there is a more intended way to accomplish that then what I have set up here. The whole quantized math system seems very much an afterthought to the rest of candle's design.


computer-whisperer:
That latest commit should have fixed the rustfmt errors -- are there any other checklist items I need to complete before this can be reviewed?

PR #2730, opened 2025-01-21
Fix bug for asort kernel & faster sampler with GPU sorting

guoqingbao:
This PR addresses an issue with the `asort` kernel, which can lead to _CUDA invalid arguments_ error when the last dimension to be sorted exceeded the number of threads allowed per CUDA block. With the revised `asort` kernel, we can now partially perform logit sampling on the GPU (specifically, the sorting process). This optimization improves text generation performance by up to 25%, depending on the model and the number of tokens generated.

### Testing Cases
I conducted tests using the following models and commands:

#### GLM4-9B
```shell
cargo run --release --example glm4 --features cuda -- \
  --weight-path /home/glm-4-9b-chat/ \
  --prompt "Please talk about deep learning." \
  --sample-len 2000
```
*Note*: By default, the GLM4 model uses sampling parameters of `temperature=0.8` and `top_p=0.8`.

#### LLaMa3.1 8B
```shell
cargo run --release --example llama --features cuda -- \
  --weight-path /home/Meta-Llama-3.1-8B-Instruct/ \
  --prompt "Please talk about deep learning." \
  --temperature 0.7 \
  --top-p 0.8 \
  --sample-len 2000
```

#### Original Implementation (Sampling Entirely on CPU)
**GLM4-9B:**
- 100 tokens generated (26.52 tokens/s)
- 500 tokens generated (25.43 tokens/s)
- 2000 tokens generated (20.41 tokens/s)

**LLaMa3.1 8B:**
- 100 tokens generated (40.14 tokens/s)
- 500 tokens generated (38.62 tokens/s)
- 607 tokens generated (38.28 tokens/s)
- 2000 tokens generated (34.12 tokens/s)

#### After `asort` Kernel Fix & Faster Sampler with GPU Sort
**GLM4-9B:**
- 100 tokens generated (29.93 tokens/s)
- 500 tokens generated (28.52 tokens/s)
- 2000 tokens generated (22.30 tokens/s)

**LLaMa3.1 8B:**
- 100 tokens generated (50.68 tokens/s)
- 500 tokens generated (48.81 tokens/s)
- 712 tokens generated (47.88 tokens/s)
- 2000 tokens generated (41.84 tokens/s)

_Note: You can adjust the `temperature` parameter to influence the model's behavior and potentially generate more than 2000 tokens for the above comparisons. While `sample-len` sets the maximum number of tokens that can be generated, the actual output length may vary depending on the model (and sampling process, parameters, etc.)._

### DIFF
--- a/candle-transformers/src/models/quantized_gemma3.rs
+++ b/candle-transformers/src/models/quantized_gemma3.rs
@@ -14,15 +14,18 @@
 //! - [Gemma 3 Models](https://blog.google/technology/developers/gemma-3/)
 //!
 
-use std::collections::HashMap;
-
 use crate::quantized_nn::RmsNorm;
 use candle::quantized::gguf_file;
 use candle::quantized::QTensor;
+use candle::D;
 use candle::{DType, Device, IndexOp, Result, Tensor};
 use candle_nn::{Embedding, Module};
 
 pub const MAX_SEQ_LEN: usize = 131072; // Gemma 3 supports 128K context window
+pub const DEFAULT_SLIDING_WINDOW_TYPE: usize = 6;
+pub const DEFAULT_ROPE_FREQUENCY: f32 = 1_000_000.;
+pub const DEFAULT_ROPE_FREQUENCY_SLIDING: f32 = 10_000.;
+pub const DEFAULT_ROPE_FREQUENCY_SCALE_FACTOR: f32 = 1.;
 
 #[derive(Debug, Clone)]
 struct QMatMul {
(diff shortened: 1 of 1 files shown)


LaurentMazare:
I'm not sure to understand the change properly and testing them they seem to produce weird results. If you run the tests with the cuda feature enabled, this will actually fail, did you try it?
```
$ cargo test --features cuda
---- asort_gpu stdout ----
thread 'asort_gpu' panicked at candle-core/tests/tensor_tests.rs:133:5:
  left: [[1, 3, 0, 2, 4], [1, 0, 2, 3, 4]]
 right: [[1, 3, 0, 2, 4], [1, 4, 0, 2, 3]]
```

Below is also another example I tried where the results seem wrong. Fwiw the original implementation comes from llama.cpp [argsort.cu](https://github.com/ggerganov/llama.cpp/blob/f35726c2fb0a824246e004ab4bedcde37f3f0dd0/ggml/src/ggml-cuda/argsort.cu#L11) so if you get the kernel to work on arbitrary sizes I would suggest to also make a PR there as it would be of interest to them.

Finally, could you make only the kernel changes in this PR, and defer the sampling bits to a PR for later once this has stabilized a bit?

```rust
use anyhow::Result;

fn main() -> Result<()> {
    let values: Vec<u32> = (0..10u32).rev().collect();
    let a = Tensor::new(values, &dev)?;
    let a = a.arg_sort_last_dim(true)?;
    let a = a.to_vec1::<u32>()?;
    println!("{a:?}");
    Ok(())
}
```

guoqingbao:
> I'm not sure to understand the change properly and testing them they seem to produce weird results. If you run the tests with the cuda feature enabled, this will actually fail, did you try it?
> 
> ```
> $ cargo test --features cuda
> ---- asort_gpu stdout ----
> thread 'asort_gpu' panicked at candle-core/tests/tensor_tests.rs:133:5:
> assertion `left == right` failed
>   left: [[1, 3, 0, 2, 4], [1, 0, 2, 3, 4]]
>  right: [[1, 3, 0, 2, 4], [1, 4, 0, 2, 3]]
> ```
> 
> Below is also another example I tried where the results seem wrong. Fwiw the original implementation comes from llama.cpp [argsort.cu](https://github.com/ggerganov/llama.cpp/blob/f35726c2fb0a824246e004ab4bedcde37f3f0dd0/ggml/src/ggml-cuda/argsort.cu#L11) so if you get the kernel to work on arbitrary sizes I would suggest to also make a PR there as it would be of interest to them.
> 
> Finally, could you make only the kernel changes in this PR, and defer the sampling bits to a PR for later once this has stabilized a bit?
> 
> ```rust
> use anyhow::Result;
> use candle_core::{Device, Tensor};
> 
> fn main() -> Result<()> {
>     let dev = Device::new_cuda(0)?;
>     let values: Vec<u32> = (0..10u32).rev().collect();
>     let a = Tensor::new(values, &dev)?;
>     let a = a.arg_sort_last_dim(true)?;
>     let a = a.to_vec1::<u32>()?;
>     println!("{a:?}");
>     Ok(())
> }
> ```

Thanks for the comments. I'm currently working on a fix.

guoqingbao:
Hi @LaurentMazare,
I have fixed the bug in the asort kernel. The issue was caused by the following factors:

1. Limited shared memory: The available shared memory is insufficient to cache even a single row of the input array (maximum ~40k f32 elements), which complicates thread synchronization.
2. Improper global memory access: Bitonic sort requires the input size to be a power of 2 (log2-based), and this was not handled correctly.
3. Padding issues: Some values were not padded appropriately, which is necessary given the nature of bitonic sort.

These issues have been addressed by creating temporary row and indice buffers (sized to a power of 2, padded) and copying the valid elements back.

guoqingbao:
We have introduced a more elegant approach for GPU token sampling in candle-vllm: [Faster token sampling](https://github.com/EricLBuehler/candle-vllm/pull/109). This solution resolves the sort kernel crash on A100 and older-generation GPUs while delivering a **20% to 45% performance increase** when sampling parameters such as `temperature` and `top-k`/`top-p` are applied in candle-vllm.

The current bitonic sort kernel in this PR is not optimally designed due to candle's lack of FFI support for GPU kernels, making it complex to allocate temporary buffers (as the bitonic kernel requires 2^n space) and perform thread sync. @LaurentMazare  I would like this PR remains open as a reference for future implementations of more elegant GPU-based token sampling strategies in candle.

PR #2735, opened 2025-01-23
Add inpainting to the stable diffusion example

mneilly:
This change updates the stable diffusion example with inpainting for 1.5, 2 and XL. It adds a --mask-path command line argument to set the path to the mask image and it uses the existing img2img argument to set the input image.

The following is an example of the output of 10 samples from SDXL using:
```
cargo run --example stable-diffusion --features cuda,cudnn,flash-attn -- --prompt "a cute baby lion sitting on a park bench" --sd-version xl-inpaint --use-f16 --use-flash-attn --img2img overture-creations-5sI6fQgYIuo_big.png --mask-path overture-creations-5sI6fQgYIuo_mask_big.png --guidance-scale 7 --num-samples 10
```

montage

### DIFF
--- a/candle-examples/examples/stable-diffusion/main.rs
+++ b/candle-examples/examples/stable-diffusion/main.rs
@@ -4,11 +4,13 @@ extern crate accelerate_src;
 #[cfg(feature = "mkl")]
 extern crate intel_mkl_src;
 
+use std::ops::Div;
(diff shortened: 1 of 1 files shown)


LaurentMazare (approved):

LaurentMazare:
Nice addition, thanks!

PR #2745, opened 2025-01-27
Implement DeepSeek V3

EricLBuehler:
- [x] Model
- [ ] FP8 weight dequantize
- [ ] Tensor parallelism


zackangelo:
Seems like most of this PR is adding the f8 dtype, perhaps worth separating that out from the deepseek model addition so its easier to review?

super-fun-surf:
this Would be great. And the F8 type is really helpful for other projects.

super-fun-surf:
it looks like the working branch is behind the main branch now. maybe the fp8 work should be pulled out. it looks like a lot of good work by @EricLBuehler

EricLBuehler:
Interesting idea. I'll take a look at doing that once a few other PRs are merged - specifically the Metal MM speedup one.

AlbertMarashi:
Is this supposed to be able to support training?

PR #2748, opened 2025-01-27
use moondream1 model/revision for moondream example

slckl:
`moondream` example seems to be broken now.

The revision in repo fails to load.
```
Error: path: "../models--vikhyatk--moondream2/snapshots/30c7cdf3fa6914f50bee3956694374143f5cc884/model.safetensors" MetadataIncompleteBuffer
```

I think upstream may have broken it.

Changing it to the latest revision of  `vikhyatk/moondream1` fixes the example.

Quantized version works just fine.


LaurentMazare:
Thanks!

PR #2762, opened 2025-02-05
Refactor Shape From<Tuple> implementations by using macros, add tests

philipfabianek:
I noticed that the From<Tuple> implementations were quite repetitive and could be refactored using macros. I also noticed there were no tests for them so I added a few basic tests. I see other things which could be refactored using macros but I will create different PRs for them.


LaurentMazare (approved):

LaurentMazare:
Thanks.

PR #2780, opened 2025-02-21
Make sorted_nodes pub function

viirya:
While debugging backpropagation path of a Tensor, I would need to look at the nodes of it. But `sorted_nodes` is not public function for now. This patch makes it a public function for easy debugging.


viirya:
Thank you @LaurentMazare

PR #2785, opened 2025-02-24
Implement I-quants (IQ4XS, IQ4NL)

EricLBuehler:
This PR refactors the quantization parts of candle-core a bit and integrates some new I-quants!

There is no CUDA or Metal support yet; perhaps we could add that in a later PR. I have added Metal support on a local branch, and I'm working on syncing the latest GGML CUDA kernels, which should also give a nice performance boost!

### DIFF
--- a/candle-core/tests/quantized_tests.rs
+++ b/candle-core/tests/quantized_tests.rs
@@ -173,7 +173,7 @@ fn quantized_matmul_neg(device: &Device) -> Result<()> {
     if device.is_cuda() {
         assert!(diff < 0.1);
     } else {
-        assert!(diff < 0.96);
+        assert_eq!(diff, 0.);
     }
     Ok(())
 }


candle-core/src/quantized/iq_quants/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+
+        quantize_iq4_xs(xs, ys, 1, k, None)?;
+
+        Ok(())
+    }
+
+    #[allow(unreachable_code)]
+    fn vec_dot(n: usize, xs: &[Self], ys: &[Self::VecDotType]) -> Result<f32> {
+        // #[cfg(target_feature = "avx")]
+        // todo!();
```
EricLBuehler, line 83: TODO: No AVX vec-dot implementation for IQ4xs to IQ4nl yet.

candle-core/src/quantized/k_quants/utils.rs
```diff
@@ -0,0 +1,352 @@
...
+    let block_size = T::BLCK_SIZE;
+    let dtype = T::DTYPE;
+
+    let expected_blocks = xs.len() / block_size;
+    let actual_blocks = ys.len();
+
+    // Validate that the input is the right size
+    if expected_blocks != actual_blocks {
+        crate::bail!("quantize {dtype:?}: expected {expected_blocks} blocks but only {actual_blocks} were provided!")
+    }
```
ivarflakstad, line 23: I am not opposed to doing this check every time, as that is safest, but this may be a decent case for [debug_assert](https://doc.rust-lang.org/std/macro.debug_assert.html). What do you think @LaurentMazare?

candle-core/tests/quantized_tests.rs
```diff
@@ -169,11 +169,11 @@ fn quantized_matmul_neg(device: &Device) -> Result<()> {
     let lhs2 = Tensor::stack(&[&lhs, &lhs], 0)?;
     let res2 = matmul.forward(&lhs2)?;
     let res2 = res2.i(1)?;
-    let diff = (res - res2)?.abs()?.sum_all()?.to_vec0::<f32>()?;
+    let diff = (&res - res2)?.abs()?.mean_all()?.to_vec0::<f32>()? / res.elem_count() as f32;
     if device.is_cuda() {
         assert!(diff < 0.1);
     } else {
-        assert_eq!(diff, 0.);
+        assert!(diff < 0.96);
```
ivarflakstad, line 176: The `assert_eq` version actually passes on my machine for both the neon and default impl. Is this increase in tolerance because of the avx impl perhaps?

ivarflakstad (commented):
Reviewing in chunks. Around half way there.

candle-core/src/quantized/iq_quants/utils.rs
```diff
@@ -0,0 +1,252 @@
...
+    for x in x.iter().take(super_block_size) {
+        sigma2 += x * x;
+    }
+    sigma2 *= 2.0 / (super_block_size as f32);
+
+    // 2. zero out q4, set dh to 0
+    for qi in q4.iter_mut() {
+        *qi = 0;
+    }
+    *dh = f16::from_f32(0.0);
```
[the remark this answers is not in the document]
EricLBuehler: Sounds good - fixed :)

candle-core/tests/quantized_tests.rs
EricLBuehler: Hmm, I think that was a remnant from when I did the other neon/default impls. I undid this change (I'm still working out a bug in the iq4xs avx impl, iq4nl works nicely).

PR #2809, opened 2025-03-16
Add Gemma 3 1b IT toe Gemma examples

andreban:
- Updates the Gemma example to include Gemma 3 1b instruction tuned.

### DIFF
--- a/candle-examples/examples/gemma/main.rs
+++ b/candle-examples/examples/gemma/main.rs
@@ -50,6 +50,8 @@ enum Which {
     InstructV2_9B,
     #[value(name = "3-1b")]
     BaseV3_1B,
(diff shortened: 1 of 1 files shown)


andreban:
@LaurentMazare I added the Gemma 3 1b instruction tuned. Both the base model and instruction tuned crash with `kv-cache: above max-seq-len 512+1>51`, but thought it was nice to add the instruction model, as it mostly works, and the same fix should apply to both.

EricLBuehler:
@andreban to fix that you could use RotatingKvCache for the sliding window layers, and all others use KvCache with the normal max seq len.

LaurentMazare:
Thanks!

PR #2810, opened 2025-03-16
Allow for growing the default KV cache when needed.

LaurentMazare:


andreban:
Just tried it and can confirm it fixes the "kv-cache: above max-seq-len" error with Gemma 3. Thank you!

LaurentMazare:
Thanks for checking, I also checked it on my side while making the change as it was the main purpose there :)

PR #2813, opened 2025-03-17
Implement the Gemma 3 vision models!

EricLBuehler:
- [x] Add vision part - mmproj
- [x] Implement forward pass
- [ ] Add inputs processor (pan & scan, normalize, etc)
- [ ] Example

### DIFF
--- a/candle-transformers/src/models/gemma3.rs
+++ b/candle-transformers/src/models/gemma3.rs
@@ -450,7 +450,20 @@ impl Model {
             .to_dtype(self.dtype)
     }
 
+    pub fn embed_tokens(&self, input_ids: &Tensor) -> Result<Tensor> {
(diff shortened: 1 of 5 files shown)


super-fun-surf:
when i run  the 4b and 12b like 
```
cargo run --example gemma_vision --release --features cuda,cudnn -- --prompt "describe this image" --which v3_4b --image /home/pro/Desktop/sd_final.png 
```
it downloads the models and then
i get error
```
temp: 0.00 repeat-penalty: 1.10 repeat-last-n: 64
retrieved the files in 98.089819ms
Error: data did not match any variant of untagged enum Gemma3Config
```

AlpineVibrations:
is this working? its been sitting here for a while would be great to have Gemma vision models in candle especially if the work is done..?

PR #2820, opened 2025-03-22
Integrate MLX SDPA kernels with mask

EricLBuehler:
This PR integrates kernel developments from: https://github.com/ml-explore/mlx/pull/1924.

Specifically, our `candle_nn::ops::sdpa` function now dispatches to optimized implementations for with and without prompts. There is also an option for causal masking, removing the necessity for mask materialization.

**Overall, this means that we can fuse the attention operation on Metal for prompt and decode phases!**

I will update this PR further with benchmarks, but it is tested and working in my fork through mistral.rs.

### DIFF
--- a/candle-metal-kernels/src/lib.rs
+++ b/candle-metal-kernels/src/lib.rs
@@ -323,6 +323,7 @@ impl Kernels {
         device: &Device,
         source: Source,
         name: &str,
+        name: &str,
         constants: Option<FunctionConstantValues>,
     ) -> Result<Function, MetalKernelError> {
         let func = self
@@ -357,7 +358,7 @@ impl Kernels {
             let pipeline = device
                 .new_compute_pipeline_state_with_function(&func)
                 .map_err(|e| MetalKernelError::FailedToCreatePipeline(e.to_string()))?;
-            pipelines.insert((name, constants), pipeline.clone());
+            pipelines.insert((name.to_string(), constants), pipeline.clone());
 
             Ok(pipeline)
         }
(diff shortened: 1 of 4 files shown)


candle-kernels/build.rs
```diff
@@ -1,4 +1,6 @@
 fn main() {
+    std::env::set_var("NVCC_PREPEND_FLAGS", "-D_USE_MATH_DEFINES");
+
```
ivarflakstad, line 3: Remnant from other work? ☺️

candle-metal-kernels/src/lib.rs
```diff
@@ -1570,174 +1570,201 @@ pub fn call_sdpa_full(
...
     k_offset: usize,
+    k_shape: &[usize],
+    k_strides: &[usize],
     k_buffer: &Buffer,
     v_offset: usize,
     v_buffer: &Buffer,
+    v_strides: &[usize],
+    mask_type: Option<SdpaDType>,
+    mask_buffer: Option<&Buffer>,
+    m_strides: Option<&[usize]>,
```
ivarflakstad, line 1584: If these are either all present together or all `None`, maybe they could be sent in as a struct wrapping them ala `Option<&SdpaMask>`?

candle-nn/src/ops.rs
```diff
@@ -1149,27 +1151,70 @@ impl candle::CustomOp3 for Sdpa {
...
-                )
+            command_buffer.set_label("full_attention");
+            if self.softcapping != 1. {
+                candle::bail!("SDPA full requires softcapping to be disabled (1.0)");
             }
 
-            command_buffer.set_label("full_attention");
+            let mask_s_l = self.mask.as_ref().map(|m| m.storage_and_layout());
+
+            let (mask_type, mask_buffer, mask_strides) = if let Some(mask) = &self.mask {
```
ivarflakstad, line 1161: Guess my assumption was right hehe

candle-metal-kernels/src/lib.rs
ivarflakstad: I know we haven't done that for the other parameters. Maybe we should have. The list is getting pretty long.

ivarflakstad: Let's have a look at that at some other time. No need to introduce that concept in this PR.

candle-metal-kernels/src/scaled_dot_product_attention.metal
```diff
@@ -5,6 +5,262 @@
...
+  _MLX_BFloat16() threadgroup = default;
+  _MLX_BFloat16() device = default;
+  _MLX_BFloat16() constant = default;
+
+  struct bits_to_bfloat_struct {};
+  static constexpr METAL_FUNC bits_to_bfloat_struct bits_to_bfloat() {
+    return bits_to_bfloat_struct();
+  }
+  constexpr METAL_FUNC _MLX_BFloat16(uint16_t bits, bits_to_bfloat_struct)
+      : bits_(bits) {}
```
ivarflakstad, line 71: Oh boy. Yeah we've been putting off this for a while now.

I'll do a refactoring of the metal backend with building kernels etc so that we actually have a structured project where we can include headers etc.

I've already done it when I was implementing reduce kernels, but since it wasn't needed right then we delayed it.

I'll get the PR back up fairly soon.

ivarflakstad (approved):
Great work 🎉 

I only have 2 comments really.

One is that at this point it's really time to start precompiling kernels. Not for performance, just for the sake of project structure and maintainability.
I have that ready to go so I'll make a PR soon.

The second is that I wonder if it makes sense to move more of the logic in `candle-nn` into `candle-metal-kernels`.
The standard asserts etc is fine, but maybe figuring out the correct kernel to call is actually a concept that belongs inside `candle-metal-kernels`?
That's open for discussion obviously.

In any case this is ready to merge in my opinion - if all tests pass and it runs smoothly ☺️

candle-metal-kernels/src/scaled_dot_product_attention.metal
ivarflakstad: By "fairly soon" I apparently meant ~1 hour
https://github.com/huggingface/candle/pull/2906 🤷 

Don't worry about precompiling in this PR btw. I'll work my way through all the kernels where it's useful 🫡

candle-kernels/build.rs
EricLBuehler: Yeah 😄! Removed

candle-metal-kernels/src/lib.rs
EricLBuehler: Ok, sounds good. Might be good for a medium-sized refactor in the future!

PR #1656, opened 2024-02-04
feat: add flash_forward and flag to toggle

drbh:
This PR adds flash attention to whisper and can be run with the following command.

```bash
cargo run --example whisper --release --features flash-attn -- --use-flash-attn
```

*note to enable candle must be built with `flash-attn` and the `--use-flash-attn` must bet set

PR #2821, opened 2025-03-22
feat: `use_flash_attn` for whisper  and fix `candle-flash-attn` MSVC build

xkeyC:
1. Resolve conflicts for https://github.com/huggingface/candle/pull/1656
2. In MSVC environments, linking stdc++ will result in the error ` LINK : fatal error LNK1181: cannot open input file 'stdc++.lib'` .  According to information from [StackOverflow](https://stackoverflow.com/questions/73836369), we simply need to remove this linking item in MSVC.

I'm working on a[ real-time subtitle software](https://github.com/xkeyC/fl_caption) using candle-based Whisper and have made great progress. While looking for ways to improve performance, I came across PR#1656.

I'm new to Rust, please let me know if I did anything wrong, thanks.

### DIFF
--- a/candle-examples/examples/whisper/main.rs
+++ b/candle-examples/examples/whisper/main.rs
@@ -473,6 +473,10 @@ struct Args {
     /// Print the full DecodingResult structure rather than just the text.
     #[arg(long)]
     verbose: bool,
+
+    /// Use flash attention.
+    #[arg(long)]
+    use_flash_attn: bool,
 }
 
 fn main() -> Result<()> {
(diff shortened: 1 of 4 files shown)


xkeyC:
After careful testing, `flash_attn` has shown poor performance and I will turn it off.

super-fun-surf:
currently with stable diffusion (at least) flash_attn reduces gpu memory needed which helps it run on smaller cards. for example my 20GB rex 4000 won't run SD3.5 unless flash_attn is on because it reduces the memory needed.

xkeyC:
In my tests, I got more transcription delay, probably because of the delay, and the quality of the results decreased.

Since my requirement is real-time processing, I'm not going to go any further with this feature.

super-fun-surf:
> In my tests, I got more transcription delay, probably because of the delay, and the quality of the results decreased.
> 
> Since my requirement is real-time processing, I'm not going to go any further with this feature.

as a dide note. we would like to proceede without Flash_attn but we are unsure how to get the memory savings that flash attn is providing. any hints or ideas?
thank you

PR #2824, opened 2025-03-22
Pickle decoder fix and Long1 opcode addition.

computer-whisperer:
Added Long1 Opcode, required for pulling tensors with offset or size > 32-bit maximum

Fixed tensor offset calculation, which has to be scaled by the data width.


LaurentMazare (approved):

LaurentMazare:
Thanks!

Issue #2816, opened 2025-03-19
Flash Attn Problem in linux


AlpineVibrations:
Its been working fine. And still works without flash Attention.
some of my debian system files might have been updated and All of a sudden i cannot build on linux with an RTX 4000 ADA using flash attn.
```
CUDA_COMPUTE_CAP=80 cargo run --example stable-diffusion-3 --features cuda,cudnn,flash-attn -- --prompt "something cool" --which 3.5-large --use-flash-attn
```

gets errors
```
 #$ gcc -std=c++17 -D__CUDA_ARCH__=800 -D__CUDA_ARCH_LIST__=800 -D__NV_LEGACY_LAUNCH -E -x c++  -DCUDA_DOUBLE_MATH_FUNCTIONS -D__CUDACC__ -D__NVCC__ -D__CUDACC_EXTENDED_LAMBDA__ -D__CUDACC_RELAXED_CONSTEXPR__  -O3 -I"cutlass/include" "-I/usr/local/cuda/bin/../targets/x86_64-linux/include"    -U "__CUDA_NO_HALF_OPERATORS__" -U "__CUDA_NO_HALF_CONVERSIONS__" -U "__CUDA_NO_HALF2_OPERATORS__" -U "__CUDA_NO_BFLOAT16_CONVERSIONS__" -D__CUDACC_VER_MAJOR__=12 -D__CUDACC_VER_MINOR__=6 -D__CUDACC_VER_BUILD__=85 -D__CUDA_API_VER_MAJOR__=12 -D__CUDA_API_VER_MINOR__=6 -DCUDA_API_PER_THREAD_DEFAULT_STREAM=1 -D__NVCC_DIAG_PRAGMA_SUPPORT__=1 -include "cuda_runtime.h" -m64 "kernels/flash_fwd_hdim64_bf16_causal_sm80.cu" -o "/tmp/tmpxft_00005461_00000000-7_flash_fwd_hdim64_bf16_causal_sm80.cpp1.ii" 

```

```
[...]

PR #2829, opened 2025-03-25
fix: `candle-flash-attn` linux and `msvc` build

xkeyC:
1. In MSVC environments, linking stdc++ will result in the error  LINK : fatal error LNK1181: cannot open input file 'stdc++.lib' . According to information from [StackOverflow](https://stackoverflow.com/questions/73836369), we simply need to remove this linking item in MSVC.
2. In gcc (or clang) , If the compiler target is a dynamic library, we need to add the `-fPIC` flag to generate position-independent code.  will fix: https://github.com/huggingface/candle/issues/2816
https://stackoverflow.com/questions/5311515/gcc-fpic-option
https://gcc.gnu.org/onlinedocs/gcc-4.0.4/gccint/PIC.html


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2835, opened 2025-03-25
Added readmes to examples 

greenrazer:


candle-examples/examples/custom-ops/README.md
```diff
@@ -0,0 +1,14 @@
+# candle-custom-ops
```
pcuenca, line 1: Maybe include a short description summarizing from https://github.com/huggingface/candle/blob/cb02b389d53a1cf5547dfa69b5168bdc1a50d325/candle-examples/examples/custom-ops/main.rs#L1-L4

candle-examples/examples/llama/README.md
```diff
@@ -0,0 +1,9 @@
+# candle-llama
```
pcuenca, line 1: same, I'd add a one-line description here

candle-examples/examples/glm4/README.org
```diff
@@ -12,7 +12,7 @@ GLM-4-9B is the open-source version of the latest generation of pre-trained mode
 
 ** Running with ~cpu~
 #+begin_src shell
-  cargo run --example glm4 --release -- --cpu--prompt "Hello world"
+  cargo run --example glm4 --release -- --cpu --prompt "Hello world"
```
pcuenca, line 15: 👍

candle-examples/examples/mamba/README.md
```diff
@@ -12,6 +12,6 @@ would only work for inference.
 ## Running the example
 
 ```bash
-$ cargo run --example mamba-minimal --release -- --prompt "Mamba is the"
+$ cargo run --example mamba --release -- --prompt "Mamba is the"
```
pcuenca, line 15: 👍

candle-examples/examples/reinforcement-learning/README.md
```diff
@@ -2,12 +2,32 @@
...
+$ uv venv
+$ source .venv/bin/activate
+$ uv pip install gymnasium
+$ which python3.xx 
+> /path/to/.venv/bin/python3.xx
+$ find ~/.local/share/uv/python -name "libpython3.xx.dylib"
+> /path/to/pythondir/lib/libpython3.xx.dylib
+$ cp /path/to/pythondir/lib/libpython3.xx.dylib ~/libpython3.xx.dylib.backup
+$ install_name_tool -id "/path/to/pythondir/lib/libpython3.xx.dylib" "/path/to/pythondir/lib/libpython3.xx.dylib"
+$ otool -L /path/to/pythondir/lib/libpython3.xx.dylib
```
pcuenca, line 25: oh wow, why is this necessary? (just curious)

pcuenca (commented):
Looks great! I'd suggest we always try to include a one-line description of what each example does, I provided suggestions for a couple of cases.

Another idea would be to include an overview README in the `examples` folder itself, but this can be addressed in another PR.

candle-examples/examples/reinforcement-learning/README.md
greenrazer: I'm honestly not sure. This was a headache for a while. For some reason the rust linker looks for the python libs in `/install/lib` which can't exist because it's in the read only file system on mac. The only way I've found to fix this is to change the install name of the dll to the location where it's supposed to be instead of `/install/lib/`.

LaurentMazare: That's odd, I never had to do something like this and wouldn't expect candle to need anything specific here compared to some standard pyo3 code.
If it's a common issue I would expect pyo3 folks to have encountered it somehow.

greenrazer: My guess is that it's a problem with using venv on [UV](https://github.com/astral-sh/uv). It's pretty new and does some different things with symlinks.

pcuenca: Some potentially related discussion here: https://github.com/PyO3/pyo3/issues/4813. In particular, [this looks promising](https://github.com/PyO3/pyo3/issues/4813#issuecomment-2617037553).

If an easy config is not possible, I'd maybe suggest to provide simple steps for standard venv and a warning that `uv` is not compatible with `pyo3`.

greenrazer: The 

`DYLD_FALLBACK_LIBRARY_PATH=/Users/kb/.local/share/uv/python/cpython-3.12.7-macos-aarch64-none/lib/ CARGO_BUILD_RUSTFLAGS="-L /Users/kb/.local/share/uv/python/cpython-3.12.7-macos-aarch64-none/lib/" cargo run --example reinforcement-learning --features=pyo3 --package candle-examples -- pg` 

was the first thing I tried and I could not get it to work. 

It seems to be a problem with PYTHONHOME and PYTHONPATH with uv.

I'll remove the procedure I used to get it to work and just provide a warning. I can attach it as a comment to this thread for archival purposes.

greenrazer:
I'm not sure why this is failing. The error message seems to be about compiling `candle-book` on macOS:

```
---- tests::book_hub_1 stdout ----

thread 'tests::book_hub_1' panicked at candle-book/src/lib.rs:21:60:
called `Result::unwrap()` on an `Err` value: LockAcquisition("/Users/runner/.cache/huggingface/hub/models--bert-base-uncased/blobs/68d45e234eb4a928074dfd868cead0219ab85354cc53d20e772753c6bb[916](https://github.com/huggingface/candle/actions/runs/14205028581/job/39800666025?pr=2835#step:6:917)9d3.lock")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Where `candle-book/src/lib.rs:21:60` refers to:

```
async fn book_hub_1() {
        // ANCHOR: book_hub_1
        use candle::Device;
        use hf_hub::api::tokio::Api;
        
        let api = Api::new().unwrap();
        
        //THIS LINE vvvvvvvvvvvvvvv
        let weights_filename = repo.get("model.safetensors").await.unwrap();
        //THIS LINE ^^^^^^^^^^^^^
        
        let weights = candle::safetensors::load(weights_filename, &Device::Cpu).unwrap();      
        // ANCHOR_END: book_hub_1
        assert_eq!(weights.len(), 206);
    }
```

I get the error when running the test on main using `cargo test -p candle-book --lib`, so I don't think it's a problem with this PR.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2834, opened 2025-03-25
Fixed rand imports for whisper-microphone example.

greenrazer:
Changed
`rand::distributions::Distribution` to `rand::distr::Distribution`
and
`rand::distributions::WeightedIndex` to `rand::distr::weighted::WeightedIndex`

PR #2839, opened 2025-03-26
Fixed errors in the reinforcement-learning example.

greenrazer:
- Fixed import errors with the `rand` package similar to #2833 and #2834.
- Updated readme with my procedure to make the example work with my python virtual environment.


greenrazer:
Whoops, just fixed by #2837

PR #2842, opened 2025-03-28
Added Deepseekr1 Llama8b variant to quantized example

greenrazer:
From [unsloth/DeepSeek-R1-Distill-Llama-8B-GGUF](https://huggingface.co/unsloth/DeepSeek-R1-Distill-Llama-8B-GGUF)

`cargo run --example quantized --release -- --which deepseekr1-llama8b --prompt "Lebesgue's integral is superior to Riemann's because "`

### DIFF
--- a/candle-examples/examples/quantized/main.rs
+++ b/candle-examples/examples/quantized/main.rs
@@ -76,7 +76,7 @@ enum Which {
     #[value(name = "SmoLM2-1.7B-Instruct")]
     SmolLM2_1BInstruct,
     #[value(name = "deepseekr1-llama8b")]
-    DeepseekR1Llama8b
+    DeepseekR1Llama8b,
 }
 
 impl Which {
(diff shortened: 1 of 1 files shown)


candle-examples/examples/quantized/main.rs
```diff
@@ -106,65 +93,21 @@ impl Which {
             | Self::Mistral7b
             | Self::Mistral7bInstruct
             | Self::Mistral7bInstructV02 => true,
+            _ => false,
```
LaurentMazare, line 96: The long explicit pattern matching was on purpose, this forces someone adding a new case to decide whether this function should return false or true rather than defaulting to one of the two cases. Same for the other functions.

greenrazer: I will revert.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2843, opened 2025-03-28
Added DeepseekR1 Qwen7B variant to quantized-qwen2-instruct example

greenrazer:
From [unsloth/DeepSeek-R1-Distill-Qwen-7B-GGUF](https://huggingface.co/unsloth/DeepSeek-R1-Distill-Qwen-7B-GGUF)

`cargo run --example quantized-qwen2-instruct  --release -- --which deepseekr1-qwen7b --prompt "Write a function to count prime numbers up to N.`

### DIFF
--- a/candle-examples/examples/quantized-qwen2-instruct/main.rs
+++ b/candle-examples/examples/quantized-qwen2-instruct/main.rs
@@ -27,6 +27,8 @@ enum Which {
     W2_7b,
     #[value(name = "72b")]
     W2_72b,
+    #[value(name = "deepseekr1-qwen7b")]
+    DeepseekR1Qwen7B,
 }
 
 #[derive(Parser, Debug)]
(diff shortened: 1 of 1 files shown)


candle-examples/examples/quantized-qwen2-instruct/main.rs
```diff
@@ -211,11 +223,18 @@ fn main() -> anyhow::Result<()> {
...
-    let prompt_str = args.prompt.unwrap_or_else(|| DEFAULT_PROMPT.to_string());
-    let prompt_str = format!(
-        "<|im_start|>user\n{}<|im_end|>\n<|im_start|>assistant\n",
-        prompt_str
-    );
+    let prompt_str = args
+        .prompt
+        .clone()
+        .unwrap_or_else(|| DEFAULT_PROMPT.to_string());
+    let prompt_str = if args.is_deepseek() {
```
LaurentMazare, line 230: Below there is a pattern matching to check if the model is deepseek rather than a `if` test, we should probably decide between one of the two styles.

greenrazer: I think pattern matching is more readable, but I'm fine with changing it if you'd prefer the `if` approach.

LaurentMazare (approved):

LaurentMazare:
Thanks!

Issue #2632, opened 2024-11-19
Cannot download CIFAR-10


VillSnow:
```rust
fn main() -> Result<(), Box<dyn Error>> {
    let cifar = candle_datasets::vision::cifar::load()?;

    dbg!(&cifar.train_images); // Tensor[dims 50000, 3, 32, 32; u8]
    dbg!(&cifar.train_labels); // Tensor[dims 50000; u8]

    dbg!(cifar
        .train_images
        .flatten_all()?
        .to_vec1::<u8>()?
        .into_iter()
        .any(|x| x != 0)); // false

    dbg!(cifar
        .train_labels
        .flatten_all()?
        .to_vec1::<u8>()?
        .into_iter()
        .any(|x| x != 0)); // true

    Ok(())
}
```

I was able to see download progress bars for `plain_text/test/0000.parquet` and `lain_text/train/0000.parquet`, but no for images.

BerserkerMother:
what is the problem here?

cschin:
There are two issues in the 0.8.0 codebase for downloading CIFAR-10:

1. The data in the Parquet file, when read through `image::load_from_memory`, has the shape `(W, H, C)`. However, the code converts the image data to `(C, W, H)`.
2. The data in the Parquet file is of type `u8` with values ranging from `0` to `255`. It needs to be cast to `f32` before being divided by `255.0_f32`.

Here is a modification that works with the latest 0.8.0 code.
(I am not sure if Candle has something similar to PyTorch's `moviedom`, so I have do the shape change with a Vec.)

```
diff --git a/candle-datasets/src/vision/cifar.rs b/candle-datasets/src/vision/cifar.rs
index 4b403a2e..71b2e735 100644
--- a/candle-datasets/src/vision/cifar.rs
+++ b/candle-datasets/src/vision/cifar.rs
@@ -67,13 +67,20 @@ fn load_parquet(parquet: SerializedFileReader<std::fs::File>) -> Result<(Tensor,
     let samples = parquet.metadata().file_metadata().num_rows() as usize;
     let mut buffer_images: Vec<u8> = Vec::with_capacity(samples * 1_024);
     let mut buffer_labels: Vec<u8> = Vec::with_capacity(samples);
+    let mut rgb_buffer: Vec<u8> = vec![0; 3 * W * H];
     for row in parquet.into_iter().flatten() {
         for (_name, field) in row.get_column_iter() {
             if let parquet::record::Field::Group(subrow) = field {
[...]

PR #2845, opened 2025-03-29
Fix CIFAR10 dataset types and dimension ordering

brylee10:
Currently, `candle_datasets::vision::cifar::load()` returns images of all 0s because of a type mismatch error. The image dimensions also need to be permuted. This is noted in https://github.com/huggingface/candle/issues/2632, an issue that I encountered recently as well. @cschin proposed a solution, which is implemented here. The permutation dimensions are slightly different to uphold the `(C, H, W)` conv layer convention, but doesn't matter given the symmetry of CIFAR10 square images.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2861, opened 2025-04-01
Add `ReduceLROnPlateau` learning rate scheduler

brylee10:
I encountered a need for a learning rate scheduler analogous to the PyTorch `ReduceLROnPlateau` so I rolled my own. I figured others may find it useful as well so made a PR with it here. 

There is an open thread about adding this feature from last year (https://github.com/huggingface/candle/discussions/2224). It looks like there was code written to add a core `Scheduler` trait to set up for learning rate schedulers which was written but never merged (https://github.com/huggingface/candle/pull/2225).  

I think a core trait like that makes sense and I think adding this feature first is still compatible with that design, if it is eventually merged. One note is that `Scheduler` trait from the PR does not take additional arguments in the `step()` function, but `ReduceLROnPlateau` is an exception to the [other PyTorch LR schedulers ](https://github.com/pytorch/pytorch/blob/v2.6.0/torch/optim/lr_scheduler.py#L1363) since it takes a metric as argument. 

For now, `ReduceLROnPlateau` is standalone, but I think the `Scheduler` trait proposed can be generalized to support this, or the API of `ReduceLROnPlateau` could be modified to be consistent: 
```rust
scheduler.set_metric(loss);
scheduler.step();
```

### DIFF
--- a/candle-nn/src/optim.rs
+++ b/candle-nn/src/optim.rs
@@ -199,3 +199,334 @@ impl AdamW {
         self.params = params;
     }
 }
+
+/// Specifies whether the metric should be minimized or maximized. Used for [`ReduceLROnPlateau`]
+// Named with a prefix to avoid naming conflicts with similar types in future. Modules are an alternative,
+// but this did not feel more ergonomic than a name prefix.
+#[derive(Clone, Copy, Debug)]
+pub enum LRPlateauMode {
+    Min,
+    Max,
+}
+
+/// Specifies how the threshold is applied: relative or absolute. Used for [`ReduceLROnPlateau`]
+// Named with a prefix to avoid naming conflicts with similar types in future.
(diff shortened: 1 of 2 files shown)


brylee10:
FYI @EricLBuehler

brylee10:
Got it. PyTorch has a plethora which I understand would be onerous to all support. The motivation was I noticed no other schedulers existed and it seems others had opened discussions about a similar need for at least one of the common schedulers like this. I thought giving one common scheduler would be a useful baseline, similar to how Candle supports SGD and AdamW as two examples of optimizers but stops there leaving the rest to [candle-optimizers](https://docs.rs/candle-optimisers/latest/candle_optimisers/). 

However, I see how it could be out of scope for the main repo given it is a "core" framework, so maybe schedulers can instead be brought into a separate crate.

Issue #2812, opened 2025-03-16
model request: DistilBertForMaskedLM


VoVAllen:
Implement MaskedLM variants for DistilBert

VoVAllen:
Thanks!

PR #2866, opened 2025-04-04
Implementing DistilBertForMaskedLM.

greenrazer:
## Overview

DistilBertForMaskedLM is DistilBert with a small MLP head at the end for predicting a masked token([see here](https://github.com/huggingface/transformers/blob/41b9b92b52215bed472c9a534a06abbc3a9a95cd/src/transformers/models/distilbert/modeling_distilbert.py#L810)). 
The weights for the masked head are in the same repo [`distilbert-base-uncased`](https://huggingface.co/distilbert/distilbert-base-uncased).

### From
#2812 

## ~~Not Working (2025/4/4)~~
- ~~Example code not displaying the top k predictions.~~
- ~~Example code and model code needs cleanup.~~
- ~~The position mask isn't implemented.~~

## ~~Next~~

- ~~Right now I've separated both models into different files in `candle-transformers/src/models`. However, since they are from the same repo, and this new model has to call the standard `distilbert` anyway, perhaps both models should go in the same file?~~
- ~~Similarly for the example, perhaps they should both go in the same example project.~~


LaurentMazare:
I feel that this model should indeed be in the same file as distilbert, and probably the examples can be folded together too. You may want to grep for `BertForMaskedLM` or `ModernBertForMaskedLM` in the repo to check a couple related examples and get a feel of the current grouping/naming conventions.

VoVAllen:
Thanks! DistilBert is used for multiple sparse vector models like SPLADE v3 and opensearch's latest models.

candle-examples/examples/distilbert/main.rs
```diff
@@ -42,94 +67,254 @@ struct Args {
     #[arg(long, default_value = "1")]
     n: usize,
 
-    /// L2 normalization for embeddings.
+    /// L2 normalization for embeddings. (Not Implemented)
```
LaurentMazare, line 70: Let's remove this if it's unused.

candle-examples/examples/distilbert/main.rs
```diff
@@ -42,94 +67,254 @@ struct Args {
...
+    fn load_variables(&self, weights_path: &PathBuf, device: &Device) -> Result<VarBuilder> {
+        if self.use_pth {
+            Ok(VarBuilder::from_pth(weights_path, DTYPE, device)?)
+        } else {
+            Ok(unsafe { VarBuilder::from_mmaped_safetensors(&[weights_path], DTYPE, device)? })
+        }
+    }
+
+    fn create_model(&self, config: &Config, vb: VarBuilder) -> Result<ModelType> {
+        if self.prompt.contains("[MASK]") {
```
LaurentMazare, line 138: Matching on the string seems a bit error-prone, I would be in favor of having an explicit command line flag to trigger this behavior.

candle-examples/examples/distilbert/main.rs
```diff
@@ -42,94 +67,254 @@ struct Args {
...
+fn process_masked_output(
+    output: &Tensor,
+    token_ids: &Tensor,
+    tokenizer: &Tokenizer,
+    args: &Args,
+) -> Result<()> {
+    let input_ids_vec = token_ids.to_vec2::<u32>()?;
+    let mask_token_id = tokenizer
+        .token_to_id("[MASK]")
+        .ok_or(anyhow::anyhow!("No mask token found"))?;
```
LaurentMazare, line 230: It doesn't matter much here but `anyhow::Context` would be a bit more idiomatic there + avoid allocating the error even on the non-error path.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2872, opened 2025-04-08
Fix dtype mismatch in BERT attention mask handling

msminhas93:
This PR addresses a critical issue with dtype compatibility in the BERT implementation. Previously, the attention mask was hardcoded to use DType::F32, which caused dtype mismatch errors when using models with different precision types (such as BF16 causing Error: dtype mismatch in mul, lhs: BF16, rhs: F32)

Changes:
Modified the get_extended_attention_mask function to convert the f32::MIN tensor to the same dtype as the attention mask

Updated BertModel::forward to use the model's dtype (obtained from embedding_output) instead of hardcoding DType::F32

This change ensures that the attention mask uses the same dtype as the model weights, preventing errors when using non-F32 precision types like BF16.

### DIFF
--- a/candle-transformers/src/models/bert.rs
+++ b/candle-transformers/src/models/bert.rs
@@ -504,8 +504,9 @@ impl BertModel {
             Some(attention_mask) => attention_mask.clone(),
             None => input_ids.ones_like()?,
         };
+        let dtype = embedding_output.dtype();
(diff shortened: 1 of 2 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2879, opened 2025-04-11
fix: missing shared tensor in bert::BertLMPredictionHead

silver-ymz:
(`cls.predictions.decoder.weight`, `bert.embeddings.word_embeddings.weight`) and (`cls.predictions.decoder.bias`, `cls.predictions.bias`) are shared tensor in `BertForMaskedLM` model. They may not exist in `model.safetensors` file.

### DIFF
--- a/candle-transformers/src/models/bert.rs
+++ b/candle-transformers/src/models/bert.rs
@@ -569,7 +569,21 @@ pub struct BertLMPredictionHead {
 impl BertLMPredictionHead {
     pub fn load(vb: VarBuilder, config: &Config) -> Result<Self> {
         let transform = BertPredictionHeadTransform::load(vb.pp("transform"), config)?;
-        let decoder = linear(config.hidden_size, config.vocab_size, vb.pp("decoder"))?;
+
+        let decoder_weight = if vb.contains_tensor("decoder.weight") {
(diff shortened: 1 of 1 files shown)


candle-transformers/src/models/bert.rs
```diff
@@ -569,7 +569,21 @@ pub struct BertLMPredictionHead {
...
+            vb.root().get(
+                (config.vocab_size, config.hidden_size),
+                "bert.embeddings.word_embeddings.weight",
+            )?
+        };
+        let decoder_bias = if vb.contains_tensor("decoder.bias") {
+            vb.get(config.vocab_size, "decoder.bias")?
+        } else {
+            vb.get(config.vocab_size, "bias")?
+        };
```
alvarobartt, line 585: Hi, not sure and better to double check with a maintainer but maybe this syntax is preferred to avoid the `contains_tensor` call (?)

```suggestion
        let decoder_weight = vb
            .get((config.vocab_size, config.hidden_size), "decoder.weight")
            .or_else(|_| {
                vb.root().get(
                    (config.vocab_size, config.hidden_size),
                    "bert.embeddings.word_embeddings.weight",
                )
            })?;
        let decoder_bias = vb
            .get(config.vocab_size, "decoder.bias")
            .or_else(|_| vb.get(config.vocab_size, "bias"))?;
```

PR #2883, opened 2025-04-12
Avoid using batched-matmul in nn::Linear.

LaurentMazare:


EricLBuehler:
Hi @LaurentMazare - do you have numbers on this change? Seems like a very interesting change!

LaurentMazare:
For a batch size of 1024, input of (1, 1024) and linear (1024, 48000) which is the case in one of our recent models.
- H100 (bf16): 32ms -> 162us.
- macbook M2Pro (bf16): 880ms -> 26ms.
- M2Pro cpu (f32): 4.4s -> 153ms.

EricLBuehler:
Wow, super exciting! Will test if this improves performance noticeably in non-quantized models.

PR #2903, opened 2025-04-17
Add qwen3 on candle-transformers

maximizemaxwell:
Add qwen3.rs to candle-transformers.
- fixed mod.rs
- add qwen3.rs
I got reference from transformers's Qwen3 python codes, which is https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3/configuration_qwen3.py

@LaurentMazare Could you review the code?


LaurentMazare:
Let's wait until the weights are out so that we can test it out a bit and ensure that the python and rust implementations are in line.

PR #2906, opened 2025-04-17
Add metal precompilation

ivarflakstad:
To improve the maintainability of `candle-metal-kernels` I think it is about time we introduce a `build.rs` to the project.
This won't necessarily affect performance, but it will let us reuse components between kernels. Like `get_strided_index`, mlx's `bfloat16` impl (when there is no native bfloat), etc.

Starting with reduce kernels and iteratively adding more to `candle.metallib`.

Note: When running `build.rs` I got an exciting new error.
> xcrun: error: unable to find utility "metal", not a developer tool or in PATH

I solved it by simply running `sudo xcode-select --switch /Applications/Xcode.app/Contents/Developer`.
This should be documented somehow - but where? Should I add it to the error description? Or in the `README.md` perhaps?


zackangelo:
Just a heads up, I wrote a build.rs script a while back to do this: 
https://github.com/huggingface/candle/pull/2335

EricLBuehler:
> Note: When running build.rs I got an exciting new error.
>
>  > xcrun: error: unable to find utility "metal", not a developer tool or in PATH
>
> I solved it by simply running sudo xcode-select --switch /Applications/Xcode.app/Contents/Developer.
> This should be documented somehow - but where? Should I add it to the error description? Or in the README.md perhaps?

Perhaps both. Adding it to the error would probably be the most helpful, as all downstream users will see it.

ivarflakstad:
> Just a heads up, I wrote a build.rs script a while back to do this: 
> 
> https://github.com/huggingface/candle/pull/2335
> 
> 

Nice, thanks for sharing.
(Beat you to it [~1 year ago](https://github.com/huggingface/candle/commit/a9e26b5ad5d58ef2f12d15b73e1976ef43d6291a) though 😉)

zackangelo:
> > Just a heads up, I wrote a build.rs script a while back to do this:
> > #2335
> 
> Nice, thanks for sharing. (Beat you to it [~1 year ago](https://github.com/huggingface/candle/commit/a9e26b5ad5d58ef2f12d15b73e1976ef43d6291a) though 😉. Ended up not using it because it didn't improve performance)

lol march to july is a whole year now? :)

candle-metal-kernels/build.rs
```diff
@@ -0,0 +1,122 @@
...
+        Err(e) => panic!("Compiling metal -> air failed: {:?}", e),
+    }
+
+    // Compile air to metallib
+    let lib_name = match platform {
+        Platform::MacOS => "candle.metallib",
+        Platform::IOS => "candle_ios.metallib",
+    };
+    let metallib = out_dir.join(lib_name);
+    let mut compile_metallib_cmd = Command::new("xcrun");
```
EricLBuehler, line 79: I think the link step should pass `--sdk platform.sdk()` too.

candle-metal-kernels/build.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+        Err(e) => panic!("Compiling air -> metallib failed: {:?}", e),
+    }
+
+    Ok(())
+}
+
+fn main() -> Result<(), String> {
+    compile(Platform::MacOS)?;
```
EricLBuehler, line 118: Can we maybe select which target to build for from the compilation target?

candle-metal-kernels/src/kernel.rs
```diff
@@ -115,12 +116,19 @@ impl Kernels {
...
             Ok(lib.clone())
         } else {
-            let lib = {
-                let source_content = self.get_library_source(source);
-                let compile_options = get_compile_options();
-                device
-                    .new_library_with_source(source_content, Some(&compile_options))
-                    .map_err(|e| MetalKernelError::LoadLibraryError(e.to_string()))?
+            let lib = match source {
+                Source::Candle => device.new_library_with_data(CANDLE).map_err(|e| {
```
EricLBuehler, line 120: Maybe we can make an opt-out to using the precompiled binaries? That may be best suited to be implemented in the build stages though.

candle-metal-kernels/src/metal_src/reduce.metal
```diff
@@ -1,52 +1,8 @@
 #include <metal_stdlib>
 #include <metal_limits>
+#include "utils.metal"
```
EricLBuehler, line 3: Runtime compilation `new_library_with_source(REDUCE)` has no include preprocessing, so I think this will break for users who disable runtime compilation (if we add that)

EricLBuehler (changes requested):
Initial review. Not sure if we want to make precompilation something that can be toggled on/off, a few of my comments are based on that.

candle-metal-kernels/src/kernel.rs
ivarflakstad: Why would you want to opt out? Binary size? Shouldn't be massive.

candle-metal-kernels/build.rs
ivarflakstad: Like `#[cfg(target_os = "macos")]`?
A user on mac wanting to compile this for ios would add `cargo build --target=aarch64-apple-ios` themselves then. Not sure if there are any issues there.

candle-metal-kernels/src/metal_src/reduce.metal
ivarflakstad: My plan was to move all existing metal code to precompiled. It simplifies JIT code as well because we can compile the shader source against a dynamic metal library - so we have all our existing concepts available.

Issue #2891, opened 2025-04-13
Example for google/gemma-3-4b-it-qat-q4_0-gguf


anidotnet:
Need an example to load and generate text for google/gemma-3-4b-it-qat-q4_0-gguf model. None of the existing gemma example or quantize example work for this model. The config file for gemma-3-4b does not match with any example.

pcuenca:
Addressed by @greenrazer in #2902 🙌

LaurentMazare:
The output definitely looks weird, even with the default repeat penalty. @greenrazer would you be able to check this against other implementations? A typical thing to check if you haven't already done so is whether the default prompt formatting is supported by this version of the model.

super-fun-surf:
the other gemma3 examples in the repo produce results like that as well
```
cargo run --example gemma -r -- --prompt 'write me a happy poem about butterflies and waterfalls ' --which 3-1b-it
```
and the output is very positive and i like it but it sure does ramble on...
```
    Finished `release` profile [optimized] target(s) in 5.68s
     Running `target/release/examples/gemma --prompt 'write me a happy poem about butterflies and waterfalls ' --which 3-1b-it`
avx: true, neon: false, simd128: false, f16c: true
temp: 0.00 repeat-penalty: 1.10 repeat-last-n: 64
retrieved the files in 27.758226ms
Running on CPU, to run on GPU, build this example with `--features cuda`
loaded the model in 4.561032005s
write me a happy poem about butterflies and waterfalls 

The sunbeams dance, a golden hue,
On petals soft, so bright and new.
Butterflies flutter, light and free,
A joyful scene, you see!

Waterfalls cascade, a silver stream,
Whispering secrets, sweet and low,
Of summer's grace, a dream.

Tiny wings, they fly, a dance,
Across the mossy stones, so green,
With colors bright, a hue.

The butterflies, a swirl, a dream,
A happy scene, a joy,
In the forest deep.

Sunbeams warm, a glow,
Of summer's grace, you see!

Waterfalls cascade,
Whispering secrets, sweet,
Of summer's dream.

Butterflies, a swirl, a dance,
With colors bright, a hue.

The forest, a dream,
A happy scene,
In the heart, you feel.

Sunbeams, warm,
Of summer.

Let your spirit soar,
And let it flow.

Butterfly,
A joyful,
A moment,
A perfect.

**Why is this poem?**

It's a simple and sweet, and evokes feelings of happiness and peace.**

[...]

PR #2914, opened 2025-04-23
fixed quantized-gemma example

greenrazer:
- Scaled down RoPE base frequency by the scale_factor in the GGUF.
- Fixed the chat prompt

## From

#2891


LaurentMazare:
Thanks!

PR #2918, opened 2025-04-24
Fixed Quantized Gemma3 Model and example

greenrazer:
While working on the fixes for #2917, I got curious about whether those changes could also benefit the quantized version of the model.

Implementing the correct sliding masking led to a slight performance improvement, so I started experimenting further. I noticed that the [GGUF file](https://huggingface.co/google/gemma-3-4b-it-qat-q4_0-gguf?show_file_info=gemma-3-4b-it-q4_0.gguf) doesn’t include a sliding window RoPE frequency or a sliding window type (the parameter that describes the ratio of sliding layers to global layers). So I tried manually setting it to the value used by the non-quantized Gemma3, ignoring the scale_factor from the GGUF metadata.

Surprisingly, this made a significant difference, the model became noticeably less rambly, terminated faster, repeated itself less, and hallucinated less often.

That said, this approach is a bit hacky, since it involves overriding values that are technically supposed to come from GGUF metadata. So I'm not quite sure how best to proceed with this PR.

Open to input on whether this is good direction or if there's a cleaner solution I’m missing.

### DIFF
--- a/candle-transformers/src/models/quantized_gemma3.rs
+++ b/candle-transformers/src/models/quantized_gemma3.rs
@@ -14,11 +14,10 @@
 //! - [Gemma 3 Models](https://blog.google/technology/developers/gemma-3/)
 //!
 
-use std::collections::HashMap;
-
 use crate::quantized_nn::RmsNorm;
 use candle::quantized::gguf_file;
 use candle::quantized::QTensor;
+use candle::D;
 use candle::{DType, Device, IndexOp, Result, Tensor};
 use candle_nn::{Embedding, Module};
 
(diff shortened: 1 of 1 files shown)


LaurentMazare:
> Implementing the correct sliding masking led to a slight performance improvement, so I started experimenting further. I noticed that the [GGUF file](https://huggingface.co/google/gemma-3-4b-it-qat-q4_0-gguf?show_file_info=gemma-3-4b-it-q4_0.gguf) doesn’t include a sliding window RoPE frequency or a sliding window type (the parameter that describes the ratio of sliding layers to global layers). So I tried manually setting it to the value used by the non-quantized Gemma3, ignoring the scale_factor from the GGUF metadata.

I think this might be hardcoded in llama.cpp based on the model type, e.g. see this [line](https://github.com/ggml-org/llama.cpp/blob/56304069599f4dd9749d94b9bca2c2c65bb27c02/src/llama-model.cpp#L930). This certainly seems hacky compared to using the explicit parameters from the gguf but as they do so, we should probably also hardcode these parameters too.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2919, opened 2025-04-24
candle-onnx: Implement layer normalization operator

BrunoSienkiewicz:
Added Layer Normalization operator with tests.


A2va:
Why not using [LayerNorm](https://github.com/huggingface/candle/blob/82def7ae3826cf140a239f22db6a3b4609918f1c/candle-nn/src/layer_norm.rs#L108) ? 

Or that is a different thing ? (I'm not that familiar with ml things)

BrunoSienkiewicz:
> Why not using [LayerNorm](https://github.com/huggingface/candle/blob/82def7ae3826cf140a239f22db6a3b4609918f1c/candle-nn/src/layer_norm.rs#L108) from candle-nn ?
> 
> Or that is a different thing ? (I'm not that familiar with ml things)

Thank you for your comment. Honestly I didn't see the implementation in candle-nn, maybe it can be used in this case. However I see difference in ONNX version of LayerNorm, that is additional `axis` parameter. I will mark this PR as draft until i clear this out.

BrunoSienkiewicz:
I have changed the implementation to use built-in candle-nn layer normalization. All tests are passing so I think everything should be alright with this approach.

candle-onnx/src/eval.rs
```diff
@@ -549,6 +549,48 @@ fn simple_eval_(
...
+                    if i < axis {
+                        row_number *= xs.dims()[i] as usize;
+                    } else {
+                        col_number *= xs.dims()[i] as usize;
+                    }
+                }
+
+                let x_mat = xs.reshape((row_number, col_number))?;
+
+                let y_mat = candle_nn::ops::layer_norm_slow(
```
LaurentMazare, line 584: Why use layer-norm slow here rather than the optimized variants?

candle-onnx/tests/ops.rs
```diff
@@ -838,6 +838,182 @@ fn test_flatten_operation() -> Result<()> {
...
+        let eval = candle_onnx::simple_eval(&manual_graph, inputs)?;
+        assert_eq!(eval.len(), 1);
+
+        let z = eval
+            .get(OUTPUT_Z)
+            .expect("Output 'z' not found")
+            .to_dtype(DType::F32)?;
+
+        let expected = Tensor::new(expected, &Device::Cpu)?.to_dtype(DType::F32)?;
+        match expected.dims().len() {
```
LaurentMazare, line 1003: Why split these cases and not compare the tensors directly?

BrunoSienkiewicz: At the start i thought it would be a good idea to test out different dimensionality of tensors, but they in the end they are cast to 2D so it would not matter. I changed the test case to compare tensors directly.

vorporeal:
Curious if there are any updates here; I just ran into this issue when trying to integrate `candle-onnx` into my application.

PR #2922, opened 2025-04-25
Added tracing page to the candle book.

greenrazer:


candle-book/src/tracing.md
```diff
@@ -0,0 +1,48 @@
+# Tracing
+
+Tracing is a powerful tool for identifying performance issues and bottlenecks in code.
```
EricLBuehler, line 3: Might be good to highlight that this isn't necessarily the best for identifying issues using CUDA and Metal devicds due to asynchronous execution - perhaps mentioning CUDA_LAUNCH_BLOCKING would be good?

greenrazer: I don't think Metal has a CUDA_LAUNCH_BLOCKING type environment variable. There's the instruments app, but that isn't terribly specific about what the GPU is working on.

candle-book/src/tracing.md
```diff
@@ -0,0 +1,51 @@
...
+
+> When using CUDA, Metal, or other devices with asynchronous execution, be aware that tracing might not accurately identify performance issues. GPU operations are typically queued and executed asynchronously, which can lead to misleading timing information in traces.
+> 
+> If are debugging CUDA specifically, you can set the environment variable `CUDA_LAUNCH_BLOCKING=1` which forces synchronous execution. This makes trace timings more accurate, at the cost of reduced performance.
```
LaurentMazare, line 7: `CUDA_LAUNCH_BLOCKING=1` should be mentioned but I don't think we should recommend using it + tracing to do performance analysis of the code as the traces are commonly very different from what you would see in a non-blocking world. Instead one should probably use nvidia `nsys profile` command + `nsys-ui`. This produces some profiles for async cuda executions using the cuda kernel names.

LaurentMazare (approved):

LaurentMazare:
Thanks!

OlivierDehaene:
@LaurentMazare do you happen to know the required profile settings / rust flags to make the rust symbols show up in `nsys`?

PR #2930, opened 2025-04-29
Updating `Add qwen3` (PR 2903) to use HF weights

greenrazer:
Builds off of @maximizemaxwell 's [PR](#2903), who wrote it before the weights were released. 

This PR:
- Loads the weights
- Fixes modeling code
- Adds Multiple Qwen3 models to the qwen example.

## ~~Issues 2025/4/28~~
- ~~Qwen3 doesn't work with qwen example code by default~~
- ~~Qwen3 modeling code needs slight rework~~


EricLBuehler:
@greenrazer looks like the MoE model has the same architecture as the dense one and the MoE block from the V2!

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,436 @@
...
+    fn forward(&self, xs: &Tensor) -> Result<Tensor> {
+        // xs: (B*L*H, D)
+        let orig_dtype = xs.dtype();
+        let xs = xs.to_dtype(DType::F32)?;
+        let var = (xs.clone() * &xs)?.mean_keepdim(D::Minus1)?;
+        let eps_tensor = Tensor::new::<f32>(self.eps as f32, xs.device())?;
+        let rms = var.broadcast_add(&eps_tensor)?.powf(-0.5)?;
+        let xs = xs.broadcast_mul(&rms)?;
+        let ws = self.weight.unsqueeze(0)?;
+        Ok((xs.broadcast_mul(&ws))?.to_dtype(orig_dtype)?)
```
asukaminato0721, line 107:
```suggestion
        xs.broadcast_mul(&ws)?.to_dtype(orig_dtype)
```

I think.

greenrazer:
I couldn't test the MoE model completely because it couldn't run on my computer. So if anyone has a beefier rig that'd be great.

Also the Model never terminates. I tried adding an EOS token to `<im_end>`. It's probably an issue with rotary embeddings like last time. I'll have to take a look later.

EricLBuehler:
> I couldn't test the MoE model completely because it couldn't run on my computer. So if anyone has a beefier rig that'd be great :).

Will test!

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+            .to_dtype(dtype)?
+            .reshape((max_seq_len, 1))?;
+        let freqs = t.matmul(&inv_freq)?;
+        Ok(Self {
+            sin: freqs.sin()?,
+            cos: freqs.cos()?,
+        })
+    }
+
+    /// RoPE 적용 (q, k shape: B x H x L x D)
```
LaurentMazare, line 52: Please keep the comments in english.

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+        let cos = self.cos.narrow(0, offset, seq_len)?;
+        let sin = self.sin.narrow(0, offset, seq_len)?;
+        let q_embed = candle_nn::rotary_emb::rope(&q.contiguous()?, &cos, &sin)?;
+        let k_embed = candle_nn::rotary_emb::rope(&k.contiguous()?, &cos, &sin)?;
+        Ok((q_embed, k_embed))
+    }
+}
+
+#[derive(Debug, Clone)]
+struct Qwen3RmsNorm {
```
LaurentMazare, line 64: Why not use the standard `RmsNorm` module here?

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+        let var = (xs.clone() * &xs)?.mean_keepdim(D::Minus1)?;
+        let rms = (var + self.eps)?.powf(-0.5)?;
+        let xs = xs.broadcast_mul(&rms)?;
+        let ws = self.weight.unsqueeze(0)?.unsqueeze(1)?;
+        xs.broadcast_mul(&ws)?.to_dtype(orig_dtype)
+    }
+}
+
+#[derive(Debug, Clone)]
+struct Qwen3HeadRmsNorm {
```
LaurentMazare, line 87: Same question as above, should this be the standard `RmsNorm` module? Or maybe just a wrapper around it? The `forward` below is likely to be much less efficient.

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+        let xs = xs.to_dtype(DType::F32)?;
+        let var = (xs.clone() * &xs)?.mean_keepdim(D::Minus1)?;
+        let rms = (var + self.eps)?.powf(-0.5)?;
+        let xs = xs.broadcast_mul(&rms)?;
+        let ws = self.weight.unsqueeze(0)?;
+        xs.broadcast_mul(&ws)?.to_dtype(orig_dtype)
+    }
+}
+
+fn repeat_kv(kv: &Tensor, n_rep: usize) -> Result<Tensor> {
```
LaurentMazare, line 110: ` crate::utils::repeat_kv` ?

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+        rotary_emb: Arc<Qwen3RotaryEmbedding>,
+        layer_idx: usize,
+        vb: VarBuilder,
+    ) -> Result<Self> {
+        let head_dim = cfg.head_dim;
+        let num_heads = cfg.num_attention_heads;
+        let num_kv_heads = cfg.num_key_value_heads;
+        let num_kv_groups = num_heads / num_kv_heads;
+
+        let (q_proj, k_proj, v_proj, o_proj) = if cfg.attention_bias {
```
LaurentMazare, line 182: Use `linear_b` rather than having two different code blocks based on `cfg.attention_bias`.

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+    q_norm: Qwen3HeadRmsNorm,
+    k_norm: Qwen3HeadRmsNorm,
+    // hyper params
+    num_heads: usize,
+    num_kv_heads: usize,
+    num_kv_groups: usize,
+    head_dim: usize,
+    hidden_size: usize,
+    // sliding window
+    _sliding_window: Option<usize>,
```
LaurentMazare, line 164: Why is this prefixed with an `_`, is it unused at the moment, if so a comment should probably explain why it doesn't matter.

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+        let q_flat = self.q_norm.forward(&q_flat)?;
+        let k_flat = self.k_norm.forward(&k_flat)?;
+        let q = q_flat.reshape((b, self.num_heads, l, self.head_dim))?;
+        let k = k_flat.reshape((b, self.num_kv_heads, l, self.head_dim))?;
+
+        // 4. RoPE
+        let (q, k) = self.rotary_emb.apply(&q, &k, offset)?;
+
+        // 5. KV 캐시 누적
+        let (k, v) = match &self.kv_cache {
```
LaurentMazare, line 259: Try using one of the standard `kv_cache` module if possible.

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+impl DecoderLayer {
+    fn new(
+        cfg: &Config,
+        rotary: Arc<Qwen3RotaryEmbedding>,
+        idx: usize,
+        vb: VarBuilder,
+    ) -> Result<Self> {
+        Ok(Self {
```
LaurentMazare, line 307: The usual style in candle is to define most of the values in variables then use them in `Self`, it's not a hard requirement but here it feels that there are a bit too much things defined inline.

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+                cfg.rms_norm_eps,
+                vb.pp("post_attention_layernorm"),
+            )?,
+        })
+    }
+
+    fn forward(&mut self, x: &Tensor, mask: Option<&Tensor>, offset: usize) -> Result<Tensor> {
+        let h = self.ln1.forward(x)?;
+        let h = self.self_attn.forward(&h, mask, offset)?;
+        let x = x.broadcast_add(&h)?;
```
LaurentMazare, line 322: Does this addition really have to be `broadcast`?

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+        self.self_attn.clear_kv_cache();
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Model {
+    embed_tokens: candle_nn::Embedding,
+    layers: Vec<DecoderLayer>,
+    norm: Qwen3RmsNorm,
+    _rotary: Arc<Qwen3RotaryEmbedding>,
```
LaurentMazare, line 338: Why is there an `_` for this field too?

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,441 @@
...
+#[derive(Debug, Clone)]
+pub struct ModelForCausalLM {
+    base: Model,
+    lm_head: Linear,
+}
+
+impl ModelForCausalLM {
+    pub fn new(cfg: &Config, vb: VarBuilder) -> Result<Self> {
+        let base = Model::new(cfg, vb.clone())?;
+        let lm_head = if vb.contains_tensor("lm_head.weight") {
```
LaurentMazare, line 422: Using `tie_words_embeddings` from the config feels a bit safer than looking at whether some weights do exist.

candle-transformers/src/models/qwen3_moe.rs
LaurentMazare, line 1: Please apply the same comments as the ones from `qwen3.rs` where applicable.
Also if some parts are unchanged between the two versions, it would seem good to make them `pub(crate)` in `qwen3.rs` and re-using them here.

EricLBuehler:
> > I couldn't test the MoE model completely because it couldn't run on my computer. So if anyone has a beefier rig that'd be great :).
> 
> Will test!

@greenrazer seems that it is indeed broken for the MoE, while the dense model rougly works. Let me know if there is any information you need.

candle-transformers/src/models/qwen3.rs
greenrazer: Sliding window is null in all the configs I've seen on HF (like [here](https://huggingface.co/Qwen/Qwen3-8B/blob/main/config.json#L23)), But the option is still there. Not sure whether I should add a comment or just remove it.

candle-transformers/src/models/qwen3.rs
greenrazer: I suppose not, I'll avoid broadcasting explicitly.

candle-transformers/src/models/qwen3.rs
greenrazer: oops

candle-transformers/src/models/qwen3.rs
LaurentMazare: In that case, let's remove the `_` and have something like:
```
if sliding_window.is_some() {
    candle::bail!("sliding window is not suppored")
}
```
That way users will know if somehow they use a config with a sliding window.

greenrazer:
> > > I couldn't test the MoE model completely because it couldn't run on my computer. So if anyone has a beefier rig that'd be great :).
> > 
> > 
> > Will test!
> 
> @greenrazer seems that it is indeed broken for the MoE, while the dense model roughly works. Let me know if there is any information you need.

Thank you @EricLBuehler! 

Since the MoE model isn’t currently working and I don’t have the resources to debug it efficiently, I’m planning to remove it from this PR and move it to a draft instead. That way, someone else can build on it if they’d like, or I can revisit it later, especially if smaller MoE variants are released.

candle-examples/examples/qwen/main.rs
```diff
@@ -85,6 +88,10 @@ impl TextGeneration {
             Some(token) => token,
             None => anyhow::bail!("cannot find the <|endoftext|> token"),
         };
+        let eos_token2 = match self.tokenizer.get_token("<|im_end|>") {
```
LaurentMazare, line 91: If this applies to the old qwen models too, did you check that all of them have this token in their tokenizers? If not we should probably keep an `Option<u32>` instead.

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,387 @@
...
+            num_heads * head_dim,
+            cfg.hidden_size,
+            cfg.attention_bias,
+            vb.pp("o_proj"),
+        )?;
+
+        let q_norm = RmsNorm::new(head_dim, cfg.rms_norm_eps, vb.pp("q_norm"))?;
+        let k_norm = RmsNorm::new(head_dim, cfg.rms_norm_eps, vb.pp("k_norm"))?;
+
+        // Necessary because the hidden_size in the cofig isn't always accurate
```
LaurentMazare, line 157:
```suggestion
        // Necessary because the hidden_size in the config isn't always accurate
```

candle-transformers/src/models/qwen3.rs
```diff
@@ -0,0 +1,387 @@
...
+#[derive(Debug, Clone)]
+pub struct ModelForCausalLM {
+    base: Model,
+    lm_head: Linear,
+}
+
+impl ModelForCausalLM {
+    pub fn new(cfg: &Config, vb: VarBuilder) -> Result<Self> {
+        let base = Model::new(cfg, vb.clone())?;
+        let lm_head = if cfg.tie_word_embeddings {
```
LaurentMazare, line 368: I think this is supposed to be the other way around, if `tie_word_embeddings` is set then we should reuse the `embed_tokens` weights. I'm surprised that testing did not catch this, did you run the model after your changes?

greenrazer: It should have failed testing, but oddly enough it doesn't change anything. Perhaps the trained output LM head learned to mimic the input embeddings, or they just copied it and set `tie_word_embeddings` to false?

### Original (Incorrect)
```
Hello there 2024, I'm a student in the field of computer science. I have to write an essay on my experience with programming and how it has impacted me as a student. Please help me start this essay.
Okay, so I need to write an essay about my experience with programming and its impact on being a student. Let me think... First, what are some common themes in such essays? Maybe starting with an introduction that sets the stage for the topic.

I should probably mention specific experiences or projects where programming was used. For example, maybe coding a program to solve a problem, or using it for a project like a website or app. Also, how did programming help me learn new skills?

Wait, I need to make sure my essay flows well. Maybe start with an introduction that states the purpose of the essay, then move into specific examples of programming experiences and their impact on learning and personal growth.

I should also consider including some reflection at the end, showing how
...
```

### Fixed
```
Hello there 2024, I'm a student in the field of computer science. I have to write an essay on my experience with programming and how it has impacted me as a student. Please help me start this essay.
Okay, so I need to write an essay about my experience with programming and its impact on being a student. Let me think... First, what are some common themes in such essays? Maybe starting with an introduction that sets the stage for the topic.

I should probably mention specific experiences or projects where programming was used. For example, maybe coding a program to solve a problem, or using it for a project like a website or app. Also, how did programming help me learn new skills?

Wait, I need to make sure my essay flows well. Maybe start with an introduction that states the purpose of the essay, then move into specific examples of programming experiences and their impact on learning and personal growth.

I should also consider including some reflection at the end, showing how
...
```
Screenshot 2025-05-01 at 13 38 31

candle-examples/examples/qwen/main.rs
LaurentMazare: Great thanks!

candle-transformers/src/models/qwen3.rs
LaurentMazare: That's odd indeed, looking at the 0.6B it seems to have `tie_word_embeddings` set and to also have the `lm_head.weight` tensor in the safetensors file. I guess you're right and it's just a copy of the embeddings. This would also explain why the model appears with 752m parameters in the huggingface hub UI, the `lm_head.weight` shape is `[151 936, 1 024]` so ~156m additional parameters.

LaurentMazare: Also I guess it would have failed testing on large models where `tie_word_embeddings = false` such as the 32b as it would have used the incorrect output matrix, but I imagine you're running tests mostly with the smaller variants.

LaurentMazare (approved):

PR #2939, opened 2025-05-03
Qwen3 quantized implementation

ljt019:
I was wanting to use the quantized Qwen3 in a project I've been working on, so gave implementing it a shot. I've tested it and it seems to work fine. Let me know if there are any adjustments that need to be made to better fit your guys style! :)

### DIFF
--- a/candle-transformers/src/models/mod.rs
+++ b/candle-transformers/src/models/mod.rs
@@ -90,6 +90,7 @@ pub mod quantized_mpt;
 pub mod quantized_phi;
 pub mod quantized_phi3;
 pub mod quantized_qwen2;
+pub mod quantized_qwen3;
 pub mod quantized_recurrent_gemma;
 pub mod quantized_rwkv_v5;
 pub mod quantized_rwkv_v6;
--- a/candle-transformers/src/models/quantized_phi3.rs
+++ b/candle-transformers/src/models/quantized_phi3.rs
@@ -136,6 +136,9 @@ impl LayerWeights {
         let q = self.apply_rotary_emb(&q, index_pos)?.contiguous()?;
         let k = self.apply_rotary_emb(&k, index_pos)?;
 
+	if index_pos == 0 {
+            self.kv_cache.reset();
+        }
         let (k, v) = self.kv_cache.append(&k.contiguous()?, &v.contiguous()?)?;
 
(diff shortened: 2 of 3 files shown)


ljt019:
oops i noticed that my change to phi3 in my last pr slipped through, I'll adjust that!

ljt019:
Fixed! It just has the quantized qwen3 stuff now :)

AlpineVibrations:
is there an example for running qwen3 quantized?

ljt019:
No but I can add one to this PR, it's the same as the other quantized models

AlpineVibrations:
> No but I can add one to this PR, it's the same as the other quantized models

so using the qwen2 quantized as an example?

ljt019:
More or less, the bulk of the difference is within the quantized_qwen3 and quantized_qwen2 actual implementation files respectively. So the existing example shouldn't need much adjustment. I'll put up an example as part of this Pull Request later tonight! :)

ljt019:
@AlpineVibrations Added an example to candle-examples :)

AlpineVibrations:
> @AlpineVibrations Added an example to candle-examples in the PR :)

Nice work. I tried it and its working. even the 0.6B model does thinking mode and produces some amazing stuff.
Do you know how to enable or disable the thinking mode?

ljt019:
@AlpineVibrations Afaik there isn't stuff related to chat templating in this repo, so you'd have to do that yourself using minjinja or smth like that. 

For an easy solution you can just add '/no_think' (or '/think' to re-enable it) to the system_prompt or any individual prompt and the model won't think for responses in that conversation, if you go read it's model card on hugging face it will tell you more about how that works :)

ljt019:
@greenrazer Hi, I'm sure you're busy working on other things, but if you get the time I'd appreciate it if you could review this pr, I'm happy to make any changes if you find any problems! thanks :)

alvinometric:
Probably @LaurentMazare could review too

AlpineVibrations:
looks like adding /nothink or /think can disable / enable thinking mode

> Probably @LaurentMazare could review too

so far its working for us on RTX 4000 linux and linux CPU even goes fast. and mac cpu is ripper.

ljt019:
Happy to hear it's working well :D

LaurentMazare (approved):

Issue #2876, opened 2025-04-10
Unsupported op_type Resize with ONNX


AlpineVibrations:
Trying to run an image upressing model with ONNX.

getting error
```
Error: unsupported op_type Resize for op NodeProto 
```

```
    let (input_tensor, h, w) = load_image(&args.image, &device, None)?;
    let input_tensor = input_tensor.to_device(&device)?;

    // Load ONNX model
    let model_path = PathBuf::from(&args.model);

    let model = candle_onnx::read_file(model_path)?;
    let graph = model.graph.as_ref().unwrap();


    let mut inputs = std::collections::HashMap::new();
    inputs.insert(graph.input[0].name.to_string(), input_tensor);
    let mut outputs = candle_onnx::simple_eval(&model, inputs)?;

    let output = outputs.remove(&graph.output[0].name).unwrap();
    save_tensor_to_image(output, &args.output)?;

```

greenrazer:
#2876 should fix this

AlpineVibrations:
> [#2876](https://github.com/huggingface/candle/issues/2876) should fix this

it sure does fix it. Great work. Its working doing some 4x-ultrasharp upscale models. thank you

PR #2946, opened 2025-05-06
Add Resize to onnx ops

greenrazer:
> [!Warning]
> Incomplete

From https://github.com/huggingface/candle/issues/2876

### DIFF
--- a/candle-examples/examples/onnx/main.rs
+++ b/candle-examples/examples/onnx/main.rs
@@ -5,8 +5,8 @@ extern crate intel_mkl_src;
 extern crate accelerate_src;
 
 use candle::{IndexOp, D};
+use candle_examples::save_image;
 use clap::{Parser, ValueEnum};
-use candle_examples::{save_image};
 
 #[derive(Clone, Copy, Debug, ValueEnum)]
 enum Which {
@@ -31,7 +31,9 @@ struct Args {
 pub fn main() -> anyhow::Result<()> {
     let args = Args::parse();
     let image = match args.which {
-        Which::SqueezeNet | Which::EfficientNet => candle_examples::imagenet::load_image224(&args.image)?,
+        Which::SqueezeNet | Which::EfficientNet => {
+            candle_examples::imagenet::load_image224(&args.image)?
+        }
         Which::EsrGan => candle_examples::imagenet::load_image(&args.image, 128)?,
     };
     let image = match args.which {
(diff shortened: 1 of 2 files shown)


candle-onnx/src/eval.rs
```diff
@@ -1960,6 +1960,71 @@ fn simple_eval_(
...
+                } else if let Some(scales_tensor) = scales {
+                    // Calculate from scales
+                    let scale_values = scales_tensor.to_vec1::<f32>()?;
+                    input.dims()
+                        .iter()
+                        .enumerate()
+                        .map(|(i, &d)| (d as f32 * scale_values[i]) as usize)
+                        .collect::<Vec<_>>()
+                } else {
+                    unreachable!("Either scales or sizes should be present");
```
LaurentMazare, line 2004: I would use `bail!` rather than unreachable here just in case, we do the check with `is_none()` above but it's possible that the check will be removed through some refactoring. Actually I would even remove the check above so that it's entirely handled here.
Also what about the case where both `scales` and `sizes` are set, should we have an error or is there a reason to choose `sizes` first?

greenrazer: Yeah, It should be an error: according to [onnx documentation](https://onnx.ai/onnx/operators/onnx__Resize.html#inputs):
> One of ‘scales’ and ‘sizes’ MUST be specified and it is an error if both are specified.

candle-examples/examples/onnx/main.rs
```diff
@@ -57,21 +73,42 @@ pub fn main() -> anyhow::Result<()> {
...
+            let out = (prs * max_pixel_val)?.i(0)?.to_dtype(candle::DType::U8)?;
+
+            let pb = std::path::PathBuf::from(args.image);
+            let input_file_name = pb.file_name().unwrap();
+            let mut output_file_name = std::ffi::OsString::from("super_");
+            output_file_name.push(input_file_name);
+            let mut output_path = std::path::PathBuf::from("./");
+            output_path.push(output_file_name);
+
+            save_image(&out, output_path).unwrap();
```
LaurentMazare, line 110: Is there a reason to use `unwrap` here rather than `?`?

candle-examples/examples/onnx/main.rs
```diff
@@ -57,21 +73,42 @@ pub fn main() -> anyhow::Result<()> {
...
+            let max_pixel_val = candle::Tensor::try_from(255.0f32)?
+                .to_device(prs.device())?
+                .broadcast_as(prs.shape())?;
+            let out = (prs * max_pixel_val)?.i(0)?.to_dtype(candle::DType::U8)?;
+
+            let pb = std::path::PathBuf::from(args.image);
+            let input_file_name = pb.file_name().unwrap();
+            let mut output_file_name = std::ffi::OsString::from("super_");
+            output_file_name.push(input_file_name);
+            let mut output_path = std::path::PathBuf::from("./");
```
LaurentMazare, line 107: Why create a path of the form `./foo`, I would have thought that passing `output_filename` directly to `save_image` would do the same thing.

greenrazer: This wasn't working for me before, but now it its 🤷.

LaurentMazare (approved):

Issue #2950, opened 2025-05-11
quantized_qwen3 CUDA_ERROR_OUT_OF_MEMORY with Irrespective of Size


nosnakeob:
I'm encountering a consistent CUDA Out of Memory (OOM) error when running inference with quantized Qwen3 models using `candle-transformers`, regardless of the model size. Even the smallest Qwen3 models trigger this issue on my  `NVIDIA GeForce RTX 2070 (8GB VRAM)`.

```cmd
running 1 test
BaseConfig { sample_len: 1000, temperature: 0.8, top_p: None, seed: 299792458, device: Cuda(CudaDevice(DeviceId(1))), repeat_penalty: 1.1, repeat_last_n: 64, which: W3_0_6b }
Error: DriverError(CUDA_ERROR_OUT_OF_MEMORY, "out of memory")
   0: std::backtrace_rs::backtrace::win64::trace
             at /rustc/2da29dbe8fe23df1c7c4ab1d8740ca3c32b15526/library\std\src\..\..\backtrace\src\backtrace\win64.rs:85
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/2da29dbe8fe23df1c7c4ab1d8740ca3c32b15526/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2: std::backtrace::Backtrace::create
             at /rustc/2da29dbe8fe23df1c7c4ab1d8740ca3c32b15526/library\std\src\backtrace.rs:331
   3: std::backtrace::Backtrace::capture
             at /rustc/2da29dbe8fe23df1c7c4ab1d8740ca3c32b15526/library\std\src\backtrace.rs:296
   4: enum2$<candle_core::error::Error>::bt
             at C:\Users\snake\.cargo\git\checkouts\candle-5b4d092929d18d36\36508a2\candle-core\src\error.rs:255
   5: candle_core::cuda_backend::error::impl$1::w::closure$0<cudarc::driver::safe::core::CudaSlice<f32>,cudarc::driver::result::DriverError>
             at C:\Users\snake\.cargo\git\checkouts\candle-5b4d092929d18d36\36508a2\candle-core\src\cuda_backend\error.rs:60
[...]

PR #2951, opened 2025-05-12
Fixed Quantized Qwen3 Model

nosnakeob:
## Description
This PR optimizes the KV cache implementation in the quantized Qwen3 model by adopting the same approach used in Qwen2. The changes address the OOM issue reported in #2950.

## Changes
- Replaced the `KvCache` type with `Option<(Tensor, Tensor)>` for more efficient memory management
- Removed dependency on `candle_nn::kv_cache::KvCache`
- Simplified cache initialization and reset logic
- Implemented direct tensor concatenation for cache updates

## References
Implementation based on the KV cache approach in quantized Qwen2 model

### DIFF
--- a/candle-transformers/src/models/quantized_qwen3.rs
+++ b/candle-transformers/src/models/quantized_qwen3.rs
@@ -10,7 +10,7 @@ use super::with_tracing::QMatMul;
 use crate::{quantized_nn::RmsNorm, utils::repeat_kv};
 use candle::quantized::{gguf_file, QTensor};
 use candle::{DType, Device, Result, Tensor};
-use candle_nn::{kv_cache::KvCache, Activation, Embedding, Module};
+use candle_nn::{Activation, Embedding, Module};
 use std::io::{Read, Seek};
 use std::sync::Arc;
 
@@ -136,7 +136,7 @@ struct AttentionWeights {
     num_kv_groups: usize,
     head_dim: usize,
     rotary_emb: Arc<RotaryEmbedding>,
-    kv_cache: KvCache,
+    kv_cache: Option<(Tensor, Tensor)>,
     span_attn: tracing::Span,
 }
 
@@ -160,13 +160,6 @@ impl AttentionWeights {
         let q_norm = gg.rms_norm(&format!("{prefix}.attn_q_norm.weight"), rms_norm_eps)?;
         let k_norm = gg.rms_norm(&format!("{prefix}.attn_k_norm.weight"), rms_norm_eps)?;
 
-        let max_position_embeddings = gg
-            .metadata()
-            .get("qwen3.context_length")
-            .and_then(|v| v.to_u32().ok())
-            .unwrap_or(4096) as usize;
-        let kv_cache = KvCache::new(2, max_position_embeddings);
-
         let span_attn = tracing::span!(tracing::Level::TRACE, "attn");
 
         Ok(Self {
@@ -181,7 +174,7 @@ impl AttentionWeights {
             num_kv_groups,
             head_dim,
             rotary_emb,
-            kv_cache,
+            kv_cache: None,
             span_attn,
         })
     }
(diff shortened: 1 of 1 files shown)


AlpineVibrations:
```
this is also effecting the Mac. the 0.6B model Is using 16BG of memory.
and its super slow with Metal and without compared to candle qwen2.5 and Ollama qwen3.
```
I tested your branch on Mac with Metal and without and it's fixed! 
works just about as fast as Ollama. Nice job.

EricLBuehler:
@nosnakeob the OOM comes from [here](https://github.com/huggingface/candle/blob/36508a2c935cbe40584b85cc95b68532dba43b2d/candle-nn/src/kv_cache.rs#L64). The current KvCache code doesn't grow the cache and instead allocates it in one shot. If the context length is large this will balloon memory usage.

This PR implements a growable cache by using `cat`, but if we use `KvCache` + grow the cache every ~512 tokens, we can use the even faster `slice_set` op!

LaurentMazare (approved):

b0r3k:
Hey, I noticed a big performance difference between the two versions in this PR. Running [unsloth/Qwen3-32B-Q4_K_M.gguf](https://huggingface.co/unsloth/Qwen3-32B-GGUF?show_file_info=Qwen3-32B-Q4_K_M.gguf) on RTX 3090, with the initial `Option<(Tensor, Tensor)>` [version](https://github.com/huggingface/candle/pull/2951/commits/217984626f2866261cf18ef81fa4a1f6f32ea78c), I got around **20 tokens/s**, while with the merged [version](https://github.com/huggingface/candle/pull/2951/commits/d0c066cb1edcb046af1d3c048d34e7b358d95ba9) I'm at around **10 tokens/s**. I autoregressively generated around 2000 tokens, so I understand the initial version had the whole cache allocated while the new version had to be grown ~4 times. That brings an overhead of course, but half the speed? Is that expected?

My autoregressive loop for reference (model is prefilled and 1 token is generated before the loop):
```rust
println!("Starting generation loop...");
let time_start = Instant::now();
let initial_context_len = context_token_ids.len(); // Store length before generation loop

// --- Generation Loop ---
// Loop until EOS is generated or max tokens reached
while next_token_id != eos_token_id
{
    context_token_ids.push(next_token_id);

    // --- Incremental Forward Pass ---
    // Process only the *last* token, using the KV cache implicitly managed by the model
    let input_token_tensor = Tensor::new(&[[next_token_id]], &device)?;

    let logits = model_weights
        .forward(
            &input_token_tensor,
            // The position is the index of the *next* token we want to predict
            context_token_ids.len() - 1,
        )?
        .squeeze(0)?; // Remove batch dim, shape: [vocab_size]

    // Sample the *next* token ID using the LogitsProcessor
    next_token_id = sampler.sample(&logits)?;
}

println!("\nGeneration finished.");

let elapsed = time_start.elapsed();
```

LaurentMazare:
Could you try tweaking the 512 value to see how it impacts performance?

b0r3k:
🤔 even when initializing the cache to 4096 ([fork](https://github.com/huggingface/candle/commit/53a072f766769d08cd94b90684eabbff38df974d)) I'm at 10 tokens/s

LaurentMazare:
Ah sorry I meant more using smaller values rather than larger ones.

b0r3k:
Same result with 128.

LaurentMazare:
That's odd, I've just tried on my 4080 super for the 14B with the example code and get roughly the same speed with the current merged code and with the kv-cache being a `Option<(Tensor, Tensor)>`, ~205 tok/s on both sides for a sample len of 256.

LaurentMazare:
Also one thing you may want to try is to make the tensor contiguous before the `repeat_kv` calls (this avoids some strided copies).

```rust
        let k = k.contiguous()?;
        let v = v.contiguous()?;
        let k = repeat_kv(k, self.num_kv_groups)?.contiguous()?;
        let v = repeat_kv(v, self.num_kv_groups)?.contiguous()?;
```

LaurentMazare:
Yeah go ahead, it's a bit sad to have to do this but that's probably reasonable for now.

PR #2952, opened 2025-05-12
candle-onnx: Implement Trilu and ScatterND ops

greenrazer:
Added more onnx ops and the ability to run [SmolLM-135M](https://huggingface.co/HuggingFaceTB/SmolLM-135M) example.

## Changes

- Added [ScatterND](https://github.com/onnx/onnx/blob/main/docs/Operators.md#scatternd) and [Trilu](https://github.com/onnx/onnx/blob/main/docs/Operators.md#trilu) Onnx ops.
- Added Tests for ScatterND and Trilu onnx ops.
- Added new example `onnx-llm`.

### DIFF
--- a/candle-examples/examples/onnx-llm/main.rs
+++ b/candle-examples/examples/onnx-llm/main.rs
@@ -147,13 +147,13 @@ pub fn main() -> Result<()> {
             let position_ids_tensor = Tensor::new(position_ids, &device)?.unsqueeze(0)?;
             inputs.insert("position_ids".to_string(), position_ids_tensor);
 
+            // Create empty key and value tensors
             for i in 0..num_layers {
                 let batch_size = 1;
                 let num_heads = 3;
                 let head_dim = 64;
                 let seq_len = 0;
 
-                // Create empty key and value tensors
                 let empty_key = Tensor::zeros(
                     &[batch_size, num_heads, seq_len, head_dim],
                     DType::F32,
@@ -170,7 +170,6 @@ pub fn main() -> Result<()> {
             }
         }
 
-        // Perform inference
         let outputs = candle_onnx::simple_eval(&model, inputs)?;
 
         let logits = outputs.get("logits").unwrap();
--- a/candle-examples/examples/onnx-llm/main.rs
+++ b/candle-examples/examples/onnx-llm/main.rs
@@ -6,6 +6,7 @@ extern crate accelerate_src;
 
 use anyhow::{Result, bail};
 use candle::{DType, Device, Tensor};
+use candle_transformers::generation::{LogitsProcessor, Sampling};
 use clap::{Parser, ValueEnum};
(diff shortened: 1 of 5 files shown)


candle-examples/examples/onnx-llm/main.rs
```diff
@@ -0,0 +1,209 @@
...
+    top_k: Option<usize>,
+
+    /// The seed to use when generating random samples.
+    #[arg(long, default_value_t = 299792458)]
+    seed: u64,
+}
+
+pub fn main() -> Result<()> {
+    let args = Args::parse();
+    let device = if args.cpu {
```
LaurentMazare, line 61: We tend to use the following in the examples lately so that it can also use metal if available.
```
let device = candle_examples::device(args.cpu)?;
```

candle-examples/examples/onnx-llm/main.rs
```diff
@@ -0,0 +1,209 @@
...
+                (Some(k), None) => Sampling::TopK { k, temperature },
+                (None, Some(p)) => Sampling::TopP { p, temperature },
+                (Some(k), Some(p)) => Sampling::TopKThenTopP { k, p, temperature },
+            }
+        };
+        LogitsProcessor::from_sampling(args.seed, sampling)
+    };
+
+    let mut past_key_values: Option<Vec<(Tensor, Tensor)>> = None;
+    let num_layers = 30;
```
LaurentMazare, line 114: Does this mean that this will only work for a model with 30 layers? Is it possible to make it more generic?

candle-examples/examples/onnx-llm/main.rs
```diff
@@ -0,0 +1,209 @@
...
+            let attention_mask_tensor = Tensor::new(attention_mask, &device)?;
+            inputs.insert("attention_mask".to_string(), attention_mask_tensor);
+
+            let position_ids: Vec<i64> = (0..seq_len as i64).collect();
+            let position_ids_tensor = Tensor::new(position_ids, &device)?.unsqueeze(0)?;
+            inputs.insert("position_ids".to_string(), position_ids_tensor);
+
+            // Create empty key and value tensors
+            for i in 0..num_layers {
+                let batch_size = 1;
```
LaurentMazare, line 152: The parameters below seem to depend on the model itself, any better way to extract them? (same as for `num_layers`)

greenrazer: I was hoping there was a way to extract this and `num_layers` from the onnx model itself, but the best way I could find that didn't involve hard-coding it was to get the hyperparameters from the `config.json` in the model repo.

LaurentMazare: Neat, thanks. I guess the best would be if the people exporting the onnx would include the `config.json` content as metadata in the onnx file but that doesn't seem to be the case here.

PR #2953, opened 2025-05-14
Optimize Quantized Qwen3 KvCache 

b0r3k:
Further optimize Qwen3 KvCache as discussed in PR #2951 by making the tensors contiguous before the `repeat_kv` calls. Two times speedup on autoregressive inference in my tests (rtx3090, qwen3:32b). Thanks @LaurentMazare for coming up with the fix.


LaurentMazare:
Thanks!

Issue #2949, opened 2025-05-10
Using the `--model` parameter with the quantized-qwen3 example fails


azw413:
Using the `--model` parameter with the quantized-qwen3 model results in an error e.g. 
```
cargo run --example quantized-qwen3 --release --features cuda -- --model 4b --prompt "A train is travelling at 120mph, how far does it travel in 3 minutes 30 seconds?"
Finished `release` profile [optimized] target(s) in 0.42s
     Running `target/release/examples/quantized-qwen3 --model 4b --prompt 'A train is travelling at 120mph, how far does it travel in 3 minutes 30 seconds?'`
avx: false, neon: true, simd128: false, f16c: false
temp: 0.80 repeat-penalty: 1.10 repeat-last-n: 64
Error: No such file or directory (os error 2)
``` 

It works fine with no `--model` parameter using the default 0.6b model. 

Running on ARM Linux on an nVidia Jetson Orin AGX.

azw413:
It seems the argument is called `--which 4b` and not `--model` as indicated in the doc.

PR #2955, opened 2025-05-15
Fix docs quantized qwen3

maximizemaxwell:
# What does this PR do?
- Fixed wrong usage in README.md of quantized-qwen3 and quantized-qwen2-instruct

## Part of
Issue #2949 

## Details
Those two models' example works with parameter `which` not `model`.
I added additional example of using different size of model and replaced `model` with `which`

## Who can review?
@LaurentMazare 
Could you review these changes?


LaurentMazare (approved):

LaurentMazare:
Thanks!

Issue #2850, opened 2025-03-31
Unsupported ONNX operator: RNN


BrunoSienkiewicz:
https://onnx.ai/onnx/operators/onnx__RNN.html

PR #2964, opened 2025-05-22
candle-onnx: Implement RNN operator

BrunoSienkiewicz:
Added RNN operator with test. Implementation is mostly based on previous implementation of LSTM operator, which is very similar.
Related issue: https://github.com/huggingface/candle/issues/2850


candle-onnx/src/eval.rs
```diff
@@ -1934,6 +1934,137 @@ fn simple_eval_(
...
+                    None => {
+                        seq_lens_default =
+                            Tensor::full(seq_length as i64, (batch_size,), x.device())?;
+                        &seq_lens_default
+                    }
+                };
+                let seq_lens_is_default =
+                    (seq_lens.to_vec1::<i64>()?.iter()).all(|e| *e as usize == seq_length);
+                if !seq_lens_is_default {
+                    bail!("RNN currently only supports default value of seq_lens");
```
greenrazer: Perhaps change to something like:

bail!("RNN currently does not support variable-length sequences. All sequences must use the full sequence length of {}", seq_length);

greenrazer (commented):
Otherwise looks good!

greenrazer:
Thanks!

PR #2972, opened 2025-05-28
feat: enhance linear algebra operations

### LINKED ISSUE #2971: Enhance the usability of the tensor struct
Hello,

I’m currently learning how to use Candle with the book Dive into Deep Learning, but implementing the code in Candle. I noticed that Candle is missing some practical utility functions, such as:

* The Frobenius norm 
* dot product (vector or matrix dot product)
* matrix-vector multiplication
While these functions aren’t overly complex to implement manually, having them natively supported by the Tensor struct would significantly improve usability.

I’ve tried adding some of these functions myself to extend Candle’s functionality (to make it more user-friendly).

ssfdust:
- Add `dot()` for vector/matrix products
- Implement the `Frobenius` norm
- Add `mv()` for matrix-vector multiply

Fix #2971

### DIFF
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -1235,6 +1235,83 @@ impl Tensor {
         Ok(from_storage(storage, (n, c, h_out, w_out), op, false))
     }
 
+    /// Computes the dot product of two 1D tensors.
+    ///
+    /// - If inputs are 1D vectors (`[n]`), returns their scalar dot product.
+    /// - Panics if shapes are not compatible
+    /// - Not supported for integer dtypes
+    ///
+    /// # Example (vectors)
+    /// ```rust
+    /// use candle_core::{Tensor, Device};
+    /// let t1 = Tensor::new(&[1.0, 2.0, 3.0], &Device::Cpu)?;
+    /// let t2 = Tensor::new(&[4.0, 5.0, 6.0], &Device::Cpu)?;
+    /// let res = t1.dot(&t2)?;
+    /// assert_eq!(res.to_scalar::<f64>()?, 32.);
+    /// # Ok::<(), candle_core::Error>(())
+    /// ```
+    pub fn dot(&self, rhs: &Self) -> Result<Self> {
+        if self.dims().len() != 1 || rhs.dims().len() != 1 {
+            return Err(Error::ShapeMismatchBinaryOp {
+                lhs: self.shape().clone(),
+                rhs: rhs.shape().clone(),
+                op: "dot",
+            });
+        }
+
+        (self * rhs).and_then(|ret| ret.sum_all())
+    }
+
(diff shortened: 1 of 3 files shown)


AlpineVibrations:
cool. i think more simple tools like this could help candle adoption.

ssfdust:
@LaurentMazare could you please have a review on this, thanks a lot.

EricLBuehler (changes requested):
Looks good; however for new ops could you please add some more [tests](https://github.com/EricLBuehler/candle/tree/main/candle-core/tests)?

candle-core/src/tensor.rs
```diff
@@ -1235,6 +1235,74 @@ impl Tensor {
         Ok(from_storage(storage, (n, c, h_out, w_out), op, false))
     }
 
+    /// Computes the dot product between two tensors
```
LaurentMazare, line 1238: The pytorch version only supports 1d tensor and mention doing this intentiionally so I would suggest doing the same here as we try to be as close as possible as their api.
https://docs.pytorch.org/docs/stable/generated/torch.dot.html

candle-core/src/tensor.rs
```diff
@@ -1235,6 +1235,74 @@ impl Tensor {
...
+                lhs: self.shape().clone(),
+                rhs: rhs.shape().clone(),
+                op: "dot",
+            });
+        }
+        (self * rhs)
+            .and_then(|ret| ret.sum_all())
+    }
+
+    /// Computes the **Frobenius norm** (L2 norm of all elements) of the tensor.
```
LaurentMazare, line 1263: Returning an int result on an int tensor seems pretty error prone as the output will be rounded + it's easy to do computations that would result in overflowing on u8 or even u32. I would suggest also being in line with the pytorch behavior here so bail on integer type tensors.

candle-core/src/tensor.rs
ssfdust: Ok.

candle-core/src/tensor.rs
```diff
@@ -1235,6 +1235,87 @@ impl Tensor {
...
+    /// ```rust
+    /// use candle_core::{Tensor, Device};
+    /// let t = Tensor::new(&[[3., 4.], [0., 0.]], &Device::Cpu)?;
+    /// let norm = t.norm()?;
+    /// assert_eq!(norm.to_scalar::<f64>()?, 5.);
+    /// # Ok::<(), candle_core::Error>(())
+    /// ```
+    pub fn norm(&self) -> Result<Self> {
+        if self.dtype().is_int() {
+            bail!("dot not supported for integer dtypes");
```
LaurentMazare, line 1280: This op is `norm` and not `dot`.

ssfdust: sorry.

LaurentMazare (approved):

PR #2973, opened 2025-05-29
Fused CPU attention kernels (~4x performance increase)

EricLBuehler:
This introduces fused CPU attention kernels for optimized CPU inference. This removes the necessity to materialize the attention matrices, thereby dramatically improving throughput.

On an M3 Max with Llama 3.2 3b at 4-bit quantization, I am measuring a **_4x_ increase** in decode T/s. This is faster than llama.cpp, even with llama.cpp CPU FlashAttention enabled.

These kernels are loosely based on the work in FlashAttention and CPU implementations in vLLM and ggml, but have been modified for higher performance.
<details>
<summary>Algorithm</summary>

## `run_flash_attn_cpu`

1. **Choose execution path**  
   - **Decode path**: if the query length `S_q == 1`, invoke a specialized “single-Q” routine  
   - **Batched path**: otherwise, invoke the general batched attention routine

2. **Compute attention**  
   - **Parallel setup**  
     - Uses a **custom Rayon thread-pool** (`FLASH_ATTN_POOL`) with macOS QoS hints  
     - Installs the pool via `FLASH_ATTN_POOL.install(...)` to isolate flash-attention tasks  
   - **Work distribution**  
     - **Batched**: flattens the output into chunks of size `D` and calls  
       ```rust
       out.par_chunks_mut(dv)
          .with_min_len(64)
          .enumerate()
          .for_each(...)
       ```  
       to assign each `(batch, head, query_pos)` row to a Rayon worker  
     - **Decode**: further splits the KV axis into cache-friendly tiles, then does  
       ```rust
       (0..kv_tiles)
         .into_par_iter()
         .map(...)     // per-tile map
         .reduce(...)  // numerically-stable softmax reduce
       ```  
       achieving **nested parallelism** for long KV sequences
   - **Per-row computation**  
     1. Gather the query vector  
     2. Loop over all key/value positions:  
        - Apply mask and positional bias  
        - Compute dot-product between query and key  
        - Update an **online softmax** (log-sum-exp) in a streaming fashion  
        - Weight and accumulate the value vectors  
     3. Normalize the accumulated value sum by the softmax denominator

3. **Assemble result**  
   - Collect all per-row outputs into a flat buffer  
   - Reshape into the final tensor of shape `(B, S_q, H, D)`  
   - Return the result on the CPU device  

</details>

### DIFF
--- a/candle-nn/Cargo.toml
+++ b/candle-nn/Cargo.toml
@@ -21,6 +21,7 @@ safetensors = { workspace = true }
 serde = { workspace = true }
 metal = { workspace = true, optional = true }
 candle-metal-kernels = { workspace = true, optional = true }
+libc = { workspace = true }
 
 [dev-dependencies]
 anyhow = { workspace = true }
--- a/candle-nn/src/cpu_flash_attention.rs
+++ b/candle-nn/src/cpu_flash_attention.rs
@@ -59,6 +59,19 @@ fn vec_dot<T: WithDType + Sum + Copy + std::ops::Mul<Output = T>>(a: &[T], b: &[
     sum
 }
 
+/// Fused attention optimized for CPU.
+///
+/// Computes softmax(qk^T*scale)v.
+///
+/// **Inputs shapes:**
+/// - `q`: (bs, seq, qhead, hidden)
+/// - `k`: (bs, kv_seq, v_head, hidden)
+/// - `k`: (bs, kv_seq, kv_head_seq, v_hidden)
+/// - `scale` is applied before softmax.
+/// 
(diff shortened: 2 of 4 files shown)


EricLBuehler:
@LaurentMazare could you please review this PR?

EricLBuehler:
This PR doesnt integrate it into any models yet; but that would be relatively easy.

Once that is done, **yes**. I saw ~4x T/s increase for CPU inference.

AlpineVibrations:
how's this looking. ? is it done? would be great to merge in if is done. and if its not done what is still needed on it?

greenrazer:
Closing + Reopening to rerun workflows

PR #2980, opened 2025-06-01
candle-onnx: Implement Hard Swish operator

Michall00:
Added Hard Swish operator with tests.


greenrazer:
Looks good! Just rebase it to main, and rustfmt.

greenrazer:
Thanks!

PR #2987, opened 2025-06-07
Fix cuda memory error for Qwen3 non-quantized

akshayballal95:
Instead of initializing the `KvCache` with the full position embedding length, initialize it with 512, similar to what is done in the quantized version [here](https://github.com/huggingface/candle/blob/0224a749f0b2082f19831256ced6afe284c56457/candle-transformers/src/models/quantized_qwen3.rs#L163-L165) This resolves large memory allocation issues.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #2994, opened 2025-06-14
[DRAFT] Add `candle-utils` helper crate for common patterns

matthewhaynesonline:
## Overview

This draft PR proposes adding a new `candle-utils` crate that provides higher-level abstractions for common patterns found throughout the Candle examples. The goal is to reduce boilerplate for common use cases.

## Motivation

While working with Candle, I noticed several patterns that are common across many examples (and throughout the ecosystem - Mistral.rs, Text Embedding Inference, etc.). For example:

1. **Device selection logic** - CPU/CUDA/Metal detection with fallbacks
2. **Model loading workflows** - Tokenizer + config + weights loading from HF Hub
3. **Tokenization helpers** - Batch encoding and attention mask creation
4. **Common tensor operations** - L2 normalization and similar utilities

## Questions

1. **Interest level**: Are you open to this kind of higher-level utility crate in Candle?
2. **Scope**: What utilities would be most valuable? I've focused on some of the patterns I've come across, but there may be others.
3. **API design**: Any preferences on the module organization or function signature (builder pattern)? I'm happy to adjust based on your feedback / conventions / requirements.
4. **Integration**: Should this be:
   - Another crate in the Candle project (current draft)?
   - Integrated into an existing crate like `candle-core`, `candle-transformers`, etc.?
   - A separate `candle-utils` third party crate?

## Current Status

This is a very minimal viable implementation just to demonstrate the concept. If there's interest, I'm happy to iterate on this based on your feedback and requirements and keep working on this (tests, builders, more utils, check the examples for compatibility, etc.).

If this direction isn't aligned with your vision for Candle, no worries at all - I just wanted to check before investing more time in the implementation.

Thanks for considering this!

### DIFF
--- a/candle-utils/src/device.rs
+++ b/candle-utils/src/device.rs
@@ -11,7 +11,7 @@ use candle::{Device, Result};
 ///
 /// # Example  
 /// ```
-/// let device = candle_utils::device::get(true, false).unwrap();
+/// let device = candle_utils::device::get_device(true, false).unwrap();
 /// ```
 pub fn get_device(use_cpu: bool, quiet: bool) -> Result<Device> {
     if use_cpu {


AlpineVibrations:
this sorta thing is great. i wonder if there is enough people with merge authority to handle this inside of the candle repo...

matthewhaynesonline:
Just an quick note: I'm working on some projects that could benefit from the helpers, so I've added them as a [separate repo](https://github.com/matthewhaynesonline/candle-utils) so I can use the helpers without having to actively maintain a fork of candle (i.e. standard HF candle core + my helper crate instead of my forked candle)

If I end up coding more helpers, I'll try to keep both this PR and the other helper repo in sync.

PR #3005, opened 2025-06-26
Qwen3: fix quality loss due to rope freq precision

zackangelo:
When running Qwen3 at bf16, I noticed a substantial loss in quality and bugs (skipping tokens) especially when dealing with multi-piece tokens/multi-byte characters. Fixed by doing all intermediate rope freq and sin/cos computation in F32, then casting down to the requested dtype. 

Also changed the candle example to run in bf16 by default on metal.


greenrazer:
Thanks!

Issue #3011, opened 2025-06-28
LayerNorm Gradient Flow Issue in candle-nn


tymat:
# LayerNorm Gradient Flow Issue in candle-nn

## Summary

LayerNorm in candle-nn does not properly propagate gradients through all parameters during backpropagation, causing only 33% of parameters to receive gradients. This prevents models from training correctly.

## Environment

- **candle-core version**: 0.8.0
- **candle-nn version**: 0.8.0
- **Platform**: macOS with Metal backend (also affects CUDA)
- **Rust version**: 1.75+

## Minimal Reproducible Example

```rust
use candle_core::{DType, Device, Result, Tensor};
use candle_nn::{Module, VarBuilder, VarMap, Optimizer};

#[test]
fn test_candle_layernorm_gradient_flow() -> Result<()> {
    // Setup
    let device = Device::Cpu; // Also fails on Metal/CUDA
    let varmap = VarMap::new();
    let vb = VarBuilder::from_varmap(&varmap, DType::F32, &device);
    
    // Create a simple model: Linear -> LayerNorm -> Linear
    let hidden_size = 64;
    let batch_size = 4;
    
    // Build model
    let linear1 = candle_nn::linear(hidden_size, hidden_size, vb.pp("linear1"))?;
    let layer_norm = candle_nn::layer_norm(
        hidden_size, 
        candle_nn::LayerNormConfig::default(), 
        vb.pp("layer_norm")
    )?;
    let linear2 = candle_nn::linear(hidden_size, hidden_size, vb.pp("linear2"))?;
    
    // Create input and target
    let input = Tensor::randn(0f32, 1.0, (batch_size, hidden_size), &device)?;
    let target = Tensor::randn(0f32, 1.0, (batch_size, hidden_size), &device)?;
    
    // Forward pass
    let x1 = linear1.forward(&input)?;
    let x_norm = layer_norm.forward(&x1)?;
    let output = linear2.forward(&x_norm)?;
    
    // Compute loss (MSE)
    let loss = (output.sub(&target))?.sqr()?.mean_all()?;
    
    // Backward pass
    let grads = loss.backward()?;
    
    // Check gradient flow
    let vars = varmap.all_vars();
    let mut params_with_gradients = 0;
    let mut params_without_gradients = 0;
    
    for var in &vars {
        if let Some(grad) = grads.get(var) {
            let grad_norm = grad.sqr()?.sum_all()?.sqrt()?.to_scalar::<f32>()?;
[...]

PR #3012, opened 2025-06-28
Fix LayerNorm gradient flow issue

tymat:
- Fix LayerNorm.forward() to use tensor operations instead of scalar operations
- Replace sum_keepdim()/size with mean_keepdim() to preserve gradients
- Use broadcast_add() with epsilon tensor instead of scalar addition
- Fix ops::layer_norm_slow() with same gradient-preserving changes
- Update ops::layer_norm() to use slow implementation for proper gradients
- Add comprehensive gradient flow test (now passes with 100% gradient flow)
- Add numerical equivalence test to ensure accuracy is maintained
- Fixes training issues where LayerNorm parameters weren't being updated

Resolves gradient propagation bug where only 33% of parameters received gradients during backpropagation, preventing proper model training. #3011

### DIFF
--- a/candle-nn/src/layer_norm.rs
+++ b/candle-nn/src/layer_norm.rs
@@ -117,16 +117,25 @@ impl Module for LayerNorm {
             DType::F16 | DType::BF16 => DType::F32,
             d => d,
         };
-        let hidden_size = x.dim(D::Minus1)?;
         let x = x.to_dtype(internal_dtype)?;
         let x = if self.remove_mean {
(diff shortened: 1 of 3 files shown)


AlpineVibrations:
great. it would be awesome to have more training code examples and workflows with candle

ivarflakstad:
Hey! Thanks for this :)

I think we'll have to implement this in the optimized kernels as well before we can merge.
I assume all the variants (cpu, cuda, metal) suffer from the same issue?

PR #3021, opened 2025-07-15
candle-datasets: add fashion-mnist

slckl:
This PR adds support for Fashion MNIST dataset for validating (tiny) vision models. It's drop-in compatible with MNIST so reuses existing MNIST dataset code.

Fashion MNIST is a slightly more difficult drop-in replacement for MNIST that can't be solved to 97% with two matchsticks and a napkin blah blah.


greenrazer:
Thanks!

Issue #3028, opened 2025-07-19
Feaute Request: Voxtral


jorge-menjivar:
Are there any plans to add support for the new Voxtral models?

Thanks!

maximizemaxwell:
I am currently implementing voxtral.rs.

jorge-menjivar:
@maximizemaxwell Awesome! Thank you! Lmk if there is anything easy I can help with.

jorge-menjivar:
@maximizemaxwell I was eager to try voxtral on rust and I got things working with your fork up to the tokenizer point. At that point, I am stuck because of the tekken format. Do you have anything in mind to process the tekken file? I don't see any simple solutions to read the tekken file in rust.

jorge-menjivar:
Got it fully working on my side and cleaning things up now, but it required lots of changes to the existing codebase. I can make a pr but would like some feedback on code before I make a bad pr. Anyone wants to check it out?

maximizemaxwell:
Maybe you may take reference my PR. (like in case of qwen3)
Or I could check if you'd like to make new one.

jorge-menjivar:
Let's continue in your PR

PR #3030, opened 2025-07-22
Feature Request: voxtral implementation

maximizemaxwell:
# What does this PR do?
Implemented voxtral and examples in candle.

## Issue
Part of #3028

## Requirements
- [ ] Need fixes to fully run the code


jorge-menjivar:
## Proposal
The Candle code should replicate the output of the [Transformers](https://github.com/huggingface/transformers/pull/39429) implementation of Voxtral, word by word, when run with deterministic values. My current test consists of three different audio files. 

## Challenges
I have encountered the following challenges while working on the current implementation:

### Tekken
The Tekken tokenizer only seems to be officially supported in Python so far.

My current solution has been to reimplement the tokenizer in Rust completely to remove the Python deps. I hope that we can find a better solution for this; otherwise, I will publish my code as its crate.

### Mel Spectrogram
The current implementation of Whisper's `pcm_to_mel()` in Candle does not seem to precisely match Python's WhisperFeatureExtractor, causing it to output different values.

My current solution is to rewrite the whole `pcm_to_mel` function externally (in the example) to better align with Python's. This is where I am struggling to understand what's going on, since Whisper is working fine in Candle.

eugenehp:
Thank you for the good work! I'm curious if you had a chance to work further on this model, and if you have some drafts in the Tekken and PCM to MEL, or insights on what you've seen so far on the reasons why the output diverges between the Python and Rust implementations. 

Thanks again and happy holidays!

jorge-menjivar:
@eugenehp Yes! Support for voxtral ended up being merged via #3036 and should produce the exact same output as python. I ended up creating the [tekken-rs](https://github.com/jorge-menjivar/tekken-rs) crate to support tekken in rust.

PR #3035, opened 2025-07-26
fix kv cache capacity bug

jhqxxx:
kv cache max_seq_len should max than  self.current_seq_len + input seq_len


ivarflakstad (approved):
Lgtm!
Slightly increased chance of OOM, but only because that's an inherent quality of KV cache

greenrazer:
Thanks!

PR #3036, opened 2025-07-27
[FEAT] Voxtral Support

### LINKED ISSUE #3028: Feaute Request: Voxtral

jorge-menjivar:
- Adds support for `mistralai/Voxtral-Mini-3B-2507`.
- Should also support `mistralai/Voxtral-Small-24B-2507`, but this is not tested.

These models only have support for tekken tokenizer files right now, so I created and added an optional tekken tokenizer crate to the examples workspace. (tekken-rs). This gets enabled with the `tekken` feature.

fixes #3028

Run example with
```sh
cargo run --example voxtral --features tekken,symphonia,cuda,cudnn --release
```


jorge-menjivar:
@maximizemaxwell Thank you for your earlier work! Here is the working code if you want to try it.

maximizemaxwell:
@greenrazer Could you review this feature implementations?

benedikt-schaber:
Hey, thank you for the implementation. I had a rough one myself but was missing the tekken part among some other flaws. I went through the code and I think I might have some pointers for improvement. I am new to audio transformers, however, so please have some leniency with any mistakes :)

For the example:
- unused code: the example has some unused code, it loads the mel filters when creating the model, but does not use them. Further, the reimplementation of  pcm_to_filer in audo_processing.rs is unused
- `audo_utils.rs` only contains a resample function, `pcm_decode` is its own module, so I think this could be named more specifically
- seperating `transcribe_audio_with_tokens` from `transcribe_audio_internal` seems unecessary
- `post_process_transcription` seems very specific, especially with the word combinators, is this necessary / should this be generalized? I only did a small number of tests (around 100 audio files of 30secs each) without the post processing but I also did not encounter the issues there.

For the main code:
- `voxtral_llama.rs` might be mergable in `llama.rs`, it only adds the explicit `head_dim` option, voxtral defaults and "fixes" dtype consistency, single sequence support

candle-transformers/src/models/voxtral/model.rs
```diff
@@ -0,0 +1,1059 @@
...
+            }
+        }
+
+        // Forward through language model using forward_input_embed
+        self.language_model
+            .forward_input_embed(&inputs_embeds, index_pos, &mut cache.cache)
+    }
+
+    /// Generate text given audio input
+    pub fn generate(
```
greenrazer, line 873: Clippy is upset because this function has too many arguments. Consider moving device and cache as fields of the struct.

candle-examples/examples/voxtral/pcm_decode.rs
```diff
@@ -0,0 +1,74 @@
...
+
+fn conv<T>(samples: &mut Vec<f32>, data: std::borrow::Cow<symphonia::core::audio::AudioBuffer<T>>)
+where
+    T: symphonia::core::sample::Sample,
+    f32: symphonia::core::conv::FromSample<T>,
+{
+    samples.extend(data.chan(0).iter().map(|v| f32::from_sample(*v)))
+}
+
+pub(crate) fn pcm_decode<P: AsRef<std::path::Path>>(path: P) -> anyhow::Result<(Vec<f32>, u32)> {
```
greenrazer: Would it be possible to use [Whisper's pcm_decode function](https://github.com/huggingface/candle/blob/42bd33e0a6c114838e623e5e11fb466628888bbf/candle-examples/examples/whisper/pcm_decode.rs#L13) here?

candle-examples/examples/voxtral/audio_utils.rs
```diff
@@ -0,0 +1,39 @@
+/// Simple linear interpolation resampling
+/// For production use, consider using a proper resampling library
+#[must_use]
+#[allow(
+    clippy::cast_possible_truncation,
+    clippy::cast_precision_loss,
+    clippy::cast_sign_loss
+)]
+pub fn resample_audio(audio: &[f32], from_rate: u32, to_rate: u32) -> Vec<f32> {
```
greenrazer: It seems snac also has a [resample audio function](https://github.com/huggingface/candle/blob/42bd33e0a6c114838e623e5e11fb466628888bbf/candle-examples/examples/snac/audio_io.rs#L248). If compatible, could you use the existing one instead?

greenrazer (changes requested):
Thank you for the contribution, great job!

Just fix the clippy errors and we should be good:

```
cargo clippy --workspace --tests --examples --fix -- -D warnings
```

jorge-menjivar:
Thank you @greenrazer and @benedikt-schaber for the suggestions.

I have made the requested changes.

I have also tested whisper's and snac's examples to make sure they run correctly after moving the `pcm_decode` and `resample` functions to the shared module.

greenrazer (approved):
Looks good, Thanks!

Issue #2962, opened 2025-05-21
Surprising lack of gradient clipping


grimtin10:
I'm noticing a _surprising_ lack of gradient clipping, as it is usually a core feature.
What's the reason for its lack of existence, and how could I add it myself?

Given the wealth of functions on tensors it wouldn't be that hard to add, I'm just not sure where it would go.

tsubakisakura:
I'm training a model with somewhat unstable gradients, and I've observed occasional gradient explosions (e.g., norm > 10^6). I implemented the following `clip_gradients_global` function to handle this.

~~~
fn clip_gradients_global(grad_store: &mut GradStore, var_map: &nn::VarMap, max_norm: f32) -> TrainResult<f32> {
    let mut total_norm_sq: f32 = 0.0;
    let mut grads = vec![];

    for var in var_map.all_vars() {
        let tensor = var.as_tensor();
        if let Some(grad) = grad_store.get(tensor) {
            let norm_sq = grad.sqr()?.sum_all()?.to_scalar::<f32>()?;
            total_norm_sq += norm_sq;
            grads.push((var, grad.clone()));
        }
    }

    let total_norm = total_norm_sq.sqrt();
    if total_norm > max_norm {
        let scale = max_norm / total_norm;
        for (var, grad) in grads {
            let tensor = var.as_tensor();
            let scale_t = Tensor::new(scale, &tensor.device())?;
            let clipped = grad.broadcast_mul(&scale_t)?;
            grad_store.insert(tensor, clipped);
        }
    }

    Ok(total_norm)
}
~~~

- In early epochs (0–10), total_norm was mostly under 2.0.
- After epoch 10, I started to observe spikes such as:
    - total_norm: 446137.56
    - total_norm: 11965370.0
- Despite this, the model continued to train stably thanks to the clipping.

This highlights the importance of having a `GradStore::insert_id(id, grad)` method for manipulating gradients without needing access to original `Tensor` instances.

PR #3038, opened 2025-07-29
Add GradStore::insert_id for inserting gradients by tensor ID

EthanAlmloff:
This PR adds a simple insert_id method to GradStore, allowing insertion of gradients directly by TensorId. This is motivated by #2962 on supporting gradient clipping. Without this function, it is impossible to clip gradients without access to the original Tensor instances.


ivarflakstad (approved):
👍

Issue #2140, opened 2024-04-29
"Illegal Instruction" on Older CPUS


jett06:
Hey there! I'm reaching out from a machine running the Intel i3-2120 processor.

The `llm-rs` library (https://github.com/rustformers/llm) can handle inference just fine on this processor (previously failed on older CPUs, works fine nowadays), but `candle` fails. I tried:
```sh
$ cargo run --example quantized-phi --release
    Finished release [optimized] target(s) in 3.59s
     Running `/home/jett/devel/candle/target/release/examples/quantized-phi`
avx: true, neon: false, simd128: false, f16c: false
temp: 0.80 repeat-penalty: 1.10 repeat-last-n: 64
Running on CPU, to run on GPU, build this example with `--features cuda`
loaded 325 tensors (1.79GB) in 0.16s
model built
[1]    13404 illegal hardware instruction  cargo run --example quantized-phi --release
```

You can see the failure output. An ML/AI library throwing "Illegal instruction" usually means your program needs features that your CPU doesn't offer, so I tried recompiling without AVX2 instructions enabled (https://github.com/huggingface/candle/issues/622#issuecomment-1694726330), but it still failed:
```sh
$ RUSTFLAGS="-Ctarget-cpu=native -Ctarget-feature=-avx2" cargo run --example quantized-phi
    Finished dev [unoptimized + debuginfo] target(s) in 9.35s
     Running `/home/jett/devel/candle/target/debug/examples/quantized-phi`
avx: true, neon: false, simd128: false, f16c: false
temp: 0.80 repeat-penalty: 1.10 repeat-last-n: 64
Running on CPU, to run on GPU, build this example with `--features cuda`
loaded 325 tensors (1.79GB) in 0.21s
model built
[1]    18202 illegal hardware instruction  RUSTFLAGS="-Ctarget-cpu=native -Ctarget-feature=-avx2" cargo run --example 
```

Any ideas? I realize it might be a code-specific issue with the way `candle` is written, but I'm hoping it would have a simple fix (with something like a feature flag to check for CPU feature presence that enables certain code so this doesn't throw "Illegal instruction"). Thanks in advance for any reply, and thanks to all the maintainers for maintaining such an awesome ML/AI library!

**TL;DR:** *Candle uses unsupported CPU features in its quantized LLM inference that is encountered when testing the prewritten examples. This causes execution to fail on older CPUs, such as the Intel i3-2120.*

LaurentMazare:
[...]

Issue #1327, opened 2023-11-13
Old CPU, trouble with FMA/gemm (and a workaround)


misutoneko:
Hi, and thank you for this project. Amazing stuff!

I had a lot of trouble getting _anything_ going, so here's my notes.
Hopefully it's useful to somebody else along the GPU-poor road.
Note that this is only a workaround, not a real fix to the underlying issue.
Using Linux Mint 21.2 and rustc 1.73.0.

So as I said, not using GPU at all, and my CPU is an old AMD FX (with AVX/FMA/FMA4/SSE4.2).
Not great I know, but I can run 7B stuff with llama.cpp, so I should be able to do something with candle too, right?
First of all, this CPU doesn't have AVX2, so I had to use **RUSTFLAGS='-C target-feature=-avx2'** to avoid the dreaded Illegal instruction.
(I knew this because I had sone a similar thing with llama.cpp.)

But even then, the simple example in the "get started" section wouldn't work for me. It just exited with the message

> called `Option::unwrap()` on a `None` value

I noticed that in the backtrace there's a reference to gemm_f32::gemm::f32::fma::gemm_basic and the file gemm-common-0.16.14/src/gemm.rs.
So what I had to do was to **comment out all the references to fma** in that file, and that seems to have done the trick :D
I guess the gemm library is incomplete somehow?

Anyways, the end result is that I can now run madlad400-3b-mt, which was my goal.
Maybe not the speediest thing in the world, but I'm happy that it now works :D

greenrazer:
Should be fixed by #3040. Please reopen if this is still an issue.

PR #3040, opened 2025-07-31
Support building on CPUs with AVX but not AVX2

jncraton:
This change corrects an issue causing AVX2 intrinsics to be called on CPUs that do not support them. These intrinsics were conditionally compiled in behind an `avx` feature gate. This change correctly uses them only if `avx2` is available. This change should have no impact on modern CPUs, but it allows older CPUs to work properly using the unoptimized code path.

An alternative to this change would be to rework the AVX code paths to avoid AVX2 intrinsics, but that would be a larger change and likely less desirable as AVX2 seems like a very reasonable baseline these days. Dropping AVX2 intrinsics would likely be a performance regression on modern CPUs.

This change corrects an issue that I was facing related to an `Illegal Instruction` after building the crate, and should also address similar issues that have been encountered by others such as #1327 and #2140.

### DIFF
--- a/candle-core/src/cpu/mod.rs
+++ b/candle-core/src/cpu/mod.rs
@@ -60,10 +60,10 @@ trait CpuBF16<const ARR: usize> {
 use half::{bf16, f16};
 
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-#[cfg(target_feature = "avx")]
+#[cfg(target_feature = "avx2")]
 pub mod avx;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-#[cfg(target_feature = "avx")]
+#[cfg(target_feature = "avx2")]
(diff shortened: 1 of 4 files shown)


greenrazer:
Thanks for the contribution!

According to the [Steam Hardware Survey](https://store.steampowered.com/hwsurvey), ~97% of users have AVX and ~95% have AVX2 support. Given this data, I agree that reworking the code to avoid AVX2 instructions isn't worth the effort just to provide optimizations for the ~2% of users on older hardware.

For anyone looking at this in the future, at least one of the AVX2 instructions is found [here](https://github.com/huggingface/candle/blob/42bd33e0a6c114838e623e5e11fb466628888bbf/candle-core/src/quantized/avx.rs#L472): `_mm256_cvtepu8_epi16`, which [according to Intel's documentation](https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/mm256-cvtepu8-epi16-32-64.html) is AVX2-exclusive.

PR #3042, opened 2025-08-01
Load safetensors with i8

chadvoegele:
Hi,

I have a small ML-enabled [app](https://www.voegele.me/posts/suppalist/) that uses candle with wasm under the hood. Downloading the model takes a long time so I quantized it to 8-bit integers. Candle doesn't have an I8 `DType` and the safetensors load doesn't support I8.

This PR adds the ability to load a safetensors file with I8 by upcasting it to I64, similar to the existing I32 to I64 conversion.

Based on this [PR](https://github.com/huggingface/candle/pull/2432) for adding a `DType::I32`, I skipped adding a `DType::I8` for now, but would consider that if requested.

Thanks!

### DIFF
--- a/candle-core/src/safetensors.rs
+++ b/candle-core/src/safetensors.rs
@@ -209,6 +209,10 @@ impl Tensor {
 
 fn convert(view: &st::TensorView<'_>, device: &Device) -> Result<Tensor> {
     match view.dtype() {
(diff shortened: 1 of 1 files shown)


greenrazer:
Thanks!

PR #3045, opened 2025-08-04
pick seed <= u32::MAX when using metal

rgbkrk:
## Problem

Running the stable diffusion example with Metal results in an error with the seed (possibly) being out of range.

```
$ cargo run --example stable-diffusion --release --features=metal -- --prompt "a cosmonaut on a horse (hd, realistic, high-def)" --sd-version turbo
   Compiling candle-transformers v0.9.1 (/Users/kylekelley/code/src/github.com/huggingface/candle/candle-transformers)
   Compiling candle-examples v0.9.1 (/Users/kylekelley/code/src/github.com/huggingface/candle/candle-examples)
    Finished `release` profile [optimized] target(s) in 13.34s
     Running `target/release/examples/stable-diffusion --prompt 'a cosmonaut on a horse (hd, realistic, high-def)' --sd-version turbo`
Using seed 14094623835642992148
Error: Metal error Metal seed must be less than or equal to u32::MAX

Caused by:
    Metal seed must be less than or equal to u32::MAX
```

The user can manually set `--seed` but we can make this easier.

## Solution

Generate the seed up to `u32::MAX` when metal is enabled.

### DIFF
--- a/candle-examples/examples/stable-diffusion/main.rs
+++ b/candle-examples/examples/stable-diffusion/main.rs
@@ -617,7 +617,18 @@ fn run(args: Args) -> Result<()> {
     let mut scheduler = sd_config.build_scheduler(n_steps)?;
     let device = candle_examples::device(cpu)?;
     // If a seed is not given, generate a random seed and print it
-    let seed = seed.unwrap_or(rand::rng().random_range(0u64..u64::MAX));
+    let seed = seed.unwrap_or_else(|| {
+        #[cfg(feature = "metal")]
+        {
+            // Metal backend requires seed to be within u32 range
+            rand::rng().random_range(0u64..u32::MAX as u64)
+        }
+        #[cfg(not(feature = "metal"))]
+        {
+            rand::rng().random_range(0u64..u64::MAX)
+        }
+    });
+
     println!("Using seed {seed}");
     device.set_seed(seed)?;
     let use_guide_scale = guidance_scale > 1.0;


candle-examples/examples/stable-diffusion/main.rs
```diff
@@ -617,7 +617,18 @@ fn run(args: Args) -> Result<()> {
     let mut scheduler = sd_config.build_scheduler(n_steps)?;
     let device = candle_examples::device(cpu)?;
     // If a seed is not given, generate a random seed and print it
-    let seed = seed.unwrap_or(rand::rng().random_range(0u64..u64::MAX));
+    let seed = seed.unwrap_or_else(|| {
+        #[cfg(feature = "metal")]
```
greenrazer, line 621: I think [cfg_if](https://docs.rs/cfg-if/latest/cfg_if/) would be slightly more readable here.

greenrazer (changes requested):
Thanks for the contribution!

rgbkrk:
Hi @greenrazer! I've addressed your feedback by refactoring the code to use `cfg_if` instead of nested `#[cfg]` attributes. The changes include:

1. Added `cfg-if = "1.0"` to the workspace dependencies in `Cargo.toml`
2. Added `cfg-if` as a dependency in `candle-examples/Cargo.toml`
3. Refactored the seed generation logic in `main.rs` to use the `cfg_if!` macro for better readability

The code now uses `cfg_if!` which is indeed more readable and idiomatic for conditional compilation. Thanks for the suggestion!

rgbkrk:
Addressed review feedback. I had to add `cfg_if` as a dependency as well.

rgbkrk:
No problem, thanks for maintaining the package.

PR #3049, opened 2025-08-09
Rev safetensors to 0.6, float8 to 0.3

zackangelo:
Fixes fp8 tensor loading


fchengjin:
It seems you missed a change，in `candle_nn/src/var_map.rs:35`， should be：
`safetensors::tensor::serialize_to_file(data, None, path.as_ref())?;`

ivarflakstad:
Fixed in recent PRs 🫡

PR #3051, opened 2025-08-10
add impl for mish activation function

oa-root:
Adding an implementation of the Mish activation function. However mish uses natural logarithm and here the log base 10 is used which skews the actual function  by a bit and converges to zero on the negative side faster than natural log.


greenrazer:
Thanks!

PR #3059, opened 2025-08-19
build: Make build.rs candle-kernels compatible with Nix and sandboxed builds

joeldsouzax:
This PR fixes an issue where the build.rs script writes generated files directly to the src/ directory. This practice violates Cargo's guidelines and causes build failures in sandboxed environments like Nix, where the source directory is read-only.

Changes:

build.rs: The script is updated to write ptx.rs to the Cargo-designated $OUT_DIR instead of src/.

src/lib.rs: The ptx module is now loaded using the include!(concat!(env!("OUT_DIR"), "/ptx.rs")); macro, which is the standard pattern for including code generated by a build script.

This makes the crate compliant with Cargo best practices and significantly improves its compatibility with reproducible build systems.

### DIFF
--- a/candle-kernels/build.rs
+++ b/candle-kernels/build.rs
@@ -4,8 +4,10 @@ fn main() {
     println!("cargo:rerun-if-changed=src/cuda_utils.cuh");
     println!("cargo:rerun-if-changed=src/binary_op_macros.cuh");
 
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let ptx_path = out_dir.join("ptx.rs");
     let builder = bindgen_cuda::Builder::default();
     println!("cargo:info={builder:?}");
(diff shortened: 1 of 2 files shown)


candle-kernels/build.rs
ahadnagy, line 1: This file is missing some imports:
```
use std::env;
use std::path::PathBuf;
```

joeldsouzax: That is soo stupid of me :/

ahadnagy (approved):

candle-kernels/build.rs
ahadnagy: Happens to me more times than I dare to admit :)

ivarflakstad (approved):
Thanks for this 👍

greenrazer:
Thanks!

PR #3061, opened 2025-08-21
Dev main staging

AlpineVibrations:
Add recent HF Candle changes


ivarflakstad:
There's a lot of stuff here that we would want to get into candle, but unfortunately I think we would have to do it the boring way where we split this up into several separate PRs.

AlpineVibrations:
yeah makes sense at this point. EricLBuehler has a fork on the dev_main branch with some features like FlashAttn v3 and maybe others..

ivarflakstad:
FA3 was just merged.
I'll close this PR, but it's a nice compilation of what we want to update so I'll refer to it down the road 👍

PR #3062, opened 2025-08-22
Fix broken slice_scatter example in basics.rs

davenpi:
## Problem
The [`basics.rs`](https://github.com/huggingface/candle/blob/main/candle-core/examples/basics.rs) example in `candle-core/examples` fails to run due to a shape mismatch error:
```zsh
$ cargo run --example basics -p candle-core
Error: shape mismatch in slice-scatter (self, src), lhs: [3, 2], rhs: [2, 1]
```
because
```rust
let a = Tensor::new(&[[0.0f32, 1.0, 2.0], [3.0, 4.0, 5.0]], &Device::Cpu)?;  // [2, 3]
let b = Tensor::new(&[[88.0f32, 99.0]], &Device::Cpu)?;                      // [1, 2] ❌
let new_a = a.slice_scatter(&b, 1, 2)?;  // Trying to scatter into column 2
```

## Fix

```rust
// Before
let b = Tensor::new(&[[88.0f32, 99.0]], &Device::Cpu)?;  // [1, 2] row vector

// After  
let b = Tensor::new(&[[88.0f32], [99.0]], &Device::Cpu)?;  // [2, 1] column vector
```

Also updated the assertion to match the expected result after column replacement.

### DIFF
--- a/candle-core/examples/basics.rs
+++ b/candle-core/examples/basics.rs
@@ -9,9 +9,12 @@ use candle_core::{Device, Tensor};
 
 fn main() -> Result<()> {
     let a = Tensor::new(&[[0.0f32, 1.0, 2.0], [3.0, 4.0, 5.0]], &Device::Cpu)?;
-    let b = Tensor::new(&[[88.0f32], [99.0]], &Device::Cpu)?; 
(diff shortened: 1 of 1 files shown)


ivarflakstad (approved):
Thanks :)

PR #3067, opened 2025-08-25
Metal device.set_seed full u64 support

ivarflakstad:
By pretending the `u64` buffer pointer is actually two successive `u32` buffer pointers we can use `atomic<uint>` operations to load and store the `ulong` seed value ✨


greenrazer (approved):
Looks Good!

PR #3071, opened 2025-08-28
Fix iOS app store validation issues

### LINKED ISSUE #2749: iOS app store validation issues
FYI with current master I am getting:

> ITMS-90755: Invalid Binary - The following binaries in your app contains prohibited instructions: (my binary with candle in it). Remove the instructions from the binaries, rebuild and resubmit.

We are reverting to 0.8.1 for now, will update if I find out more.

greenrazer:
From #2749, when uploading to app store some users get an error:
> ITMS-90755: Invalid Binary - The following binaries in your app contains prohibited instructions: (my binary with candle in it). Remove the instructions from the binaries, rebuild and resubmit.

This seems to be because the `ug` crate uses `gemm` ops that are unsupported on iOS. This PR hides any `ug` functionality behind a cfg flag similar to `wasm`.

### DIFF
--- a/candle-core/Cargo.toml
+++ b/candle-core/Cargo.toml
@@ -37,15 +37,6 @@ ug = { workspace = true }
 ug-cuda = { workspace = true, optional = true }
 ug-metal = { workspace = true, optional = true }
 
-[target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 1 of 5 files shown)


ykhrustalev:
confirm it works with AppStore validation

PR #3074, opened 2025-08-30
feat: Add Kontext support to FLUX transformers

tosi-n:
- Add kontext_image_ids parameter to WithForward trait
- Update Flux forward method for dynamic sequence concatenation
- Add denoise_kontext function for conditional image generation
- Add create_kontext_image_ids helper for positional embeddings
- Support variable sequence lengths in transformer blocks
- Extract only generation noise (skip static image latents)

Based on mflux architecture analysis for proper Kontext implementation.


AlpineVibrations:
would be great to have Kontext support is this working? why was it closed? @tosi-n

tosi-n:
I missed this, I will re-open by the end of this week. I was doing some private test.

Issue #3002, opened 2025-06-23
Is there a roadmap or intention to support CUDA Graph?


guoqingbao:
vLLM v1 uses CUDA Graph to capture the execution workflow of the entire model, resulting in significant performance improvements compared to the previous version. I'm wondering if there are any plans to support CUDA Graph in Candle. Would it be possible to add `start_capture`, `end_capture`, and `replay` to the `Module` so that the captured graph can be replayed within the forward method? @LaurentMazare 

Eric may also be interested in this @EricLBuehler

guoqingbao:
Some references for CUDA graph usage:

vLLM: https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/gpu_model_runner.py#L2190
Mini implementation: https://github.com/GeeeekExplorer/nano-vllm/blob/main/nanovllm/engine/model_runner.py#L208

guoqingbao:
**Implementation Skeleton**

Add a `CudaGraphModule`

```rust
use candle_core::{Result, Tensor, Shape};
use std::collections::HashMap;

pub trait Module {
    fn forward(&self, xs: &Tensor) -> Result<Tensor>;
}

pub trait CudaGraphModule: Module {
    fn start_capture(&mut self, shape: &Shape) -> Result<()>;
    fn end_capture(&mut self) -> Result<()>;
    fn replay(&self, xs: &Tensor) -> Result<Tensor>;
}
```

Use a CUDA graph wrapper to capture multiple shapes

```rust

use std::collections::HashMap;
use candle_core::{Device, Result, Shape, Tensor};

pub struct CudaGraphWrapper<M: Module> {
    module: M,
    captured_graphs: HashMap<Shape, CudaGraphHandle>,
    capturing: bool,
    current_shape: Option<Shape>,
}

pub struct CudaGraphHandle {
    // Native CUDA graph handles, pointers, etc.
    // For now, placeholder. In real code, bind to C/CUDA.
}

impl CudaGraphHandle {
    pub fn replay(&self, input: &Tensor) -> Result<Tensor> {
        // Launch the graph on the GPU
        // Placeholder logic
        Ok(input.clone()) // Replace with actual graph result
    }
}

impl<M: Module> CudaGraphWrapper<M> {
    pub fn new(module: M) -> Self {
        Self {
            module,
            captured_graphs: HashMap::new(),
            capturing: false,
            current_shape: None,
        }
    }
}

[...]

PR #3083, opened 2025-09-10
[WIP] CUDA graph support

ahadnagy:
This PR adds CUDA graph support.
Currently it's under development, I'll add more details here when the structure seems final, until then this PR is just for tracking progress.

Related issue: https://github.com/huggingface/candle/issues/3002


sempervictus:
@guoqingbao has graph working in `candle-vllm` and `vllm.rs` and at least on older HW it does bump throughput by 5-10%.

PR #3085, opened 2025-09-15
In the BERT example: apply the attention mask from tokenization during pooling

kshitijl:
So the results match those from Python examples on HuggingFace.

This will help people who want to migrate from Python `sentence_transformers` or `transformers` to candle. I think it's also more correct, in that it is more consistent with how the models are trained and evaluated: no padding or other masked token outputs (but I'm less confident about that).

It certainly helped me when I was comparing a reference Python implementation to my new rust implementation.

### DIFF
--- a/candle-examples/examples/bert/main.rs
+++ b/candle-examples/examples/bert/main.rs
@@ -186,9 +186,7 @@ fn main() -> Result<()> {
             // tokens, including padding. This was the original behavior of this
             // example, and we'd like to preserve it for posterity.
             let (_n_sentence, n_tokens, _hidden_size) = embeddings.dims3()?;
-            let embeddings = (embeddings.sum(1)? / (n_tokens as f64))?;
-
-            embeddings
+            (embeddings.sum(1)? / (n_tokens as f64))?
         } else {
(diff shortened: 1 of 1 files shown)


ivarflakstad:
Hey @kshitijl . Thanks for this!
Could you add a flag to the args so that the behaviour is optional? I'm happy to have it be the default behaviour, just want to keep the original variant for posterity ☺️

ivarflakstad:
@kshitijl bump!
Not sure if you saw my responses on your PRs ☺️

kshitijl:
Hi! Yes, I saw your replies. Thanks for taking a look, your suggestions sound good. I'm having a busy week with some personal things but should have some time today to work on it!

kshitijl:
@ivarflakstad Apologies for the delay! I added the flag you suggested and verified that it produces the same values as the current behavior (at the head of huggingface/candle:main).

ivarflakstad:
If you could fix the clippy lint as well that would be great :)

kshitijl:
Ah, sorry, I was running `cargo clippy` but should have been running `cargo clippy --examples`. 

Fixed!

PR #3086, opened 2025-09-15
re-export prost::Message

dvdsk:
To decode a model into ModelProto you need to have the prost::Message trait in scope. If you depend on a different version of prost then candle decoding will fail. By re-exporting prost users can depend directly on our version.

### DIFF
--- a/candle-onnx/src/lib.rs
+++ b/candle-onnx/src/lib.rs
@@ -1,4 +1,5 @@
 use candle::Result;
+pub use prost;
 use prost::Message;
 
 pub mod onnx {


yara-blue:
Unfortunately this does not work on all platforms (breaking on mac & windows). How do you all feel about shipping pre-'compiled' protocol buffers? That should help compile time & get rid of the dependency on protoc?

ivarflakstad:
> Unfortunately this does not work on all platforms (breaking on mac & windows). How do you all feel about shipping pre-'compiled' protocol buffers? That should help compile time & get rid of the dependency on protoc?

Not opposed at all actually. With the change I'd like a github action that ensures correctness though :)

PR #3078, opened 2025-09-07
Add CUDA 13 support

jfernandez:
Update cudarc to v0.17.3 which has support for CUDA 13.

PR #3089, opened 2025-09-18
Upgrade ug dependency for CUDA 13 support

grahamking:
ug 0.5.0 pulls in cudarc 0.17.3, which supports CUDA 13

Extension of https://github.com/huggingface/candle/pull/3078


greenrazer:
Thanks!

grahamking:
The runner upgraded to Rust 1.90 which came out an hour or so ago, hence the clippy failures.

grahamking:
We should probably do the clippy's from that Rust upgrade in a different PR.

PR #3090, opened 2025-09-20
[WIP] Generic Tensor<B>

ivarflakstad:
Makes tensor backend generic. Will allow us to support any number of backends, and the backend can even be defined outside of the main repo - as opposed to the existing enum approach.


greenrazer:
Some things I found for examples that use metal:

- based 
  - `Metal conv1d BF16 not implemented`
- biet
  - `Metal strided affine I64 not implemented`
- Deepseek
	- `thread '<unnamed>' panicked at /Users/kb/Documents/webclones/candle/candle-transformers/src/models/deepseek2.rs:1071:37:
	  called `Result::unwrap()` on an `Err` value: Metal error Could not lock kernel map: Command buffer map  
	  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace  `
- Depth Anything has some bigger issues, but I'll work on that

ivarflakstad:
> candle-transformers/src/models/deepseek2.rs:1071:37: called Result::unwrap()on anErrvalue: Metal error Could not lock kernel map: Command buffer map   note: run withRUST_BACKTRACE=1environment variable to display a backtrace 

Oh woops this was not supposed to be included in this PR. I was testing using rayon inside model code to launch several pieces simultaneously. Cut the initial loading time in half so we should explore, but not in this PR :)

Just for completeness the fix is simply to ensure the current thread actually _has_ a command buffer
```rust
pub fn command_buffer(&mut self) -> Result<(bool, CommandBuffer), MetalKernelError> {
    let mut command_buffers = self.command_buffers.lock()?;
    let command_buffer = match command_buffers.get_mut(&thread::current().id()) {
        Some(command_buffer) => command_buffer,
        None => {
            let command_buffer = create_command_buffer(&self.command_queue)?;
            command_buffers.insert(thread::current().id(), command_buffer);
            command_buffers.get_mut(&thread::current().id()).unwrap()
        }
    };
    ...
```

PR #3091, opened 2025-09-22
[WIP] Implement Tensor::unfold(dim, size, step): 1-dim unfolded window views.

crutcher:
This manipulates layout to unfold a window along a dimension.

See: https://docs.pytorch.org/docs/2.8/generated/torch.Tensor.unfold.html

TODO: I need help with `backprop.rs::Tensor::backward()`

### DIFF
--- a/candle-core/src/backprop.rs
+++ b/candle-core/src/backprop.rs
@@ -496,8 +496,32 @@ impl Tensor {
                         let sum_grad = grads.or_insert(arg)?;
                         *sum_grad = sum_grad.add(&arg_grad.broadcast_as(sum_grad.dims())?)?;
                     }
-                    Op::Unfold(_, _, _, _) => {
-                        todo!()
+                    &Op::Unfold(ref arg, dim, size, step) => {
+                        assert!(false, "never runs");
+
+                        let arg_dims = arg.dims();
+                        let node_dims = node.dims();
+
+                        println!("arg.id {:?}", arg.id());
+                        println!("node.id {:?}", node.id());
+
+                        let sum_grad = grads.or_insert(arg)?;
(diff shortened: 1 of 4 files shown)


crutcher:
I'm working to thread `unfold` through rust tensor backends, to support lower-memory spatial transforms; particularly those being used by `burn`.

I'm new to candle, and don't know how to write the backprop code; could use a suggested impl or some pointers.

crutcher:
A heavier handed approach would be something like Tensor::with_layout(); which I understand is a scary API to provide, though it is actually pretty simple to validate. Any given layout has an easy to compute max and min ravel offset; and the test is just "is that inside the storage size?"

crutcher:
I have a broken impl of backprop; I'm trying to debug it, but it never runs? there's an assert in it that never triggers.

crutcher:
@ivarflakstad ping?

greenrazer:
Hi @crutcher,

In your test instead of:

```
let data = &[[0f32, 1., 2., 3., 4.], [5f32, 6., 7., 8., 9.]];
    let x = Tensor::new(data, device)?;
```
you should do:
```
    let data = &[[0f32, 1., 2., 3., 4.], [5f32, 6., 7., 8., 9.]];
    let v = Var::new(data, device)?;
    let x = v.as_tensor();
 ```
 
This sets the `variable` field on the tensor to true, which in turn makes `x.track_op()` return true. As a result, `BackpropOp::new1(self, |arg| Op::Unfold(arg, dim, size, step))` returns `Some` instead of `None`, which allows `sorted_nodes` to return a non-empty list.

With this change`assert!(false, "never runs");` should execute

PR #3092, opened 2025-09-22
[Metal] various minor improvements

ivarflakstad:


greenrazer:
The CI is failing on clippy, 

`cargo clippy --workspace --tests --examples  -- -D warnings`

It seems to want you to write `.is_multiple_of` instead of using mod? It seems to be in files you haven't touched,  but since this PR is clippy-related, would you mind addressing those as well while you're at it?

PR #3093, opened 2025-09-22
[Metal] Buffer improvements

ivarflakstad:
Approaching the same storage mode as [mlx](https://github.com/ml-explore/mlx/blob/ec2ab428883790d14ce8ca719abafdbe380453bb/mlx/backend/metal/allocator.cpp#L14). We cannot use untracked yet, which is because we do not use fences and atomics in the same manner as mlx.

This should allow us to run candle on other apple devices like iphones.

### DIFF
--- a/candle-core/src/metal_backend/device.rs
+++ b/candle-core/src/metal_backend/device.rs
@@ -5,7 +5,7 @@ use candle_metal_kernels::{
     },
     Kernels,
 };
-use objc2_foundation::{NSRange, NSURL};
+use objc2_foundation::NSURL;
 use objc2_metal::{MTLCaptureDescriptor, MTLCaptureDestination, MTLCaptureManager};
 use std::path::Path;
 use std::sync::{Arc, Mutex, RwLock};
@@ -202,30 +202,6 @@ impl MetalDevice {
             .map_err(MetalError::from)?;
         let new_buffer = Arc::new(new_buffer);
         subbuffers.push(new_buffer.clone());
-        Ok(new_buffer)
-    }
-
-    /// Allocate a buffer with data, reusing existing buffers if possible
-    fn allocate_buffer_with_data<T>(&self, data: &[T], name: &str) -> Result<Arc<Buffer>> {
-        let size = core::mem::size_of_val(data);
-        let mut buffers = self.buffers.write().map_err(MetalError::from)?;
-        if let Some(b) = find_available_buffer(size, &buffers) {
-            unsafe {
-                std::ptr::copy_nonoverlapping(data.as_ptr(), b.data() as *mut T, data.len());
-            }
-            b.did_modify_range(NSRange::new(0, size));
-            // Cloning also ensures we increment the strong count
-            return Ok(b.clone());
-        }
-        let subbuffers = buffers.entry(size).or_insert(vec![]);
-
-        let new_buffer = self
-            .device
-            .new_buffer_with_data(data.as_ptr().cast(), size, RESOURCE_OPTIONS)
-            .map_err(MetalError::from)?;
-        let new_buffer = Arc::new(new_buffer);
-        subbuffers.push(new_buffer.clone());
-
         Ok(new_buffer)
     }
 


candle-core/src/metal_backend/device.rs
```diff
@@ -141,61 +147,37 @@ impl MetalDevice {
...
-    }
-
-    /// Creates a new buffer (not necessarily zeroed).
-    /// The buffer is [MTLManaged](https://developer.apple.com/documentation/metal/mtlstoragemode)
-    /// This means the buffer can be read on the CPU but will require manual
-    /// synchronization when the CPU memory is modified
-    /// Used as a bridge to gather data back from the GPU
-    pub fn new_buffer_managed(&self, size: usize) -> Result<Arc<Buffer>> {
-        self.allocate_buffer(size, MTLResourceOptions::StorageModeManaged, "managed")
+        self.allocate_buffer(size)
```
McPatate, line 157: don't you want to use the default resource options here?

candle-core/src/metal_backend/device.rs
```diff
@@ -58,6 +58,12 @@ pub struct MetalDevice {
     pub(crate) seed: Arc<Mutex<Buffer>>,
 }
 
+// Resource options used for creating buffers. Shared storage mode allows both CPU and GPU to access the buffer.
+pub const RESOURCE_OPTIONS: MTLResourceOptions =
+    objc2_metal::MTLResourceOptions(MTLResourceOptions::StorageModeShared.bits());
```
McPatate, line 63: Not sure I understand the motivation here, could you explain?
EDIT: I saw that's what they're doing within mlx, but unsure why we would want to do the same?

candle-core/src/metal_backend/mod.rs
```diff
@@ -2053,7 +2058,7 @@ impl BackendDevice for MetalDevice {
                 .new_buffer_with_data(
                     [299792458u64].as_ptr() as *const c_void,
                     4,
-                    MTLResourceOptions::StorageModeManaged,
+                    RESOURCE_OPTIONS,
```
McPatate, line 2061: why do you pass it as an arg here? you could just use `RESOURCE_OPTIONS` since it's in the module scope where `new_buffer_with_data` is defined

candle-metal-kernels/examples/metal_benchmarks.rs
```diff
@@ -1,12 +1,11 @@
 use anyhow::Result;
 use candle_metal_kernels::{
     metal::{create_command_buffer, Device},
-    GemmDType,
+    GemmDType, RESOURCE_OPTIONS,
```
McPatate, line 4: not sure I understand why we expose it publicly

ArthurZucker (approved):
Did not have time to deep dive yet but this works (allocated some tensor on metal on my iphone while it was failing before! 

trying to do a forward with a gemma model got me a `"Metal error Could not lock kernel map: Command buffer map"` but might be unrelated

candle-core/src/metal_backend/device.rs
```diff
@@ -205,28 +187,21 @@ impl MetalDevice {
     }
 
     /// The critical allocator algorithm
-    fn allocate_buffer(
-        &self,
-        size: usize,
-        option: MTLResourceOptions,
-        _name: &str,
-    ) -> Result<Arc<Buffer>> {
+    pub fn allocate_buffer(&self, size: usize) -> Result<Arc<Buffer>> {
```
McPatate, line 190: I don't have all the context, so lmk if I'm misunderstanding something, but it feels strange here to remove the configurable option param. Either we "hardcode it" (use const) everywhere, either we leave configurable.
But here it feels like we're in a little in between.
If we leave it configurable but provide a "should work in all scenarios" default, perhaps we can rename `RESOURCE_OPTIONS` to `DEFAULT_RESOURCE_OPTIONS` or something of the like. Or pass the param as `Option<MTLResourceOptions>` and `unwrap_or_default(RESOURCE_OPTIONS)`.

ivarflakstad:
> Did not have time to deep dive yet but this works (allocated some tensor on metal on my iphone while it was failing before! 

❤️

> trying to do a forward with a gemma model got me a `"Metal error Could not lock kernel map: Command buffer map"` but might be unrelated.


Should be fixed in #3100 :)

candle-core/src/metal_backend/device.rs
ivarflakstad: We want to use shared storage mode (resource is available for both cpu and gpu) because it is required for ios support.

Additionally for our uses there is no downside to using this mode (like performance), so it has the added bonus of simplifying our metal backend slightly.

candle-core/src/metal_backend/mod.rs
ivarflakstad: Well this is in the fn that creates a new `MetalDevice`.
It is calling the objc binding directly, so here we have to provide the option.
Same name, different fn :)

candle-metal-kernels/examples/metal_benchmarks.rs
ivarflakstad: Same reason as the other one. Calls objc device directly instead of the candle `MetalDevice`.

If the urge to tidy up this later takes over feel free 😉

candle-metal-kernels/src/metal_src/indexing.metal
ivarflakstad, line 1: Poor PR etiquette of me to include this in this one. Apologies 🙇

candle-core/src/metal_backend/device.rs
ivarflakstad: You're right. It should be hardcoded and it is. It's just a little bit of confusion because I use the same fn names as the objc bindings :)

candle-core/src/metal_backend/mod.rs
```diff
@@ -2026,8 +2032,7 @@ impl MetalStorage {
 
     pub(crate) fn to_cpu<T: Clone>(&self) -> Result<Vec<T>> {
         let size = self.count * self.dtype.size_in_bytes();
-
-        let buffer = self.device.new_buffer_managed(size)?;
+        let buffer = self.device.allocate_buffer(size)?;
```
McPatate, line 2035:
```suggestion
        let buffer = self.device.new_buffer(self.count, self.dtype, "")?;
```

candle-metal-kernels/src/lib.rs
```diff
@@ -21,6 +21,11 @@ use source::Source;
 pub use utils::BufferOffset;
 use utils::{get_block_dims, linear_split, EncoderParam, EncoderProvider};
 
+pub const RESOURCE_OPTIONS: MTLResourceOptions =
+    objc2_metal::MTLResourceOptions(MTLResourceOptions::StorageModeShared.bits());
```
McPatate, line 25: I sort of understand why it's here, but at the same time it feels like an abstraction leak. But I don't have the broader picture of how candle-core and this crate interact so just ignore me 😄

candle-core/src/metal_backend/mod.rs
ivarflakstad: Yeah I agree. I just need to either use the  `_name` variable for something or get rid of it.

candle-metal-kernels/src/lib.rs
ivarflakstad: It shouldn't be pub. Soon it won't be :)
But for now I'll merge

filipw:
🥳 🥳 🥳 🥳

PR #3098, opened 2025-09-29
Updated candle-book

little-world:
I did write a book on candle 
I did try to get it published, but it was rejected. 
I will use it in my own courses, but I think it can be useful to the community.


little-world:
Updated candle-book

ivarflakstad:
Hey @little-world, thanks for the contribution!
This is a lot of writing! Impressive.
Let me have a little think before we decide how we want to approach this.

little-world:
Hi Ivar, Thank you
You may do with it what you (as a person but also other people) think is feasible
I can understand that not all chapters fit in the way you would present candle to the world
I did write the book for readers who do not know PyTorch and a little about neural networks
the examples are tested; maybe they should be part of the GitHub website

PR #3100, opened 2025-09-30
[Metal] Improve command buffer map

ivarflakstad:
Hide the `ThreadId` usage inside `CommandBufferThreadMap`.

Remove some failure points from `wait_until_completed`.

### DIFF
--- a/candle-metal-kernels/src/metal/commands.rs
+++ b/candle-metal-kernels/src/metal/commands.rs
@@ -94,10 +94,10 @@ impl Commands {
                     command_buffer.commit();
                     command_buffer.wait_until_completed();
                 }
-                MTLCommandBufferStatus::Committed => {
+                MTLCommandBufferStatus::Committed | MTLCommandBufferStatus::Scheduled => {
                     command_buffer.wait_until_completed();
                 }
-                MTLCommandBufferStatus::Scheduled | MTLCommandBufferStatus::Completed => {}
+                MTLCommandBufferStatus::Completed => {}
                 _ => {}
             }
             *command_buffer = create_command_buffer(&self.command_queue)?;
--- a/candle-metal-kernels/src/metal/commands.rs
+++ b/candle-metal-kernels/src/metal/commands.rs
@@ -1,19 +1,13 @@
-use crate::metal::CommandBuffer;
+use crate::metal::{CommandBuffer, CommandBufferThreadMap};
 use crate::MetalKernelError;
 use objc2::{rc::Retained, runtime::ProtocolObject};
 use objc2_metal::{MTLCommandBufferStatus, MTLCommandQueue, MTLCounterSet};
-use std::{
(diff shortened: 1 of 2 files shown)


candle-metal-kernels/src/metal/commands.rs
```diff
@@ -69,12 +64,14 @@ impl Commands {
 
     pub fn command_buffer(&mut self) -> Result<(bool, CommandBuffer), MetalKernelError> {
         let mut command_buffers = self.command_buffers.lock()?;
-        let command_buffer =
-            command_buffers
-                .get_mut(&thread::current().id())
-                .ok_or(MetalKernelError::LockError(
-                    "Command buffer map".to_string(),
-                ))?;
+        let command_buffer = match command_buffers.get_mut() {
```
drbh, line 67: nit: might be possible to avoid blocking when checking the status/wait_until_completed - not sure if this has much of an impact

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -89,23 +86,25 @@ impl Commands {
...
+            match command_buffer.status() {
+                MTLCommandBufferStatus::NotEnqueued | MTLCommandBufferStatus::Enqueued => {
+                    command_buffer.commit();
+                    command_buffer.wait_until_completed();
+                }
+                MTLCommandBufferStatus::Committed | MTLCommandBufferStatus::Scheduled => {
+                    command_buffer.wait_until_completed();
+                }
+                MTLCommandBufferStatus::Completed => {}
+                _ => {}
```
drbh, line 101: is there a Error case? could that be bubbled up in `MetalKernelError`?

drbh (approved):
looks good, added some nits/thoughts

candle-metal-kernels/src/metal/commands.rs
ivarflakstad: There is! I could extract the value etc, sure.

candle-metal-kernels/src/metal/commands.rs
ivarflakstad: I don't think it would have much of a performance impact, but I agree that it would theoretically be slightly better so I'll give it a go (later 🫡 )

candle-metal-kernels/src/metal/commands.rs
ivarflakstad: Indeed there is. I'll do it in another PR 👍

ivarflakstad: https://github.com/huggingface/candle/pull/3106

candle-metal-kernels/src/metal/commands.rs
ivarflakstad: https://github.com/huggingface/candle/pull/3107

PR #3103, opened 2025-10-02
[Metal] Add conv ops for more dtypes

ivarflakstad:
Now we can run models that use convolutions on metal with bf16. Need to update our metal gemm though as the current version does not handle all the odd strides / dims that for example `based` creates.

Also added backtracing to metal kernel errors for easier debugging


drbh (approved):
LGTM!

PR #3104, opened 2025-10-02
Rough example of inlining model files into binary

### LINKED ISSUE #3096: [Question] Minimal documentation/example on including weights in compiled executable
Just what the title says: Is there a minimal code example on including weights in the compiled executable using include_bytes. Nervous to implement this without understanding best practices and end up with a suboptimal solution.

matthewhaynesonline:
Fixes #3096

Not sure if this is the best approach, but getting the model files inlined using `include_bytes!` had some tough IMO limitations. Using a step in build.rs seemed like the least worst option, though this does require hard coding some things.

Beyond that, initializing the model / tokenizer from bytes vs from file path (see `build_model_and_tokenizer_from_bytes`) seemed to be a uncommon usecase, so I'm not sure if that's the best approach either:

```rust
    let config_string = std::str::from_utf8(config_data)?;
    let config: BertConfig = serde_json::from_str(config_string)?;
    let tokenizer = Tokenizer::from_bytes(tokenizer_data).map_err(anyhow::Error::msg)?;
```


ivarflakstad:
Thanks @matthewhaynesonline this is really nice!

I've figured out an approach that makes the downloading etc completely isolated from the other examples :)

Create a new emtpy lib crate within `candle-examples` called `bert-single-file-binary-builder` (or whatever you prefer).
Here you place your `build.rs`, and update `Cargo.toml` with `ureq` and `anyhow = { workspace = true }` as build dependencies.
I think you'll probably want the download dest path in the build script to be simply `files`.

Then in `candle-examples`'s `Cargo.toml` you add a new dependency `bert-single-file-binary-builder = { path = "bert-single-file-binary-builder", optional = true }`, and a new feature `bert-single-file-binary-builder = ["dep:bert-single-file-binary-builder"]`.
Update your example to use the new files location, and when you build your example you now use `--features=bert-single-file-binary-builder`.

ivarflakstad:
> Beyond that, initializing the model / tokenizer from bytes vs from file path (see build_model_and_tokenizer_from_bytes) seemed to be a uncommon usecase, so I'm not sure if that's the best approach either

Nothing wrong with this way of loading the "files". Data is data is data 👍

matthewhaynesonline:
@ivarflakstad thanks so much! Great point about segmenting the build into a crate (agreed much cleaner) - if it sounds okay, I'll take a pass at implementing your feedback and then set the pr as ready for review?

ivarflakstad:
Sounds good :)

candle-examples/bert_single_file_binary_builder/build.rs
```diff
@@ -0,0 +1,77 @@
...
+    let dest_path = Path::new("files");
+    let files = ["config.json", "tokenizer.json", "model.safetensors"];
+
+    let all_files_exist = files
+        .iter()
+        .all(|filename| dest_path.join(filename).exists());
+
+    if all_files_exist {
+        println!(
+            "cargo:warning=All {} files already exist, skipping download",
```
matthewhaynesonline, line 25: Switched from eprintln to cargo:warning (also, since existing candle examples build.rs was using single cargo: print statement not cargo::, I stuck with it here too)

matthewhaynesonline:
@ivarflakstad  thanks again! Updated in https://github.com/huggingface/candle/pull/3104/commits/2bbd2735e458a3dd06d55fe921503680ea4a0b4f

candle-examples/bert_single_file_binary_builder/build.rs
ivarflakstad: Ah, right. It's actually candle examples that is outdated. We're well past rust v1.77 so `cargo::` is now the correct prefix.
Using `cargo:` hasn't been deprecated yet though, so no worries.

matthewhaynesonline: @ivarflakstad want me to update this PR to use the new :: syntax and / or open a follow up PR to update the main build rs to use :: as well?

ivarflakstad: Feel free to update them in a later PR :)

ivarflakstad (approved):
👌

PR #3105, opened 2025-10-02
bump gemm dependency to 0.18.2 to match ug

slckl:
`ug` uses `gemm 0.18`, while `candle-core` and `candle-nn` still depend on `gemm 0.17`.

This causes for 2 `gemm` versions to be compiled currently (visible in `Cargo.lock`).

This PR bumps candle's dependency on `gemm` to `0.18.2`, resulting in just one `gemm` dependency.

I ran the benches on my local machine, the changes are within noise threshold mostly, no big gains or losses.


ivarflakstad (approved):
Thanks! :)

PR #3111, opened 2025-10-03
make cuda benches run again

slckl:
Looks like `cargo bench` with `--features cuda,cudnn` has not been touched for a while, and fails to compile these days.

Fortunately, this is easily fixed by adding a missing trait import and removing some error wrapping.

Leaving the wrap `w()` produced more compile errors, but removing it compiles and works just fine, so idk? `candle-nn` benches already skipped the `w()`, so I skipped it as well.


ivarflakstad (approved):
Thanks!

PR #3113, opened 2025-10-04
[Metal] Improved get_strided_index

ivarflakstad:
Between 6.7% and 11.4% increased thrpt for strided metal kernels.

We should add either pre-compilation or load kernels with a [dynamic library](https://developer.apple.com/documentation/metal/metal-dynamic-libraries) so we don't have to repeat the same impl for each kernel.


ivarflakstad:
Superseded by #3285

PR #3117, opened 2025-10-06
Adding inference for GraniteMoeHybrid models from IBM

atilag:
There's also an example of how to run the [Granite 4 Micro model](https://huggingface.co/ibm-granite/granite-4.0-micro)

### DIFF
--- a/candle-transformers/src/models/granitemoehybrid.rs
+++ b/candle-transformers/src/models/granitemoehybrid.rs
@@ -1,0 +1,582 @@
+//! GraniteMoeHybrid is a Long Context Transformer Language Model.
+//!
+//! A high performance transformer model optimized for efficient processing
+//! of very long context sequences
+
+use super::with_tracing::{linear_no_bias as linear, Linear, RmsNorm};
+use candle::{DType, Device, IndexOp, Result, Tensor, D};
+use candle_nn::{embedding, Embedding, Module, VarBuilder};
+use std::{collections::HashMap, f32::consts::PI};
+
+pub const DEFAULT_MAX_SEQ_LEN: usize = 4096;
+
+#[derive(Debug, Clone, serde::Deserialize, Default)]
+pub enum GraniteMoeHybridRopeType {
+    #[serde(rename = "granite")]
+    Granite,
+    #[default]
+    #[serde(rename = "default")]
+    Default,
+}
+
+#[derive(Debug, Clone, serde::Deserialize, Default)]
+pub struct GraniteMoeHybridRopeConfig {
+    pub factor: f32,
+    pub low_freq_factor: f32,
+    pub high_freq_factor: f32,
+    pub original_max_position_embeddings: usize,
+    pub rope_type: GraniteMoeHybridRopeType,
+}
+
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct GraniteMoeHybridConfig {
+    pub hidden_size: usize,
+    pub intermediate_size: usize,
+    pub vocab_size: usize,
+    pub num_hidden_layers: usize,
+    pub num_attention_heads: usize,
+    pub num_key_value_heads: Option<usize>,
+    pub rms_norm_eps: f64,
+    #[serde(default = "default_rope")]
+    pub rope_theta: f32,
+    pub bos_token_id: Option<u32>,
+    pub eos_token_id: Option<u32>,
+    pub rope_scaling: Option<GraniteMoeHybridRopeConfig>,
+    pub max_position_embeddings: usize,
+    #[serde(default)]
+    pub layer_types: Vec<GraniteMoeHybridLayerType>,
+    #[serde(default = "default_one")]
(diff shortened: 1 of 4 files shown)


candle-transformers/src/models/granitemoehybrid.rs
```diff
@@ -0,0 +1,582 @@
...
+        })
+    }
+
+    fn mask(&mut self, t: usize) -> Result<Tensor> {
+        if let Some(mask) = self.masks.get(&t) {
+            Ok(mask.clone())
+        } else {
+            let mask: Vec<_> = (0..t)
+                .flat_map(|i| (0..t).map(move |j| u8::from(j > i)))
+                .collect();
```
ivarflakstad: I realize that this is inherited from the `granite` impl, and we're also talking about microseconds here so not the bottleneck of the pipeline, but I'll mention it anyways :)

Preferably we don't want to call cpu specific code inside a tensor forward pass where the storage is not explicitly cpu.
I see that this is only called when flash attention is not available, but there's also metal to consider.
Did you try `candle_nn::ops::sdpa`? If that for some reason didn't work then this is fine.

In any case I would rather have this PR merged and then potentially iterate on it, but how about we use a sligthly different implementation that is ~35x faster? ☺️ (fastest safe variant I could come up with).
```suggestion
            let mut mask: Vec<u8> = Vec::with_capacity(t * t);
            (0..t).for_each(|i| {
                mask.extend(core::iter::repeat(0).take(i + 1));
                mask.extend(core::iter::repeat(1).take(t - i - 1));
            });
```

candle-transformers/src/models/granitemoehybrid.rs
```diff
@@ -0,0 +1,582 @@
...
+}
+
+impl MultiLayerPercepton {
+    fn forward(&self, x: &Tensor) -> Result<Tensor> {
+        let _enter = self.span.enter();
+        let projected = self.input_linear.forward(x)?;
+        let mut chunks = projected.chunk(2, D::Minus1)?;
+        let right = chunks.pop().expect("mlp chunk right");
+        let left = chunks.pop().expect("mlp chunk left");
+        let gated = (candle_nn::ops::silu(&left)? * right)?;
```
ivarflakstad: Could get rid of `.expect` with something like this
```suggestion
        let (left, right) = (&chunks[0], &chunks[1]);
        let gated = (candle_nn::ops::silu(left)? * right)?;
```

ivarflakstad (commented):
Thank you for this! This all looks very good, but I have one gripe - there's perhaps a bit too much identical code between this and good ol' `granite.rs`.
I'd prefer if we reused as much as possible so that we keep bloat to a minimum.

So first thing I'd ask is that we simply use the standard `Granite*` types when they are identical to the ones you have in this new impl. Like `GraniteRopeType` vs `GraniteMoeHybridRopeType`.
For those types that are strict supersets of those in `granite` (or where they are otherwise easily mergeable) I'd prefer if we embedded them in the types you have here. For example 
```rust
#[derive(Debug, Clone, serde::Deserialize)]
pub struct GraniteMoeHybridConfig {
    #[serde(flatten)]
    pub granite_config: granite::GraniteConfig,
     #[serde(default)]
    pub layer_types: Vec<GraniteMoeHybridLayerType>,
    ... etc
```

How does that sound to you? :)

candle-transformers/src/models/granitemoehybrid.rs
atilag: Sure thing!

atilag:
Thanks for the quick review @ivarflakstad !!! ❤️ 

Yeah, there's a lot of code repetition because I wasn't sure in the beginning how this new Hybrid arch was going to look like once implemented. For this specific model that I tried first: `granite-4.0-micro`, things look pretty similar to the old architecture, just adding some more normalization, and scaling factors here and there. And this is the case because there are no Mamba layers (all are good all Attention layers). The rest of the Granite 4 models require implement the Mamba arch, this is the Hybrid arch, and this is where things are way different as you can imagine.
 Anyway, I hate unnecessary code repetition. so what do you think about these 2 options:

 1. We merge this PR with some code repetition so people can start using at least the granite-4.0-micro (and only this one) and  once I finish the Mamba support, I PR and if everything looks good we merge, and then I work on generalizing things to avoid that much repetition (now that we have all supported architectures and I will know better how to do this)
 
 2. I'm ok working on generalizing now, and then move on with the Mamba support. Even though I don't know how the Hybrid support is going to ending look like, I'm not expecting anything crazy complex, so this can be a valid option as well

Your choice :)

Thanks again for everything!

atilag:
Oh, BTW. I did have plans to work on performance improvements once all the models work.
I'm eager to see what can we do to speed things up and compare with Python Transformers ;)

ivarflakstad:
I'm happy with option 1 👍

atilag:
I fixed some issues with clipper and applied the suggestions from the comments.
I think we should be ready to go! :)

PR #3123, opened 2025-10-09
[Metal] Fix quantized to_float calls

ivarflakstad:


candle-core/src/quantized/metal.rs
```diff
@@ -48,63 +48,63 @@ impl QMetalStorage {
         match self.dtype {
             GgmlDType::F32 => {
                 let vec: Vec<f32> = read_to_vec(&buffer, block_len);
-                f32::to_float(&vec, &mut out)?;
+                f32::to_float(&vec, &mut out);
```
EricLBuehler, line 51: Are we ignoring the error?

Might be my misunderstanding.

ivarflakstad: Sorry, no. I broke it in #3109

LLukas22: I think `T::to_float` was changed so it no longer returns a `Result`.

But I’m also wondering where does an error go if one occurs? This conversion shouldn’t really be safe, right? 🤔

ivarflakstad: In rust terms it is safe regardless of the changes.

I'll just change the term to "correct" for simplicity.
It is correct as long as the quantized condition (typically `n.is_multiple_of(block size)` etc) holds. Those checks are now in `debug_assert`s instead of if-statement + `bail!`.
In other words incorrect usage will be detected in tests and if you run your code in debug mode. It will not be detected in release builds.

EricLBuehler: Seems fine to me, makes sense. In that case maybe we should remove the `Result` from `to_float` entirely?

PR #3125, opened 2025-10-09
feat(candle-nn): Add RwLock-based VarMap for concurrent inference

artem1984A:


candle-nn/src/var_builder.rs
```diff
@@ -466,6 +482,12 @@ impl SimpleBackend for candle::safetensors::SliceSafetensors<'_> {
 }
 
 impl<'a> VarBuilder<'a> {
+    /// Initializes a `VarBuilder` using a custom backend.
+    ///
+    /// It is preferred to use one of the more specific constructors. This
+    /// constructor is provided to allow downstream users to define their own
+    /// backends.
+
```
ivarflakstad:
```suggestion
```

ivarflakstad: Duplicate

candle-nn/tests/var_map_compatibility.rs
```diff
@@ -0,0 +1,488 @@
+use candle::{DType, Device, Result, Tensor, Var};
+use candle_nn::var_map::ConcurrentVarMap;
+use candle_nn::{Init, VarMap};
+use std::sync::{Arc, Barrier};
+use std::thread;
+
+#[test]
+fn test_basic_operations_compatibility() -> Result<()> {
```
ivarflakstad, line 8: I think I'd prefer we simply verify results against a hardcoded expected value.
Like `assert_eq!(t1.shape(), /* expected shape */);` etc.

So that we don't need this original varmap concept within the test :)

candle-nn/tests/var_map_compatibility.rs
```diff
@@ -0,0 +1,488 @@
...
+        let name = format!("layer_{}.weight", i);
+        let shape = (64, 64);
+        // Use a deterministic init for comparison
+        original.get(shape, &name, Init::Const(i as f64), DType::F32, &device)?;
+        updated.get(shape, &name, Init::Const(i as f64), DType::F32, &device)?;
+    }
+
+    // Save both
+    let original_path = "/tmp/test_original_varmap.safetensors";
+    let updated_path = "/tmp/test_updated_varmap.safetensors";
```
ivarflakstad: This will fail on windows (and android etc but less relevant).
Use `std::env::temp_dir` to get an OS agnostic temporary directory 👍

candle-nn/tests/var_map_compatibility.rs
```diff
@@ -0,0 +1,488 @@
...
+            DType::F32,
+            &device,
+        )?;
+    }
+
+    // Test batch operations
+    let names: Vec<&str> = (0..10)
+        .map(|i| Box::leak(format!("weight_{}", i).into_boxed_str()) as &str)
+        .collect();
+    let batch_vars = concurrent.get_vars_batch(&names);
```
ivarflakstad: You can avoid the manual leaking of memory here.

Just create a vec of strings, then  map with `as_str`, and finally send in as slice

candle-nn/tests/var_map_compatibility.rs
ryzhov-artem: Thank You for review!
I was applied this suggestion at branch.

ryzhov-artem:
Applied all  suggestions:
 **Simplified test to use hardcoded expected values** (7e3adae)
- Removed comparison with original VarMap implementation
- Tests now directly verify against known expected values
- Makes tests clearer and more maintainable

 **Used `std::env::temp_dir()` for cross-platform compatibility** (4bc438e)  
- Replaced hardcoded `/tmp/` paths with OS-agnostic temporary directory
- Now should works correctly on Windows, macOS, Linux, and other platforms

**Removed manual memory leak in batch operations** (5930869)
- Changed from `Box::leak()` to  `Vec<String>` + `.as_str()` pattern

ivarflakstad:
Apologies for the delay!

Now that I've taken a step back and had a second look at this, I'm actually a bit unsure if this has a real usecase.
For training we use the `Mutex` variant, which makes sense.
But for inference we shouldn't really use varmaps in the first place. You should use something like `VarBuilder::from_mmaped_safetensors`, or one of the other variants that have no locking at all.

ryzhov-artem:
Thank You for the thorough review and feedback.
I'm just  enthusiast in mathematics and AI models (officially a backend developer), 
and I initially thought `ConcurrentVarMap` could help with multi-threaded inference 
scenarios involving mutable weights.
However, after analyzing my own quantization experiments, I realize I'm still need only
direct SafeTensors access (memory-mapped, thread-safe), which is the correct approach 
for this use cases.
I apologize if this contribution doesn't address a real need at this time. 
If I will complete improvements to my works in quantization tools in the future, I'd be grateful 
for Your review of those contributions.
Thank You once more for time and insights!

ivarflakstad:
Shame I didn't catch the issue sooner. Apologies for wasting your time. At least we both learned some valuable details 🤗

PR #3128, opened 2025-10-11
feat: manual release workflow

drbh:
This PR adds a new github workflow that aims to improve the release workflow. 

The workflow requires manually triggering and checks that the intended release semver version matches the version in the cargo workspace to avoid unintended releases. 

High level process
- check version
- run cargo test/check
- create new release on github
- publish latest version to crates

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,8 +61,17 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       
+      # helper tool to publish all workspace crates in correct order
+      - name: Install cargo-workspaces
+        run: cargo install cargo-workspaces
+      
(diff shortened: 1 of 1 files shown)


ivarflakstad (approved):
Ready to give this a go 👌

ivarflakstad:
Though with that said, [actions/create-release](https://github.com/actions/create-release) is unmaintained and we should probably use [softprops/action-gh-release](https://github.com/softprops/action-gh-release) instead as that seems the most well maintained of the alternatives.

drbh:
updated the deps and ran some sanity test with [act](https://github.com/nektos/act) locally. The workflow seems to run as expected 

```bash
act workflow_dispatch -W .github/workflows/release.yml --input version=0.9.1 --container-architecture linux/amd64
```

PR #3131, opened 2025-10-12
Add more conv2d bench cases to candle-nn benches

slckl:
Adds more bench cases to `candle-nn` conv2d benches. This will be useful, when evaluating alternative conv2d implementations in more details as present benches mask the difference on cpu thanks to +bias being super slow.
- made input size (M, K) and kernel size ( K_SIZE) dynamic
- made bias optional - bias essentially adds `+ constant` to runtime, which just murks the real difference in conv implementation
- a total of 4 cases per dtype: 
    - input size 128, kernel 3x3 + bias (aka the existing bench case)
    - input size 128, kernel 1x1, no bias
    - input size 128, kernel 5x5, no bias
    - input size 512, kernel 3x3, no bias
- two dtypes as before: f32 and f16

### DIFF
--- a/candle-nn/benches/benchmarks/conv.rs
+++ b/candle-nn/benches/benchmarks/conv.rs
@@ -18,18 +18,21 @@ fn run_conv2d_benchmark(
     k_size: usize,
     m: usize,
     bias: bool,
-    name: &str,
 ) {
     let weight = Tensor::ones((1, C, k_size, k_size), dtype, device)
         .unwrap()
         .to_dtype(dtype)
         .unwrap();
-    let bias = if bias {
+    let bias_t = if bias {
         Some(Tensor::zeros(m, dtype, device).unwrap())
     } else {
         None
     };
     let input = Tensor::ones((B, C, m, m), dtype, device).unwrap();
+    let name = format!(
+        "conv2d_{dtype:?}_i{m}_k{k_size}x{k_size}_{}",
+        if bias { "b" } else { "nb" }
+    );
 
     let mut group = c.benchmark_group(device.bench_name(name));
     group.bench_function("iter", move |b| {
@@ -39,7 +42,7 @@ fn run_conv2d_benchmark(
                 run(
                     black_box(input.clone()),
                     black_box(weight.clone()),
-                    black_box(bias.clone()),
+                    black_box(bias_t.clone()),
                     Default::default(),
                 );
             }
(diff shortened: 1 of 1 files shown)


candle-nn/benches/benchmarks/conv.rs
```diff
@@ -6,23 +6,30 @@ use std::time::Instant;
...
         .to_dtype(dtype)
         .unwrap();
-    let bias = Tensor::zeros(K, dtype, device).unwrap();
-    let input = Tensor::ones((B, C, M, K), dtype, device).unwrap();
+    let bias = if bias {
+        Some(Tensor::zeros(m, dtype, device).unwrap())
+    } else {
+        None
+    };
+    let input = Tensor::ones((B, C, m, m), dtype, device).unwrap();
```
ivarflakstad, line 31:
```suggestion
    let input = Tensor::ones((B, C, m, m), dtype, device).unwrap();
    let name = format!("conv2d_{dtype:?}_i{m}_k{k}x{k}_{}", if bias { "b" } else { "nb" });
```

candle-nn/benches/benchmarks/conv.rs
```diff
@@ -6,23 +6,30 @@ use std::time::Instant;
...
-fn run_conv2d_benchmark(c: &mut Criterion, device: &Device, dtype: DType, name: &str) {
-    let weight = Tensor::ones((1, 1, K_SIZE, K_SIZE), dtype, device)
+fn run_conv2d_benchmark(
+    c: &mut Criterion,
+    device: &Device,
+    dtype: DType,
+    k_size: usize,
+    m: usize,
+    bias: bool,
+    name: &str,
```
ivarflakstad:
```suggestion
```

candle-nn/benches/benchmarks/conv.rs
```diff
@@ -46,8 +53,14 @@ fn run_conv2d_benchmark(c: &mut Criterion, device: &Device, dtype: DType, name:
...
-        run_conv2d_benchmark(c, &d, DType::F32, "conv2d_f32");
-        run_conv2d_benchmark(c, &d, DType::F16, "conv2d_f16");
+        run_conv2d_benchmark(c, &d, DType::F32, 3, 128, true, "conv2d_f32_i128_k3x3_b");
+        run_conv2d_benchmark(c, &d, DType::F32, 1, 128, false, "conv2d_f32_i128_k1x1_nb");
+        run_conv2d_benchmark(c, &d, DType::F32, 5, 128, false, "conv2d_f32_i128_k5x5_nb");
+        run_conv2d_benchmark(c, &d, DType::F32, 3, 512, false, "conv2d_f32_i512_k3x3_nb");
+        run_conv2d_benchmark(c, &d, DType::F16, 3, 128, true, "conv2d_f16_i128_k3x3_b");
+        run_conv2d_benchmark(c, &d, DType::F16, 1, 128, false, "conv2d_f16_i128_k1x1_nb");
+        run_conv2d_benchmark(c, &d, DType::F16, 5, 128, false, "conv2d_f16_i128_k5x5_nb");
+        run_conv2d_benchmark(c, &d, DType::F16, 5, 512, false, "conv2d_f16_i512_k3x3_nb");
```
ivarflakstad:
```suggestion
        run_conv2d_benchmark(c, &d, DType::F32, 3, 128, true);
        run_conv2d_benchmark(c, &d, DType::F32, 1, 128, false);
        run_conv2d_benchmark(c, &d, DType::F32, 5, 128, false);
        run_conv2d_benchmark(c, &d, DType::F32, 3, 512, false);
        run_conv2d_benchmark(c, &d, DType::F16, 3, 128, true);
        run_conv2d_benchmark(c, &d, DType::F16, 1, 128, false);
        run_conv2d_benchmark(c, &d, DType::F16, 5, 128, false);
        run_conv2d_benchmark(c, &d, DType::F16, 5, 512, false);
```

candle-nn/benches/benchmarks/conv.rs
slckl: Included dynamic name gen, your freestyled version *almost* compiled.

candle-nn/benches/benchmarks/conv.rs
slckl: Included in dynamic name gen

candle-nn/benches/benchmarks/conv.rs
slckl: Yup, it's now dynamic

candle-nn/benches/benchmarks/conv.rs
ivarflakstad: So close! 😄

candle-nn/benches/benchmarks/conv.rs
```diff
@@ -6,23 +6,33 @@ use std::time::Instant;
...
+    bias: bool,
+) {
+    let weight = Tensor::ones((1, C, k_size, k_size), dtype, device)
         .unwrap()
         .to_dtype(dtype)
         .unwrap();
-    let bias = Tensor::zeros(K, dtype, device).unwrap();
-    let input = Tensor::ones((B, C, M, K), dtype, device).unwrap();
+    let bias_t = if bias {
+        Some(Tensor::zeros(m, dtype, device).unwrap())
```
ivarflakstad, line 27: Just remembered. In certain situations the hardware/runtime may detect that data is empty, and understands that adding zeroes is a noop. So we should use `Tensor::rand` or similar instead.

slckl: Would `tensor::ones` be good enough? I think I remember the article you think of :D  

this one? https://www.thonking.ai/p/strangely-matrix-multiplications

ivarflakstad: Yes, like that article - and I remember we actually had this happen in our matmul benchmark. And I was the one that added the `zero` tensors 😅 

The bias being deterministic will obviously not be as impactful as all zeros/ones matmul.
So in that sense both `zeros` and `ones` is good enough.
It's more about having a "standard" approach of generating bench data in the project so that it is less likely that we accidentally get imprecise results

PR #3132, opened 2025-10-12
Update cargo build instructions to use double colon syntax

matthewhaynesonline:
Follow up of https://github.com/huggingface/candle/pull/3104#discussion_r2409288382

Use `cargo::` syntax for build instructions per https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script.

### Notes
- This just simply replaces any of the existing `cargo:` build commands with `cargo::`. 
- `candle-examples/*` passes `$ cargo check`
- `candle-flash-attn` and `candle-kernels` seem to rely on having NVIDIA drivers installed, so I couldn't run `$ cargo check` (I use a m4 macbook)

### Questions
1. I see a `cargo::info` build command used (and passes check in `candle-examples`) but couldn't find reference to it in the docs. When I tried updating the bert binary example to use it instead of `cargo::warn` that seemed to fail. Should `cargo::info` be used instead of `cargo::warn` in the bert binary builder?

```sh
Compiling bert-single-file-binary-builder v0.9.1 (/Users/matt/Code/matthewhaynes/candle/candle-examples/bert_single_file_binary_builder)
error: invalid output in build script of `bert-single-file-binary-builder v0.9.1 (/Users/matt/Code/matthewhaynes/candle/candle-examples/bert_single_file_binary_builder)`: `cargo::info=All bert-single-file-binary-builder files already exist, skipping download`
Unknown key: `info`.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```
 
2. I don't think I can build / check `candle-flash-attn` and `candle-kernels` on my mac, but CI should ensure I didn't break anything build wise for CUDA - is that right?

### DIFF
--- a/candle-kernels/src/compatibility.cuh
+++ b/candle-kernels/src/compatibility.cuh
@@ -7,7 +7,7 @@
 
 // FIXME: the minimum compute capabilities are just guesses since the table is not specific enough
 
-#if (__CUDACC_VER_MAJOR__ < 12 || __CUDACC_VER_MINOR__ < 2) && __CUDA_ARCH__ < 800
+#if (__CUDACC_VER_MAJOR__ < 12 || __CUDACC_VER_MINOR__ < 2) && __CUDA_ARCH__ < 750
 __device__ __forceinline__ __half __hmax_nan(__half a, __half b) {
     return __hisnan(a) ? a : (__hisnan(b) ? b : __hmax(a, b));
 }


ivarflakstad:
> 1. I see a cargo::info ...

No documentation because it doesn't exist. Should be `cargo::warning` if you actually want to display a warning. For example I wouldn't show a warning for downloading files when that is the expected behaviour.

> 2. I don't think I can build 

I don't think you will be able to break it by just using the correct amount of `:` in the build script instructions. When the PR is ready I can check on a cuda machine.

matthewhaynesonline:
@ivarflakstad  rebased off latest main, so I think this is ready. Candle examples pass cargo check on Mac, but still not sure about flash-attn / kernels (really simple diff I know, but just wanted to mention I wasn't able to check those personally).

Also, this is one of the `cargo:info` lines that caught my eye:

https://github.com/huggingface/candle/blob/d312da234440ca64fb64c38ba57588a692f58caf/candle-examples/build.rs#L31

Would this just behave like a normal println and not a build specific command? Not a biggie, just genuinely curious.

Lastly, https://github.com/huggingface/candle/pull/3147 looks really nice! Thanks again!

ivarflakstad:
> Would this just behave like a normal println and not a build specific command? Not a biggie, just genuinely curious.

It actually does nothing.
Feel free to simply remove them :)

ivarflakstad:
Alternatively change to warnings. They're only shown locally or when you build with `-vv`for very verbose

matthewhaynesonline:
> Alternatively change to warnings. They're only shown locally or when you build with `-vv`for very verbose

Sounds good! Swapped the two info statements with warnings in 4ace53034199cd35ada75f2824bb9c60161dc7fc and made sure to rebase

Issue #3138, opened 2025-10-19
ModernBERT head.forward fails on Metal with Invalid matmul arguments … (CPU OK) — missing contiguous() before Linear


whitebox2:
## Summary

On Apple Metal, `ModernBertHead::forward` intermittently fails at the first `Linear` with:

```
Metal error Invalid matmul arguments [17920, 1] [1, 256] (2, 256, 256)
```

The same inputs run fine on CPU. The root cause is that the tensor passed into `Linear` is a **non-contiguous view** (produced upstream by slicing/aggregation).  Adding `contiguous()` in `ModernBertHead::forward` fixes the issue. 

## Environment

* Device: Apple Silicon (Metal backend)
* Candle: `candle-core`/`candle-nn`/`candle-transformers` version 0.9.1 (with `metal` feature)
* Model: ModernBERT for sequence classification
* CPU backend: **no error**
* Metal backend: **fails**

## Root Cause
  ```rust
impl Module for ModernBertHead {
    fn forward(&self, xs: &Tensor) -> Result<Tensor> {
        let xs = xs.apply(&self.dense)?.gelu_erf()?.apply(&self.norm)?;
        Ok(xs)
    }
}
  ```

## Proposed Fix

Materialize a contiguous buffer before applying `Linear`:

```rust
impl Module for ModernBertHead {
    fn forward(&self, xs: &Tensor) -> Result<Tensor> {
        let xs = xs.contiguous()?.apply(&self.dense)?.gelu_erf()?.apply(&self.norm)?;
        Ok(xs)
    }
}
```

ivarflakstad:
Thanks!
I'm not getting the error from the modernbert example. I tried manually changing the classifier config to trigger the bug, but no luck.
Could you share a way to reproduce?

whitebox2:
I created the code based on the modern bert example, but an similar error occurred. My environment is Apple M1.
#Error
```bash
Error: Metal error Invalid matmul arguments [9216, 1] [1, 768] (3, 768, 768)

Caused by:
    Invalid matmul arguments [9216, 1] [1, 768] (3, 768, 768)
```

#Code
```rust
use anyhow::{Error as E, Result};
use candle_core::{DType, Device, Tensor};
use candle_nn::VarBuilder;
use candle_transformers::models::modernbert;
use hf_hub::{Repo, RepoType, api::sync::Api};
use std::collections::HashMap;
use std::path::PathBuf;
use tokenizers::{PaddingParams, Tokenizer};

[...]

PR #3139, opened 2025-10-19
Fix Metal matmul failure in `ModernBertHead::forward` by ensuring contiguous input 

whitebox2:
## Summary

On Apple Metal, `ModernBertHead::forward` may receive a **non-contiguous view** from upstream pooling (CLS/MEAN). Passing that view directly into `Linear` causes the Metal matmul path to misinterpret shapes and fail with:

```
Invalid matmul arguments [17920, 1] [1, 256] (2, 256, 256)
```

CPU succeeds because it tolerates non-contiguous inputs. We **materialize** the input with `contiguous()` before applying `Linear`. This resolves the Metal-only failure and aligns with prior guidance that GPU matmul expects contiguous tensors. ([GitHub][1])

## Patch

```diff
 impl Module for ModernBertHead {
     fn forward(&self, xs: &Tensor) -> Result<Tensor> {
-        let xs = xs.apply(&self.dense)?.gelu_erf()?.apply(&self.norm)?;
+        let xs = xs.contiguous()?.apply(&self.dense)?.gelu_erf()?.apply(&self.norm)?;
         Ok(xs)
     }
 }
```

## Issue
https://github.com/huggingface/candle/issues/3138#issue-3529976234


ivarflakstad:
Thanks for this!
I'm thinking it might make more sense to apply contiguous on the cls variant only, like`ClassifierPooling::CLS => output.i((.., 0, ..))?.contiguous()?,`.
What do you think?

BectorVoom:
I agree.

ivarflakstad:
> I agree.

Nice!
If you update your PR with the changes we discussed I'll get it merged 👍 
And seems CI wants you to do a quick `cargo fmt` as well

whitebox2:
I updated the PR.

ivarflakstad (approved):
Thanks!

PR #3142, opened 2025-10-21
fix: build errors for compute cap 7.5

neksodebe:
this pull request fixes build issues for GPU's with a compute cap of `7.5`, such as the RTX 2060.

the steps i took to implement the fix can be found [here](https://github.com/huggingface/candle/issues/353#issuecomment-3428523368).


ivarflakstad (approved):
Thanks 👌

PR #3144, opened 2025-10-22
add padding support for max pool

donjuanplatinum:


ivarflakstad:
Nice!
If you want this to work on cuda / metal you'll have to add padding functionality to the actual kernels as well.
You can see the cuda kernel [here](https://github.com/huggingface/candle/blob/main/candle-kernels/src/conv.cu#L443) and the metal kernel [here](https://github.com/huggingface/candle/blob/main/candle-metal-kernels/src/metal_src/conv.metal#L330).
And update the code where the kernels are called of course, like [here](https://github.com/huggingface/candle/blob/main/candle-metal-kernels/src/kernels/convolution.rs#L138) for metal.

donjuanplatinum:
> Nice! If you want this to work on cuda / metal you'll have to add padding functionality to the actual kernels as well. You can see the cuda kernel [here](https://github.com/huggingface/candle/blob/main/candle-kernels/src/conv.cu#L443) and the metal kernel [here](https://github.com/huggingface/candle/blob/main/candle-metal-kernels/src/metal_src/conv.metal#L330). And update the code where the kernels are called of course, like [here](https://github.com/huggingface/candle/blob/main/candle-metal-kernels/src/kernels/convolution.rs#L138) for metal.

I have modified the code :)

candle-core/src/tensor.rs
```diff
@@ -1228,10 +1228,61 @@ impl Tensor {
...
+    /// * `padding` - The amount of zero-padding to add to both sides of the height and width
+    /// # Examples
+    /// ```rust
+    /// use candle_core::{Tensor,Device,Shape};
+    /// let t = Tensor::rand(0.0f32, 1.0, (1, 1, 10, 10), &device)?;
+    /// let (kernel_size,stride,padding) = (2,2,2)
+    /// let t = t.max_pool2d_with_stride_padding(kernel_size,stride,padding)?;
+    /// assert_eq!(t.shape().dims(),[1,1,7,7]);
+    /// ```
+    pub fn max_pool2d_with_stride_padding<T: crate::ToUsize2>(
```
ivarflakstad, line 1239: If you look at `max_pool2d` you can see how it simply calls `max_pool2d_with_stride` but with stride equal to kernel size (aka no stride).

If we make `max_pool2d_with_stride_padding` the base implementation, then `max_pool2d_with_stride` and `max_pool2d` can call it without padding.

candle-metal-kernels/src/kernels/convolution.rs
```diff
@@ -167,6 +167,40 @@ pub fn call_pool2d(
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
+pub fn call_pool2d_with_padding(
```
ivarflakstad: I don't think we need a whole new function, can just add padding as a parameter to the original.

candle-core/src/metal_backend/mod.rs
```diff
@@ -1300,6 +1300,7 @@ impl BackendStorage for MetalStorage {
         let buffer = self.device.new_buffer(dst_el, self.dtype, "avg_pool2d")?;
         let command_buffers = self.device.command_buffer()?;
         candle_metal_kernels::call_pool2d(
+        candle_metal_kernels::call_pool2d_with_padding(
```
ivarflakstad: This won't compile.

candle-core/src/tensor.rs
donjuanplatinum: i have moddified the code :let maxpool_with_stride and max_pool to call the maxpool_with_stride_padding;change two functions into the original maxpool function; and fix the compile error

PR #3148, opened 2025-10-25
rust-ci: add --benches to clippy, fix warnings

slckl:
- adds `--benches` to clippy in `rust-ci` CI step
- fixes the few straggling clippy warnings in benches

### DIFF
--- a/candle-core/benches/benchmarks/where_cond.rs
+++ b/candle-core/benches/benchmarks/where_cond.rs
@@ -23,7 +23,7 @@ const M: usize = 1024;
 const K: usize = 1024;
 const SIZE: usize = B * M * K;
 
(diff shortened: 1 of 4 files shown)


candle-core/benches/benchmarks/where_cond.rs
```diff
@@ -23,7 +23,7 @@ const M: usize = 1024;
 const K: usize = 1024;
 const SIZE: usize = B * M * K;
 
-const DATA: [u8; SIZE] = create_cond_arr::<SIZE>();
+static DATA: [u8; SIZE] = create_cond_arr::<SIZE>();
```
slckl, line 26: on cpu, this clippy suggestion ends up slightly improving the bench performance, but within noise levels :D  

obviously doesn't matter for non-bench code.

PR #3156, opened 2025-10-29
Add unsafe from_storage apis

EricLBuehler:
These APIs are useful for low-level creation of tensors.

### DIFF
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -2909,8 +2909,8 @@ impl std::ops::Div<&Tensor> for f64 {
     }
 }
 
-impl From<(Storage, Shape)> for Tensor {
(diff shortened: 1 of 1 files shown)


ivarflakstad (commented):
These functions are not unsafe, so you can remove the keyword.
Also they could both use `from_storage` fn further up.

I don't mind the extra utility methods on `Tensor`, but I'd prefer if we removed the `new_` prefix from them.

EricLBuehler:
Addressed those comments, so: `Tensor::new_from_storage_op` -> `Tensor::from_storage` and `Tensor::new_from_storage` -> `<Tensor as From<Storage, S>>::from`!

ivarflakstad (approved):
👌

PR #3161, opened 2025-11-03
Add `nll_with_ignore` and `cross_entropy_with_ignore`

embedding-shapes:
## Summary
Adds `ignore_index` support to cross-entropy and NLL loss functions via new `*_with_ignore` functions, enabling proper handling of padding tokens in sequence modeling.

## Motivation
Sequence modeling (NLP, time series) requires excluding padding tokens from loss computation. Currently users must manually mask losses, which is error-prone and less efficient.

## Changes
- Add `loss::nll_with_ignore(inp, target, ignore_index)` function
- Add `loss::cross_entropy_with_ignore(inp, target, ignore_index)` function
- Implement safe target sanitization to prevent OOB errors
- Add 4 tests including safety and PyTorch parity checks
- **Fully backward compatible** (new functions, existing ones unchanged)

## PyTorch Compatibility
Matches PyTorch functionality with two documented divergences:
1. **All-ignored case**: Returns `0.0` (PyTorch returns `nan`) for better ergonomics
2. **Type**: Uses `u32` (PyTorch uses `i32`) to match Candle's target dtype

## Open Questions

- Is adding two new functions better for you rather than changing the signature of existing functions? I'm personally fine with both, opted for the approach that reduces API churn for now, but OK to change it if you want to.
- Are there too many comments? I opted for adding "too much" over "too little" right now, let me know if it's too verbose and you want to adjust. I'm still a beginner ML enthusiast, so the comment were more for helping me during the implementation, we might want to remove some of them if deemed "too basic" maybe.


embedding-shapes:
Hello maintainers, any possibility of getting this checked, merged or otherwise some advice on anything that is missing/incorrect for a merge? I'm still using a franken-fork of various PRs merged in my own codebase and would love to be able to start using upstream Candle again without my own patches on top.

candle-nn/src/loss.rs
```diff
@@ -29,6 +29,59 @@ pub fn nll(inp: &Tensor, target: &Tensor) -> Result<Tensor> {
...
+    // Convert mask to same dtype as `inp` for arithmetic masking.
+    let mask_float = keep_mask.to_dtype(inp.dtype())?;
+
+    // Zero-out contributions from ignored positions.
+    let masked = gathered.mul(&mask_float)?;
+
+    // Count non-ignored elements.
+    let count = mask_float
+        .sum_all()?
+        .to_dtype(candle::DType::F64)?
```
ivarflakstad, line 76: Is the intention here to sum in f64 for precision?
If so the cast should be prior to `sum_all`.
If not then feel free to ignore.
I think using `f32` here would be fine though, and is more likely to hit a noop.

candle-nn/src/loss.rs
```diff
@@ -29,6 +29,59 @@ pub fn nll(inp: &Tensor, target: &Tensor) -> Result<Tensor> {
...
+        .to_scalar::<f64>()?;
+    if count == 0.0 {
+        return Tensor::new(0.0f64, inp.device())?.to_dtype(inp.dtype());
```
ivarflakstad, line 79: Could use `Tensor::zeros` to avoid the cast :)

candle-nn/tests/loss.rs
```diff
@@ -86,3 +86,85 @@ fn binary_cross_entropy_with_logit() -> Result<()> {
     assert_eq!(to_vec0_round(&loss, 4)?, 0.8224);
     Ok(())
 }
+
+#[test]
+fn cross_entropy_with_ignore_matches_subset() -> Result<()> {
+    let cpu = Device::Cpu;
```
ivarflakstad, line 92: Could you use the test device macro and send in device as input to the test instead? :)

Issue #3151, opened 2025-10-27
Tensor conversion to_vec1() failing on 0.9.2-alpha.1 - Metal


si-harps:
Dependencies

```toml
candle-core = { git = "https://github.com/huggingface/candle", rev = "df618f8", features = ["metal"] }
candle-nn = { git = "https://github.com/huggingface/candle", rev = "df618f8", features = ["metal"] }
candle-transformers = { git = "https://github.com/huggingface/candle", rev = "df618f8", features = ["metal"] }
```

Running on Macbook M2 Pro - Metal - Tahoe 26.0.1

Since upgrading to 0.9.2-alpha.1, BERT operations on Metal have started hanging when converting rank-1 tensor to Vec<32>. This seems to be affecting any ops that attempt to synchronize or move data from GPU to CPU. Not sure if this is directly related to the update but rolling back to 0.9.1 or using CPU as device fixes the issue. 

Some example of ops that are failing...

```rust
tensor.device().synchronize()
tensor.to_device()
tensor.to_vec1()
```

Actual code being run...

```rust
let (token_ids, token_type_ids, attention_mask) = self.encode_text(text)?;

let hidden_states = self
    .forward_model(&token_ids, &token_type_ids, &attention_mask)
    .await
    .map_err(|e| {
        log::error!("Failed to forward to model: {}", e);
        e
    })?;

let embeddings = self
    .apply_mean_pooling(&hidden_states, &attention_mask)
    .map_err(|e| {
        log::error!("Failed to apply mean pooling: {}", e);
        e
    })?;

...

fn apply_mean_pooling(
      &self,
      hidden_states: &Tensor,
      attention_mask: &Tensor,
  ) -> Result<Vec<f32>> {
      log::info!("Applying mean pooling to hidden states...");

      let attention_mask_for_pooling = attention_mask
          .to_dtype(hidden_states.dtype())?
          .unsqueeze(2)?;
      let sum_mask = attention_mask_for_pooling.sum(1)?;

      let pooled = (hidden_states.broadcast_mul(&attention_mask_for_pooling)?).sum(1)?;
      let sum_mask_safe = sum_mask.clamp(NUMERICAL_STABILITY_EPSILON, f32::MAX)?;
      let pooled = pooled.broadcast_div(&sum_mask_safe)?;

      let denom = pooled
          .sqr()?
          .sum_keepdim(1)?
          .sqrt()?
[...]

PR #3162, opened 2025-11-04
update commands to use shared command buffer fixes (#3151)

anonenity:


ivarflakstad:
Thanks!
As discussed on discord this unfortunately doesn't address all cases, but I just merged a PR that does 👍

PR #3164, opened 2025-11-05
[Metal] Ensure metal backend is send/sync via status semaphore

ivarflakstad:
Uses `CommandSemaphore` to ensure correct safe/correct usage of command buffers, encoders, etc.

Classic `Mutex` + `Condvar` implementation. Wraps `CommandStatus` which indicates what state we are currently in.
For example threads have to wait for an encoder to end encoding to the command buffer before they can create a new encoder, or commit the command buffer, etc.

For single thread usage there is no performance change, but obviously the more threads you use the more they will be waiting on other threads to finish interacting with a command buffer.

`device.command_buffer()` is no longer pub, and instead you should use methods like `device.command_encoder()` directly. This is to avoid having a thread commit the active command buffer before you get the encoder out of it, which leads to `SIGABRT` when you try to create the encoder.

### DIFF
--- a/candle-metal-kernels/src/metal/commands.rs
+++ b/candle-metal-kernels/src/metal/commands.rs
@@ -6,7 +6,7 @@ use objc2::{rc::Retained, runtime::ProtocolObject};
 use objc2_metal::{MTLCommandBufferStatus, MTLCommandQueue, MTLCounterSet};
 use std::sync::{
     atomic::{AtomicUsize, Ordering},
-    Arc, Mutex, MutexGuard, RwLock, RwLockWriteGuard,
+    Arc, RwLock,
 };
 
 // Use Retained when appropriate. Gives us a more elegant way of handling memory (peaks) than autoreleasepool.
--- a/candle-core/benches/benchmarks/affine.rs
+++ b/candle-core/benches/benchmarks/affine.rs
@@ -38,8 +38,7 @@ fn criterion_benchmark(c: &mut Criterion) {
         run_affine_benchmark(c, &device, DType::F32, "affine_f32");
         run_affine_benchmark(c, &device, DType::F16, "affine_f16");
         run_affine_benchmark(c, &device, DType::BF16, "affine_bf16");
-        #[cfg(feature = "metal")]
-        continue;
+        #[cfg(not(feature = "metal"))]
         run_affine_benchmark(c, &device, DType::F8E4M3, "affine_fp8");
     }
 }
--- a/candle-core/benches/benchmarks/mod.rs
+++ b/candle-core/benches/benchmarks/mod.rs
@@ -32,7 +32,7 @@ impl BenchDevice for Device {
             }
             Device::Metal(device) => {
                 #[cfg(feature = "metal")]
-                return Ok(device.wait_until_completed()?);
+                return device.wait_until_completed();
                 #[cfg(not(feature = "metal"))]
                 panic!("Metal device without metal feature enabled: {:?}", device)
             }
--- a/candle-core/src/custom_op.rs
+++ b/candle-core/src/custom_op.rs
@@ -436,7 +436,7 @@ impl InplaceOp1 for UgIOp1 {
         let device = sto.device();
         let encoder = device.command_encoder()?;
         encoder.set_compute_pipeline_state(&self.func);
-        let (g, b) = if elem_count % 32 == 0 {
+        let (g, b) = if elem_count.is_multiple_of(32) {
             (elem_count / 32, 32)
         } else {
             (elem_count, 1)
--- a/candle-core/src/quantized/metal.rs
+++ b/candle-core/src/quantized/metal.rs
@@ -128,7 +128,7 @@ impl QMetalStorage {
     }
 
     pub fn storage_size_in_bytes(&self) -> usize {
-        self.buffer.length() as usize
+        self.buffer.length()
     }
 
     fn fwd_mv(
@@ -289,7 +289,7 @@ impl QMetalStorage {
             blit.end_encoding();
         }
         self.device.wait_until_completed()?;
-        Ok(read_to_vec::<u8>(&buffer, self.buffer.length() as usize))
+        Ok(read_to_vec::<u8>(&buffer, self.storage_size_in_bytes()))
     }
 }
 
--- a/candle-nn/benches/benchmarks/mod.rs
+++ b/candle-nn/benches/benchmarks/mod.rs
@@ -25,7 +25,7 @@ impl BenchDevice for Device {
             }
             Device::Metal(device) => {
                 #[cfg(feature = "metal")]
-                return Ok(device.wait_until_completed()?);
+                return device.wait_until_completed();
                 #[cfg(not(feature = "metal"))]
                 panic!("Metal device without metal feature enabled: {:?}", device)
             }


candle-core/tests/tensor_tests.rs
```diff
@@ -1705,6 +1705,21 @@ fn tensor_send_sync(device: &Device) -> Result<()> {
             assert_eq!(result, vec![2.0f32, 4.0, 6.0]);
         });
     }
+    let result: Vec<f32> = tensor.to_vec1().unwrap();
+    assert_eq!(result, vec![1.0f32, 2.0, 3.0]);
+
+    let tensor = Tensor::new(vec![1.0f32, 2.0, 3.0], device)?;
+    tensor.device().synchronize().unwrap();
```
EricLBuehler, line 1712: Are these .synchronize calls necessary for correctness across threads?

EricLBuehler (approved):
Looks good!

candle-core/tests/tensor_tests.rs
ivarflakstad: No, but they're necessary for the test to catch the bug (at least the one inside the thread spawn below)

ivarflakstad:
Regarding slightly higher latency when using multiple threads:
We can solve this by having a pool of command buffers and lock them individually.
Something like 
```rust
struct CommandBufferEntry {
  id: usize,
  cb: CommandBuffer,
  compute_count: usize,
  semaphore: CommandSemaphore,
}

struct CommandBufferPool {
  pool: Vec<Arc<Mutex<CommandBufferEntry>>>
}

impl CommandBufferPool {
  fn get_command_buffer_entry(&self) ->MutexGuard<'_, CommandBufferEntry> {
      // Try non-blocking first
      for entry in self.pool.iter() {
        if let Some(cb) = entry.try_lock() {
          return cb
        }
      }
      // If none were available we lock and wait until the lock is acquired
      for entry in self.pool.iter() {
        return entry.lock().unwrap()
      }
    }
}
```
Perhaps the pool entries should be organized in a priority queue by their `compute_count`. In which case we would extract that variable out of the lock and use `AtomicUsize` instead, so that we could sort the list without locking the contents.
The list will be fixed size and small, like 10-20 entries (so arguably we should use an array), so it'll be faster to loop through and check the `compute_count`s manually than using a sorting algorithm / binary heap for the priority queue.

The slightly more complex part is tracking locks between command buffers and their encoders. My first idea was to use a shared unique id (ref `id: usize` above), and use that to select which semaphore to use.
However that means the structure needs to be more like
```rust
struct CommandBufferEntry {
  id: usize,
  cb: Mutex<CommandBuffer>, // moved mutex inside
  compute_count: AtomicUsize,
  semaphore: CommandSemaphore,
}

struct CommandBufferPool {
  pool: Vec<Arc<CommandBufferEntry>>
}
```
We need the mutex guard to be in scope of whatever fn uses it. When the guard is dropped the lock is released.
Since `MutexGuard::map` is only available on nightly we have to manually return some representation of `(usize, MutexGuard<'_, CommandBuffer>, AtomicUsize, CommandSemaphore)` from `get_command_buffer_entry`.


But this is for another day!
Just noting this down for posterity / in case someone feels like giving it a go.

candle-metal-kernels/src/metal/command_buffer.rs
```diff
@@ -2,32 +2,89 @@ use crate::{BlitCommandEncoder, ComputeCommandEncoder};
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum CommandStatus {
+    Available,
+    Encoding,
+    Done,
+}
+
+#[derive(Debug)]
+pub struct CommandSemaphore {
```
McPatate, line 16:
```suggestion
pub struct CommandMonitor {
```
Saw this a bit late sry, small nit on naming!

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -21,107 +26,153 @@ pub struct Commands {
...
+            let mut command_buffer = self.command_buffer.write()?;
             command_buffer.commit();
-            *command_buffer = create_command_buffer(&self.command_queue)?;
-            self.command_buffer_index = 0;
-            flushed = true;
+            *command_buffer = self.create_command_buffer()?;
+            self.compute_count.store(1, Ordering::Relaxed);
+            Ok((true, command_buffer.into()))
+        } else {
+            self.compute_count.fetch_add(1, Ordering::Relaxed);
```
McPatate, line 90: Are you sure `Ordering::Relaxed` works here?

candle-metal-kernels/src/metal/command_buffer.rs
ivarflakstad: Why is monitor an improvement?

candle-metal-kernels/src/metal/commands.rs
ivarflakstad: Yeah. As long as it's approximately >= `computer_per_buffer` we're good

candle-metal-kernels/src/metal/command_buffer.rs
McPatate: It's just a naming convention, conceptually `CommandSemaphore` isn't a semaphore, cf [Monitor wiki page](https://en.wikipedia.org/wiki/Monitor_%28synchronization%29)

ivarflakstad: It is definitely a semaphore. In fact you can implement monitors using semaphores, so the question then becomes if it is a monitor as well.

Just while we're talking about semaphores; truly one the topics with the most frequent misconceptions I've ever seen online. Regularly people confuse the generic semaphore concept with the most common semaphore implementations; counting and binary, and are unaware that you can actually put any state inside it to express your desired behaviour.
They get corrected yet respond with "no, I am correct" and show a source with the same misconception - and so the confusion continues.
Like look at this [stackoverflow](https://stackoverflow.com/questions/7335950/semaphore-vs-monitors-whats-the-difference). Basically all of the answers are incorrect/incomplete. A lot of truth there, but some are only describing counting semaphores as if that's the only semaphore in existence, while others are doing the same just with binary semaphores. I assume because in their language of choice that is what their `Semaphore` does.

Anyways! Wtr monitor: my understanding is that a monitor encapsulates the data you want to access, and that only a single thread can interact with its functions at a time. Seems smooth, perhaps worth considering, but not exactly how we've implemented it currently.

ivarflakstad: No wait a second. I think you're right. There are two definitions of a monitor.

The one I learned is the encapsulating kind, but I see the more modern definition doesn't have that requirement.

(It's still a semaphore as well though 😤)

McPatate: > (It's still a semaphore as well though 😤)

I can't find anything corroborating your statement 😅
The only thing I found on this being a semaphore is chatgpt talking about a "generalised semaphore on abstract state machines" but when prompted a bit further it can't find anything in the literature referencing this.
Actually, I've found a [reference](https://pure.tue.nl/ws/files/4279816/344354178746665.pdf) of "The general semaphore" but it is what you can find in [wikipedia](https://en.wikipedia.org/wiki/Semaphore_(programming)) as the "counting semaphore":
> The semaphores are essentially non-negative integers; when only used
to solve the mutual exclusion problem, the range of their values will even
be restricted to "0" and "1 ". It is the merit of the Dutch physicist and
computer designer Drs.C.S.Scholten to have demonstrated a considerable field
of applicability for semaphores that can also take on larger values. When
there is a need for distinction, we shall talk about ''binary semaphores" and
"general semaphores" respectively. The definition of the P- and V-operation 
EWD123- 29
that I shall give ~ow, is insensitive to tr.is distiction. 

In the definitions I've found, semaphores are an integer based synchronisation primitive, with two methods, P() & V() in Dijkstra's paper, that respectively wait/decrement and signal/increment said counter.
Your implementation conflicts with this and I've found multiple references to your implementation being something akin to [monitors, conditional critical regions](https://www4.cs.fau.de/Lehre/WS14/PS_KVBK/papers/conditionvariables.pdf) or [guarded commands](https://www.cs.toronto.edu/~chechik/courses05/csc410/readings/dijkstra.pdf).

ivarflakstad: This has become way more complicated than it needs to be. A semaphore is not as concretely and neatly defined as it's specific implementations. It's simple and generic both in definition and function, hence "synchronisation primitive". The existence of both counting and binary semaphores proves that the concept is inherently generic. For example you can easily imagine a ternary semaphore. It simply has three possible states instead of two (bonus points if you see why I specifically mention ternary)

Additionally you've actually provided a reference that corroborates my claim yourself (wikipedia).

> A trivial semaphore is a plain variable that is changed (for example, incremented or decremented, or toggled) depending on programmer-defined conditions.

In that sentence is all the information you need.

McPatate: I’m not sure I follow why the ternary example would imply a broader definition of semaphores. The definition I previously referenced (Dijkstra) frames them as integer-based primitives with P/V operations; so a ternary would still fit that definition ~, even if `{-1, 0, 1}`~. _edit: reread the definition and it states non negative integers, which makes sense, it becomes something different otherwise_

You can definitely build higher level primitives with Semaphores, probably even any other synchronisation primitive, but that doesn't make the resulting structure a semaphore. You refer to it by its semantics.

I may be misunderstanding though, do you have a reference that I can read up on? I'm genuinely interested in understanding what you mean.

PR #3173, opened 2025-11-08
Improve sample_argmax

ivarflakstad:
In general for all these ops we should use tensor ops instead of moving to the cpu first.

Several models (like llama) are implemented to use `sample_argmax` when temperature is < 0.0. So you can verify correctness by comparing output with main.


EricLBuehler (approved):
Looks great!

PR #3174, opened 2025-11-09
Draft: QuantizedDType for normal tensors

LLukas22:
### Disclaimer

This PR is still a **work in progress** and needs some polishing.
The main goal of opening it early is to gather feedback on whether this approach is heading in the right direction.

---

### Overview

The primary goal of this PR is to introduce **`QuantizedDType`** support.
These data types allow regular tensors to store and operate on quantized data, making it much easier to add new quantization schemes to Candle in the future without relying on custom ops.

---

#### 1. New `candle-macros` crates

* **`candle-macros`**: contains the procedural macros that generate the dispatch code for `QuantizedDTypes`.
* **`candle-macros-types`**: defines the traits that quantized types can implement to provide backend-specific support.

#### 2. `QuantizedType` trait

Each quantization implements the `QuantizedType` trait, which defines:

* a static `NAME`
* functions to calculate storage size for the quantized format

Quantizations can optionally implement one or more of the following backend traits:

* `QuantizedCpuOps`
* `QuantizedCudaOps`
* `QuantizedMetalOps`

These traits define the de/quantization logic and backend-specific matmul implementations (e.g., `f32 × quantized`).

#### 3. The `register_quantized_types!` macro

This macro generates:

* a `QuantizedDType` enum
* all required dispatch functions to call backend ops efficiently (minimizing runtime overhead)

The enum is then integrated into Candle Core as a new `DType::Quantized(QuantizedDType)` variant.

Tensors using this type:

* support most operations through **implicit dequantization** (currently to `f32`)
* dispatch directly to backend-specific matmul implementations when data types match

#### 4. External Quantization Support

A `register_external_quantized_type!` macro is also included.
This will allow **external crates** to register their own quantization types without modifying Candle Core directly.

---

### Current Limitations

* Quantized tensors can currently only be created via `.to_dtype()` — I haven’t yet found a clean way to load them from files.
* Using `f32` as the intermediate type isn’t ideal for some backends (like CUDA) and may need refinement.
* The Metal backend implementation is not yet complete, as I don’t have access to Apple hardware for testing.

### DIFF
--- a/candle-core/examples/quantize_basics.rs
+++ b/candle-core/examples/quantize_basics.rs
@@ -131,7 +131,7 @@ fn cuda_examples(device: &Device) -> Result<()> {
         (1, 128),
         device,
     )?;
-    let weights_q4k = weights.to_dtype(DType::Quantized(QuantizedDType::GgmlQ4K))?;
+    let weights_q4k = weights.to_dtype(DType::Quantized(QuantizedDType::GgmlQ8_0))?;
     let result = activations.matmul(&weights_q4k)?;
 
     let result_cpu = result.to_device(&Device::Cpu)?;
@@ -142,8 +142,9 @@ fn cuda_examples(device: &Device) -> Result<()> {
         result_cpu.shape()
     );
     println!(
-        "First 3 values: [{:.4}, {:.4}, {:.4}]",
-        result_vec[0][0], result_vec[0][1], result_vec[0][2]
+        "  Result sample: [{:.4}, ..., {:.4}]",
+        result_vec[0][0],
+        result_vec[0][result_vec[0].len() - 1]
     );
 
     Ok(())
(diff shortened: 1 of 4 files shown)


candle-core/src/display.rs
```diff
@@ -63,6 +63,17 @@ impl std::fmt::Debug for Tensor {
...
+                // Dequantize to f32 for display
+                match self.to_dtype(DType::F32) {
+                    Ok(f32_tensor) => f32_tensor.fmt_dt::<f32>(f),
+                    Err(_) => write!(
+                        f,
+                        "Quantized Tensor[dtype={}; error converting to f32]",
+                        self.dtype().as_str()
+                    ),
+                }
+            }
```
ivarflakstad: Is this easier or harder to read? 🤔 
```suggestion
            DType::Quantized(_) => self
                .to_dtype(DType::F32)
                .map(|t| t.fmt_dt::<f32>(f))
                .unwrap_or_else(|_| {
                    write!(
                        f,
                        "Quantized Tensor[dtype={}; error converting to f32]",
                        self.dtype().as_str()
                    )
                }),
```

PR #3175, opened 2025-11-09
Add command buffer pool for improved multi-threaded Metal performance

anonenity:
This PR introduces a command buffer pool that maintains multiple command buffers, enabling threads to encode concurrently with minimal contention. The pool distributes work across buffers and automatically manages buffer recycling and completion.

**Key changes**

- Low-contention selection path
  - Non-blocking scan for immediately available entries via a per-entry semaphore
  - Fallback to the least-loaded entry (tracked with atomics) and wait
- Encoder/buffer coordination
  - Per-entry `CommandSemaphore` encoders flip status back to Available on end/Drop
  - Exclusive encoder per entry during encoding
- Automatic recycling
  - After a configurable operation threshold, the current buffer is committed and replaced
  - Committed buffers are kept in an `in_flight` list so we can wait on them later
- Flushing
  - `flush()` commits pending work without blocking
  - `flush_and_wait()` waits for all in-flight buffers to complete
  -   Pool Drop calls `flush()` (non-blocking) to avoid long destructor stalls

**Performance impact**

- Multi-threaded: reduces contention by avoiding a single global lock on encoder acquisition. Selection is usually non-blocking
- Single-threaded: minimal overhead (pool size 1 behaves like the original)
- Memory: small constant increase (≈N command buffers; N defaults to 5)
- CPU: very low overhead; selection uses a small linear scan with atomics

**Compatibility**

- No API changes for callers using `Commands`. Internal wiring now uses the pool
- Tests cover pool creation and encoder acquisition
- Existing tests continue to pass


candle-metal-kernels/src/metal/command_buffer_pool.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// Commits any pending work on the current buffers, moves them to in-flight,
+    /// then waits on all in-flight buffers including those from prior recycles.
+    pub fn flush_and_wait(&self) -> Result<(), MetalKernelError> {
+        for entry in &self.pool {
+            // Ensure no active encoder is still encoding on this entry.
+            {
+                let _guard = entry
+                    .semaphore
+                    .wait_until(|s| matches!(s, CommandStatus::Available));
+            }
```
ivarflakstad: The guard is released at `}`. It feels to me like we should acquire the entry state lock before dropping the guard, no?

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -1,178 +1,62 @@
...
 impl Commands {
     pub fn new(command_queue: CommandQueue) -> Result<Self, MetalKernelError> {
-        let semaphore = Arc::new(CommandSemaphore::new());
-        let command_buffer = create_command_buffer(&command_queue, Arc::clone(&semaphore))?;
-        let command_buffer = Arc::new(RwLock::new(command_buffer));
-
         let compute_per_buffer = match std::env::var("CANDLE_METAL_COMPUTE_PER_BUFFER") {
             Ok(val) => val.parse().unwrap_or(50),
-            _ => 50,
+            _ => DEFAULT_CANDLE_METAL_COMPUTE_PER_BUFFER,
```
ivarflakstad, line 64: Let's use the const on both values of 50.

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -1,178 +1,62 @@
...
-                .semaphore
-                .wait_until(|s| matches!(s, CommandStatus::Available));
-            // Set status as encoding to block other threads from encoding to this commmand buffer
-            *guard = CommandStatus::Encoding;
-        }
-        // Notify after command status lock is released
-        self.semaphore.cond.notify_one();
+        let pool_size = match std::env::var("CANDLE_METAL_COMMAND_POOL_SIZE") {
+            Ok(val) => val.parse().unwrap_or(4),
+            _ => DEFAULT_CANDLE_METAL_COMMAND_POOL_SIZE,
```
ivarflakstad: Same here. Also if you set a value that can't be parsed you even get 4 instead of 5.

candle-metal-kernels/src/metal/command_buffer_pool.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        // Phase 2: Select the buffer with the least work and wait for it
+        let entry = self
+            .pool
+            .iter()
+            .min_by_key(|e| e.compute_count.load(Ordering::Acquire))
+            .ok_or(MetalKernelError::FailedToCreateResource(
+                "Command buffer pool is empty".to_string(),
+            ))?;
```
ivarflakstad: I wonder if it is technically more correct with `max_by_key` here (or perhaps a heuristic like closest to `compute_per_buffer / 2`), but there's too many moving parts for me to be sure.

My thinking is simply that In an optimal scenario (not factoring in waiting locks/semaphores) we hit the `compute_count >= compute_per_buffer` as quickly as possible, commit, and move on.
The opposite (bad) scenario is that we fill `N` command buffers `compute_count / N` full and only commit when `compute_count / N >= compute_per_buffer`.

The semaphores and locks completely change this dynamic however, so I think this is probably good. Just wanted to note it down in case it comes up later.

candle-metal-kernels/src/metal/command_buffer_pool.rs
```diff
@@ -0,0 +1,255 @@
...
+    ) -> Result<(bool, E), MetalKernelError>
+    where
+        F: FnOnce(&mut CommandBuffer) -> E,
+    {
+        let mut state = entry.state.lock()?;
+
+        let count = entry.compute_count.fetch_add(1, Ordering::Relaxed);
+        let flushed = count >= self.compute_per_buffer;
+
+        if flushed {
```
ivarflakstad: Nit. To me `flushed` in past tense implies "we have flushed" or "has been flushed", so for if-statements like this I prefer "should we flush?"
```suggestion
        let flush = count >= self.compute_per_buffer;

        if flush {
```

candle-metal-kernels/src/metal/command_buffer_pool.rs
```diff
@@ -0,0 +1,255 @@
...
+                    state.current.commit();
+                    let new_cb =
+                        create_command_buffer(&self.command_queue, Arc::clone(&entry.semaphore))?;
+                    let old_cb = std::mem::replace(&mut state.current, new_cb);
+                    state.in_flight.push(old_cb);
+                    entry.compute_count.store(0, Ordering::Release);
+                }
+
+                // Drain in_flight into a local vec to wait without holding the lock.
+                std::mem::take(&mut state.in_flight)
```
ivarflakstad: Nice 👌 
Could you also add a comment about how this leaves a default `Vec` in it's place? (I don't expect every reader to be aware of that)

candle-metal-kernels/src/metal/command_buffer_pool.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        Ok(())
+    }
+}
+
+impl Drop for CommandBufferPool {
+    fn drop(&mut self) {
+        let _ = self.flush();
+    }
+}
```
ivarflakstad: Two minor things here.
1. It seems a both a bit wasteful to create new command buffers etc here when we're going to be immediately dropping them.
2. Is `flush` enough, or should we `flush_and_wait`? I don't want to wait if we don't have to, but it also seems to me like dropping the pool implies that everything in the pool should be done.

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -1,178 +1,62 @@
...
+const DEFAULT_CANDLE_METAL_COMMAND_POOL_SIZE: usize = 5;
+
+/// Commands manager backed by a command buffer pool for improved CPU-GPU parallelism.
+///
+/// This struct maintains a pool of command buffers, allowing
+/// the pool to balance load across multiple buffers and improve GPU utilization.
+/// The API remains the same as the original single-buffer implementation, but the
+/// implementation now uses the pool internally.
+///
+/// Thread-safe: Can be shared across threads safely.
```
ivarflakstad: I'm not opposed more detailed documentation, but it shouldn't be a description of the changes from this PR.

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -1,178 +1,62 @@
...
-    /// command buffer
-    compute_count: AtomicUsize,
-    /// The maximum amount of [compute command encoder](https://developer.apple.com/documentation/metal/mtlcomputecommandencoder?language=objc) per [command buffer](https://developer.apple.com/documentation/metal/mtlcommandbuffer?language=objc)
-    compute_per_buffer: usize,
-    semaphore: Arc<CommandSemaphore>,
-    //capture: Option<Retained<MTLCaptureManager>>,
-    //timestamp_counter_set: Option<CounterSet>,
+    pool: Arc<CommandBufferPool>,
 }
+
```
ivarflakstad, line 54: `MetalDevice` already has `Commands` in an `Arc`, so we don't really need the buffer pool to be in an arc as well.
When we remove `Arc` here, it kinda becomes obvious that conceptually `CommandBufferPool` == `Commands`.

So I'd prefer if we moved the logic back into this file, as it's better wrt history, and I don't see a reason to have the actual implementation of `Commands` in another file.

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -1,178 +1,62 @@
...
+use objc2_metal::{MTLCommandQueue, MTLCounterSet};
+use std::sync::Arc;
 
 // Use Retained when appropriate. Gives us a more elegant way of handling memory (peaks) than autoreleasepool.
 // https://docs.rs/objc2/latest/objc2/rc/struct.Retained.html
 pub type CommandQueue = Retained<ProtocolObject<dyn MTLCommandQueue>>;
 pub type CounterSet = Retained<ProtocolObject<dyn MTLCounterSet>>;
 
+const DEFAULT_CANDLE_METAL_COMPUTE_PER_BUFFER: usize = 50;
+const DEFAULT_CANDLE_METAL_COMMAND_POOL_SIZE: usize = 5;
```
ivarflakstad, line 15: Is there a reason for chosing 5?
Testing / benches, seen in othe projects, or just gut feeling? ☺️

ivarflakstad: For reference we chose compute count = 50 because it performed well. We've gotten this verified later by mlx as they have a similar approach (though more detailed, they use different values for different hardware).

candle-metal-kernels/src/metal/command_buffer_pool.rs
```diff
@@ -0,0 +1,255 @@
...
+        &self,
+        entry: Arc<CommandBufferEntry>,
+        create_encoder: F,
+    ) -> Result<(bool, E), MetalKernelError>
+    where
+        F: FnOnce(&mut CommandBuffer) -> E,
+    {
+        let mut state = entry.state.lock()?;
+
+        let count = entry.compute_count.fetch_add(1, Ordering::Relaxed);
```
McPatate: Shouldn't ordering be `AcqRel` here? Or do we not care about ordering in this context?

McPatate: I'm asking since I see the use `Release` and `Acquire` on other stores and loads respectively

candle-metal-kernels/src/metal/command_buffer_pool.rs
McPatate: `flush` feels safer to be used in a `Drop`. If you were to need things to be completed you'd call `flush_and_wait` manually, so imo this makes sense

candle-metal-kernels/src/metal/command_buffer_pool.rs
ivarflakstad: For the compute count I think it should be as relaxed / non-blocking as possible.

It's used to auto-commit when we get >= 50 commands in a command buffer, so we don't care that much if the actual count is 50 or 51.

We also use it when we select the next available command buffer, but again doesn't matter that much if we select the least full or almost least full command buffer.

candle-metal-kernels/src/metal/command_buffer_pool.rs
anonenity: @ivarflakstad agreed this does feel wasteful. I'm wondering whether to avoid creating the replacement on flush and lazily allocate the buffer on the next acquisition. I guess there are trade-offs to both approaches. What are your thoughts?

ivarflakstad: I'd like this merged so I'm thinking slap a `// TODO: <something about avoiding redundant allocations before drop>` on it and we'll come back to it at a later point ☺️

ivarflakstad: > If you were to need things to be completed you'd call flush_and_wait manually, so imo this makes sense

Yeah and since we fully control when it is dropped I don't think we'll have any issues. It's inside an `Arc` in the device so it won't be dropped until we drop all tensors.

candle-metal-kernels/src/metal/command_buffer_pool.rs
anonenity: Good catch! Updated scope to hold guard during commit/swap and  `in_flight` drain

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -52,127 +27,242 @@ fn create_command_buffer(
...
             }
-            MTLCommandBufferStatus::Completed => {} // No action needed
+            MTLCommandBufferStatus::Completed => {}
             MTLCommandBufferStatus::Error => {
-                if let Some(error) = current.error() {
-                    return Err(MetalKernelError::CommandBufferError(error.to_string()));
-                }
+                let msg = cb
+                    .error()
+                    .map(|e| e.into_owned())
```
ivarflakstad:
```suggestion
                    .map(|e| e.to_string())
```

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -52,127 +27,242 @@ fn create_command_buffer(
...
-            // Ensure command buffer available.
-            let mut guard = self
+            let mut guard = entry
                 .semaphore
                 .wait_until(|s| matches!(s, CommandStatus::Available));
-            // Set status as encoding to block other threads from encoding to this commmand buffer
             *guard = CommandStatus::Encoding;
         }
-        // Notify after command status lock is released
-        self.semaphore.cond.notify_one();
```
ivarflakstad, line 105: At first I thought this was incorrect, but it's not. Why would we notify another thread that it is currently blocked?
We should only notify when we set the state to available or done.

Good.

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -52,127 +27,242 @@ fn create_command_buffer(
...
 
-        let (flush, command_buffer) = self.command_buffer()?;
-        let blit_command_encoder = command_buffer.blit_command_encoder();
+            let mut state = entry.state.lock()?;
 
-        Ok((flush, blit_command_encoder))
+            if entry.compute_count.load(Ordering::Acquire) > 0 {
+                self.commit_swap_locked(&entry, &mut state, 0)?;
```
ivarflakstad, line 215: Sleek

candle-metal-kernels/src/tests.rs
```diff
@@ -2400,3 +2421,73 @@ fn const_fill() {
...
+    let queue = device.new_command_queue().unwrap();
+    let commands = Commands::new(queue).unwrap();
+
+    let (_flush, encoder) = commands.command_encoder().unwrap();
+    drop(encoder);
+}
+
+#[test]
+fn commands_rotation_threshold() {
+    std::env::set_var("CANDLE_METAL_COMPUTE_PER_BUFFER", "2");
```
ivarflakstad, line 2437: iirc `set_var` is per process so it shouldn't pollute other tests etc, right?

candle-metal-kernels/src/tests.rs
```diff
@@ -2400,3 +2421,73 @@ fn const_fill() {
...
+        let (flush, encoder) = commands.command_encoder().unwrap();
+        flush_count += flush as usize;
+        drop(encoder);
+    }
+
+    assert!(flush_count >= 2);
+}
+
+#[test]
+fn commands_wait_flushes_all() {
```
ivarflakstad: Does this test cover more than the one above? Doesn't seem like it

candle-metal-kernels/src/tests.rs
```diff
@@ -2400,3 +2421,73 @@ fn const_fill() {
...
+            let (_flush, encoder) = c.command_encoder().unwrap();
+            drop(encoder);
+        }));
+    }
+
+    for h in handles {
+        h.join().unwrap();
+    }
+
+    commands.wait_until_completed().unwrap();
```
ivarflakstad, line 2480: I like the test. Not sure I understand the significance of this line.
Is it to ensure there are no errors reported via the command buffer status?
If so worth noting that I've never actually gotten an error through that channel hehe

candle-metal-kernels/src/metal/commands.rs
anonenity: haha thanks! Saw an opportunity to dedupe some logic

candle-metal-kernels/src/tests.rs
anonenity: Yes, env is set per process

candle-metal-kernels/src/tests.rs
anonenity: The idea here was to cover different behaviour...
`commands_rotation_threshold` validates rotation policy where `commands_wait_flushes_all` validates that the `flush_and_wait` path drains `in_flight` buffers correctly. Having said that, there is some overlap so happy to remove if you think it's redundant?

candle-metal-kernels/src/tests.rs
anonenity: Yes, if an encoder didn't end or command buffer failed, this line would hang or return an error rather than silently failing. Maybe not worth adding though?

candle-metal-kernels/src/tests.rs
ivarflakstad: Hmm no actually tests are ran in parallel so per process is still problematic.
It's a minor inconvenience though so I can come back to this later.

candle-metal-kernels/src/tests.rs
ivarflakstad: Yeah I understand. How about we simply added `commands.wait_until_completed().unwrap();` at the end of the one above? Maybe add a comment about what it verifies.

candle-metal-kernels/src/tests.rs
ivarflakstad: No I agree, we keep it 👍

ivarflakstad: Looking at these tests as a whole again they could technically be a single test if you use an atomic usize to track flushes 🤔 
I like the idea of verifying correct flushing behaviour in a multithreaded test scenario, but the test could become a bit noisy / hard to read.

anonenity: I guess my only concern with using a single test is that we lose clarity and isolation. For example, the concurrency test can hang if semaphore logic regresses, keeping it isolated prevents masking other failures. This way, when one fails, we know what broke. But, happy to use a single test if you think it will improve readability / noise?

candle-metal-kernels/src/tests.rs
anonenity: ah, good catch! There's a `temp_env` crate that provides scoped environment variable settings but i'm guessing you're not keen to add a new dependency for a few tests here?

ivarflakstad: Exactly

candle-metal-kernels/src/tests.rs
ivarflakstad: I'm happy with two tests (ref the other discussion)

candle-metal-kernels/src/metal/command_buffer_pool.rs
anonenity: I think you're right. Picking the entry with the minimum `compute_count` evens work across the pool which delays hitting `compute_per_buffer` on any single buffer. If our goal is to commit buffers as soon as possible (to keep the GPU queue fed), maybe we should add work to the buffer closest to threshold? I guess that approach also comes with its own drawbacks though.

ivarflakstad: Exactly. That's why a heuristic may be better.

But closest to cap seems logical to me, and good enough for this iteration.

candle-metal-kernels/src/tests.rs
anonenity: To confirm, add `commands.wait_until_completed().unwrap();` (with comment) to `commands_rotation_threshold` and drop `commands_wait_flushes_all`?

ivarflakstad: Yes please :)

candle-metal-kernels/src/metal/commands.rs
ivarflakstad: Just for clarity - I didn't mean to imply that the pool size should be 50, or any other number for that matter, I was simply curious if 5 was a special number in some way.

anonenity: Thanks for clarifying. I originally chose 5 as it seemed like a good balance, providing a few thread entries to reduce contention without lots of tiny command buffers. I did think 50 was excessive but I clearly misunderstood your comment completely! :) 

Also, interesting to know that compute count of 50 performs well. I've reset the pool size to 5, does that make sense as a default?

ivarflakstad: 5 seems reasonable 👌

PR #3188, opened 2025-11-14
Generic KvCache

ivarflakstad:
This should let us (and users of candle) easily switch out KvCache implementations in models.

I explored using `dyn KvCache` instead, but it quickly became a pain, so generics it is.

### DIFF
--- a/candle-core/src/cuda_backend/device.rs
+++ b/candle-core/src/cuda_backend/device.rs
@@ -94,6 +94,18 @@ impl CudaDevice {
         self.stream.memcpy_dtod(src, dst).w()
     }
 
+    pub fn memcpy_dtoh<
+        T: cudarc::driver::DeviceRepr,
+        Src: cudarc::driver::DevicePtr<T>,
+        Dst: cudarc::driver::HostSlice<T>,
+    >(
+        &self,
+        src: &Src,
+        dst: &mut Dst,
+    ) -> Result<()> {
+        self.stream.memcpy_dtoh(src, dst).w()
+    }
+
     pub fn memcpy_stod<
         T: cudarc::driver::DeviceRepr,
         Src: cudarc::driver::HostSlice<T> + ?Sized,
--- a/candle-core/src/metal_backend/device.rs
+++ b/candle-core/src/metal_backend/device.rs
@@ -125,7 +125,7 @@ impl MetalDevice {
     }
 
     pub fn command_encoder(&self) -> Result<ComputeCommandEncoder> {
-        let mut commands = self.commands.write().map_err(MetalError::from)?;
+        let commands = self.commands.write().map_err(MetalError::from)?;
         let (flush, command_encoder) = commands.command_encoder().map_err(MetalError::from)?;
         if flush {
             self.drop_unused_buffers()?
@@ -134,7 +134,7 @@ impl MetalDevice {
     }
 
     pub fn blit_command_encoder(&self) -> Result<BlitCommandEncoder> {
-        let mut commands = self.commands.write().map_err(MetalError::from)?;
+        let commands = self.commands.write().map_err(MetalError::from)?;
         let (flush, command_encoder) = commands.blit_command_encoder().map_err(MetalError::from)?;
         if flush {
             self.drop_unused_buffers()?
@@ -143,7 +143,7 @@ impl MetalDevice {
     }
 
     pub fn wait_until_completed(&self) -> Result<()> {
-        let mut commands = self.commands.write().map_err(MetalError::from)?;
+        let commands = self.commands.write().map_err(MetalError::from)?;
         commands.wait_until_completed().map_err(MetalError::from)?;
         Ok(())
     }
--- a/candle-core/src/quantized/cuda.rs
+++ b/candle-core/src/quantized/cuda.rs
@@ -46,24 +46,57 @@ fn pad(p: usize, q: usize) -> usize {
 fn quantize_q8_1(
     src: &CudaView<f32>,
     dst: &mut CudaSlice<u8>,
-    elem_count: usize,
+    k: usize,
     ky: usize,
     dev: &CudaDevice,
 ) -> Result<()> {
-    let kx = elem_count;
-    let kx_padded = pad(kx, MATRIX_ROW_PADDING);
+    let kx_padded = pad(k, MATRIX_ROW_PADDING);
     let num_blocks = ceil_div(kx_padded, CUDA_QUANTIZE_BLOCK_SIZE);
+
+    let total_rows = ky;
+    // Get Q8_1 metadata.
+    let q8_1_block_size = GgmlDType::Q8_1.block_size();
+    let q8_1_type_size = GgmlDType::Q8_1.type_size();
+
+    // Calculate the size of the output buffer in bytes.
+    let num_blocks_per_row = kx_padded / q8_1_block_size;
+    let dst_row_size_bytes = num_blocks_per_row * q8_1_type_size;
+
+    const CHUNK_SIZE: usize = 65535; // gridDim.y limit
     let func = dev.get_or_load_func("quantize_q8_1", &candle_kernels::QUANTIZED)?;
-    let cfg = cudarc::driver::LaunchConfig {
-        grid_dim: (num_blocks as u32, ky as u32, 1),
-        block_dim: (CUDA_QUANTIZE_BLOCK_SIZE as u32, 1, 1),
-        shared_mem_bytes: 0,
-    };
-    let mut builder = func.builder();
-    builder.arg(src);
-    builder.arg(dst);
-    barg!(builder, kx as i32, kx_padded as i32);
-    unsafe { builder.launch(cfg) }.w()?;
+
+    let mut rows_processed = 0;
+    while rows_processed < total_rows {
+        // --- calculate the number of rows for this chunk ---
+        let remaining_rows = total_rows - rows_processed;
+        // This is our gridDim.y, now <= 65535
+        let rows_in_chunk = std::cmp::min(CHUNK_SIZE, remaining_rows);
+
+        // --- slice the source (f32) tensor by elements ---
+        let src_start_elem = rows_processed * k;
+        let src_num_elems = rows_in_chunk * k;
+        let src_chunk = src.slice(src_start_elem..(src_start_elem + src_num_elems));
+
+        // --- slice the destination (u8) tensor by bytes ---
+        let dst_start_byte = rows_processed * dst_row_size_bytes;
+        let dst_num_bytes = rows_in_chunk * dst_row_size_bytes;
+        let dst_chunk = dst.slice(dst_start_byte..(dst_start_byte + dst_num_bytes));
+
+        let cfg = cudarc::driver::LaunchConfig {
+            grid_dim: (num_blocks as u32, rows_in_chunk as u32, 1),
+            block_dim: (CUDA_QUANTIZE_BLOCK_SIZE as u32, 1, 1),
+            shared_mem_bytes: 0,
+        };
+
+        let mut builder = func.builder();
+        builder.arg(&src_chunk);
+        builder.arg(&dst_chunk);
+        barg!(builder, k as i32, kx_padded as i32);
+        unsafe { builder.launch(cfg) }.w()?;
+
+        rows_processed += rows_in_chunk;
+    }
+
     Ok(())
 }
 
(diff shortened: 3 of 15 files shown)


candle-nn/src/kv_cache.rs
```diff
@@ -2,26 +2,43 @@
 //!
 use candle::{DType, Device, Result, Tensor};
 
+pub type DefaultKvCache = ConcatKvCache;
+
+pub trait KvCache {
+    type Mask;
+    fn new(dim: usize, max_seq_len: usize) -> Self;
+    fn append(&mut self, k: &Tensor, v: &Tensor) -> Result<(Tensor, Tensor)>;
```
McPatate, line 10: Does `append` insert into the cache?

McPatate: Probably worth adding some doc strings here

candle-nn/src/kv_cache.rs
```diff
@@ -2,26 +2,43 @@
...
+pub struct InnerCache {
     // all_data is an option on a Tensor, this makes it possible to only create the actual tensor
     // on the first call where the batch size is easily known.
     // Also this makes it safe to clone a KvCache that has been reset (as in it will not share
     // its internal state with the cloned instance).
     all_data: Option<Tensor>,
     dim: usize,
     current_seq_len: usize,
-    grow_by: usize,
+    increment: usize,
```
McPatate, line 31: why the change?

candle-nn/src/kv_cache.rs
```diff
@@ -117,8 +149,8 @@ impl KvCache {
     }
 
     pub fn append(&mut self, k: &Tensor, v: &Tensor) -> Result<(Tensor, Tensor)> {
-        self.k.append(k)?;
-        self.v.append(v)?;
+        self.k.append(&k.contiguous()?)?;
```
McPatate, line 152: Curious to know what the rationale for adding `contiguous` here is 👀

candle-nn/src/kv_cache.rs
```diff
@@ -2,26 +2,43 @@
...
+pub trait KvCache {
+    type Mask;
+    fn new(dim: usize, max_seq_len: usize) -> Self;
+    fn append(&mut self, k: &Tensor, v: &Tensor) -> Result<(Tensor, Tensor)>;
+    fn append_with_mask(
+        &mut self,
+        k: &Tensor,
+        v: &Tensor,
+        _mask: Option<&Self::Mask>,
+    ) -> Result<(Tensor, Tensor)> {
```
McPatate, line 16: What's the point of having `mask` being an `Option` here when we already have `append[_without_mask]`?

candle-nn/src/kv_cache.rs
```diff
@@ -414,34 +461,90 @@ impl IndicesAndMask {
...
 }
+impl KvCache for ScatteredKvCache {
+    type Mask = IndicesAndMask;
+
+    fn new(dim: usize, max_seq_len: usize) -> Self {
+        ScatteredKvCache::new(dim, max_seq_len)
+    }
+
+    fn append(&mut self, k: &Tensor, v: &Tensor) -> Result<(Tensor, Tensor)> {
+        self.append_with_mask(k, v, None)
```
McPatate, line 477: see here we could just call `candle::bail!` instead of having to do that in `append_with_mask`. feels a bit clunky

candle-nn/src/kv_cache.rs
```diff
@@ -2,26 +2,43 @@
...
+        _mask: Option<&Self::Mask>,
+    ) -> Result<(Tensor, Tensor)> {
+        self.append(k, v)
+    }
+    fn reset(&mut self);
+}
+
 #[derive(Debug, Clone)]
-pub struct Cache {
+pub struct InnerCache {
```
McPatate, line 23: I think `Cache` was ok as a name, although I get the sentiment

candle-transformers/src/models/quantized_gemma3.rs
```diff
@@ -133,14 +136,14 @@ struct LayerWeights {
...
-    kv_cache: Option<(Tensor, Tensor)>,
+    kv_cache: C,
 
     // Tracing
     span_attn: tracing::Span,
     span_mlp: tracing::Span,
 }
 
-impl LayerWeights {
+impl<C: KvCache> LayerWeights<C> {
```
McPatate, line 146: Does it need to be generic? It feels like we're expecting a specific behaviour from the cache

candle-transformers/src/models/quantized_gemma3.rs
```diff
@@ -247,17 +238,17 @@ impl LayerWeights {
 }
 
 #[derive(Debug, Clone)]
-pub struct ModelWeights {
+pub struct ModelWeights<C: KvCache = DefaultKvCache> {
```
McPatate, line 241: I assume this means "if not specified, use `DefaultKvCache`"?

candle-nn/src/kv_cache.rs
ivarflakstad: True.
I initially added optional mask to `append` and later realized it is better expressed as a separate fn.

candle-nn/src/kv_cache.rs
ivarflakstad: Yeah it's only used in one kv cache variant, and not even the default one, so it feels wrong that it should be _the_ `Cache`

candle-transformers/src/models/quantized_gemma3.rs
ivarflakstad: Sometimes you want the `KvCache` variant that strictly has the highest throughput. Sometimes you want one that is more careful about how it consumes memory. Etc

candle-transformers/src/models/quantized_gemma3.rs
ivarflakstad: Yes!
By default you get `ConcatKvCache`, which has the highest throughput but grows indefinitely (until you `reset()` in the model impl)

candle-nn/src/kv_cache.rs
ivarflakstad: It's the default behaviour in the model code I used as reference. Is there any reason it shouldn't call contiguous?
I could omit it but it improves performance for the cache impls I tested, and calling `contiguous` multiple times has almost zero cost (if tensor is already contiguous it just clones an `Arc`)

candle-nn/src/kv_cache.rs
ivarflakstad: I guess it can depend on the cache implementation, but typically you append to the kv cache. It's not like your typical cache ala hashmap where you insert by key.

candle-nn/src/kv_cache.rs
ivarflakstad: Aligns better with the new name

sempervictus:
Don't know how visible [attention.rs](https://github.com/guoqingbao/attention.rs) might be to the larger ecosystem but the KVCache implementation therein appears to work quite well for older and modern (NV at least) hardware and already has variable data type and offloaded scaling support. Given the tight coupling between attention and kvcache, is it worth considering the entire library as a candle ecosystem candidate (would need some "genericization" for ease of adoption by other projects)?

ivarflakstad:
Yep! I actually mentioned this PR to @guoqingbao just the other day for this very reason ☺️

guoqingbao:
> Yep! I actually mentioned this PR to @guoqingbao just the other day for this very reason ☺️

Thanks for including me.

I'm wondering whether we can pre-allocate a large KV-cache tensor and copy each new KV-cache slice into the existing tensor. This would avoid repeatedly concatenating and destroying tensors, especially as the context length grows significantly.

A skeleton:

```rust
use candle::{Result, Tensor, Device, DType};
use candle::index::IndexOp;

pub struct KvCache {
    pub k: Tensor,              // [max_seq_len, num_heads, head_dim]
    pub v: Tensor,              // [max_seq_len, num_heads, head_dim]
    pub max_seq_len: usize,
    pub cur_len: usize,
}

impl KvCache {
    /// Preallocate empty KV cache for one layer
    pub fn new(
        max_seq_len: usize,
        num_heads: usize,
        head_dim: usize,
        device: &Device,
        dtype: DType,
    ) -> Result<Self> {
        let shape = (max_seq_len, num_heads, head_dim);

        Ok(Self {
            k: Tensor::zeros(shape, dtype, device)?,
            v: Tensor::zeros(shape, dtype, device)?,
            max_seq_len,
            cur_len: 0,
        })
    }

    /// Insert a new KV slice for the next token.
    /// new_k/new_v shape: [num_heads, head_dim]
    pub fn insert(&mut self, new_k: &Tensor, new_v: &Tensor) -> Result<()> {
        let pos = self.cur_len;

        if pos >= self.max_seq_len {
            candle::bail!(
                "KV cache full — need resize or rotary window",
            );
        }

        // k[pos] = new_k
        self.k = self.k.assign((pos, .., ..), new_k)?;

        // v[pos] = new_v
        self.v = self.v.assign((pos, .., ..), new_v)?;

        self.cur_len += 1;
        Ok(())
    }
}
```
I think we may need to update the current SDPA attention to support this layout, or use the existing FlashAttention implementation, which accepts tensors shaped as `[num_blocks, block_size, num_kv_heads, head_size]`. Since `num_blocks * block_size == max_seq_len`, the `[max_seq_len, num_heads, head_dim]` layout can be viewed as `[num_blocks, block_size, num_kv_heads, head_size]`.

DrJesseGlass:
Just to add some context here — some of the impetus for this generic KvCache trait came from my benchmarks showing ConcatKvCache/Existing custom cache for Llama3 (no pre-allocation) was 2-5x faster on CUDA compared to the previous pre-allocated + slice approach, with the gap growing as sequence length increased (tested 300-2000 tokens). On CPU and WASM32, concat matched pre-allocation performance.
I was also surprised by this — pre-allocation should be better in theory. Just wanted to flag this before we move toward a pre-allocation design that might regress CUDA performance.
Separately, I'm working on a CPU-optimized interleaved attention implementation and want to make sure the trait stays flexible enough to support different layouts/strategies per backend.

sempervictus:
> Just to add some context here — some of the impetus for this generic KvCache trait came from my benchmarks showing ConcatKvCache/Existing custom cache for Llama3 (no pre-allocation) was 2-5x faster on CUDA compared to the previous pre-allocated + slice approach, with the gap growing as sequence length increased (tested 300-2000 tokens). On CPU and WASM32, concat matched pre-allocation performance. I was also surprised by this — pre-allocation should be better in theory. Just wanted to flag this before we move toward a pre-allocation design that might regress CUDA performance. Separately, I'm working on a CPU-optimized interleaved attention implementation and want to make sure the trait stays flexible enough to support different layouts/strategies per backend.

Interesting - i wonder how that concat performance thing pans out on older HW like V100s where random/async access isn't the same.

At a high level (with software architect hat on sideways while i grok machinery of this ecosystem), my hope is that _attention_ implementations [would reside within attention.rs](https://github.com/guoqingbao/attention.rs/issues/2) and the various KV cache implementations being nearly integral to attention mechanics would live there or in a tightly-coupled sister crate, providing a common abstraction to consumers in the ecosystem. CUDA and [BlackHole RISC-V accelerators](https://tenstorrent.com/hardware/blackhole) don't handle data in the same layouts or blocks necessarily so having an interoperable KV construct which would map to attention implementations relevant to the layout of different hardware classes/their compiler or input abstractions should allow the ecosystem to matrix-test/benchmark/qualify accuracy across those as new hardware classes become supported and attn implementations and cache layouts are added. Since @guoqingbao works with all sorts of hardware, that library is less likely (IMO) to suffer from the "new and shiny causes performance regressions on actually available" problem we often see w/ CUDA itself while wider adoption would provide the FOSS focus needed to help keep it in sync w upstream for `cudarc` and the like (@guoqingbao has graph working and working well enough to matter for at least our own internal use - upstream seems to have some issues bearing integration effort).

Reasoning behind that train of thought is that _attention efficacy_ is absolutely critical to output quality and no amount of performance benefit adds up to the time loss (or in the worst case, loss of life - critical paths exist and they are adopting) resulting from "workslop." We just did a panel w/ NV and one of the CSPs in Boston talking about what it will take for adoption in _healthcare_ and everyone at the event with industry background concluded that scientific-grade reproducibility and accuracy are prerequisite to any formal adoption of language models and attention-based architectures in decision-making paths of research or healthcare in lifesci. Some use-cases effectively don't care about performance when the criticality of results mandates getting it right the first time.

guoqingbao:
> Just to add some context here — some of the impetus for this generic KvCache trait came from my benchmarks showing ConcatKvCache/Existing custom cache for Llama3 (no pre-allocation) was 2-5x faster on CUDA compared to the previous pre-allocated + slice approach, with the gap growing as sequence length increased (tested 300-2000 tokens

Thanks for the additional context — that’s really helpful.
My earlier comment was a bit off-topic for the current PR. I agree that the generic KvCache trait is a good step toward supporting multiple backends and layouts cleanly.
The pre-allocation idea I mentioned was more of a long-term thought about potential strategies we might revisit later, not something I’m suggesting we shift to right now — especially given your CUDA benchmarks showing clear advantages for the concat-based approach as sequence length grows. It’s good to know the empirical results don’t match the theoretical expectation there.
For the purposes of this PR, keeping the trait flexible enough to accommodate future backend-specific optimizations (like your interleaved CPU attention work) seems like the right priority. Happy to move forward with the current design.

ivarflakstad:
> ConcatKvCache/Existing custom cache for Llama3 (no pre-allocation) was 2-5x faster on CUDA compared to the previous pre-allocated + slice approach

Just for full transparency - we have already had concat based kv cache in several models for a while ([ref](https://github.com/huggingface/candle/blob/95ea4538c3527f7cf001559281f45fe3ddaab359/candle-transformers/src/models/llama.rs#L302)), it just wasn't moved into its own reusable struct. Which was very much needed! 👍 

Most of the latency of the current `KvCache` is due to repeatedly allocating new buffers. If we pre-allocate a large enough buffer then using a in-place approach is going to outpace concat. The bottleneck will move from memory allocation to the (for example) `slice_set` kernel.

PR #3191, opened 2025-11-17
Add Provence model implementation

matthewhaynesonline:
Add implementation for https://huggingface.co/naver/provence-reranker-debertav3-v1. This is still a WIP, but I wanted to gauge interest before going too far.

### Notes
- Provence has a [CC Non Commercial license](https://huggingface.co/naver/provence-reranker-debertav3-v1/blob/main/Provence_LICENSE.txt)
  - Not sure if that's against candle conventions to implement models with restricted licenses
- Since it's based on DeBERTa-v3, the core implementation is pretty easy (forward) _however_ the original py transformers implementation provides a [custom process function](https://huggingface.co/naver/provence-reranker-debertav3-v1/blob/main/modeling_provence.py#L104) (and thus requires remote code)
  - Since by convention the candle models implement forward but not full pipelines, I wasn't sure if adding a process function in the model would be a bad choice.
  - For now it's there, but feature flagged
- I started by looking at the DeBERTa example and the XLM-RoBERTa reranking and tried to go from there.
- There's other versions ([multilingual](https://huggingface.co/naver/xprovence-reranker-bgem3-v1)) and [an LLM approach](https://huggingface.co/papers/2504.07109), but this particular model was a good fit for a use case of mine, so that's why I've focused on it
- If there's potential for approving this, I can reach out to the original authors on HF to get their feedback as well 

Thank you!

### DIFF
--- a/candle-examples/examples/provence/main.rs
+++ b/candle-examples/examples/provence/main.rs
@@ -152,7 +152,7 @@ fn main() -> Result<()> {
             // Forward only
             println!("Running forward pass only");
 
-            let input_text = model.format_input(question, context);
+            let input_text = ProvenceModel::format_input(question, context);
 
             let encoding = tokenizer
                 .encode(input_text, true)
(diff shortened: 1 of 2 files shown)


matthewhaynesonline:
https://github.com/matthewhaynesonline/provence-rs

PR #3192, opened 2025-11-17
Add `cublas_handle` api, update safetensors

EricLBuehler:
- Add `cublas_handle` api
- Update safetensors 0.4.0 -> 0.6.0
- Add more quantized `from_data` apis


.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
+    "rust-analyzer.cargo.features": ["cuda"],
```
ivarflakstad: We should probably just gitignore `.vscode`

candle-core/src/quantized/mod.rs
```diff
@@ -32,6 +32,22 @@ use half::{bf16, f16};
 
 pub use k_quants::GgmlType;
 
+fn as_t_slice<T>(data: Cow<'_, [u8]>) -> &[T] {
```
ivarflakstad, line 35: This looks sound to me, especially since we're dealing with `u8`, but in the future I'd like to use [zerocopy](https://docs.rs/zerocopy/latest/zerocopy/) for such transmutes. Not just here of course.

I can't imagine this would compile unless rust correctly infers that the slice has the same lifetime as the borrowed variant of `Cow<'_, ...>`.

candle-core/src/quantized/mod.rs
```diff
@@ -63,6 +79,46 @@ pub enum QStorage {
 }
 
 impl QStorage {
+    pub fn from_data(data: Cow<'_, [u8]>, device: &Device, dtype: GgmlDType) -> Result<Self> {
+        match device {
+            Device::Cpu => Ok(Self::Cpu(dtype.from_data(data))),
+            Device::Metal(d) => match dtype {
+                GgmlDType::F32 => metal::load_quantized(d, as_t_slice::<f32>(data)),
```
ivarflakstad, line 86: It would be nice if we could use an associated type here and get rid of the match, but meh.
This is fine.

ivarflakstad (approved):
Looks good. I'd drop the vscode setting before merging though :)

.vscode/settings.json
EricLBuehler: Fixed! Removed .vscode (it is a gitignore now)

candle-core/src/quantized/mod.rs
EricLBuehler: Sounds great. We should absolutely use `zerocopy` for some of these transmutes. Will put in a PR later today for that!

> I can't imagine this would compile unless rust correctly infers that the slice has the same lifetime as the borrowed variant of Cow<'_, ...>.
👍

PR #3193, opened 2025-11-17
Add initial support for imatrix quantization

EricLBuehler:
Support:
- quantizing with an imatrix
- loading imatrix files

Does not support any of the newer i-quants, only imatrix quantization of existing k-quants.

### DIFF
--- a/candle-core/tests/quantized_tests.rs
+++ b/candle-core/tests/quantized_tests.rs
@@ -1,5 +1,9 @@
 use candle_core::{
-    DType, Device, IndexOp, Module, Result, Tensor, Var, bail, quantized::{self, GgmlDType}, test_device, test_utils::to_vec2_round
+    bail,
+    quantized::{self, GgmlDType},
+    test_device,
+    test_utils::to_vec2_round,
+    DType, Device, IndexOp, Module, Result, Tensor, Var,
 };
 use quantized::{k_quants, GgmlType};
 use rand::prelude::*;


candle-core/src/quantized/cuda.rs
```diff
@@ -44,26 +44,59 @@ fn pad(p: usize, q: usize) -> usize {
 }
 
 fn quantize_q8_1(
-    src: &CudaView<f32>,
+    src: &CudaSlice<f32>,
```
ivarflakstad: If src is not mutated nor owned then `CudaView` is fine, no?
`CudaSlice<T>` -> `Vec<T>`
`CudaView<T>`-> `&[T]`

ivarflakstad: I guess same question applies for all the view -> slice changes.
Admittedly weird naming by `cudarc`. Could have used `CudaVec<T>` and `CudaSlice<T>` instead.

candle-core/src/quantized/cuda.rs
```diff
@@ -477,6 +510,87 @@ impl QCudaStorage {
         Ok(())
     }
 
+    pub fn quantize_imatrix(
+        &mut self,
+        src: &CudaStorage,
+        imatrix_weights: &[f32],
+        n_per_row: usize,
+    ) -> Result<()> {
+        // Run the quantization on cpu.
```
ivarflakstad, line 519: Ideally we would do imatrix quantization on gpu as well, or am I missing something?

candle-core/src/quantized/cuda.rs
```diff
@@ -643,7 +770,7 @@ mod test {
         xs.quantize(&CudaStorage::wrap_cuda_slice(y.clone(), dev.clone()))?;
         let cuda_storage = mul_mat_vec_via_q8_1(
             &xs.data,
-            &y.slice(..),
+            &y,
```
ivarflakstad: Ok, this could make sense. Does `.slice(..)` have a cost?
If so we could have used `.as_view()` instead, no range bounds needed.

candle-core/src/quantized/gguf_file.rs
```diff
@@ -1,5 +1,6 @@
 //! Support for the [GGUF file format](https://github.com/philpax/ggml/blob/gguf-spec/docs/gguf.md).
 //!
+//! Spec: https://github.com/philpax/ggml/blob/gguf-spec/docs/gguf.md
```
ivarflakstad: Deleted. Let's use the official instead?
```suggestion
//! Spec: https://github.com/ggml-org/ggml/blob/master/docs/gguf.md
```

candle-core/src/quantized/imatrix_file.rs
```diff
@@ -0,0 +1,85 @@
+use std::collections::HashMap;
+use std::fs::File;
+use std::io::{Cursor, Read};
+use std::path::Path;
+
+use byteorder::{LittleEndian, ReadBytesExt};
+
+use crate::Result;
+
+pub fn load_imatrix<P: AsRef<Path>>(fname: P) -> Result<HashMap<String, Vec<f32>>> {
```
ivarflakstad, line 10: 👌

candle-core/src/quantized/metal.rs
```diff
@@ -35,8 +35,11 @@ impl QMetalStorage {
...
     pub fn dequantize(&self, elem_count: usize) -> Result<MetalStorage> {
         use crate::quantized::k_quants::GgmlType;
-        let buffer = self.device.allocate_buffer(self.buffer.length())?;
-        let blit = self.device.blit_command_encoder()?;
+
+        let buffer = self.device.new_buffer_managed(self.buffer.length())?;
+        let command_buffer = self.device.command_buffer()?;
+        command_buffer.set_label("to_cpu");
+        let blit = command_buffer.new_blit_command_encoder();
         blit.set_label("blit_to_cpu");
```
ivarflakstad: I think this change should simply be reverted

candle-core/src/quantized/metal.rs
```diff
@@ -46,63 +49,63 @@ impl QMetalStorage {
         match self.dtype {
             GgmlDType::F32 => {
                 let vec: Vec<f32> = read_to_vec(&buffer, block_len);
-                f32::to_float(&vec, &mut out);
+                f32::to_float(&vec, &mut out)?;
```
ivarflakstad: These as well it seems

candle-core/src/quantized/metal.rs
```diff
@@ -127,8 +130,62 @@ impl QMetalStorage {
...
+        }
+
+        let buffer = self.device.new_buffer_with_data(&qcpu_storage.data()?)?;
+        self.buffer = buffer;
+        Ok(())
+    }
+
     pub fn storage_size_in_bytes(&self) -> usize {
-        self.buffer.length()
+        self.buffer.length() as usize
```
ivarflakstad: This and all below in this file should also be reverted

candle-core/tests/quantized_tests.rs
```diff
@@ -20,9 +20,7 @@ fn test_matmul(
     (b, m, n, k): (usize, usize, usize, usize),
     dtype: GgmlDType,
 ) -> Result<()> {
-    if (device.is_cuda() || device.is_metal())
-        && (dtype == GgmlDType::Q8_1 || dtype == GgmlDType::Q8K)
-    {
+    if device.is_metal() && (dtype == GgmlDType::Q8_1 || dtype == GgmlDType::Q8K) {
```
ivarflakstad: Does this PR add cuda matmul for these two types? Otherwise the check should remain.

candle-core/src/quantized/mod.rs
```diff
@@ -32,6 +32,22 @@ use half::{bf16, f16};
 
 pub use k_quants::GgmlType;
 
+fn as_t_slice<T>(data: Cow<'_, [u8]>) -> &[T] {
```
ivarflakstad, line 37: déjà vu 👀

candle-core/src/quantized/cuda.rs
EricLBuehler: I think `ggml` also runs it on the cpu

candle-core/src/quantized/cuda.rs
EricLBuehler: Nice! Addressed.

`.slice(..)` doesnt have a cost but it's cleaner to use `.as_view()` for sure.

candle-core/tests/quantized_tests.rs
EricLBuehler: Fixed!

candle-core/src/quantized/cuda.rs
ivarflakstad: It's not a one off calculation, is it?

candle-core/src/quantized/metal.rs
```diff
@@ -35,9 +35,9 @@ impl QMetalStorage {
 
     pub fn dequantize(&self, elem_count: usize) -> Result<MetalStorage> {
         use crate::quantized::k_quants::GgmlType;
+
         let buffer = self.device.allocate_buffer(self.buffer.length())?;
         let blit = self.device.blit_command_encoder()?;
-        blit.set_label("blit_to_cpu");
```
ivarflakstad, line 40: Do we want to get rid of this label?

candle-core/src/quantized/cuda.rs
EricLBuehler: Yeah it's a one time operation, just on the weights.

candle-core/src/quantized/k_quants.rs
```diff
@@ -31,6 +32,17 @@ pub trait GgmlType: Sized + Clone + Send + Sync {
     }
     fn to_float(xs: &[Self], ys: &mut [f32]);
     fn from_float(xs: &[f32], ys: &mut [Self]);
+    fn from_float_imatrix(
+        _xs: &[f32],
+        _ys: &mut [Self],
+        _imatrix_weights: &[f32],
+        _n_per_row: usize,
+    ) -> Result<()> {
```
ivarflakstad: We don't need `Result` here if there is no `Err` case.

candle-core/src/quantized/k_quants.rs
```diff
@@ -1690,7 +2028,93 @@ impl GgmlType for BlockQ6K {
...
+    fn from_float_imatrix(
+        xs: &[f32],
+        ys: &mut [Self],
+        imatrix_weights: &[f32],
+        n_per_row: usize,
+    ) -> Result<()> {
+        debug_assert_eq!(
+            xs.len(),
+            ys.len() * Self::BLCK_SIZE,
+            "quantize_row_q6k: size mismatch {} {} {}",
```
ivarflakstad: Wrong name?

ivarflakstad:
Could we add a imatrix kquant test?
Doesn't need to be complicated and cover all edge cases, just prove that it works as expected.

candle-core/src/quantized/k_quants.rs
EricLBuehler: Fixed

candle-core/src/quantized/metal.rs
EricLBuehler: Nope, fixed!

candle-core/src/quantized/k_quants.rs
EricLBuehler: Made a more detailed error

EricLBuehler:
> Could we add a imatrix kquant test?

Added it!

The tests are modeled after the algorithm used by `llama.cpp`. For a complete rust implementation, see: https://github.com/EricLBuehler/mistral.rs/blob/2bcf0e9e3a0c688786cfa3142e2b18844a7f3447/mistralrs-quant/src/imatrix.rs#L43

This might be good to integrate as a standalone feature at some time.

PR #3196, opened 2025-11-17
Add more misc. changes from candle fork

EricLBuehler:
- `indexed_moe_forward` (fast path for ggml quants)
- Improved usability of `Context`
- Add full attn support for Metal SDPA
- Fix bug w/ FlashAttn f16
- Add necessary metal `Device` apis


candle-core/src/metal_backend/device.rs
```diff
@@ -167,6 +167,23 @@ impl MetalDevice {
         self.allocate_buffer(size)
     }
 
+    /// Creates a new private buffer (not necessarily zeroed).
+    ///
+    /// This is intentionally not in the Metal buffer pool to allow the efficient implementation of persistent buffers.
+    pub fn new_private_buffer(
```
ivarflakstad, line 181: I agree that this is nice to have, but I think we should name it something other than private buffer since that already means something for metal buffers (only available on gpu, [ref](https://developer.apple.com/documentation/metal/mtlstoragemode/private)).
We don't want to use actual metal private buffers as that isn't supported on iOS.

How about `new_unpooled_buffer` or `new_persistent_buffer`? :)

candle-core/src/quantized/cuda.rs
```diff
@@ -373,7 +373,125 @@ fn mul_mat_via_q8_1(
...
+                input_l.shape(), //[batch, topk or 1, k]
+                &ids_storage.slice(0..),
+                ids_l.shape(), //[batch, topk]
+                &self.device,
+            )
+        } else {
+            crate::bail!(
+                "The given quantized dtype {:?} is not supported for indexed_moe_forward!",
+                self.dtype()
+            );
```
ivarflakstad, line 524: Just thinking out loud here. It would be nice to have automatic fallback to an approach that isn't as optimized, but still valid. Perhaps returning `Result<Option<(CudaStorage, crate::Shape)>>` is a decent starting point?
If `None` then fallback?

Not thinking we add this in this PR ofc.

candle-core/src/error.rs
```diff
@@ -299,40 +308,85 @@ pub fn zip<T, U>(r1: Result<T>, r2: Result<U>) -> Result<(T, U)> {
...
+    fn with_context<C, F>(self, f: F) -> std::result::Result<T, Error>
     where
-        C: std::fmt::Display + Send + Sync + 'static,
+        C: Display + Send + Sync + 'static,
         F: FnOnce() -> C;
 }
 
-impl<T> Context<T> for Option<T> {
-    fn context<C>(self, context: C) -> Result<T>
+impl<T, E> Context<T, E> for std::result::Result<T, E>
```
ivarflakstad, line 335: I'm a big fan of this 👌 
sleek

candle-core/src/tensor.rs
```diff
@@ -2768,6 +2813,49 @@ impl Tensor {
         }
         Ok(result)
     }
+
+    /// Returns a view of which contains all slices of size `size` from self tensor in the dimension
+    /// `dim` and stepped by `step`.
+    pub fn unfold<D: Dim>(&self, dim: D, size: usize, step: usize) -> Result<Self> {
```
ivarflakstad, line 2819: 👌

candle-flash-attn/src/lib.rs
```diff
@@ -487,9 +487,9 @@ impl FlashAttnVarLen {
             None => candle::bail!("seqlens_k has to be contiguous"),
         };
 
-        let q = q.as_cuda_slice::<f16>()?;
-        let k = k.as_cuda_slice::<f16>()?;
-        let v = v.as_cuda_slice::<f16>()?;
+        let q = q.as_cuda_slice::<T>()?;
+        let k = k.as_cuda_slice::<T>()?;
+        let v = v.as_cuda_slice::<T>()?;
```
ivarflakstad, line 492: So many tasty bits in this PR! 🧑‍🍳

candle-kernels/src/quantized.cu
```diff
@@ -4329,3 +4329,209 @@ extern "C" __global__ void
...
+ * by an index.
+ *
+ * Parallelization Strategy:
+ * - The grid is 2D: gridDim.y corresponds to the task index, and gridDim.x corresponds to the row blocks of the output matrix.
+ * - `blockIdx.y`: Identifies which task to perform from the batch (`0` to `total_tasks - 1`).
+ * - `blockIdx.x`: Used internally by `mul_mat_vec_q` to parallelize the dot products across the rows of the weight matrix.
+ *
+ * @author
+ *   Guoqing Bao
+ *   Part of the project: https://github.com/guoqingbao/vllm.rs/
```
ivarflakstad, line 4348: @guoqingbao 🐐 

If you don't mind the extra work I'd actually like if you recommitted this file with him as co-author 👍 
Seems fair ;)

candle-metal-kernels/src/kernels/sdpa.rs
ivarflakstad, line 1: ❤️

candle-metal-kernels/src/metal/device.rs
```diff
@@ -93,4 +93,12 @@ impl Device {
...
         Ok(raw)
     }
+
+    pub fn recommended_max_working_set_size(&self) -> usize {
+        self.as_ref().recommendedMaxWorkingSetSize() as usize
+    }
+
+    pub fn current_allocated_size(&self) -> usize {
+        self.as_ref().currentAllocatedSize()
+    }
```
ivarflakstad, line 103: These will be very useful in conjunction with #3197 👌

candle-core/src/error.rs
```diff
@@ -299,40 +308,85 @@ pub fn zip<T, U>(r1: Result<T>, r2: Result<U>) -> Result<(T, U)> {
...
     where
-        C: std::fmt::Display + Send + Sync + 'static,
+        C: Display + Send + Sync + 'static,
         F: FnOnce() -> C,
     {
         match self {
-            Some(v) => Ok(v),
-            None => Err(Error::UnwrapNone.context(f()).bt()),
+            Some(ok) => Ok(ok),
+            None => Err(Error::msg(context())),
```
ivarflakstad: I'd like to keep the backtrace if possible. Seems to me like in this (and `fn context`) we lose bt, correct?  🤔

candle-core/src/error.rs
EricLBuehler: No credit to me, all credit to `anyhow` as per the comment :)

candle-core/src/tensor.rs
EricLBuehler: 😄

candle-flash-attn/src/lib.rs
EricLBuehler: Yeah! Surprised this one persisted for so long.

candle-kernels/src/quantized.cu
EricLBuehler: Would be awesome!

candle-metal-kernels/src/kernels/sdpa.rs
EricLBuehler: 🚀

candle-metal-kernels/src/metal/device.rs
EricLBuehler: Absolutely! Fwiw this are used to implement the automatic device mapping/checks to make sure models actually fit on metal w/ mistral.rs

candle-core/src/error.rs
EricLBuehler: Yeah, fixed for all cases.

candle-core/src/metal_backend/device.rs
EricLBuehler: Actually, this was a mistake on my part. The correct behavior that I intended for this function is to have:
- `private` if not on iOS
- `shared`/`RESOURCE_OPTIONS` if on iOS

EricLBuehler:
Addressed the review comments, the `new_private_buffer` method is now implemented correctly.

candle-core/src/quantized/cuda.rs
EricLBuehler: This might work, the issue is that effectively `indexed_moe_forward` is a grouped gemm so we'd need existing infrastructure to run a grouped gemm.

Regardless, providing a grouped gemm functionality will be very useful!

candle-core/src/metal_backend/device.rs
ivarflakstad: I see. Could I ask why you want it to be private?
According to Apple's documentation there is no performance benefit, so private is usually used when you want to ensure that the cpu does not have access to the buffer for some specific reason. I'd wager a guess this kind of behaviour is frequently used in gaming.

candle-core/src/error.rs
ivarflakstad: Credit to you for recognizing good code. And now credit to me as well.

candle-core/src/metal_backend/device.rs
```diff
@@ -65,6 +65,14 @@ pub const RESOURCE_OPTIONS: MTLResourceOptions =
 //| MTLResourceOptions::HazardTrackingModeUntracked.bits(),
 //);
 
+// Resource options used for `new_private_buffer`. This uses `private` where supported.
+#[cfg(target_os = "ios")]
+pub const PRIVATE_RESOURCE_OPTIONS: MTLResourceOptions =
+    objc2_metal::MTLResourceOptions(MTLResourceOptions::StorageModeShared.bits());
+#[cfg(not(target_os = "ios"))]
+pub const PRIVATE_RESOURCE_OPTIONS: MTLResourceOptions =
+    objc2_metal::MTLResourceOptions(MTLResourceOptions::StorageModePrivate.bits());
```
ivarflakstad: While we're still discussing private/shared I just wanted to point out that these could be written as
```suggestion
#[cfg(target_os = "ios")]
pub const PRIVATE_RESOURCE_OPTIONS: MTLResourceOptions = MTLResourceOptions::StorageModeShared;
#[cfg(not(target_os = "ios"))]
pub const PRIVATE_RESOURCE_OPTIONS: MTLResourceOptions = MTLResourceOptions::StorageModePrivate;
```

Only reason I defined `RESOURCE_OPTIONS` the way I did is because I eventually want to apply `| MTLResourceOptions::HazardTrackingModeUntracked.bits()` to the value (which is what mlx does) ☺️

EricLBuehler: Sounds good, switchted to the enum. For now it's cleaner and we can add that later.

candle-core/src/metal_backend/device.rs
ivarflakstad: Actually I seem to be spreading misinformation:
> Metal may apply additional optimizations to private resources that aren’t allowed on shared or managed resources.

ivarflakstad: https://developer.apple.com/documentation/metal/mtlstoragemode/private

haricot:
for ci ubuntu, the linker seems to have crashed due to lack of memory.
maybe possible resolution:

1)  workflow .github/workflows/rust-ci.yml:
```
      # Add lld install, optional if lld is already present on runners
      - name: Install lld (Linux only)
        if: runner.os == 'Linux'
        run: sudo apt-get update && sudo apt-get install -y lld

      # The change: Add RUSTFLAGS for Linux to use linker-features
      - name: Run tests (with lld on Linux)
        if: runner.os == 'Linux'
        env:
          RUSTFLAGS: "-Clinker-features=-lld"
        run: cargo test --workspace

      # Existing Windows and Mac steps (unchanged)
      - name: Run tests (Windows & macOS)
        if: runner.os != 'Linux'
        run: cargo test --workspace
```

2) add  : .cargo/config.toml  
`[target.x86_64-unknown-linux-gnu] rustflags = ["-Clinker-features=-lld"]`

ivarflakstad (approved):
Lgtm! 🔥
Same wrt CI here as well

ivarflakstad:
@haricot yeah we can try setting the flag in CI.
I'd prefer to not have candle be opinionated wrt the linker, so I'd rather not add it to the config if we can avoid it.
I assume there are some (possibly obscure) reasons why it is not the default.

PR #3197, opened 2025-11-17
Metal: bound temporary buffer cache and prevent runaway memory usage on large softmax/broadcast/matmul workloads

TimmyOVO:
## Description:

On the Metal backend, large transformer/VLM workloads (e.g. Dots‑style OCR with a heavy vision tower and Qwen2‑style text tower) can cause the process RSS on macOS to grow to tens or even hundreds of GiB during a single forward pass, even
  though the model’s working set should fit comfortably in memory.

I use `Instruments` traces on a Dots‑style model show that the bulk of the “resource size” comes from a chain of large tensor ops:

  - `candle_nn::ops::softmax` on 3D tensors `[batch * heads, seq_len, total_len]`
  - the underlying elementwise ops:
    - `Tensor::max_keepdim`
    - `Tensor::broadcast_sub`
    - `Tensor::exp`
    - `Tensor::sum_keepdim`
    - `Tensor::broadcast_div`
  - plus large `Tensor::matmul` and `broadcast_mul` on matching shapes

<details>

  <summary>Click to expand raw data</summary>

```
Resource Size	Self Resource Size	Symbol Names
390.07 GiB  66.5%	0 Bytes	                                       _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h56c4e8aa02cfcf3b
250.01 GiB  42.6%	0 Bytes	                                        candle_nn::ops::softmax::hf1a7e8c20702ed6b
84.00 GiB  14.3%	0 Bytes	                                         candle_core::tensor::Tensor::broadcast_div::h1f0e384f7eef2e07
84.00 GiB  14.3%	0 Bytes	                                          candle_core::tensor::Tensor::div::h9a2bcb60b7c25bce
84.00 GiB  14.3%	0 Bytes	                                           candle_core::storage::Storage::binary_impl::hfda09b175c7c7058
84.00 GiB  14.3%	0 Bytes	                                            _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h35f1c58c739baa69
84.00 GiB  14.3%	0 Bytes	                                             candle_core::metal_backend::MetalStorage::binary::h8a4fc6643c45827f
84.00 GiB  14.3%	0 Bytes	                                              _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h13fc16f7d7a15a9a
84.00 GiB  14.3%	0 Bytes	                                               candle_core::metal_backend::device::MetalDevice::allocate_buffer::h87e110be93f79a4d
84.00 GiB  14.3%	0 Bytes	                                                candle_metal_kernels::metal::device::Device::new_buffer::hdff57d7c15bf891f
84.00 GiB  14.3%	0 Bytes	                                                 objc2_metal::generated::__MTLDevice::MTLDevice::newBufferWithLength_options::hc27f4530b25513c7
84.00 GiB   0.0%	0 Bytes	                                                  _$LT$MethodFamily$u20$as$u20$objc2..__macro_helpers..msg_send_retained..MsgSend$LT$Receiver$C$Return$GT$$GT$::send_message::h2d47315393e8e979
84.00 GiB   0.0%	0 Bytes	                                                   objc2::runtime::message_receiver::MessageReceiver::send_message::h81dfd419189b6a37
84.00 GiB   0.0%	0 Bytes	                                                    objc2::runtime::message_receiver::msg_send_primitive::send::hc0961fbfcac6b629
84.00 GiB   0.0%	0 Bytes	                                                     _$LT$$LP$A$C$B$RP$$u20$as$u20$objc2..encode..EncodeArguments$GT$::__invoke::hd2941f8e31fd9e9b
84.00 GiB   0.0%	0 Bytes	                                                      -[AGXBuffer initWithDevice:length:alignment:options:isSuballocDisabled:pinnedGPULocation:]
84.00 GiB   0.0%	0 Bytes	                                                       -[AGXBuffer(Internal) initWithDevice:length:alignment:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
84.00 GiB   0.0%	0 Bytes	                                                        -[AGXBuffer(Internal) initWithDevice:length:alignment:pointerTag:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
84.00 GiB   0.0%	0 Bytes	                                                         -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:args:argsSize:deallocator:]
84.00 GiB   0.0%	0 Bytes	                                                          -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:placementSparsePageSize:placementSparseResidencyBytes:args:argsSize:deallocator:]
84.00 GiB   0.0%	84.00 GiB	                                                           __kdebug_trace64
84.00 GiB  14.3%	0 Bytes	                                         candle_core::tensor::Tensor::exp::hadf5e168faedee1f
84.00 GiB  14.3%	0 Bytes	                                          candle_core::storage::Storage::unary_impl::h736da20777c1466c
84.00 GiB  14.3%	0 Bytes	                                           _$LT$candle_core..metal_backend..MetalStorage$u20$as$u20$candle_core..backend..BackendStorage$GT$::unary_impl::he570d8aad3e1c738
84.00 GiB  14.3%	0 Bytes	                                            _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h13fc16f7d7a15a9a
84.00 GiB  14.3%	0 Bytes	                                             candle_core::metal_backend::device::MetalDevice::allocate_buffer::h87e110be93f79a4d
84.00 GiB  14.3%	0 Bytes	                                              candle_metal_kernels::metal::device::Device::new_buffer::hdff57d7c15bf891f
84.00 GiB  14.3%	0 Bytes	                                               objc2_metal::generated::__MTLDevice::MTLDevice::newBufferWithLength_options::hc27f4530b25513c7
84.00 GiB   0.0%	0 Bytes	                                                _$LT$MethodFamily$u20$as$u20$objc2..__macro_helpers..msg_send_retained..MsgSend$LT$Receiver$C$Return$GT$$GT$::send_message::h2d47315393e8e979
84.00 GiB   0.0%	0 Bytes	                                                 objc2::runtime::message_receiver::MessageReceiver::send_message::h81dfd419189b6a37
84.00 GiB   0.0%	0 Bytes	                                                  objc2::runtime::message_receiver::msg_send_primitive::send::hc0961fbfcac6b629
84.00 GiB   0.0%	0 Bytes	                                                   _$LT$$LP$A$C$B$RP$$u20$as$u20$objc2..encode..EncodeArguments$GT$::__invoke::hd2941f8e31fd9e9b
84.00 GiB   0.0%	0 Bytes	                                                    -[AGXBuffer initWithDevice:length:alignment:options:isSuballocDisabled:pinnedGPULocation:]
84.00 GiB   0.0%	0 Bytes	                                                     -[AGXBuffer(Internal) initWithDevice:length:alignment:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
84.00 GiB   0.0%	0 Bytes	                                                      -[AGXBuffer(Internal) initWithDevice:length:alignment:pointerTag:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
84.00 GiB   0.0%	0 Bytes	                                                       -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:args:argsSize:deallocator:]
84.00 GiB   0.0%	0 Bytes	                                                        -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:placementSparsePageSize:placementSparseResidencyBytes:args:argsSize:deallocator:]
84.00 GiB   0.0%	84.00 GiB	                                                         __kdebug_trace64
82.00 GiB  14.0%	0 Bytes	                                         candle_core::tensor::Tensor::broadcast_sub::heb9fb3e79a80bbd1
82.00 GiB   0.0%	0 Bytes	                                          candle_core::tensor::Tensor::sub::h571175f735b5dfbb
82.00 GiB   0.0%	0 Bytes	                                           candle_core::storage::Storage::binary_impl::h5d345597b9bcf69f
82.00 GiB   0.0%	0 Bytes	                                            _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h35f1c58c739baa69
82.00 GiB   0.0%	0 Bytes	                                             candle_core::metal_backend::MetalStorage::binary::h8a4fc6643c45827f
82.00 GiB   0.0%	0 Bytes	                                              _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h13fc16f7d7a15a9a
82.00 GiB   0.0%	0 Bytes	                                               candle_core::metal_backend::device::MetalDevice::allocate_buffer::h87e110be93f79a4d
82.00 GiB   0.0%	0 Bytes	                                                candle_metal_kernels::metal::device::Device::new_buffer::hdff57d7c15bf891f
82.00 GiB   0.0%	0 Bytes	                                                 objc2_metal::generated::__MTLDevice::MTLDevice::newBufferWithLength_options::hc27f4530b25513c7
82.00 GiB   0.0%	0 Bytes	                                                  _$LT$MethodFamily$u20$as$u20$objc2..__macro_helpers..msg_send_retained..MsgSend$LT$Receiver$C$Return$GT$$GT$::send_message::h2d47315393e8e979
82.00 GiB   0.0%	0 Bytes	                                                   objc2::runtime::message_receiver::MessageReceiver::send_message::h81dfd419189b6a37
82.00 GiB   0.0%	0 Bytes	                                                    objc2::runtime::message_receiver::msg_send_primitive::send::hc0961fbfcac6b629
82.00 GiB   0.0%	0 Bytes	                                                     _$LT$$LP$A$C$B$RP$$u20$as$u20$objc2..encode..EncodeArguments$GT$::__invoke::hd2941f8e31fd9e9b
82.00 GiB   0.0%	0 Bytes	                                                      -[AGXBuffer initWithDevice:length:alignment:options:isSuballocDisabled:pinnedGPULocation:]
82.00 GiB   0.0%	0 Bytes	                                                       -[AGXBuffer(Internal) initWithDevice:length:alignment:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
82.00 GiB   0.0%	0 Bytes	                                                        -[AGXBuffer(Internal) initWithDevice:length:alignment:pointerTag:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
82.00 GiB   0.0%	0 Bytes	                                                         -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:args:argsSize:deallocator:]
82.00 GiB   0.0%	0 Bytes	                                                          -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:placementSparsePageSize:placementSparseResidencyBytes:args:argsSize:deallocator:]
82.00 GiB   0.0%	82.00 GiB	                                                           __kdebug_trace64
7.50 MiB   0.0%	0 Bytes	                                         candle_core::tensor::Tensor::sum_keepdim::h648a11aa78a7272d
7.50 MiB   0.0%	0 Bytes	                                          candle_core::tensor::Tensor::sum_impl::h56c45f280ce49025
7.50 MiB   0.0%	0 Bytes	                                           candle_core::storage::Storage::reduce_op::ha2d9bc02bdba85cb
7.50 MiB   0.0%	0 Bytes	                                            _$LT$candle_core..metal_backend..MetalStorage$u20$as$u20$candle_core..backend..BackendStorage$GT$::reduce_op::h6c1fa34e549af913
7.50 MiB   0.0%	0 Bytes	                                             _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h13fc16f7d7a15a9a
7.50 MiB   0.0%	0 Bytes	                                              candle_core::metal_backend::device::MetalDevice::allocate_buffer::h87e110be93f79a4d
7.50 MiB   0.0%	0 Bytes	                                               candle_metal_kernels::metal::device::Device::new_buffer::hdff57d7c15bf891f
7.50 MiB   0.0%	0 Bytes	                                                objc2_metal::generated::__MTLDevice::MTLDevice::newBufferWithLength_options::hc27f4530b25513c7
7.50 MiB   0.0%	0 Bytes	                                                 _$LT$MethodFamily$u20$as$u20$objc2..__macro_helpers..msg_send_retained..MsgSend$LT$Receiver$C$Return$GT$$GT$::send_message::h2d47315393e8e979
7.50 MiB   0.0%	0 Bytes	                                                  objc2::runtime::message_receiver::MessageReceiver::send_message::h81dfd419189b6a37
7.50 MiB   0.0%	0 Bytes	                                                   objc2::runtime::message_receiver::msg_send_primitive::send::hc0961fbfcac6b629
7.50 MiB   0.0%	0 Bytes	                                                    _$LT$$LP$A$C$B$RP$$u20$as$u20$objc2..encode..EncodeArguments$GT$::__invoke::hd2941f8e31fd9e9b
7.50 MiB   0.0%	0 Bytes	                                                     -[AGXBuffer initWithDevice:length:alignment:options:isSuballocDisabled:pinnedGPULocation:]
7.50 MiB   0.0%	0 Bytes	                                                      -[AGXBuffer(Internal) initWithDevice:length:alignment:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
7.50 MiB   0.0%	0 Bytes	                                                       -[AGXBuffer(Internal) initWithDevice:length:alignment:pointerTag:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
7.50 MiB   0.0%	0 Bytes	                                                        -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:args:argsSize:deallocator:]
7.50 MiB   0.0%	0 Bytes	                                                         -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:placementSparsePageSize:placementSparseResidencyBytes:args:argsSize:deallocator:]
7.50 MiB   0.0%	7.50 MiB	                                                          __kdebug_trace64
1.50 MiB   0.0%	0 Bytes	                                         candle_core::tensor::Tensor::max_keepdim::ha1543e08396d8976
1.50 MiB   0.0%	0 Bytes	                                          candle_core::tensor::Tensor::reduce_impl::h9b9a30dfacd26553
1.50 MiB   0.0%	0 Bytes	                                           candle_core::storage::Storage::reduce_op::ha2d9bc02bdba85cb
1.50 MiB   0.0%	0 Bytes	                                            _$LT$candle_core..metal_backend..MetalStorage$u20$as$u20$candle_core..backend..BackendStorage$GT$::reduce_op::h6c1fa34e549af913
1.50 MiB   0.0%	0 Bytes	                                             _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h13fc16f7d7a15a9a
1.50 MiB   0.0%	0 Bytes	                                              candle_core::metal_backend::device::MetalDevice::allocate_buffer::h87e110be93f79a4d
1.50 MiB   0.0%	0 Bytes	                                               candle_metal_kernels::metal::device::Device::new_buffer::hdff57d7c15bf891f
1.50 MiB   0.0%	0 Bytes	                                                objc2_metal::generated::__MTLDevice::MTLDevice::newBufferWithLength_options::hc27f4530b25513c7
1.50 MiB   0.0%	0 Bytes	                                                 _$LT$MethodFamily$u20$as$u20$objc2..__macro_helpers..msg_send_retained..MsgSend$LT$Receiver$C$Return$GT$$GT$::send_message::h2d47315393e8e979
1.50 MiB   0.0%	0 Bytes	                                                  objc2::runtime::message_receiver::MessageReceiver::send_message::h81dfd419189b6a37
1.50 MiB   0.0%	0 Bytes	                                                   objc2::runtime::message_receiver::msg_send_primitive::send::hc0961fbfcac6b629
1.50 MiB   0.0%	0 Bytes	                                                    _$LT$$LP$A$C$B$RP$$u20$as$u20$objc2..encode..EncodeArguments$GT$::__invoke::hd2941f8e31fd9e9b
1.50 MiB   0.0%	0 Bytes	                                                     -[AGXBuffer initWithDevice:length:alignment:options:isSuballocDisabled:pinnedGPULocation:]
1.50 MiB   0.0%	0 Bytes	                                                      -[AGXBuffer(Internal) initWithDevice:length:alignment:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
1.50 MiB   0.0%	0 Bytes	                                                       -[AGXBuffer(Internal) initWithDevice:length:alignment:pointerTag:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
1.50 MiB   0.0%	0 Bytes	                                                        -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:args:argsSize:deallocator:]
1.50 MiB   0.0%	0 Bytes	                                                         -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:placementSparsePageSize:placementSparseResidencyBytes:args:argsSize:deallocator:]
1.50 MiB   0.0%	1.50 MiB	                                                          __kdebug_trace64
76.00 GiB  13.0%	0 Bytes	                                        candle_core::tensor::Tensor::broadcast_mul::h87612ac4e7a9271d
76.00 GiB   0.0%	0 Bytes	                                         candle_core::tensor::Tensor::mul::h0369c368b543f4c8
76.00 GiB   0.0%	0 Bytes	                                          candle_core::storage::Storage::binary_impl::h3594d7ee262558eb
76.00 GiB   0.0%	0 Bytes	                                           _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h35f1c58c739baa69
76.00 GiB   0.0%	0 Bytes	                                            candle_core::metal_backend::MetalStorage::binary::h8a4fc6643c45827f
76.00 GiB   0.0%	0 Bytes	                                             _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h13fc16f7d7a15a9a
76.00 GiB   0.0%	0 Bytes	                                              candle_core::metal_backend::device::MetalDevice::allocate_buffer::h87e110be93f79a4d
76.00 GiB   0.0%	0 Bytes	                                               candle_metal_kernels::metal::device::Device::new_buffer::hdff57d7c15bf891f
76.00 GiB   0.0%	0 Bytes	                                                objc2_metal::generated::__MTLDevice::MTLDevice::newBufferWithLength_options::hc27f4530b25513c7
76.00 GiB   0.0%	0 Bytes	                                                 _$LT$MethodFamily$u20$as$u20$objc2..__macro_helpers..msg_send_retained..MsgSend$LT$Receiver$C$Return$GT$$GT$::send_message::h2d47315393e8e979
76.00 GiB   0.0%	0 Bytes	                                                  objc2::runtime::message_receiver::MessageReceiver::send_message::h81dfd419189b6a37
76.00 GiB   0.0%	0 Bytes	                                                   objc2::runtime::message_receiver::msg_send_primitive::send::hc0961fbfcac6b629
76.00 GiB   0.0%	0 Bytes	                                                    _$LT$$LP$A$C$B$RP$$u20$as$u20$objc2..encode..EncodeArguments$GT$::__invoke::hd2941f8e31fd9e9b
76.00 GiB   0.0%	0 Bytes	                                                     -[AGXBuffer initWithDevice:length:alignment:options:isSuballocDisabled:pinnedGPULocation:]
76.00 GiB   0.0%	0 Bytes	                                                      -[AGXBuffer(Internal) initWithDevice:length:alignment:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
76.00 GiB   0.0%	0 Bytes	                                                       -[AGXBuffer(Internal) initWithDevice:length:alignment:pointerTag:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
76.00 GiB   0.0%	0 Bytes	                                                        -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:args:argsSize:deallocator:]
76.00 GiB   0.0%	0 Bytes	                                                         -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:placementSparsePageSize:placementSparseResidencyBytes:args:argsSize:deallocator:]
76.00 GiB   0.0%	76.00 GiB	                                                          __kdebug_trace64
64.00 GiB  10.9%	0 Bytes	                                        candle_core::tensor::Tensor::matmul::h78ebc324e001cd34
64.00 GiB   0.0%	0 Bytes	                                         candle_core::storage::Storage::matmul::h63e24b756488a23b
64.00 GiB   0.0%	0 Bytes	                                          _$LT$candle_core..metal_backend..MetalStorage$u20$as$u20$candle_core..backend..BackendStorage$GT$::matmul::hcdf3ad195cd136a7
64.00 GiB   0.0%	0 Bytes	                                           _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h13fc16f7d7a15a9a
64.00 GiB   0.0%	0 Bytes	                                            candle_core::metal_backend::device::MetalDevice::allocate_buffer::h87e110be93f79a4d
64.00 GiB   0.0%	0 Bytes	                                             candle_metal_kernels::metal::device::Device::new_buffer::hdff57d7c15bf891f
64.00 GiB   0.0%	0 Bytes	                                              objc2_metal::generated::__MTLDevice::MTLDevice::newBufferWithLength_options::hc27f4530b25513c7
64.00 GiB   0.0%	0 Bytes	                                               _$LT$MethodFamily$u20$as$u20$objc2..__macro_helpers..msg_send_retained..MsgSend$LT$Receiver$C$Return$GT$$GT$::send_message::h2d47315393e8e979
64.00 GiB   0.0%	0 Bytes	                                                objc2::runtime::message_receiver::MessageReceiver::send_message::h81dfd419189b6a37
64.00 GiB   0.0%	0 Bytes	                                                 objc2::runtime::message_receiver::msg_send_primitive::send::hc0961fbfcac6b629
64.00 GiB   0.0%	0 Bytes	                                                  _$LT$$LP$A$C$B$RP$$u20$as$u20$objc2..encode..EncodeArguments$GT$::__invoke::hd2941f8e31fd9e9b
64.00 GiB   0.0%	0 Bytes	                                                   -[AGXBuffer initWithDevice:length:alignment:options:isSuballocDisabled:pinnedGPULocation:]
64.00 GiB   0.0%	0 Bytes	                                                    -[AGXBuffer(Internal) initWithDevice:length:alignment:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
64.00 GiB   0.0%	0 Bytes	                                                     -[AGXBuffer(Internal) initWithDevice:length:alignment:pointerTag:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
64.00 GiB   0.0%	0 Bytes	                                                      -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:args:argsSize:deallocator:]
64.00 GiB   0.0%	0 Bytes	                                                       -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:placementSparsePageSize:placementSparseResidencyBytes:args:argsSize:deallocator:]
64.00 GiB   0.0%	64.00 GiB	                                                        __kdebug_trace64
64.00 MiB   0.0%	0 Bytes	                                        candle_core::tensor::Tensor::reshape::h6f8d982e8766aede
64.00 MiB   0.0%	0 Bytes	                                         candle_core::device::Device::alloc_uninit::h6c7d3c31740be006
64.00 MiB   0.0%	0 Bytes	                                          candle_core::metal_backend::_$LT$impl$u20$candle_core..backend..BackendDevice$u20$for$u20$candle_core..metal_backend..device..MetalDevice$GT$::alloc_uninit::h3925f8c3eca38242
64.00 MiB   0.0%	0 Bytes	                                           _$LT$core..result..Result$LT$T$C$E$GT$$u20$as$u20$core..ops..try_trait..Try$GT$::branch::h13fc16f7d7a15a9a
64.00 MiB   0.0%	0 Bytes	                                            candle_core::metal_backend::device::MetalDevice::allocate_buffer::h87e110be93f79a4d
64.00 MiB   0.0%	0 Bytes	                                             candle_metal_kernels::metal::device::Device::new_buffer::hdff57d7c15bf891f
64.00 MiB   0.0%	0 Bytes	                                              objc2_metal::generated::__MTLDevice::MTLDevice::newBufferWithLength_options::hc27f4530b25513c7
64.00 MiB   0.0%	0 Bytes	                                               _$LT$MethodFamily$u20$as$u20$objc2..__macro_helpers..msg_send_retained..MsgSend$LT$Receiver$C$Return$GT$$GT$::send_message::h2d47315393e8e979
64.00 MiB   0.0%	0 Bytes	                                                objc2::runtime::message_receiver::MessageReceiver::send_message::h81dfd419189b6a37
64.00 MiB   0.0%	0 Bytes	                                                 objc2::runtime::message_receiver::msg_send_primitive::send::hc0961fbfcac6b629
64.00 MiB   0.0%	0 Bytes	                                                  _$LT$$LP$A$C$B$RP$$u20$as$u20$objc2..encode..EncodeArguments$GT$::__invoke::hd2941f8e31fd9e9b
64.00 MiB   0.0%	0 Bytes	                                                   -[AGXBuffer initWithDevice:length:alignment:options:isSuballocDisabled:pinnedGPULocation:]
64.00 MiB   0.0%	0 Bytes	                                                    -[AGXBuffer(Internal) initWithDevice:length:alignment:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
64.00 MiB   0.0%	0 Bytes	                                                     -[AGXBuffer(Internal) initWithDevice:length:alignment:pointerTag:options:isSuballocDisabled:resourceInArgs:pinnedGPULocation:]
64.00 MiB   0.0%	0 Bytes	                                                      -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:args:argsSize:deallocator:]
64.00 MiB   0.0%	0 Bytes	                                                       -[IOGPUMetalBuffer initWithDevice:pointer:length:alignment:options:sysMemSize:gpuAddress:gpuTag:placementSparsePageSize:placementSparseResidencyBytes:args:argsSize:deallocator:]
64.00 MiB   0.0%	64.00 MiB	                                                        __kdebug_trace64
```

</details>

  Each of these ops, on the Metal backend, allocates its output via:

  - `MetalStorage::unary_impl` / `MetalStorage::binary` / `MetalStorage::matmul`
  - `MetalDevice::new_buffer` → `MetalDevice::allocate_buffer`
  - `Device::new_buffer` → `MTLDevice::newBufferWithLength_options`

  Because command buffers are only flushed based on a fixed “kernels per command buffer” counter and we don’t track allocation volume, long sequences of these ops can allocate many large temporary buffers before any flush happens. Those buffers only become reusable after a distant flush, so peak RSS grows roughly with the sum of all intermediate outputs in a stage rather than the maximum.

## Minimum exmaples to reproduce this issue

<details>

<summary>Cargo.toml</summary>

```toml
[dependencies]
candle-core = { git = "https://github.com/huggingface/candle", default-features = false,features = ["accelerate","metal"] }
candle-nn = { git = "https://github.com/huggingface/candle", default-features = false,features = ["accelerate","metal"] }
```

</details>

<details>

<summary>main.rs</summary>

```rust
use candle_core::{DType, Device, Result, Tensor, D};

fn main() -> Result<()> {
    let dev = Device::new_metal(0)?;
    let dtype = DType::BF16;

    let batch = 1usize;
    let num_heads = 12usize;
    let head_dim = 1536 / num_heads;
    let seq_len = 8192usize;
    let total_len = seq_len;

    let bh = batch * num_heads;

    let q_flat = Tensor::randn(0f32, 1f32, (bh, seq_len, head_dim), &dev)?.to_dtype(dtype)?;
    let k_tiled = Tensor::randn(0f32, 1f32, (bh, head_dim, total_len), &dev)?.to_dtype(dtype)?;
    let v_flat = Tensor::randn(0f32, 1f32, (bh, total_len, head_dim), &dev)?.to_dtype(dtype)?;

    let use_mask = true;
    let mask = if use_mask {
        Some(Tensor::randn(0f32, 1f32, (batch, 1, seq_len, total_len), &dev)?.to_dtype(dtype)?)
    } else {
        None
    };

    for _layer in 0..32 {
        let mut scores = q_flat.matmul(&k_tiled)?;
        let scale = 1.0f64 / (head_dim as f64).sqrt();
        let scale_tensor = Tensor::full(scale as f32, (), scores.device())?.to_dtype(dtype)?;
        scores = scores.broadcast_mul(&scale_tensor)?;
        if let Some(mask) = &mask {
            let expanded = mask
                .expand((batch, num_heads, seq_len, total_len))?
                .reshape((bh, seq_len, total_len))?;
            scores = scores.add(&expanded)?;
        }

        let probs = candle_nn::ops::softmax(&scores, D::Minus1)?;
        let _ctx = probs.matmul(&v_flat)?;
    }

    dev.synchronize()?;
    Ok(())
}

```

</details>
 
## Whats changed

This PR introduces a simple allocation policy for the Metal backend so that, once a configurable amount of new buffer memory has been allocated, we automatically synchronize and trim the reuse cache, giving the existing Metal buffer pooling a chance to recycle large temporaries and preventing runaway memory growth.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -11,8 +11,9 @@ jobs:
     name: Check
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
+        os: [ubuntu-latest, ubuntu-24.04, windows-latest, macOS-latest, ubuntu-24.04-arm]
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-python@v6
@@ -22,7 +23,26 @@ jobs:
         if: runner.os == 'macOS'
         run: rm -f .cargo/config.toml
       - uses: dtolnay/rust-toolchain@stable
-      - run: cargo check --workspace
+
+      - name: Run macos with metal
+        if: matrix.os == 'macOS-latest' 
+        run: cargo check --workspace --features metal
+
+      - name: Run normal cpu
+        if: matrix.os == 'ubuntu-latest' || matrix.os == 'windows-latest'
+        run: cargo check --workspace
+
+      - name: Run with avx2
+        if: matrix.os == 'ubuntu-24.04'
+        run: |
+          export RUSTFLAGS="-C target-feature=avx2"
+          cargo check --workspace 
+
+      - name: Run with arm neon
+        if: matrix.os == 'ubuntu-24.04-arm'
+        run: |
+          export RUSTFLAGS="-C target-feature=neon"
+          cargo check --workspace 
 
   test:
     name: Test Suite
@@ -78,3 +98,4 @@ jobs:
         with:
           components: clippy
       - run: cargo clippy --workspace --tests --examples --benches -- -D warnings
+ 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ exclude = [
 resolver = "2"
 
 [workspace.package]
-version = "0.9.2-alpha.2"
+version = "0.9.2"
 edition = "2021"
 description = "Minimalist ML framework."
 repository = "https://github.com/huggingface/candle"
@@ -36,20 +36,20 @@ ab_glyph = "0.2.23"
 accelerate-src = { version = "0.3.2" }
 anyhow = { version = "1", features = ["backtrace"] }
 byteorder = "1.4.3"
-candle = { path = "./candle-core", package = "candle-core", version = "0.9.2-alpha.2" }
-candle-datasets = { path = "./candle-datasets", version = "0.9.2-alpha.2" }
-candle-flash-attn-build = { path = "candle-flash-attn-build", version = "0.9.2-alpha.2"  }
-candle-flash-attn = { path = "./candle-flash-attn", version = "0.9.2-alpha.2" }
-candle-flash-attn-v3 = { path = "./candle-flash-attn-v3", version = "0.9.2-alpha.2" }
-candle-kernels = { path = "./candle-kernels", version = "0.9.2-alpha.2" }
-candle-metal-kernels = { path = "./candle-metal-kernels", version = "0.9.2-alpha.2" }
-candle-nn = { path = "./candle-nn", version = "0.9.2-alpha.2" }
-candle-onnx = { path = "./candle-onnx", version = "0.9.2-alpha.2" }
-candle-transformers = { path = "./candle-transformers", version = "0.9.2-alpha.2" }
-candle-ug = { path = "./candle-ug", version = "0.9.2-alpha.2" }
+candle = { path = "./candle-core", package = "candle-core", version = "0.9.2" }
+candle-datasets = { path = "./candle-datasets", version = "0.9.2" }
+candle-flash-attn-build = { path = "candle-flash-attn-build", version = "0.9.2" }
+candle-flash-attn = { path = "./candle-flash-attn", version = "0.9.2" }
+candle-flash-attn-v3 = { path = "./candle-flash-attn-v3", version = "0.9.2" }
+candle-kernels = { path = "./candle-kernels", version = "0.9.2" }
+candle-metal-kernels = { path = "./candle-metal-kernels", version = "0.9.2" }
+candle-nn = { path = "./candle-nn", version = "0.9.2" }
+candle-onnx = { path = "./candle-onnx", version = "0.9.2" }
+candle-transformers = { path = "./candle-transformers", version = "0.9.2" }
+candle-ug = { path = "./candle-ug", version = "0.9.2" }
 clap = { version = "4.2.4", features = ["derive"] }
-criterion = { version = "0.7.0", default-features = false }
-cudarc = { version = "0.18.2", features = [
+criterion = { version = "0.8", default-features = false }
+cudarc = { version = "0.19.0", features = [
     "std",
     "cublas",
     "cublaslt",
@@ -61,21 +61,23 @@ cudarc = { version = "0.18.2", features = [
     "cuda-version-from-build-system",
     "dynamic-linking",
 ], default-features = false }
-fancy-regex = "0.13.0"
-gemm = { version = "0.18.2", features = ["wasm-simd128-enable"] }
+fancy-regex = "0.17.0"
+gemm = { version = "0.19.0", features = ["wasm-simd128-enable"] }
 hf-hub = "0.4.1"
 half = { version = "2.5.0", features = [
     "num-traits",
     "use-intrinsics",
     "rand_distr",
 ] }
-float8 = { version = "0.5.0", features = ["num-traits", "rand_distr"] }
+float8 = { version = "0.6.0", features = ["num-traits", "rand_distr"] }
 hound = "3.5.1"
 image = { version = "0.25.2", default-features = false, features = [
     "jpeg",
     "png",
 ] }
-imageproc = { version = "0.24.0", default-features = false }
+imageproc = { version = "0.26.0", features = [
+    "text",
+], default-features = false }
 intel-mkl-src = { version = "0.8.1", features = ["mkl-static-lp64-iomp"] }
 libc = { version = "0.2.147" }
 libm = { version = "0.2.15" }
@@ -83,24 +85,24 @@ log = "0.4"
 memmap2 = { version = "0.9.3", features = ["stable_deref_trait"] }
 num_cpus = "1.15.0"
 num-traits = "0.2.15"
-parquet = { version = "51.0.0" }
+parquet = "57"
 rand = "0.9.0"
 rand_distr = "0.5.1"
 rayon = "1.7.0"
-safetensors = "0.6.0"
+safetensors = "0.7.0"
 serde = { version = "1.0.171", features = ["derive"] }
 serde_plain = "1.0.2"
 serde_json = "1.0.99"
-thiserror = "1"
-tokenizers = { version = "0.21.0", default-features = false }
+thiserror = "2"
+tokenizers = { version = "0.22.0", default-features = false }
 tracing = "0.1.37"
 tracing-chrome = "0.7.1"
 tracing-subscriber = "0.3.7"
 ug = "0.5.0"
 ug-cuda = "0.5.0"
 ug-metal = "0.5.0"
-yoke = { version = "0.7.2", features = ["derive"] }
-zip = { version = "1.1.1", default-features = false }
+yoke = { version = "0.8.1", features = ["derive"] }
+zip = { version = "7.2.0", default-features = false }
 objc2-metal = { version = "0.3.1" }
 objc2-foundation = { version = "0.3.1" }
 
--- a/candle-core/src/cuda_backend/mod.rs
+++ b/candle-core/src/cuda_backend/mod.rs
@@ -1257,6 +1257,87 @@ impl CudaStorage {
     pub fn as_cuda_slice_mut<T: CudaDType>(&mut self) -> Result<&mut CudaSlice<T>> {
         T::as_cuda_slice_mut(self)
     }
+
+    pub fn transfer_to_device(&self, dst: &CudaDevice) -> Result<Self> {
+        let dst_stream = dst.cuda_stream();
+        let storage_slice = match self.dtype() {
+            DType::U8 => {
+                let cuda_slice = self.as_cuda_slice::<u8>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::U8(result)
+            }
+            DType::U32 => {
+                let cuda_slice = self.as_cuda_slice::<u32>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::U32(result)
+            }
+            DType::I16 => {
+                let cuda_slice = self.as_cuda_slice::<i16>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::I16(result)
+            }
+            DType::I32 => {
+                let cuda_slice = self.as_cuda_slice::<i32>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::I32(result)
+            }
+            DType::I64 => {
+                let cuda_slice = self.as_cuda_slice::<i64>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::I64(result)
+            }
+            DType::BF16 => {
+                let cuda_slice = self.as_cuda_slice::<bf16>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::BF16(result)
+            }
+            DType::F16 => {
+                let cuda_slice = self.as_cuda_slice::<f16>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::F16(result)
+            }
+            DType::F32 => {
+                let cuda_slice = self.as_cuda_slice::<f32>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::F32(result)
+            }
+            DType::F64 => {
+                let cuda_slice = self.as_cuda_slice::<f64>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::F64(result)
+            }
+            DType::F8E4M3 => {
+                let cuda_slice = self.as_cuda_slice::<float8::F8E4M3>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::F8E4M3(result)
+            }
+            DType::F6E2M3 => {
+                let cuda_slice = self.as_cuda_slice::<u8>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::F6E2M3(result)
+            }
+            DType::F6E3M2 => {
+                let cuda_slice = self.as_cuda_slice::<u8>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::F6E3M2(result)
+            }
+            DType::F4 => {
+                let cuda_slice = self.as_cuda_slice::<u8>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::F4(result)
+            }
+            DType::F8E8M0 => {
+                let cuda_slice = self.as_cuda_slice::<u8>()?;
+                let result = dst_stream.clone_dtod(cuda_slice).w()?;
+                CudaStorageSlice::F8E8M0(result)
+            }
+        };
+
+        Ok(Self {
+            slice: storage_slice,
+            device: dst.clone(),
+        })
+    }
 }
 
 fn gemm_config<T>(
--- a/candle-core/src/dummy_cuda_backend.rs
+++ b/candle-core/src/dummy_cuda_backend.rs
@@ -10,15 +10,26 @@ pub struct CudaDevice;
 #[derive(Debug)]
 pub struct CudaStorage;
 
+impl CudaStorage {
+    pub fn transfer_to_device(&self, _dst: &CudaDevice) -> Result<Self> {
+        Err(Error::NotCompiledWithCudaSupport)
+    }
+}
+
 macro_rules! fail {
     () => {
         unimplemented!("cuda support has not been enabled, add `cuda` feature to enable.")
     };
 }
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct DeviceId(usize);
 
 impl CudaDevice {
     pub fn new_with_stream(_: usize) -> Result<Self> {
         Err(Error::NotCompiledWithCudaSupport)
+    }
+    pub fn id(&self) -> DeviceId {
+        DeviceId(0)
     }
 }
 
--- a/candle-core/src/quantized/cuda.rs
+++ b/candle-core/src/quantized/cuda.rs
@@ -406,6 +406,7 @@ fn mul_mat_via_q8_1(
     Ok(CudaStorage::wrap_cuda_slice(dst, dev.clone()))
 }
 
+#[allow(clippy::too_many_arguments)]
 fn indexed_moe_forward_fused_q8_1_input(
     weight: &CudaView<u8>,
     w_shape: &crate::Shape, //[num_experts, n, k]
@@ -511,7 +512,7 @@ impl QCudaStorage {
                 &self.data.inner.slice(0..),
                 self_shape, //[num_experts, n, k]
                 self.dtype(),
-                &input_storage,
+                input_storage,
                 input_l.shape(), //[batch, topk or 1, k]
                 &ids_storage.slice(0..),
                 ids_l.shape(), //[batch, topk]
@@ -620,7 +621,7 @@ impl QCudaStorage {
             data.len() + MATRIX_ROW_PADDING * self.dtype.type_size() / self.dtype.block_size();
         let mut inner = unsafe { self.device.alloc::<u8>(padded_len)? };
         self.device
-            .memcpy_htod(data.as_ref(), &mut inner.slice_mut(..data.len()))?;
+            .memcpy_htod(&*data, &mut inner.slice_mut(..data.len()))?;
         self.data = PaddedCudaSlice {
             inner,
             len: data.len(),
@@ -648,7 +649,7 @@ impl QCudaStorage {
             data.len() + MATRIX_ROW_PADDING * self.dtype.type_size() / self.dtype.block_size();
         let mut inner = unsafe { self.device.alloc::<u8>(padded_len)? };
         self.device
-            .memcpy_htod(data.as_ref(), &mut inner.slice_mut(..data.len()))?;
+            .memcpy_htod(&*data, &mut inner.slice_mut(..data.len()))?;
         self.data = PaddedCudaSlice {
             inner,
             len: data.len(),
@@ -677,7 +678,7 @@ impl QCudaStorage {
             data.len() + MATRIX_ROW_PADDING * self.dtype.type_size() / self.dtype.block_size();
         let mut inner = unsafe { self.device.alloc::<u8>(padded_len)? };
         self.device
-            .memcpy_htod(data.as_ref(), &mut inner.slice_mut(..data.len()))?;
+            .memcpy_htod(&*data, &mut inner.slice_mut(..data.len()))?;
         self.data = PaddedCudaSlice {
             inner,
             len: data.len(),
@@ -701,7 +702,7 @@ impl QCudaStorage {
             data.len() + MATRIX_ROW_PADDING * self.dtype.type_size() / self.dtype.block_size();
         let mut inner = unsafe { self.device.alloc::<u8>(padded_len)? };
         self.device
-            .memcpy_htod(data.as_ref(), &mut inner.slice_mut(..data.len()))?;
+            .memcpy_htod(&*data, &mut inner.slice_mut(..data.len()))?;
         self.data = PaddedCudaSlice {
             inner,
             len: data.len(),
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -2379,10 +2379,9 @@ impl Tensor {
                 (Storage::Cuda(storage), Device::Cpu) => Storage::Cpu(storage.to_cpu_storage()?),
                 (Storage::Metal(storage), Device::Cpu) => Storage::Cpu(storage.to_cpu_storage()?),
                 (Storage::Cuda(storage), Device::Cuda(cuda)) => {
-                    // TODO: Avoid passing through the cpu storage here, especially if the gpu ids
-                    // are the same.
-                    let cpu_storage = storage.to_cpu_storage()?;
-                    Storage::Cuda(cuda.storage_from_cpu_storage(&cpu_storage)?)
+                    // can't clone storage if it's the same device because of the underlying device ptr
+                    let dst_storage = storage.transfer_to_device(cuda)?;
+                    Storage::Cuda(dst_storage)
                 }
                 (Storage::Cpu(storage), Device::Cpu) => Storage::Cpu(storage.clone()),
                 _ => {
--- a/candle-core/tests/tensor_tests.rs
+++ b/candle-core/tests/tensor_tests.rs
@@ -1998,3 +1998,65 @@ fn tensor_norm() -> Result<()> {
     assert_eq!(norm.to_scalar::<f64>()?, 5.);
     Ok(())
 }
+
+#[cfg(feature = "cuda")]
+#[test]
+fn transfers_cuda_to_device() -> Result<()> {
+    use rand::seq::SliceRandom;
+
+    let devices = cudarc::driver::safe::CudaContext::device_count()
+        .map_err(candle_core::cuda::CudaError::from)?;
+    if devices < 2 {
+        return Ok(());
+    }
+    let first = Device::new_cuda(0)?;
+
+    let mut data: Vec<u32> = (0..262144).collect();
+    let mut rng = rand::rng();
+    data.shuffle(&mut rng);
+
+    let t1 = Tensor::from_vec(data, (512, 512), &first)?;
+    let second = Device::new_cuda(1)?;
+    let t2 = t1.to_device(&second)?;
+
+    assert_ne!(
+        t1.device().as_cuda_device()?.id(),
+        t2.device().as_cuda_device()?.id()
+    );
+    Ok(())
+}
+
+#[cfg(feature = "cuda")]
+#[test]
+fn allocates_twice_when_transferring_to_same_device() -> Result<()> {
+    use std::{ops::Deref, sync::RwLockReadGuard};
+
+    use candle_core::Storage;
+    use rand::seq::SliceRandom;
+
+    let first = Device::new_cuda(0)?;
+    let second = Device::new_cuda(0)?;
+
+    let mut data: Vec<u32> = (0..262144).collect();
+    let mut rng = rand::rng();
+    data.shuffle(&mut rng);
+
+    let t1 = Tensor::from_vec(data, (512, 512), &first)?;
+    let t2 = t1.to_device(&second)?;
+
+    let (storage1, _) = t1.storage_and_layout();
+    let (storage2, _) = t2.storage_and_layout();
+    let extract = |s: RwLockReadGuard<'_, Storage>| match &s.deref() {
+        Storage::Cuda(c) => {
+            use cudarc::driver::DevicePtr;
+            let slice = c.as_cuda_slice::<u32>().unwrap();
+            let ptr = slice.device_ptr(slice.stream()).0;
+            ptr
+        }
+        _ => unimplemented!(),
+    };
+    let id1 = extract(storage1);
+    let id2 = extract(storage2);
+    assert_ne!(id1, id2);
+    Ok(())
+}
--- a/candle-examples/Cargo.toml
+++ b/candle-examples/Cargo.toml
@@ -34,7 +34,7 @@ pyo3 = { version = "0.27", features = [
     "abi3-py311",
 ], optional = true }
 rayon = { workspace = true }
-rubato = { version = "0.15.0", optional = true }
+rubato = { version = "1", optional = true }
 safetensors = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/candle-examples/build.rs
+++ b/candle-examples/build.rs
@@ -28,7 +28,6 @@ fn main() -> Result<()> {
 
         for kdir in KERNEL_DIRS.iter() {
             let builder = bindgen_cuda::Builder::default().kernel_paths_glob(kdir.kernel_glob);
-            println!("cargo::warning={builder:?}");
             let bindings = builder.build_ptx().unwrap();
 
             // Changed: This now writes to a safe path inside $OUT_DIR.
--- a/candle-flash-attn-build/Cargo.toml
+++ b/candle-flash-attn-build/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "candle-flash-attn-build"
-version = "0.9.2-alpha.2"
+version = "0.9.2"
 edition = "2021"
 
 description = "Build utilities for candle flash attention crates"
@@ -8,4 +8,3 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 anyhow = "1"
-
--- a/candle-flash-attn-v3/Cargo.toml
+++ b/candle-flash-attn-v3/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "candle-flash-attn-v3"
-version = "0.9.2-alpha.2"
+version = "0.9.2"
 edition = "2021"
 
 description = "Flash attention v3 layer for the candle ML framework."
@@ -12,14 +12,14 @@ readme = "README.md"
 exclude = ["cutlass/docs/**", "cutlass/test/**", "cutlass/examples/**", "cutlass/tools/**", "cutlass/media/**"]
 
 [dependencies]
-candle = { path = "../candle-core", features = ["cuda"], package = "candle-core", version = "0.9.2-alpha.2" }
+candle = { path = "../candle-core", features = ["cuda"], package = "candle-core", version = "0.9.2" }
 half = { version = "2.3.1", features = ["num-traits"] }
 
 [build-dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
 num_cpus = "1.15.0"
 rayon = "1.7.0"
-candle-flash-attn-build = { path = "../candle-flash-attn-build" }
+candle-flash-attn-build = { path = "../candle-flash-attn-build", version = "0.9.2" }
 
 [dev-dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
--- a/candle-flash-attn/Cargo.toml
+++ b/candle-flash-attn/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "candle-flash-attn"
-version = "0.9.2-alpha.2"
+version = "0.9.2"
 edition = "2021"
 
 description = "Flash attention layer for the candle ML framework."
@@ -11,13 +11,13 @@ license = "MIT OR Apache-2.0"
 readme = "README.md"
 
 [dependencies]
-candle = { path = "../candle-core", features = ["cuda"], package = "candle-core", version = "0.9.2-alpha.2" }
+candle = { path = "../candle-core", features = ["cuda"], package = "candle-core", version = "0.9.2" }
 half = { version = "2.3.1", features = ["num-traits"] }
 
 [build-dependencies]
 bindgen_cuda = "0.1.5"
 anyhow = { version = "1", features = ["backtrace"] }
-candle-flash-attn-build = { path = "../candle-flash-attn-build" }
+candle-flash-attn-build = { path = "../candle-flash-attn-build", version = "0.9.2" }
 
 [dev-dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
--- a/candle-kernels/Cargo.toml
+++ b/candle-kernels/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "candle-kernels"
-version = "0.9.2-alpha.2"
+version = "0.9.2"
 edition = "2021"
 
 description = "CUDA kernels for Candle"
@@ -12,4 +12,4 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 
 [build-dependencies]
-bindgen_cuda = { git = "https://github.com/guoqingbao/bindgen_cuda.git", version= "0.1.7" }
+bindgen_cuda = "0.1.6"
--- a/candle-kernels/build.rs
+++ b/candle-kernels/build.rs
@@ -10,16 +10,20 @@ fn main() {
     // Build for PTX
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
     let ptx_path = out_dir.join("ptx.rs");
-    let mut builder = bindgen_cuda::Builder::default()
+    let builder = bindgen_cuda::Builder::default()
         .arg("--expt-relaxed-constexpr")
         .arg("-std=c++17")
         .arg("-O3");
-    println!("cargo::warning={builder:?}");
     let bindings = builder.build_ptx().unwrap();
     bindings.write(&ptx_path).unwrap();
 
     // Remove unwanted MOE PTX constants from ptx.rs
     remove_lines(&ptx_path, &["MOE_GGUF", "MOE_WMMA", "MOE_WMMA_GGUF"]);
+
+    let mut moe_builder = bindgen_cuda::Builder::default()
+        .arg("--expt-relaxed-constexpr")
+        .arg("-std=c++17")
+        .arg("-O3");
 
     // Build for FFI binding (must use custom bindgen_cuda, which supports simutanously build PTX and lib)
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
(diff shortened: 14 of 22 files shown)


ivarflakstad:
Look @pcuenca, it's the power of open source ✨ 
(Ref pointing out the memory spikes just the other day☺️)

ivarflakstad:
This looks good to me. Just want to try it out in various contexts first :)

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,61 @@ impl DeviceId {
...
+                .map(|mb| mb * 1024 * 1024)
+        }
+
+        fn system_memory_bytes() -> Option<usize> {
+            use libc::c_void;
+            let mut value: u64 = 0;
+            let mut len = core::mem::size_of::<u64>();
+            let ret = unsafe {
+                libc::sysctlbyname(
+                    b"hw.memsize\0".as_ptr() as *const libc::c_char,
```
ivarflakstad: One thing to note is that the GPU typically only has access to 2/3 of the available memory on the machine.
You can change this via `sudo sysctl iogpu.wired_limit_mb=<insert mb here>`, which brings me to my next point:
Perhaps we should use `iogpu.wired_limit_mb` instead?

pcuenca:
Indeed, I saw several softmax buffers when checking with Instruments!

candle-core/src/metal_backend/device.rs
```diff
@@ -57,6 +160,10 @@ pub struct MetalDevice {
     pub(crate) kernels: Arc<Kernels>,
     /// Seed for random number generation.
     pub(crate) seed: Arc<Mutex<Buffer>>,
+    /// Bytes allocated since the last synchronization point.
+    pub(crate) pending_allocation_bytes: Arc<AtomicUsize>,
```
ivarflakstad, line 160: Since this is an `AtomicUsize` we don't need an atomic reference counter around it.
(`Arc` is actually two `AtomicUsize`'s (strong and weak refs) + the actual data (`T`) it wraps)

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,106 @@ impl DeviceId {
...
+    /// Maximum bytes to keep cached for reuse.
+    cache_limit_bytes: usize,
+}
+
+impl Default for AllocationPolicy {
+    fn default() -> Self {
+        const DEFAULT_PENDING: usize = 4 * 1024 * 1024 * 1024; // 4 GiB
+        const MIN_PENDING: usize = 512 * 1024 * 1024; // 512 MiB
+        const MAX_PENDING: usize = 12 * 1024 * 1024 * 1024; // 12 GiB
+
```
ivarflakstad, line 53: Slightly more modern way of interacting with C strings, provided you add `use std::ffi::CStr;`
```suggestion
const HW_MEMSIZE_KEY: &CStr = c"hw.memsize";
const IOGPU_WIRED_LIMIT_MB_KEY: &CStr = c"iogpu.wired_limit_mb";

```

Then you can change `sysctl_u64` to accept `&CStr` and you need only use `.as_ptr()` because it returns `*const c_char` directly.

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,106 @@ impl DeviceId {
...
+                    &mut value as *mut u64 as *mut c_void,
+                    &mut len as *mut usize,
+                    std::ptr::null_mut(),
+                    0,
+                ) != 0
+                {
+                    return None;
+                }
+                Some(value)
+            }
```
ivarflakstad, line 81: I had a look at `man sysctlbyname`, and if I understand correctly we can simplify this a bit.
First of all I think it's ok to do the len check in the same call as the one where you retrieve the actual value. But we can actually simplify a bit more as well
```c++
int
     sysctlbyname(const char *name, void *oldp, size_t *oldlenp, void *newp, size_t newlen);

.. snip ...
If the amount of data available is greater than the size of the buffer
     supplied, the call supplies as much data as fits in the buffer provided and returns with
     the error code ENOMEM.
```

To me this means we don't actually need `len > core::mem::size_of::<u64>()` as it is covered by `err != 0`. I don't see documentation about what happens if there is no data, so let's keep `len == 0` just in case ☺️ 
```suggestion
            let mut value: u64 = 0;
            let mut len: usize = 0;
            unsafe {
                let err = libc::sysctlbyname(
                    name.as_ptr(),
                    &mut value as *mut u64 as *mut c_void,
                    &mut len as *mut usize,
                    std::ptr::null_mut(),
                    0,
                );
            }
            if err != 0 || len == 0 {
                return None;
            }
            Some(value)
        }
```

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,106 @@ impl DeviceId {
...
+            const SYSTEM_RESERVE_FRACTION: usize = 4; // Keep at least 25% for the OS.
+            const SYSTEM_RESERVE_MIN: usize = 2 * 1024 * 1024 * 1024; // 2 GiB floor.
+
+            let hw_total = sysctl_u64(b"hw.memsize\0").and_then(|bytes| {
+                if bytes == 0 || bytes > usize::MAX as u64 {
+                    None
+                } else {
+                    Some(bytes as usize)
+                }
+            })?;
```
ivarflakstad: We have to go all the way back to 2008 to worry about wether our macs/iphones are 32bit or 64bit.
I agree with the `bytes == 0` check though

```suggestion
            let hw_total = sysctl_u64(HW_MEMSIZE_KEY).and_then(|bytes| {
                if bytes == 0 {
                    None
                } else {
                    Some(bytes as usize)
                }
            })?;
```

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,106 @@ impl DeviceId {
...
+            if hw_budget == 0 {
+                return None;
+            }
+
+            let wired_limit_bytes = sysctl_u64(b"iogpu.wired_limit_mb\0").and_then(|limit_mb| {
+                if limit_mb == 0 || limit_mb > usize::MAX as u64 {
+                    return None;
+                }
+                (limit_mb as usize).checked_mul(MEBIBYTE)
+            });
```
ivarflakstad:
```suggestion
            let wired_limit_bytes = sysctl_u64(IOGPU_WIRED_LIMIT_MB_KEY).and_then(|limit_mb| {
                if limit_mb == 0 {
                    return None;
                }
                (limit_mb as usize).checked_mul(MEBIBYTE)
            });
```

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,106 @@ impl DeviceId {
...
+
+            let wired_clamped = wired_limit_bytes.map(|limit| std::cmp::min(limit, hw_total));
+
+            Some(match wired_clamped {
+                Some(wired) => std::cmp::min(wired, hw_budget),
+                None => hw_budget,
+            })
```
ivarflakstad: Un-optioning early makes the code slightly sleeker.
Also, since `let hw_budget = hw_total.saturating_sub(reserve);` we know that `hw_budget` must always be smaller than `hw_total`, so we only need one `min` call.
```suggestion
            let result = wired_limit_bytes.or_else(hw_budget);
            result.min(hw_budget)
```

I

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,106 @@ impl DeviceId {
...
+    fn default() -> Self {
+        const DEFAULT_PENDING: usize = 4 * 1024 * 1024 * 1024; // 4 GiB
+        const MIN_PENDING: usize = 512 * 1024 * 1024; // 512 MiB
+        const MAX_PENDING: usize = 12 * 1024 * 1024 * 1024; // 12 GiB
+
+        fn parse_env_mebibytes(var: &str) -> Option<usize> {
+            std::env::var(var)
+                .ok()
+                .and_then(|value| value.trim().parse::<usize>().ok())
+                .map(|mb| mb * 1024 * 1024)
```
ivarflakstad: Not that it matters much, but users could theoretically set the env var as `usize::MAX` 🤷 
```suggestion
                .and_then(|mb| mb.checked_mul(1024 * 1024))
```

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,106 @@ impl DeviceId {
     }
 }
 
+#[derive(Clone)]
+pub(crate) struct AllocationPolicy {
+    /// Total bytes we can allocate before forcing a sync to reclaim temporaries.
+    pending_limit_bytes: usize,
```
ivarflakstad: Based on the description this seems more like `allocation_limit_bytes`. Is there a reason for the word pending specifically?

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,106 @@ impl DeviceId {
...
+                None => hw_budget,
+            })
+        }
+
+        let pending_limit = parse_env_mebibytes("CANDLE_METAL_PENDING_LIMIT_MB")
+            .or_else(|| system_memory_bytes().map(|mem| (mem / 3).clamp(MIN_PENDING, MAX_PENDING)))
+            .unwrap_or(DEFAULT_PENDING);
+
+        let cache_limit = parse_env_mebibytes("CANDLE_METAL_CACHE_LIMIT_MB")
+            .unwrap_or_else(|| std::cmp::max(pending_limit / 2, 64 * 1024 * 1024));
```
ivarflakstad: Let's make `64 * 1024 * 1024` a named const. `DEFAULT_CACHE_LIMIT` or something.

candle-core/src/metal_backend/device.rs
```diff
@@ -26,6 +29,106 @@ impl DeviceId {
...
+            let wired_clamped = wired_limit_bytes.map(|limit| std::cmp::min(limit, hw_total));
+
+            Some(match wired_clamped {
+                Some(wired) => std::cmp::min(wired, hw_budget),
+                None => hw_budget,
+            })
+        }
+
+        let pending_limit = parse_env_mebibytes("CANDLE_METAL_PENDING_LIMIT_MB")
+            .or_else(|| system_memory_bytes().map(|mem| (mem / 3).clamp(MIN_PENDING, MAX_PENDING)))
```
ivarflakstad, line 113: Just verifying that I understand the intention here.
Given that we are between 512MB and 12GB, we set the `pending_limit` to 1/3 of the available memory?

candle-core/src/metal_backend/device.rs
```diff
@@ -235,6 +376,22 @@ impl MetalDevice {
             .map_err(|e| MetalError::from(e.to_string()))?;
         Ok(())
     }
+
+    fn on_new_allocation(&self, size: usize) -> Result<()> {
+        let pending = self
+            .pending_allocation_bytes
+            .fetch_add(size, Ordering::AcqRel)
+            .saturating_add(size);
+        if pending >= self.allocation_policy.pending_limit_bytes {
```
ivarflakstad: Ah, now I see where pending comes in.
I still think a name like `allocation_limit_bytes` or `max_allocations_bytes` makes more sense.

candle-core/src/metal_backend/device.rs
```diff
@@ -112,14 +219,46 @@ impl MetalDevice {
...
+        let mut empty_keys = Vec::new();
+        for (size, subbuffers) in buffers.iter_mut() {
+            if bytes_to_drop == 0 {
+                break;
+            }
+            subbuffers.retain(|buffer| {
+                if bytes_to_drop == 0 {
+                    return true;
+                }
+                if Arc::strong_count(buffer) == 1 {
```
ivarflakstad, line 252: `Arc::strong_count(buffer) == 1` means that the only strong reference of this buffer is the one in `self.buffers`, aka it is not in use anywhere.
I'm not sure if we want to keep unused buffers, even if `bytes_to_drop == 0`, but I'm open to being convinced.

candle-core/src/metal_backend/device.rs
```diff
@@ -112,14 +219,46 @@ impl MetalDevice {
...
+                } else {
+                    true
+                }
+            });
+            if subbuffers.is_empty() {
+                empty_keys.push(*size);
+            }
+        }
+        for key in empty_keys {
+            buffers.remove(&key);
```
ivarflakstad: Removing keys isn't really that important. The buffer map is `usize -> Vec<Buffer>` where the usize denotes the size of the buffer. No real danger in keeping empty vecs. At worst it's a tiny increase in cpu memory footprint, at best it means less allocations.

ivarflakstad:
It might be that `sysctl_usize` behaves correctly in all cases, if you want to get rid of the `u64 -> usize` conversions. Worth trying.
Btw, I can't guarantee the code I provided via the review is completely correct. It's all freehand :)

candle-core/src/metal_backend/device.rs
TimmyOVO: I think we still need Arc<AtomicUsize> here. MetalDevice derives `Clone`, `MetalStorage` holds a cloned `MetalDevice `If we switch to a plain `AtomicUsize`, each clone would carry an independent counter, so the pending-allocation limit would be enforced per clone instead of per physical device then the allocator could overshoot the global limit.

ivarflakstad: You're right!

candle-core/src/metal_backend/device.rs
TimmyOVO: Yes, we pick pending_limit from `CANDLE_METAL_PENDING_LIMIT_MB` if set, otherwise as one-third of the available system memory (after OS/GPU reservations) clamped between 512 MiB and 12 GiB, and fall back to 4 GiB if we can’t detect the memory.

candle-core/src/metal_backend/device.rs
TimmyOVO: Yes, agreed

candle-core/src/metal_backend/device.rs
TimmyOVO: With the current code we do keep some of those once `bytes_to_drop == 0`. That’s intentional.
Letme explain why we do this:
They’re not used by any tensor, but they are ready to be reused on the next allocation.But if we dropped all `strong_count == 1` buffers, we’d have zero cache after each trim, and next allocations would always go back to Metal to create new buffers, which is slower and causes more RSS churn

TimmyOVO:
> It might be that `sysctl_usize` behaves correctly in all cases, if you want to get rid of the `u64 -> usize` conversions. Worth trying. Btw, I can't guarantee the code I provided via the review is completely correct. It's all freehand :)

Thanks for catching that. I'll look into removing those `u64` to `usize` conversions first thing tomorrow. It's 8 AM here and I need to get some sleep, so I'd rather tackle this with a fresh mind.

candle-core/src/metal_backend/device.rs
ivarflakstad: Right. I forgot that with this new approach we run this every time we allocate a buffer. Not just when we flush (only triggered every once in a while).

pcuenca:
I did a quick test, this is what I saw:

* The memory profile is much better behaved on macOS. This is what it looks like running Stable Diffusion:

Screenshot 2025-11-21 at 19 18 16

The last spike is the VAE decoding phase; it's normal that memory grows. Memory is ~stable during the UNet forward steps. However, it increases a bit each step. This could point to some memory leak somewhere (including my test code).

* On iOS, I'm seeing deadlocks again. This may be a regression after the fixes in #3164 (not necessarily from this PR, I didn't check)
* I think we need to add `dep:libc` to `candle-core/Cargo.toml` for the `metal` feature.

I think we can merge this PR if we can determine that the iOS regression was introduced elsewhere; I'll run some more tests about that.

ivarflakstad:
> I think we need to add dep:libc to candle-core/Cargo.toml for the metal feature.

Rust `std` depends `libc`, so it's only when you're working with `no_std` targets that you would have to manually add the `libc` dependency (interesting situation where you both want `no_std` and `libc`, but it has probably happened plenty of times in the wild).

ivarflakstad:
@TimmyOVO we haven't forgotten about you! :) 

@pcuenca has been swamped recently so he hasn't had time to look into the issue he detected.
Once he has time and we figure out the issue (if there is one) this will get merged asap

ivarflakstad:
> it's only when you're working with `no_std` targets that you would have to manually add the `libc` dependency

Also I this is a big fat lie. The actual answer is that all these changes are in `candle-core` which already has the `libc` dependency.

pcuenca:
Really sorry for the big delay here! Here's what I tested:

- Merged main into my ios testing branch (it exposes a candle to Swift interface for stable diffusion) - it works as before on iOS: big memory spikes, inference proceeds until the app is killed for memory consumption.
- Merged the branch above with the changes in this PR: then I see a deadlock **on iOS**: the first forward pass through the UNet never returns.

I suspect the [`wait_until_completed()` call](https://github.com/huggingface/candle/pull/3197/files#diff-0ff83f29473b211016cbde3897f9f4536fc533c73a71efc83a3dda8782462141R398) because that's the only place in the diff that looks potentially related.

---

My testing process is a bit inconvenient:
- I compile [this branch](https://github.com/pcuenca/candle/tree/sd-ios-example) with the changes to test (i.e., the ones from this here branch): `make -f candle-sd15-bridge/Makefile`
- I copy the compiled `CandleSD15.xcframework` to a checkout of [this demo app branch](https://github.com/pcuenca/swift-candle-diffusers), and run it on my iPhone.

Not sure if there's an easy way to isolate a unit test (not even sure it fails on simulator).

pcuenca:
> > it's only when you're working with `no_std` targets that you would have to manually add the `libc` dependency
> 
> Also I this is a big fat lie. The actual answer is that all these changes are in `candle-core` which already has the `libc` dependency.

It still fails compilation for me (for iOS targets) unless I do this:

```diff
diff --git a/candle-core/Cargo.toml b/candle-core/Cargo.toml
index 2dddaf4f..09f39117 100644
--- a/candle-core/Cargo.toml
+++ b/candle-core/Cargo.toml
@@ -55,6 +55,7 @@ metal = [
     "dep:objc2-metal",
     "dep:objc2-foundation",
     "dep:candle-metal-kernels",
+    "dep:libc",
     "dep:ug-metal",
 ]
```

pcuenca:
> simulator

Does not work out of the box: 
> failed assertion `Dispatch Threads with Non-Uniform Threadgroup Size is not supported on this device'

ivarflakstad:
@TimmyOVO would you mind making this change?

> ```diff
> diff --git a/candle-core/Cargo.toml b/candle-core/Cargo.toml
> index 2dddaf4f..09f39117 100644
> --- a/candle-core/Cargo.toml
> +++ b/candle-core/Cargo.toml
> @@ -55,6 +55,7 @@ metal = [
>      "dep:objc2-metal",
>      "dep:objc2-foundation",
>      "dep:candle-metal-kernels",
> +    "dep:libc",
>      "dep:ug-metal",
>  ]
> ```

ivarflakstad:
Well.. that didn't go as planned.
I just wanted to update the PR with the changes from main, but apparently since the source of this PR is your main branch when there are changes github automatically closes the PR? And I'm unable to reopen. Great.

edit: Oh I see what happened now. I pushed candle/main to TimmyOVO/main with `git push timmyovo main` when I meant to `git push timmyovo TimmyOVO:main`.
If you add me to your fork so I have push to your main I can fix it. Apologies, I'm a bit sleep deprived 😅

TimmyOVO:
> Well.. that didn't go as planned. I just wanted to update the PR with the changes from main, but apparently since the source of this PR is your main branch when there are changes github automatically closes the PR? And I'm unable to reopen. Great.
> 
> edit: Oh I see what happened now. I pushed candle/main to TimmyOVO/main with `git push timmyovo main` when I meant to `git push timmyovo TimmyOVO:main`. If you add me to your fork so I have push to your main I can fix it. Apologies, I'm a bit sleep deprived 😅

I've been quite busy with work lately,apologize for the late reply. I should be able to find some time this weekend to open a new PR for detailed testing on the iPhone

ivarflakstad:
@TimmyOVO Could you give me push access to your candle fork main branch? ☺️
That way I can at least rectify my mistake.

TimmyOVO:
> ivarflakstad

Absolutely, invite sent.

TimmyOVO:
@ivarflakstad @pcuenca After some investigation, I noticed that on iOS, due to limited physical memory, the system is under more frequent reclaim pressure, making Metal more likely to hit “sync/flush” paths. I've ensured that allocations happen before acquiring a compute encoder, and confirmed that the deadlock regression no longer reproduces under these forced settings.

image


Repro command (`macOS`):

> CANDLE_METAL_PENDING_LIMIT_MB=1 \
    CANDLE_METAL_CACHE_LIMIT_MB=128 \
    CANDLE_METAL_COMPUTE_PER_BUFFER=1 \
    CANDLE_METAL_COMMAND_POOL_SIZE=1 \
    cargo run -p candle-nn --example metal_rss_spike --features metal

```rust
use candle::{DType, Device, Tensor, D};

#[cfg(not(feature = "metal"))]
fn main() {
    eprintln!("This example requires `--features metal`.");
}

#[cfg(feature = "metal")]
fn main() -> candle::Result<()> {
    fn env_usize(key: &str, default: usize) -> usize {
        std::env::var(key)
            .ok()
            .and_then(|v| v.trim().parse::<usize>().ok())
            .unwrap_or(default)
    }

    fn env_bool(key: &str, default: bool) -> bool {
        std::env::var(key)
            .ok()
            .and_then(|v| v.trim().parse::<u8>().ok())
            .map(|v| v != 0)
            .unwrap_or(default)
    }

    fn max_rss_mb() -> f64 {
        unsafe {
            let mut usage: libc::rusage = core::mem::zeroed();
            if libc::getrusage(libc::RUSAGE_SELF, &mut usage) == 0 {
                // On macOS, ru_maxrss is in bytes. (On Linux it's in KiB.)
                let bytes = usage.ru_maxrss as f64;
                return bytes / (1024.0 * 1024.0);
            }
        }
        0.0
    }

    let dev = Device::new_metal(0)?;
    let dtype = DType::BF16;

    let batch = env_usize("CANDLE_METAL_RSS_BATCH", 1);
    let num_heads = env_usize("CANDLE_METAL_RSS_NUM_HEADS", 12);
    let hidden = env_usize("CANDLE_METAL_RSS_HIDDEN", 1536);
    let seq_len = env_usize("CANDLE_METAL_RSS_SEQ_LEN", 8192);
    let total_len = env_usize("CANDLE_METAL_RSS_TOTAL_LEN", seq_len);
    let layers = env_usize("CANDLE_METAL_RSS_LAYERS", 32);
    let use_mask = env_bool("CANDLE_METAL_RSS_USE_MASK", true);

    let head_dim = hidden / num_heads;
    let bh = batch * num_heads;

    eprintln!(
        "metal_rss_spike: batch={batch} heads={num_heads} head_dim={head_dim} seq_len={seq_len} total_len={total_len} layers={layers} use_mask={use_mask} dtype={dtype:?}",
    );
    eprintln!("max_rss_mb(start)={:.1}", max_rss_mb());

    let q_flat = Tensor::randn(0f32, 1f32, (bh, seq_len, head_dim), &dev)?.to_dtype(dtype)?;
    let k_tiled =
        Tensor::randn(0f32, 1f32, (bh, head_dim, total_len), &dev)?.to_dtype(dtype)?;
    let v_flat =
        Tensor::randn(0f32, 1f32, (bh, total_len, head_dim), &dev)?.to_dtype(dtype)?;

    let mask = if use_mask {
        Some(Tensor::randn(0f32, 1f32, (batch, 1, seq_len, total_len), &dev)?.to_dtype(dtype)?)
    } else {
        None
    };

    for layer in 0..layers {
        eprintln!("layer {layer}: max_rss_mb(pre)={:.1}", max_rss_mb());
        let mut scores = q_flat.matmul(&k_tiled)?;
        let scale = 1.0f64 / (head_dim as f64).sqrt();
        let scale_tensor = Tensor::full(scale as f32, (), scores.device())?.to_dtype(dtype)?;
        scores = scores.broadcast_mul(&scale_tensor)?;
        if let Some(mask) = &mask {
            let expanded = mask
                .expand((batch, num_heads, seq_len, total_len))?
                .reshape((bh, seq_len, total_len))?;
            scores = scores.add(&expanded)?;
        }

        let probs = candle_nn::ops::softmax(&scores, D::Minus1)?;
        let _ctx = probs.matmul(&v_flat)?;
        eprintln!("layer {layer}: max_rss_mb(post)={:.1}", max_rss_mb());
    }

    dev.synchronize()?;
    eprintln!("max_rss_mb(done)={:.1}", max_rss_mb());
    Ok(())
}
```

Issue #3199, opened 2025-11-18
CUDA integer reductions use undefined initialization (±INFINITY) causing incorrect max/min on Blackwell (sm_100)


TimmyOVO:
## Description

`candle-kernels` integer reduction kernels (`fast_max`, `fast_min`, `fast_argmax`, `fast_argmin`) initialize __shared__ buffers with -`INFINITY`/`INFINITY` even for integer types (`int64_t`/`u32`/`u8`). Casting ±∞ to integers is UB. On NVIDIA Blackwell (sm_100) maybe even newer gpu, this UB manifests as garbage initial values in shared memory, so max_all(i64) (and similar) can return huge bogus integers after running a `conv2d` kernel. This breaks downstream logic (e.g., RoPE cache sizing) in `candle-core` users.

## Minimum exmaples to reproduce this issue

<details>

<summary>Cargo.toml</summary>

```toml
[dependencies]
anyhow = "1.0.100"
candle-core = { path = "../candle/candle-core", features = ["cuda"] }
```

</details>

<details>

<summary>main.rs</summary>

```rust
use anyhow::Result;
use candle_core::{DType, Device, Tensor};

fn main() -> Result<()> {
    let device = Device::new_cuda(0).unwrap_or_else(|e| panic!("failed to init CUDA device {e:?}"));
    let cpu = Device::Cpu;

    // input [1, 3, 1024, 1024]，BF16
    let b_size: usize = 1;
    let c_in: usize = 3;
    let i_h: usize = 1024;
    let i_w: usize = 1024;

    //  [768, 3, 16, 16]，BF16
    let c_out: usize = 768;
    let k_h: usize = 16;
    let k_w: usize = 16;

    // conv2d hp
    let padding: usize = 0;
    let stride: usize = 16;
    let dilation: usize = 1;
    let groups: usize = 1;

    println!(
        "Input shape:  [{b_size}, {c_in}, {i_h}, {i_w}], dtype=BF16, device={:?}",
        device
    );
    println!(
        "Kernel shape: [{c_out}, {c_in}, {k_h}, {k_w}], dtype=BF16, device={:?}",
        device
    );
    println!(
        "Conv2d params: padding={padding}, stride={stride}, dilation={dilation}, groups={groups}"
    );

    let x = Tensor::rand(0f32, 1f32, (b_size, c_in, i_h, i_w), &device)?.to_dtype(DType::BF16)?;

[...]

PR #3200, opened 2025-11-18
CUDA: Fix integer reductions by removing +/-INF initialization

TimmyOVO:
# Whats changed
  * Fix the CUDA reduction kernels (`fast_{min,max,argmin,argmax}`) so that integer
    specializations no longer initialize `__shared__` buffers with `+/-INFINITY`.
    Instead, use well-defined identity values (`INT64_MIN`, `INT64_MAX`, `UINT32_MAX`,
    etc.). This removes the UB that caused garbage outputs from `max_all(i64)` on
    Blackwell after running other kernels like conv2d.

Related issue #3199


TimmyOVO:
You are right, this does have better readability

candle-kernels/src/reduce.cu
```diff
@@ -301,7 +354,9 @@ fast_max(const size_t src_numel, const size_t el_to_sum_per_block,
   size_t tid = threadIdx.x;
   size_t dst_id = blockIdx.x;
 
-  shr[tid] = -INFINITY;
+  // Initialize with the lowest representable value for T so that the first
+  // comparison in the reduction always picks a real element.
+  shr[tid] = reduce_init_lowest<T>();
```
ivarflakstad, line 359: I'm saying you could use `::cuda::std::numeric_limits<T>::lowest()` directly here :)

TimmyOVO: There are several reasons why I did this
1. `T` is not just float/int, but also includes CUDA custom types like `__half`/`__nv_bfloat16`, and `cuda::std::numeric_limits<T>` may not have specializations for all of them, im not sure when writing this ;
2. We only want to change the behavior for "integer types", while keeping the original `±INFINITY` semantics for floating-point types to minimize behavioral changes;
3. The initialization logic is used in multiple kernels (fast_max/fast_min/fast_argmax/fast_argmin), so extracting it into a helper might makes it more centralized and easier to maintain.

TimmyOVO: Sorry, i double checked this https://github.com/NVIDIA/cccl/issues/525, but i think for `candle` as a crate which we might want to be buildable on a wider range of toolchains and wants the smallest possible behavior change.

ivarflakstad: Right. I don't think we're targetting a wider range of _cuda_ toolchains, but it fixes a real bug so I'll gladly accept it ☺️

PR #3201, opened 2025-11-19
metal: fix a tiny issue in blit usage

amritsingh183:
# Whats changed

There is a small typo as well as general usage error for `blit`. `blie.set_label("blit_to_cpu")?;` should be `blit.set_label("blit_to_cpu");`.

i was installing candle from git to use `qwen3-vl` and this small mistake is preventing me from using it effectively.

### DIFF
--- a/candle-core/src/quantized/metal.rs
+++ b/candle-core/src/quantized/metal.rs
@@ -38,7 +38,7 @@ impl QMetalStorage {
 
         let buffer = self.device.allocate_buffer(self.buffer.length())?;
         let blit = self.device.blit_command_encoder()?;
-        blie.set_label("blit_to_cpu")?;
+        blit.set_label("blit_to_cpu");
         blit.copy_from_buffer(&self.buffer, 0, &buffer, 0, self.buffer.length());
         blit.end_encoding();
         self.device.wait_until_completed()?;


benjamin21st:
It baffles me how this typo got merged into the repo in the first place. I love rust and am thrilled to see its progress in ML, but failing to capture a basic typo kind of slaps the language, community, and ecosystem in the face 🤦‍♂️

amritsingh183:
`cargo test --features metal` was able to catch this. so now i am wondering just like you, how did this code make it to the repo!!

amritsingh183:
@benjamin21st the issue is related to ci cd setup used in mr review and merge. It has nothing to do with the Rust language or community. Rust catches this immediately as soon as you write it. As mentioned in my last comment, cargo also catches it during testing.

These kind of issues can be avoided by improving the MR process.

ivarflakstad:
> `cargo test --features metal` was able to catch this. so now i am wondering just like you, how did this code make it to the repo!!

It definitely should not have made it into the repo. That's my mistake.
But I can explain why it wasn't caught automatically by CI:
To build the project with `--features=metal` you need a machine that actually supports it. Mac, iphone, etc.

This means that we can't run tests or clippy etc on the code behind that feature flag on other machines.
(I'm going to double check wrt clippy, but afaik it doesn't have the capability to lint without building the project)

Since we don't have an apple machine in our CI it's up to the maintainers to ensure this doesn't happen. That's me. Apologies again 😅

benjamin21st:
Ah that's a solid point! Although I wonder if the code paths can be covered in tests because typos like this can in theory be caught quite easily.

amritsingh183:
i think we can use https://github.blog/changelog/2022-10-03-github-actions-jobs-running-on-macos-latest-are-now-running-on-macos-12/ and something on lines on the following in ci

```
jobs:
  # New job specifically for Metal verification
  check-macos-metal:
    runs-on: macos-latest
    steps:
      - uses: actions/checkout@v4
      - uses: dtolnay/rust-toolchain@stable
      - name: Check Metal Feature
        # Validates metal logic on actual hardware
        run: cargo check --features metal

```

ivarflakstad:
That's a great idea. I have been so locked in on getting `cargo test --features=metal` to work in CI (which needs an actual machine, not just an image) that I totally overlooked that `cargo check --features=metal` is within reach!

There is one caveat though:
This should work now (I assume), since currently all metal kernels are compiled at runtime.
In the near future we want to add some precompilation of metal kernels at buildtime. It makes maintenance much easier, code will be easier to reason about, and adding new kernels will be simpler as well.
Hopefully the macos image includes the tools required to compile metal kernels, but it may cause some issues. That's a problem for future me though.

ivarflakstad:
Update: unfortunately the macos image isn't quite the same as the real deal. See failed run [here](https://github.com/huggingface/candle/actions/runs/19566365724/job/56029227922?pr=3207).

amritsingh183:
that;s unfortunate... so what are we going to do? i ll try to understand this and find a fix ....if there is one..

ivarflakstad:
Ah, nothing you can do unfortunately.

HF would have to set up a mac dedicated to run CI on (or rent through a provider).
I've seen other projects use circleci to achieve this.

amritsingh183:
ok then.. thanks for looking into this :-)

PR #3207, opened 2025-11-21
[Metal] Add macos clippy runner with metal feature flag

ivarflakstad:
Having CI running with `--features=metal` would be a real QOL improvement for the project.
This PR attempts to enable `cargo clippy --features=metal ...` so that we at least have CI linting on the metal specific code.


ivarflakstad:
Unfortunately that didn't work. We'll need a real machine to enable CI with `--features=metal`.
@pcuenca for visibility.

PR #3209, opened 2025-11-22
chore: switch to `cudarc`'s `dynamic-loading` from `dynamic-linking`

### LINKED ISSUE #3208: `cudarc` dynamic loading support
Currently, `candle` uses `cudarc` with the `dynamic-linking` feature, which requires the executable to find the DLLs or SOs at startup. However, it would be more convenient if `candle` also supported the `dynamic-loading` feature from `cudarc` to load DLLs or SOs at runtime.
Is it possible for `candle` to support it?

mayocream:
Fix #3208


ivarflakstad:
I was actually still planning to merge this, we're just blocked by `ug` so I need to either fork it, hide it behind a feature flag, or remove it.

Is there a specific reason (other than time passing) for why you closed this? :)

mayocream:
> I was actually still planning to merge this, we're just blocked by `ug` so I need to either fork it, hide it behind a feature flag, or remove it.
> 
> Is there a specific reason (other than time passing) for why you closed this? :)

Ah, I think i deleted this branch by accident, let me reopen this.

Also, please note that both `float8` and `ug` need to be updated to use the same feature! otherwise it will fail to compile due to both `dynamic-loading` and `dynamic-linking` are set.

mayocream:
let me reopen this!

Issue #691, opened 2023-08-31
Math is difficult to implement


timokoesters:
In Python it's possible to write just this:
`3 * x**2 - 4*x + 5`

With Candle it currently looks like this, which is difficult to read and write:
`Ok((&(&(&x.powf(2.0)? * 3.0)? - &x * 4.0)? + 5.0)?)`

1. Why does everything return a Result? I assume the reason is that the GPU connection might fail, but in that case maybe it's okay to panic the whole application?
2. Why is Tensor not Copy? It's an Arc, so it would be okay in theory.
3. A few more traits can be implemented to make it look like this: `3.0 * x * x - 4.0 * x - 5.0`

I wrote an example here: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e5abf73c9e38fb6cef2c17b5a1898a29

LaurentMazare:
Yes the syntax is cumbersome, we have some ideas that should slightly improve it though it will not be as simple as the version you suggest. To answer your points:
1. All the ops may indeed fail, the most common thing is not really the GPU going down but more the GPU running out of memory, using invalid shape, etc.
2. I don't think a `Tensor` could be copy, copy implies that you can just do copy the immediate memory representation (see [here](https://doc.rust-lang.org/std/marker/trait.Copy.html)) which isn't the case for `Arc`. Hence `Tensor` is only clone - and the clone op is pretty inexpensive.

LaurentMazare:
Both of these are actually already done :)

timokoesters:
Oh that's great, maybe I was using an old version, or it only worked one way.

timokoesters:
It looks like Tensor + Result is implemented, but not Result + Tensor or Result + Result.

I have another related problem: I'm unable to write a rusty variant for modifying a single index:
var[2, 3] += 1, is there a method that can help me with that?

LaurentMazare:
Re `Result + Result` I think this is because of the foreign trait limitation of rust so not much that we can do here.
Re assigning single values in place, we don't really have an op to do this and it's on purpose as these tend to be really inefficient so we prefer favoring more functional approaches and avoiding accessing data by index. I don't think this has been an issue for all the models we've been looking at so far but if there are some specific models for which this would be a limitation, certainly curious to hear about them.

timokoesters:
[...]

PR #3214, opened 2025-11-25
[WIP] TensorResult abstraction

ivarflakstad:
Currently when we chain math with `Result<Tensor>` we have to write code like
```rust
(&(&(&x.powf(2.0)? * 3.0)? - &x * 4.0)? + 5.0)?
```

Given that we change all our methods to return the newtype `TensorResult` instead of `Result<Tensor>`, we could simply write
```rust
(x.powf(2.0) * 3.0 - x * 4.0 + 5.0).ok()?;
```
Where `.ok()` does `TensorResult -> Result<Tensor>`

The only unanswered question is what to do when we have two `Err` results.
My proposed solution is to simply aggregate them. Either via a new `candle_core::Error` enum type that wraps two, or by simply using `Error::context` (`e1.context(e2)`).

Ref #691

### DIFF
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -2914,3 +2914,95 @@ impl<S: Into<Shape>> From<(Storage, S)> for Tensor {
         from_storage(storage, shape, BackpropOp::none(), false)
     }
 }
+
+struct TensorResult(Result<Tensor>);
+
+impl TensorResult {
+    fn ok(self) -> Result<Tensor> {
+        self.0
+    }
+}
+
+impl From<Result<Tensor>> for TensorResult {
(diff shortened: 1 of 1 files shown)


EricLBuehler:
Looks awesome.

>The only unanswered question is what to do when we have two Err results.
My proposed solution is to simply aggregate them. Either via a new candle_core::Error enum type that wraps two, or by simply using Error::context (e1.context(e2)).

I think a new `Error` variant would be better, since the `.context` might not exactly match expectations.

Maybe this can tie in with some of our specialized backend work?

sempervictus:
Big fan of explicit error-typing for maintainability and security reasons. Convenience tucks of types into others can later be abused for type-confusion logic by red team even though Rust itself is thought to be safe from such things.

PR #3216, opened 2025-11-27
Pinned memory

ahadnagy:
This PR adds support for pinned-memory backed tensors. Since the custom allocator trait is not stable in Rust yet (and I don't see that happening soon), this PR uses polyfill libraries to stay on the stable release. If there's a better approach, I'm all ears. :)

All of this is gated behind the `pinned-memory` feature which also enables `cuda` due to its nature.


candle-core/src/cpu_backend/mod.rs
```diff
@@ -10,6 +10,41 @@ mod utils;
 pub use utils::{
     binary_map, binary_map_vec, unary_map, unary_map_vec, Map1, Map1Any, Map2, Map2InPlace, Map2U8,
 };
+
+// Conditionally use polyfill Vec when pinned-memory feature is enabled
+#[cfg(not(feature = "pinned-memory"))]
+use std::vec::Vec;
+
+#[cfg(feature = "pinned-memory")]
+use allocator_api2::vec::Vec;
```
ivarflakstad: Could we perhaps move this to a `candle_core/src/vec` module, where based on the feature flag we import either the `std` or `allocator_api2` impl?

candle-core/src/cpu_backend/mod.rs
```diff
@@ -10,6 +10,41 @@ mod utils;
...
+}
+
+// Helper macro to create a Vec filled with a value (works with both std and polyfill Vec)
+macro_rules! storage_vec {
+    ($val:expr; $len:expr) => {{
+        let mut v = Vec::with_capacity($len);
+        v.resize($len, $val);
+        v
+    }};
+}
```
ivarflakstad: Kind of a nit but I'd like for this to support the same branches as the original macro.
So something like
```suggestion
#[cfg(feature = "pinned-memory")]
pub use allocator_api2::vec as storage_vec;
#[cfg(not(feature = "pinned-memory")]
pub use std::vec as storage_vec;
```

candle-core/src/cuda_backend/pinned_allocator.rs
```diff
@@ -0,0 +1,111 @@
+//! CUDA pinned memory allocator support.
+//!
+//! This module is only available when both `pinned-memory` and `cuda` features are enabled.
+//! The `pinned-memory` feature automatically enables `cuda` (see Cargo.toml), so checking
+//! for `pinned-memory` is sufficient.
+
+#[cfg(feature = "pinned-memory")]
```
ivarflakstad: I don't think we need all these as the entire file is behind the feature flag

candle-core/src/quantized/cuda.rs
```diff
@@ -577,25 +426,29 @@ impl QCudaStorage {
         let buffer = self
             .device
             .memcpy_dtov(&self.data.inner.slice(..self.data.len))?;
-        let mut out = vec![0.0; elem_count];
+        let mut out_std: std::vec::Vec<f32> = (0..elem_count).map(|_| 0.0).collect();
```
ivarflakstad: If you're specifically using `std::vec::Vec` then I don't see why you can't simply use good ol' `vec!`

candle-core/Cargo.toml
```diff
@@ -49,6 +49,8 @@ default = []
 cuda = ["cudarc", "dep:candle-kernels", "dep:ug-cuda", "float8/cuda"]
 cudnn = ["cuda", "cudarc/cudnn"]
 nccl = ["cuda", "cudarc/nccl"]
+# pinned-memory requires cuda because it uses CUDA APIs for pinned host memory allocation
+pinned-memory = ["cuda", "dep:allocator-api2"]
```
ivarflakstad: Clearly documented here, but when running across the feature flag in the wild it could be confusing as pinning memory is not a cuda specific concept.
Could we rename to `cuda-pinned-memory`. Perhaps `cuda-pin-mem` if you think it's getting tedious to read/write.

We may find uses for pinned memory in other contexts in the future, in which case I think we're approaching a structure like

```rust
pinned-memory = ["dep:allocator-api2"]
cuda-pin-mem = ["cuda", "pinned-memory"]
xxx-pin-mem = ["xxx", "pinned-memory"]
```
Where `pinned-memory` would surface a generic API and `*-pin-mem` would enable the backend specific implementations taking advantage of it.

ivarflakstad:
Officially alive again. Yay.

I think that covering every usage of `Vec` with the feature toggled `StorageVec` is kind of suboptimal. I think it would be better if we only create pinned memory where we need it.
I see how that could be difficult if you want `CpuStorage` to be backed by pinned memory, but when do you actually need that?

PR #3220, opened 2025-12-02
fix: add gguf wrapper in quantized gemma3

### LINKED ISSUE #3215: cannot find gemma3.attention.head_count in metadata
**version：** 0.9.2-alpha.1
**model：**  https://huggingface.co/unsloth/embeddinggemma-300m-GGUF 
**model file:** embeddinggemma-300M-Q8_0.gguf

my code: 
`
use candle_core::{IndexOp, Tensor, quantized::gguf_file};
use candle_transformers::models::quantized_gemma3::ModelWeights;
 
let mut file = std::fs::File::open(&model_file)?;
let model_content = gguf_file::Content::read(&mut file).map_err(|e| e.with_path(model_path))?;
let model = ModelWeights::from_gguf(model_content, &mut file, &device)?;
`
**error:
cannot find gemma3.attention.head_count in metadata**

check ModelWeights::from_gguf source code:
```
impl ModelWeights {
    pub fn from_gguf<R: std::io::Seek + std::io::Read>(
        ct: gguf_file::Content,
        reader: &mut R,
        device: &Device,
    ) -> Result<Self> {
        let md_get = |s: &str| match ct.metadata.get(s) {
            None => candle::bail!("**cannot find {s} in metadata**"),
            Some(v) => Ok(v),
        };
```

but, I use candle_transformers::models::quantized_qwen3 is fine, 

quantized_qwen3::ModelWeights::from_gguf source code:
```
impl ModelWeights {
        ct: gguf_file::Content,
        reader: &mut R,
        device: &Device,
    ) -> Result<Self> {
        let mut gg = Gguf::new(ct, reader, device.clone());
        let md_get = |s: &str| match gg.metadata().get(s) {
            None => candle::bail!("cannot find {s} in metadata"),
            Some(v) => Ok(v),
        };
```

Should  quantized_gemma3::ModelWeights::from_gguf  be changed to add this line `let mut gg = Gguf::new(ct, reader, device.clone()); ` ?

junjunjd:
Add Gguf wrapper in quantized Gemma3 loader to fix missing metadata lookups.

Fixes #3215.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ exclude = [
 resolver = "2"
 
 [workspace.package]
-version = "0.9.2-alpha.1"
+version = "0.9.2-alpha.2"
 edition = "2021"
 description = "Minimalist ML framework."
 repository = "https://github.com/huggingface/candle"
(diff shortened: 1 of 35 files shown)


ivarflakstad:
@junjunjd Thanks for this!
I tried running the embedding-gemma model mentioned in the issue and unfortunately this doesn't fix the issue. Luckily the problem isn't has complex as it may seem.
Default gemma3 uses the prefix `gemma3.*`, while the embedding-gemma model uses `gemma-embedding.*` ([ref](https://huggingface.co/unsloth/embeddinggemma-300m-GGUF/blob/main/embeddinggemma-300m-Q4_0.gguf)).
`ModelWeights` in `quantized_gemma3.rs` is hard coded to use `gemma3`.
I'll post the findings in the issue as well.

clocksmith:
@junjunjd Thanks for taking a crack at this! The GGUF wrapper is a nice cleanup for the tensor loading calls

Unfortunately this doesn't quite fix the root issue. The problem is that `md_get("gemma3.attention.head_count") `still has the prefix **hardcoded** , and the wrapper passes through to the same metadata without transforming the keys.

Like @ivarflakstad mentioned

  - Regular Gemma3 models use gemma3.* keys
  - Embedding-gemma uses gemma-embedding.* keys (https://huggingface.co/unsloth/embeddinggemma-300m-GGUF)

If the actual prefix can be detected before lookups, then the problem can be addressed there. **Luckily it can!**

## The fix

Right before this existing on line 266 (currently not as robust):
```
let md_get = |s: &str| match ct.metadata.get(s) {
      None => candle::bail!("cannot find {s} in metadata"),
      Some(v) => Ok(v),
  };
```

This can be added, along with a tweak the `let md_get` block above

```
  // NEW
  let prefix = ["gemma3", "gemma2", "gemma", "gemma-embedding"]
      .iter()
      .find(|p| ct.metadata.contains_key(&format!("{}.attention.head_count", p)))
      .copied()
      .unwrap_or("gemma3");

  // TWEAKED
  let md_get = |s: &str| {
      let key = format!("{prefix}.{s}");
      match ct.metadata.get(&key) {
          None => candle::bail!("cannot find {key} in metadata"),
          Some(v) => Ok(v),
      }
  };
```

Then replace all instance of `md_get("gemma3.attention.head_count")` with  `md_get("attention.head_count")` in the following lines 271-285.

## Notes

BONUS: This proposed solutionalso has the advantage of robustness by "auto" discovering any future Gemma variants (gemma4, gemma-vision, etc.) without code changes.

NOTE: This wrapper could still be useful alongside this for a cleaner tensor API - they're complementary changes! 

ALTS: I had this issue on something similar, but solution was not as tight, here: https://github.com/clocksmith/gamma/blob/83d0565c938b18a363026070b526885682450c79/src/core/hardware/gguf_parser.py#L129 This also works, but the hardcoded list of names is probably better for a library of this maturity.

Happy to help further if you'd like to update the PR.

ivarflakstad:
Closing since the root issue has been fixed. If you want to contribute with the intention of improving the gguf API feel free :)

PR #3224, opened 2025-12-03
Add yolo-v5 example

### LINKED ISSUE #1858: Will candle support YOLOV5？
The current project is based on YOLOV5，Pytorch.  We are required to use rust/wasm to do the same thing（Reuse existing Pytorch model）.  Now It's already supported YOLOV8，so Will candle support YOLOV5？ and when？

mayocream:
Fix #1858

### DIFF
--- a/candle-examples/examples/yolo-v5/README.md
+++ b/candle-examples/examples/yolo-v5/README.md
@@ -5,7 +5,7 @@ Candle implementation of [YOLOv5](https://github.com/ultralytics/yolov5) for obj
 ## Running an example
 
 ```bash
-$ cargo run --example yolo-v5 --release -- candle-examples/examples/yolo-v5/assets/bike.jpg
+$ cargo run --example yolo-v5 --release -- candle-examples/examples/yolo-v8/assets/bike.jpg
 ```
 
 ## Command line flags


candle-examples/examples/yolo-v5/model.rs
sdabbour, line 1: Would this model be better placed in the `candle-transformers/src/models/` dir?

candle-examples/examples/yolo-v5/convert_to_safetensors.py
sdabbour, line 1: Not sure if this is necessary to include, could maybe describe a bit how to get the model in the README.md file

sdabbour (commented):
I've been diving into the code recently to investigate implementation of an example for Qwen 3 VL. These are just some things I noticed from a layout perspective.

candle-examples/examples/yolo-v5/convert_to_safetensors.py
slckl: I think it's probably fine for a small stand-alone python script to be included, especially one that is somewhat obvious like this one.

if there's anything I'd change, is to maybe make it stand-alone executable with `uv` by including dependencies in the script itself: https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

then one can just `uv run script.py` without much thinking.

candle-examples/examples/yolo-v5/model.rs
slckl: Probably not, yolov5 is quite old, it's ok to implement your own models for whoever needs it based on `candle-examples`?

candle-examples/examples/yolo-v5/README.md
```diff
@@ -0,0 +1,18 @@
+# yolo-v5
+
+Candle implementation of [YOLOv5](https://github.com/ultralytics/yolov5) for object detection.
+
+## Running an example
+
+```bash
+$ cargo run --example yolo-v5 --release -- candle-examples/examples/yolo-v8/assets/bike.jpg
```
slckl, line 8: I can confirm this works and produces sane-looking output :smiley:

candle-examples/examples/yolo-v5/main.rs
```diff
@@ -0,0 +1,179 @@
...
+    for bbox in bboxes {
+        imageproc::drawing::draw_hollow_rect_mut(
+            image,
+            imageproc::rect::Rect::at(bbox.xmin as i32, bbox.ymin as i32).of_size(
+                (bbox.xmax - bbox.xmin) as u32,
+                (bbox.ymax - bbox.ymin) as u32,
+            ),
+            image::Rgba([255, 0, 0, 255]),
+        );
+    }
```
slckl, line 155: can you please include class labels + confidence on the output image? similar to how yolo-v8 example does it?

I would also be grateful if you println logged the class names + bbox data similar to how yolo-v8 example does it? It's very useful to validate any numerical changes in the library etc:
```rust
            println!(
                "{}: {:?}",
                candle_examples::coco_classes::NAMES[class_index],
                b
            );
```

candle-examples/examples/yolo-v5/convert_to_safetensors.py
ivarflakstad: I'd actually prefer if we did this transformation once and uploaded the weights to huggingface. Like Laurent did with v8 [here](https://huggingface.co/lmz/candle-yolo-v8/tree/main).

PR #3230, opened 2025-12-06
[Metal] unary and affine improvements

ivarflakstad:
All unary ops + `affine`, `powf`, and `elu` have the following performance improvements.
Additionally tiling is now automatic and easier to maintain/reason about.

unary/affine improvements


EricLBuehler (approved):
LGTM! Looks awesome.

PR #3234, opened 2025-12-08
replace cutlass submodule references with explicit build step

jacobgorm:
this has the benefit of not weighing down CI builds on non-CUDA platforms, and also enables shallow clone to make builds on CUDA platforms lighter weight. The cutlass checkout alone is 210MiB, which this commit saves on platforms that don't need it.

I don't have a modern enough CUDA GPU available to run with flash-attention, but I have tested on older CUDA machine as well as on MacOS, and modulo the asserts for the NVIDIA hardware version, the build completes as expected.

### DIFF
--- a/candle-kernels/build.rs
+++ b/candle-kernels/build.rs
@@ -46,7 +46,9 @@ fn main() {
     println!("cargo:rustc-link-search={}", out_dir.display());
     println!("cargo:rustc-link-lib=moe");
     println!("cargo:rustc-link-lib=dylib=cudart");
-    println!("cargo:rustc-link-lib=stdc++");
+    if !is_target_msvc {
+        println!("cargo:rustc-link-lib=stdc++");
+    }
 }
 
 fn remove_lines<P: AsRef<std::path::Path>>(file: P, patterns: &[&str]) {


candle-flash-attn-build/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    // Clone cutlass if the directory doesn't exist
+    if !cutlass_dir.exists() {
+        println!("cargo::warning=Cloning cutlass from {}", CUTLASS_REPO);
+        let status = Command::new("git")
+            .args([
+                "clone",
+                "--filter=blob:none",
+                "--sparse",
```
ivarflakstad: I had to remove this line for the build to work correctly.

ivarflakstad:
git v2.37.3 works.
Not sure if we can narrow it down more without testing each version hehe

ivarflakstad:
Oh, nvm found a good overview for linux
https://launchpad.net/git/+packages

I think we should be good if OS used is fairly new.

candle-flash-attn-v3/build.rs
```diff
@@ -190,7 +197,7 @@ fn main() -> Result<()> {
                 command.args(["--default-stream", "per-thread"]);
 
                 // Include path
-                command.arg("-Icutlass/include");
+                command.arg(&cutlass_include);
```
ivarflakstad, line 200: You have to use the same box leak trick here

jacobgorm: Pushed untested fix (untested due to not have the relevant hardware)

candle-flash-attn-build/src/lib.rs
jacobgorm: So you are saying we can keep the sparse checkout?

ivarflakstad: Well basically it means if you're on ubuntu 20 you have to update your `git`.
If keep sparse checkout then I would like a error message that explains exactly why the build step failed and how to fix it.
Aka I would do something like
```rust
if let Ok(status) = Command::new("git")
    .args([
        "clone",
        "--filter=blob:none",
        "--sparse",
        CUTLASS_REPO,
        cutlass_dir.to_str().unwrap(),
    ])
    .status()
} else {
    // check git version
    // if <= 2.25 then recommend updating.
    // Or perhaps try `git clone depth 1` as a fallback? That also worked with 2.25.1 - including the sparse usage below.
}
```

jacobgorm: I think we should go for --depth 1 for simplicity then.

candle-flash-attn-v3/Cargo.toml
```diff
@@ -19,6 +19,7 @@ half = { version = "2.3.1", features = ["num-traits"] }
 anyhow = { version = "1", features = ["backtrace"] }
 num_cpus = "1.15.0"
 rayon = "1.7.0"
+candle-flash-attn-build = { path = "../candle-flash-attn-build" }
```
ivarflakstad, line 22: We need to actually publish a version of the build crate for it to be found by the crate that depends on it when used outside of the candle repo.
Let's make it follow the same versioning pattern as the rest of candle while we're at it 👍

jacobgorm: Do you mean like this?
```
diff --git a/candle-flash-attn-build/Cargo.toml b/candle-flash-attn-build/Cargo.toml
index b41ed97d..e9c4a00a 100644
--- a/candle-flash-attn-build/Cargo.toml
+++ b/candle-flash-attn-build/Cargo.toml
@@ -1,7 +1,8 @@
 [package]
 name = "candle-flash-attn-build"
-version = "0.1.0"
+version = "0.9.2-alpha.2"
 edition = "2021"
+
 description = "Build utilities for candle flash attention crates"
 license = "MIT OR Apache-2.0"
```

ivarflakstad: Exactly - and the crates that use it must refer to the (soon to be) published crate as well.

jacobgorm: But does this not cover the last part already?

```
$ git grep flash-attn-build
Cargo.toml:    "candle-flash-attn-build",
candle-flash-attn-build/Cargo.toml:name = "candle-flash-attn-build"
candle-flash-attn-v3/Cargo.toml:candle-flash-attn-build = { path = "../candle-flash-attn-build" }
candle-flash-attn/Cargo.toml:candle-flash-attn-build = { path = "../candle-flash-attn-build" }
```

ivarflakstad: Try it yourself :) 
New project. Add your fork (and branch) of `candle-flash-attn` as a dependency. Run.

My hypothesis is that `candle-flash-attn` will not find the build crate on your system. Because in a new project you don't have `../candle-flash-attn-build`

jacobgorm: I've added it to the top-level Cargo.toml. As far as I can test without modern-enough CUDA hardware, this gets resolved and build kicks off like it should.

PR #3236, opened 2025-12-09
feat(qwen2): add KV cache management and selective attention

danielclough:
# Summary
Adds KV cache management and fixes critical causal mask bug for Qwen2 multi-turn inference. Includes numerical precision improvements for RoPE and attention.

## Changes
- **Causal mask bug fix**: Corrects mask shape for cached decoding (was `[tgt, tgt]`, now `[tgt, total]`) - critical for multi-turn conversations
- **Precision improvements**: RoPE and softmax now use F32 intermediates to match PyTorch behavior
- **KV cache API**: Adds `extract_kv_cache`/`restore_kv_cache` methods for cache manipulation and inspection
- **Selective attention**: New `prepare_4d_causal_attention_mask_with_cache_position` for non-contiguous cache positions
- **Embedding injection**: `forward_from_embeds` methods enable custom embedding workflows (e.g., multimodal)
- **Stability fix**: Replaces `NEG_INFINITY` with `f32::MIN` to avoid NaN propagation when combining masks
- **Cache manipulation**: Adds `shift_kv_cache_first_to_last` for advanced patterns (e.g., negative prompt refresh)

## Motivation
The causal mask bug prevented proper multi-turn decoding with KV cache. The new cache management APIs enable advanced inference patterns like streaming audio generation (VibeVoice) and speculative decoding while maintaining precision for F16/BF16 inference.

## Breaking Changes
None - all changes are backward compatible additions or bug fixes.

### Routine
cargo fmt --all
cargo test -p candle-transformers
cargo clippy -p candle-transformers

#### Simple Query
cargo run --example qwen --features metal --release  -- --prompt "Write a poem about butterflies. <think></think>." --model "2-1.5b"

#### Test with very short prompt to ensure single-token decode works
cargo run --example qwen --features metal --release -- --prompt "Hi" --sample-len 10 --model "2-1.5b"


candle-transformers/src/models/qwen2.rs
```diff
@@ -49,17 +49,23 @@ impl RotaryEmbedding {
         let max_seq_len = cfg.max_position_embeddings;
         let inv_freq: Vec<_> = (0..dim)
             .step_by(2)
-            .map(|i| 1f32 / cfg.rope_theta.powf(i as f64 / dim as f64) as f32)
+            .map(|i| 1f32 / (cfg.rope_theta as f32).powf(i as f32 / dim as f32))
             .collect();
         let inv_freq_len = inv_freq.len();
-        let inv_freq = Tensor::from_vec(inv_freq, (1, inv_freq_len), dev)?.to_dtype(dtype)?;
+
+        // Keep in f32 for precision (Python keeps inv_freq as float32)
```
ivarflakstad: You probably meant to write pytorch.
I don't mind documenting special behaviour with "to match pytorch behaviour" etc, but most of these comments (like this one) are good on their own.
Aka I think `// Keep in f32 for precision` is enough. It's not like pytorch invented numerical precision.

candle-transformers/src/models/qwen2.rs
```diff
@@ -49,17 +49,23 @@ impl RotaryEmbedding {
...
         let t = Tensor::arange(0u32, max_seq_len as u32, dev)?
-            .to_dtype(dtype)?
+            .to_dtype(DType::F32)? // f32, not model dtype
             .reshape((max_seq_len, 1))?;
-        let freqs = t.matmul(&inv_freq)?;
+
+        // Compute freqs, sin, cos all in f32 for precision
+        let freqs = t.matmul(&inv_freq)?; // f32 matmul
+
+        // Only cast to model dtype at the very end (matches Python's return statement)
```
ivarflakstad: Fine without `(matches Python's return statement)`

candle-transformers/src/models/qwen2.rs
```diff
@@ -200,7 +206,13 @@ impl Attention {
...
                 Some(mask) => attn_weights.broadcast_add(mask)?,
             };
-            let attn_weights = candle_nn::ops::softmax_last_dim(&attn_weights)?;
+            // CRITICAL: Match Python's behavior - compute softmax in float32 then cast back
+            // Python: attn_weights = softmax(attn_weights, dim=-1, dtype=torch.float32).to(query.dtype)
+            let attn_weights = {
+                let original_dtype = attn_weights.dtype();
+                let attn_f32 = attn_weights.to_dtype(DType::F32)?;
+                candle_nn::ops::softmax_last_dim(&attn_f32)?.to_dtype(original_dtype)?
+            };
```
ivarflakstad: This is kinda interesting.
I implemented online softmax ([ref](https://arxiv.org/pdf/1805.02867)) for the metal backend a while ago. Online softmax ensures you won't overflow in the reduction and it also slightly improves numerical precision in general.
I haven't gotten to the cuda kernels yet, but we should definitely do the same there.
IIRC is almost no extra cost in doing online softmax reduction in f32 (load as T, cast and reduce as f32), so perhaps we should just make the change there instead of casting back and forth here.

In the meantime I don't mind this approach, but let's add a TODO comment.

Oh and "python" ;)

candle-transformers/src/models/qwen2.rs
```diff
@@ -331,32 +378,163 @@ impl Model {
...
+                // Sliding window: can't attend to positions too far in the past
+                let is_too_old = j + self.sliding_window < abs_query_pos;
+
+                if is_future || is_too_old {
+                    mask_data.push(min_dtype);
+                } else {
+                    mask_data.push(0.0);
+                }
+            }
+        }
```
ivarflakstad: This looks pretty close to a `where_cond` where `t: min_dtype` and `f: 0.0`. Would be nice if we didn't have to move `cache_position` into the cpu for this, but hard to express `is_future` and `is_too_old` though.
I'll give it a think.

ivarflakstad (commented):
There's a lot of good stuff here!
A slight issue is that on cuda performance drops by 20%. Would be nice to figure out if we can easily avoid this drop before merging (I'm ok with accuracy > performance)

candle-transformers/src/models/qwen2.rs
danielclough: I checked in candle-kernels/src/reduce.cu and see:
```
// TODO: Maybe add some fast_sum_f16_f32 variant that not only accumulate in f32
// but also expect a f32 output so that this can be used for normalization e.g.
// in softmax.
```
So, it seems this dtype cast is redundant in every case as it's already being done in f32.

I have tested both qwen and vibevoice-rs and get the same output with a speedup by removing it.

ivarflakstad:
Ok, so I think to get this merged more easily we should pick out some of these improvements instead of doing it as one large PR.
So the ones I think we can easily merge are
* Causal mask bug fix
* Precision improvements (granted that it is in model code only)
* KV cache API
* Stability fix

And these I'd prefer this as a separate PR(s).
* Selective attention
* Embedding injection
* Cache manipulation

PR #3239, opened 2025-12-09
[Metal] further improve unary and binary

ivarflakstad:
Unary throughput improved by 9-25%
Affine/powf/elu throughput improved by 4-18%
Binary (contiguous) throughput improved by 2-5%
Binary (one strided, one contiguous) throughput improved by 95-98%

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,6 +106,3 @@ objc2-foundation = { version = "0.3.1" }
 [profile.release-with-debug]
 inherits = "release"
 debug = true
-
-[profile.dev]
-debug = 1


candle-metal-kernels/src/metal_src/unary.metal
```diff
@@ -18,12 +18,18 @@ METAL_FUNC uint get_strided_index(
     return strided_i;
 }
 
-#define MAX(x, y) ((x) > (y) ? (x) : (y))
+METAL_FUNC uint nonzero(uint n) {
```
EricLBuehler, line 21: Maybe we could use some other name like `div_ceil`?

candle-metal-kernels/src/metal_src/binary.metal
```diff
@@ -20,14 +20,44 @@ METAL_FUNC uint get_strided_index(
...
+        uint idx,
+        constant size_t &num_dims,
+        constant size_t *dims,
+        constant size_t *strides
+    ) {
+        return get_strided_index(idx, num_dims, dims, strides);
+    }
+};
+
+METAL_FUNC uint nonzero(uint n) {
```
EricLBuehler, line 45: Same here

EricLBuehler (approved):
🔥 

1 nit comment

candle-metal-kernels/src/metal_src/unary.metal
ivarflakstad: Indeed. I could change it to something like this.
I'll do it in a follow up.
```c++
template<uint X, uint Y>
constexpr uint div_ceil() {
    return X / Y + (X % Y > 0);
}

template<typename T>
constexpr uint work_per_thread() {
    return div_ceil<8, sizeof(T)>();
}
```

ivarflakstad: Fixed in #3241

PR #3245, opened 2025-12-16
feat: allow tokenizer to load from GGUF metadata

fffonion:
Right now a subset of pipeline are supported (qwen2, llama3 style BPE, gpt2).

Example:
```
cargo run --example gguf-tokenizer -- --model unsloth/Qwen3-4B-GGUF/Qwen3-4B-Q4_K_S.gguf --prompt "Hello"
```

To use in your code:

```rust
use candle_core::quantized::tokenizer::TokenizerFromGguf;

let content = gguf_file::Content::read(&mut reader).context("failed to load GGUF metadata")?;
let tokenizer =
        Tokenizer::from_gguf(&content).context("failed to initialize tokenizer from GGUF")?;
```

### DIFF
--- a/candle-core/Cargo.toml
+++ b/candle-core/Cargo.toml
@@ -33,6 +33,7 @@ safetensors = { workspace = true }
 thiserror = { workspace = true }
 yoke = { workspace = true }
 zip = { workspace = true }
+tokenizers = { workspace = true, features = ["onig"] }
 
 [target.'cfg(all(not(target_arch = "wasm32"), not(target_os = "ios")))'.dependencies]
 ug = { workspace = true }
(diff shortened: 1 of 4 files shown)


tardigrade-dot:
Subject: Fix Wasm compilation regression in candle-core (hardcoded onig feature)
Description
Hi, I noticed a regression in WebAssembly support due to the recently added onig feature in candle-core/Cargo.toml.

The dependency is currently hardcoded as:

`tokenizers = { workspace = true, features = ["onig"] }
`This forces onig-sys into the dependency graph for all targets, breaking wasm32-unknown-unknown builds since it cannot link against C headers.

Reproduction
```
cd candle-wasm-examples/whisper
sh build-lib.sh
```
Error:

```
error: failed to run custom build command for `onig_sys v69.9.1`
warning: onig_sys@69.9.1: oniguruma/src/regint.h:123:10: fatal error: 'stdlib.h' file not found
```
Suggested Fix
`tokenizers = { version = "0.22.0", default-features = false }`
This allows downstream projects to using features = ["onig"] or features = ["unstable_wasm"]

Thanks for the great work on Candle!

PR #3252, opened 2025-12-20
add HuberLoss

donjuanplatinum:
add HuberLoss and add Loss Trait
I guess  add a `Loss` trait  will make it easier to add more complex types of loss functions in the future


candle-nn/src/loss.rs
```diff
@@ -2,6 +2,53 @@
 //!
 use candle::{Result, Tensor};
 
+pub trait Loss: Sized {
+    type Config: Sized;
+    fn new(config: Self::Config) -> Result<Self>;
+    fn forward(&mut self, inp: &Tensor, target: &Tensor) -> Result<Tensor>;
+}
```
ivarflakstad: Traits are at their most useful when you use them to describe generic behaviour, and in rust you typically either use generics ala `fn calculate<L: Loss>(t1: Tensor, t2: Tensor, loss: L)` or as a dyn trait `fn calculate(t1: Tensor, t2: Tensor, loss: dyn Loss)`.
However this trait is not [dyn-compatible](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility), so you're left with generics.

I like generics a lot, but I don't think `Loss` is a particulary good fit as `<L: Loss>` would permeate all code that in any way relates to the usage of the loss function. I think it is better if we stick to the same approach as the other loss functions in this file.

donjuanplatinum: > Traits are at their most useful when you use them to describe generic behaviour, and in rust you typically either use generics ala `fn calculate<L: Loss>(t1: Tensor, t2: Tensor, loss: L)` or as a dyn trait `fn calculate(t1: Tensor, t2: Tensor, loss: dyn Loss)`. However this trait is not [dyn-compatible](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility), so you're left with generics.
> 
> I like generics a lot, but I don't think `Loss` is a particulary good fit as `<L: Loss>` would permeate all code that in any way relates to the usage of the loss function. I think it is better if we stick to the same approach as the other loss functions in this file.

Thank you for the clarification! i have change the Loss Trait of huber into the same approach as the other loss functions in this file

ivarflakstad (approved):
👌

PR #3250, opened 2025-12-18
feat: implementation varlen-flash-attention on cpu

michaelfeil:
I would like native support for cuda-free fallback for variable length flash attention, so that the modeling code can be written for Metal, CPU and Cuda without requiring the kernel. 

Actual Motivation: I want to get rid of the non-varlen formulation of the models. https://github.com/huggingface/text-embeddings-inference/tree/main/backends/candle/src/models Their main issue is that you need a varlen flash-attention primiteve. For that i need metal and cpu support.

The implementation is fully tested against what i have to use on cuda https://github.com/Dao-AILab/flash-attention/blob/ac9b5f107f2f19cd0ca6e01548d20d072a46335c/csrc/flash_attn/flash_api.cpp#L515

For a unfused impl, n o fused attention is used, "flash" fused, but allows for padding-free variable length attention on Metal and Cuda. 
For cpu, I implemented a similar fashion a "flash-version", that one is however more or less auto-generated and peer reviewed by me. 

I previously wrote the candle flash attn 3 bindings: https://github.com/michaelfeil/candle-flash-attn-v3.

I wrote up some benches and compare my flash-attn-cpu against the varlen formulation.

F16 Performance (milliseconds)
| Shape | Unfused | Flash | Varlen Flash | Flash vs Unfused | Varlen vs Flash | Varlen vs Unfused |
|-------|---------|-------|---------------|------------------|------------------|-------------------|
| b4_hq8_hk8_d64_noncausal | 35.11ms | 16.15ms | 1.63ms | 2.17x faster | 9.91x faster | 21.54x faster |
| b4_hq8_hk8_d64_causal | 35.11ms* | ~16.15ms* | ~1.63ms* | 2.17x faster | 9.91x faster | 21.54x faster |
| b3_hq12_hk4_d64_gqa_noncausal | 39.54ms | 18.40ms | 1.93ms | 2.15x faster | 9.53x faster | 20.49x faster |
| b2_hq8_hk8_d64_alibi_causal | ~35.11ms* | ~16.15ms* | ~1.63ms* | 2.17x faster | 9.91x faster | 21.54x faster |
F32 Performance (milliseconds)
| Shape | Unfused | Flash | Varlen Flash | Flash vs Unfused | Varlen vs Flash | Varlen vs Unfused |
|-------|---------|-------|---------------|------------------|------------------|-------------------|
[...]

PR #3254, opened 2025-12-20
Optimization for CPU Causal Flash Attention (integrated into Qwen3)

DrJesseGlass:
This PR implements a complete CPU flash attention subsystem for candle, extending the original causal attention work with varlen support, interleaved KV cache, and fused zero-allocation decode kernels. Integrated into Qwen3 (full and quantized) and SmolLM3 (quantized with interlaced GGUF weights).
                                                                                                                                                                              
**Quantized Qwen3-0.6B Q4_K_M (MacBook Air, Apple Silicon,  22 prompt / 499 generated)):**
  | | Standard | CPU Flash | Delta |                                                                                                                                                                        
  |---|---|---|---|
  | Token/s | 26.42 | 36.89 | **+39.6%** |                                                                                                                                                                  
  | Steady-state RSS | ~800 MB | ~700 MB | **-12.5%**                                                                                                                                                   
                                                                                                                                                                                                            
  **Quantized SmolLM3-3B Q4_K_M (MacBook Air, Apple Silicon, 81 prompt / 499 generated):**                                                                                                                                                                                                            
  | | Standard | CPU Flash | Delta |                                                                                                                                                                      
  |---|---|---|---|                                                                                                                                                                                         
  | Token/s | 12.23 | 15.30 | **+25.1%** |                                                                                                                                                                  
  | Steady-state RSS | ~3450 MB | ~2020 MB | **-41.4%** |   
                                                                                                                                                                                                             
What changed                                                                                                                                                                                              
                                                                                                                                                                                                            
Attention module (candle-nn/src/attention/):                                                                                                                                                              
  - Shimmed old cpu_flash_attention (two functions changed to pub(crate)) and moved to modular attention/cpu_flash/ structure                                                                             
  - AttnMask enum with None, Causal { kv_offset }, Mask(tensor) for clean dispatch                                                                                                                          
  - B=1 kernels: standard.rs (explicit mask), causal.rs (loop-bound, ~50% less work)
  - B>1: auto-packs to varlen format — integrated from https://github.com/huggingface/candle/pull/3250                                                                                                      
  - Fused f32 decode kernels with prefetch hints (NEON/AVX), pre-allocated thread-local accumulators                                                                                                        
  - Interleaved KV decode kernel: K and V adjacent per position, single prefetch per KV step                                                                                                                
                                                                                                                                                                                                            
KV cache (candle-nn/src/kv_cache.rs):                                                                                                                                                                     
  - InterleavedKvCache: tensor-based interleaved cache for prefill                                                                                                                                          
  - RawInterleavedKvCache: pre-allocated Vec<f32> buffer, zero allocation on append                                                                                                                         
  - .detach() on ConcatKvCache and InterleavedKvCache to prevent BackpropOp memory leak — addresses the same issue as https://github.com/huggingface/candle/pull/3419                                     
  - In-place RoPE (rope_i_inplace) for fused decode without tensor ops                                                                                                                                      
                                                                                                                                                                                                            
Model integrations:                                                                                                                                                                                       
  - Qwen3 (qwen3.rs): CPU/GPU standard/flash dispatch, --use-flash-attn flag                                                                                                                                
  - Quantized Qwen3 (quantized_qwen3.rs): fused decode with raw cache, interleaved KV, conditional cache allocation                                                                                         
  - Quantized SmolLM3 (quantized_smollm3.rs): fused decode with interlaced GGUF weight support (skips reconstruction, uses rope_i for native element ordering)                                            
  - All flash paths guard B>1 — fall back to standard matmul for multi-batch


DrJesseGlass:
Thinking to also include Rotary embeddings and other attention related shared processes to this component.

candle-nn/src/attention/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// `kv_offset`: Number of prior KV positions when using KV cache.
+    /// - Prefill: `kv_offset = 0`
+    /// - Decode: `kv_offset = cached_kv_len`
+    Causal { kv_offset: usize },
+
+    /// Custom mask tensor for arbitrary patterns.
+    ///
+    /// Shape: `(B, Q_LEN, KV_LEN)` or broadcastable.
+    /// Values: `0.0` to attend, `NEG_INFINITY` to mask.
+    Mask(&'a Tensor),
```
ivarflakstad: Tensor data is behind an `Arc`, so cloning is very cheap. May not be worth it to track the lifetime here

ivarflakstad (commented):
I'm waiting on https://github.com/huggingface/candle/pull/3250 but if it takes too long we'll incorporate these changes first :)

DrJesseGlass:
Tensor is Arc-backed, so cloning is cheap. Per @ivarflakstad's suggestion, AttnMask now owns the tensor, simplifying the API and removing lifetime annotations.

DrJesseGlass:
Also latest commit fixes GPU flash attention dropping KV cache during single-token decode. Now causal = l > 1.
and guards against and invalid config: use_flash_attn=true + --features flash-attn + CPU now errors explicitly.

Although Standard decode and full prefill work correctly, there is future work required. Chunked prefill (offset > 0 and l > 1) remains unsupported with GPU flash attention—the kernel can't offset the causal mask.

DrJesseGlass:
@ivarflakstad the original scope was dispatch logic only (CPU/GPU standard/flash routing for Qwen3). While waiting on the varlen PR https://github.com/huggingface/candle/pull/3250, I integrated it along with the tests and wired it into the dispatch and examples.                                                                        
   
Once varlen was in, it was actually outperforming the B=1 causal kernel on benchmarks, which raised the question of whether the specialized B=1 path was worth maintaining. So I kept pushing on it — removing nested parallelism, eliminating per-token allocations, adding the interleaved KV cache and fused decode path — until the B=1 kernel justified itself with the +40% speed result on quantized models.                                                                                                                                                                                                   
                                                                                                                                                                                                     
The interleaved KV cache work also surfaced the same BackpropOp memory leak that https://github.com/huggingface/candle/pull/3419 addresses, so the .detach() fix is included here for both ConcatKvCache and the new InterleavedKvCache.
                                                                                                                                                                                                            
The B=1 causal kernel is also the foundation for future interlaced-weight attention (reading K/V elements in native GGUF order without reconstruction), which is validated and working for SmolLM3 in this PR. The fused decode path with RawInterleavedKvCache eliminates all tensor operations between QKV projection and o_proj during token generation.

  Each step compounded:                                                                                                                                                                                     
  - Causal B=1 kernel (+6%) — eliminated mask materialization, loop-bound causal masking
  - Interleaved KV cache (+15% on top) — K and V adjacent per position, single prefetch, halved cache misses                                                                                                
  - Fused decode (+15% on top) — raw Vec<f32> cache with zero tensor allocations per token, in-place slice extraction from QMatMul output, direct kernel call   

**Speed (quantized Qwen3-0.6B Q4_K_M, MacBook Air Apple Silicon):**                                                                                                                                       
                                                                                                                                                                                                          
  | Milestone | Token/s | vs Standard |                                                                                                                                                                     
  |---|---|---|                                                                                                                                                                                             
  | Standard matmul (no flash) | 26.42 | — |                                                                                                                                                                
  | First CPU flash (causal B=1) | 28.02 | +6.1% |                                                                                                                                                        
  | Interleaved KV cache | 31.98 | +21.0% |                                                                                                                                                                 
  | Fused decode (zero alloc) | **36.89** | **+39.6%** |
                                                                                                                                                                                                            
  **Memory (RSS polling during 500-token generation):**                                                                                                                                                   
                                                                                                                                                                                                            
  | | Standard | CPU Flash | Delta |                                                                                                                                                                        
  |---|---|---|---|
  | Peak RSS | ~1558 MB | ~735 MB | **-52.8%** |                                                                                                                                                            
  | Steady-state | ~800 MB | ~700 MB | **-12.5%** |

ivarflakstad:
This is great 🙌 .

Sharing performance on my M4

**Speed (quantized Qwen3-0.6B Q4_K_M, MacBook M4 Max Apple Silicon):**                                                                                                                                       
                                                                                                                                                                                                          
  | Milestone | Token/s | vs Standard |                                                                                                                                                                     
  |---|---|---|                                                                                                                                                                                             
  | Standard matmul (no flash) | 31.77 | — |                                                                                                                                                                
  | PR CPU flash | 36.50 | +13% |
  
This is very impressive, even if it is not the 40% increase that you observed. Also output is identical 🎉 

I used cargo instruments to check allocations:
`cargo instruments --template="Allocations" --example quantized-qwen3 --release -- --prompt "Write a poem about butterflies." --which 0.6b`. (with `--use-flash-attn` on this branch ofc)

According to that trace the new approach has a _higher_ memory footprint than before. At last token it has 1.95GiB heap allocated while the old matmul approach has 1.13GiB.

It could be that using cargo instruments is a bad idea for some reason unknown to me, or perhaps that I'm calling the example with incorrect arguments. Just wanted to let you know what I found :)


Also wanted to mention that I'm considering retiring the `--use-flash-attn` argument. It should instead simply be used if available, and fall back to non-flash attn if needed. But that is for later.

Finally fyi the plan now is to first make a release of what has been merged recently, and then I want to merge this asap 👍

DrJesseGlass:
That's great. I'm excited for the integration and for your improvements. I'm happy to have independent verification of speed on different hardware. One note: the prompt "Write a poem about butterflies" likely generates <100 tokens before EOS. Both the speed and memory advantages scale with generation length because standard attention's per-token cost grows with KV cache length (mask materialization). Flash stays flat. At 500 tokens we see +40% vs +13% at ~50 tokens.

For RAM, I found RSS polling during generation gives a better picture of actual memory pressure:                                                                                                                                                                                                        
`while true; do ps -o rss=,etime= -p $(pgrep -n quantized-qwen) 2>/dev/null && sleep 1; done > bench_results/ram_log.txt`
This captures the working set over time, separating model loading from generation. On my MacBook Air with 500-token generation, flash steady-state RSS was ~700 MB vs ~800 MB for standard.               

Our discrepancy in RAM measurements is largely explained by instruments measuring preallocated versus RSS polling only see what is actually written. We can reduce the size of preallocated positions (something like dropping to 256 from 4096 for initial size) and flash should prove more efficient as we approach 256 and then again near 512 because allocation doubles and I'm not certain as the in between. Currently we preallocate 4096 positions: 4096 positions × 4 KV heads × 128 floats (2×D) × 4 bytes = 8 MB per layer × 28 layers = 224 MB total pre-allocated. At 50 generated tokens, most of that is untouched. Instruments sees 224 MB allocated, RSS polling only sees the pages actually written (~2-3 MB).

ivarflakstad:
> That's great. I'm excited for the integration and for your improvements. I'm happy to have independent verification of speed on different hardware. One note: the prompt "Write a poem about butterflies" likely generates <100 tokens before EOS. 

Try running the prompt yourself :) 
Not only would you know exactly how many tokens it generates, we would also know if you get the same result as me. 
iirc it generated ~400-450 tokens.


> For RAM, I found RSS polling during generation gives a better picture of actual memory pressure:                                                                                                                                                                                                        
> 
> `while true; do ps -o rss=,etime= -p $(pgrep -n quantized-qwen) 2>/dev/null && sleep 1; done > bench_results/ram_log.txt`

I can try this approach and see if it makes a difference :)

DrJesseGlass:
For the poem prompt on my MacBook Air (Apple Silicon), standard attention produced:
 - 14 prompt tokens processed: 53.50 token/s
 - 640 tokens generated: 23.79 token/s                                                                                                                                                                   
   
While the flash implementation produced:                                                                                                                                                                  
- 14 prompt tokens processed: 59.69 token/s
- 557 tokens generated: 38.59 token/s

Both with temperature zero. Although different responses (and token counts) — for other prompts I was getting closer matches. It is within reason that the two methods produce similar but different outputs given their different kernels (online softmax vs materialized attention matrix).
                                                                                                                                                                                                            
Capping at 500 tokens for a controlled comparison:                                                                                                                                                        
   
### Quantized Qwen3-0.6B Q4_K_M                                                                                                                                                                           
                                                                                                                                                                                                          
**Speed:**                                                                                                                                                                                                
                                                                                                                                                                                                          
  | | Standard | CPU Flash | Delta |                                                                                                                                                                        
  |---|---|---|---|                                                                                                                                                                                       
  | Prompt (14 tokens) | 58.16 t/s | 51.31 t/s | -11.8% |
  | Generation (499 tokens) | 26.93 t/s | 37.30 t/s | **+38.5%** |                                                                                                                                          
                                                                                                                                                                                                            
**Memory (RSS polling during generation):**                                                                                                                                                               
                                                                                                                                                                                                            
  | | Standard | CPU Flash | Delta |                                                                                                                                                                      
  |---|---|---|---|
  | Steady-state RSS | ~810 MB | ~775 MB | **-4.3%** |
  | Peak RSS | ~1158 MB | ~784 MB | **-32.3%** |                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                         
### How speed scales with generation length                                                                                                                                                               
                                                                                                                                                                                                            
  | Tokens generated | Standard t/s | CPU Flash t/s | Delta |                                                                                                                                               
  |---|---|---|---|
  | ~200 | 35.64 | 42.91 | +20.4% |                                                                                                                                                                         
  | ~500 | 26.93 | 37.30 | +38.5% |

DrJesseGlass:
Any updates or anything needed from me? I'm thinking to update this an remove the flag that you suggested.

DrJesseGlass:
I merged main and removed the use-flash-atten feature for CPUs -- note that as a result standard attention is never called for CPU when using the streamlined attention (except for quantized and batch_size > 1 because there is no quantized varlen CPU flash implementation yet). Cheers and GL.

candle-nn/src/attention/cpu_flash/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+use candle::{DType, Result, Tensor, WithDType};
+use std::iter::Sum;
+/// Debug-only: CANDLE_FLASH_FORCE_VARLEN=1 forces varlen path for B=1 (A/B benchmarking).
+#[cfg(debug_assertions)]
+use std::sync::LazyLock;
+#[cfg(debug_assertions)]
+static FORCE_VARLEN: LazyLock<bool> =
+    LazyLock::new(|| std::env::var("CANDLE_FLASH_FORCE_VARLEN").is_ok_and(|v| v == "1"));
+#[cfg(not(debug_assertions))]
+static FORCE_VARLEN: bool = false;
```
ivarflakstad: I assume you then use `--profile=release-with-debug`?
In just debug mode the generated binary would be so different that I suspect comparing paths would be futile.

I don't see any difference in output nor performance with the flag on or off though.

candle-nn/src/kv_cache.rs
```diff
@@ -723,26 +723,20 @@ impl ConcatKvCache {
     /// Tuple of `(full_k, full_v)` containing all cached keys and values,
     /// including the newly appended data.
     pub fn append(&mut self, k: &Tensor, v: &Tensor) -> Result<(Tensor, Tensor)> {
-        // Ensure inputs are contiguous for optimal concatenation performance
-        let k = k.contiguous()?;
-        let v = v.contiguous()?;
-        // Update K cache using concatenation
+        // Detach inputs to break BackpropOp chain — KV caches are inference-only,
+        // no gradients flow through them. Without detach, Tensor::cat stores
+        // references to prior cache buffers in the op graph, causing O(N²) memory.
```
ivarflakstad: I'd prefer if the documentation was written by a human (I'm assuming you don't casually throw `U+00B2`, `U+002D`, and `U+2192` around).
Not because AI docs are inherently bad, in fact they can be quite good at documentation.
However these tools aren't great at brevity. And as we can see some lines below, the exact same concept is repeated.
In the general case (I know you're a good dev so doesn't apply) it is also a great early sign to me as a reviewer that what I am not wasting my time reviewing full slop.

Imo it's enough to simply state `// Detach inputs to break BackpropOp chain — KV caches are inference-only`. This is a comment for a fellow developer, as opposed to `///` , and so I expect that they either know exactly what this is referring to or gather from context that `.detach()` has some performance benefit.

ivarflakstad (commented):
Thank you for your patience!
I'll have another look tomorrow 👍

candle-nn/src/kv_cache.rs
DrJesseGlass: Understood. Noted for future PRs. I will go over all the notes and documentation and simplify / write manually.

candle-nn/src/attention/cpu_flash/mod.rs
DrJesseGlass: I used CANDLE_FLASH_FORCE_VARLEN during development to A/B the single-batch and varlen paths against each other -- running cargo run --release --profile-time with debug-assertions = true in a custom profile so the flag was live. Once I'd picked the B>1 threshold it stopped being useful.

candle-nn/src/attention/cpu_flash/causal.rs
```diff
@@ -0,0 +1,867 @@
...
+        };
+
+        result.and_then(|t| t.unsqueeze(0))
+    }
+}
+
+// f32 dot product
+
+#[inline(always)]
+fn dot_f32(a: &[f32], b: &[f32]) -> f32 {
```
ivarflakstad, line 171: Compilers are pretty amazing. This simple impl is faster than the target specific `CurrentCpu::vec_dot_f32` we have.

I did manage to improve the `CurrentCpu` to where it was slightly faster, but let's look into it later.

Here's updated version.
Made `const STEP` available and used it instead of using `fn n()`.
Changed iter.step_by to while loop.
Also added `#[inline(always)]` on all the `CurrentCpu` fns called here.

```rust
#[inline(always)]
pub(crate) unsafe fn vec_dot_f32(a_row: *const f32, b_row: *const f32, c: *mut f32, k: usize) {
    let mut sum = CurrentCpu::zero_array();
    let mut ax = CurrentCpu::zero_array();
    let mut ay = CurrentCpu::zero_array();

    let mut i = 0;
    while i + CurrentCpu::STEP < k {
        for j in 0..CurrentCpu::ARR {
            ax[j] = CurrentCpu::load(a_row.add(i + j * CurrentCpu::EPR));
            ay[j] = CurrentCpu::load(b_row.add(i + j * CurrentCpu::EPR));

            sum[j] = CurrentCpu::vec_fma(sum[j], ax[j], ay[j]);
        }
        i += CurrentCpu::STEP;
    }

    CurrentCpu::vec_reduce(sum, c);

    // leftovers
    while i < k {
        *c += *a_row.add(i) * (*b_row.add(i));
        i += 1;
    }
}
```

candle-nn/src/attention/cpu_flash/causal.rs
```diff
@@ -0,0 +1,867 @@
...
+
+    let q_seq_stride = h_q * d;
+    let k_seq_stride = h_kv * d;
+    let v_seq_stride = h_v * d;
+
+    let mut out = vec![0f32; h_q * s_q * d];
+
+    FLASH_ATTN_POOL.install(|| {
+        out.par_chunks_mut(d)
+            .with_min_len(64)
```
ivarflakstad, line 491: Should look into this for the other variants as well

ivarflakstad (approved):
Great! Let's get this merged 🎉 

I'll add @michaelfeil as a co-author of the PR for the varlen work 👍

ivarflakstad:
I'd love for us to converge on a coherent design of the different cpu flash attn impls, without sacrificing performance ofc. Because with this merged we have 3 slightly different designs.
For example we have `dot_f32`, `vec_dot<T: ...>`, and candle's existing `f32::vec_dot` from `VecOps`.

DrJesseGlass:
I'll work on streamlining it to a single design next. Was trying for a small PR original mostly of dispatch logic but it grew due to varlen implementation conflict and then need to prove that non-varlen design was worth while.

DrJesseGlass:
@ivarflakstad I standardized the dot products behind a DotF32 trait in candle-nn/src/attention/cpu_flash/dot.rs. The f32 impl uses the compiler-vectorized body we both found fastest; the f16/bf16 impls accumulate in f32 (correctness fix over the old vec_dot::<T> which lost precision in T). I also factored the online softmax step into a shared helper at candle-nn/src/attention/cpu_flash/online_softmax.rs for cleaner kernel code #3513. 

Following on from that, I opened #3516 to bring the same compiler-vectorized body into candle-core::cpu::vec_dot_f32 itself, replacing the cfg-gated CurrentCpu intrinsics path. That makes conv1d / conv2d / convtranspose2d on the f32 CPU path use the same simpler implementation. The cfg gates and the unsafe pointer-arithmetic codepath go away; CurrentCpu and the f16/bf16/vec_sum paths are untouched. Conv2d microbench on M1 NEON shows it at least matching the intrinsics version (~4 percent faster on median, within noise) so the clear win there is maintenance. The design question you raised about whether a tweaked intrinsics version could win is still open and worth revisiting on AVX.

PR #3260, opened 2025-12-23
Adds onnx ops to support debertav3/piiranha

skeet70:
These are the changes I needed to run the ONNX version of Piiranha, which is debertaV3 based. I've since noticed that the debertaV2 model already in Candle also supports debertaV3, so I'm likely going to switch to using that, but thought the changes may be useful. Let me know if a subset is desired, I can cut it down to a smaller PR.

Adds operations:
- Add
- Or
- Tile
- LessOrEqual
- GreaterOrEqual

Also adds and uses helpers `to_scalar_flexible` and `to_vec0_flexible` that both allow for more broad input definitions that fit what some onnx models export and are still scalar/vec0 but not "true" versions of them because the model didn't squeeze them.


ivarflakstad:
If you fix formatting I’ll merge ☺️

Issue #3257, opened 2025-12-21
Add Mamba2 model support


Anri-Lombard:
Candle has Mamba1 but not Mamba2 from ["Transformers are SSMs"](https://arxiv.org/abs/2405.21060). Mamba2 is 2-8x faster via structured state space duality.

PR #3264, opened 2025-12-25
Mamba2 implementation

Anri-Lombard:
Addressing #3257

Adding the mamba2 architecture.

### DIFF
--- a/candle-examples/examples/mamba2/README.md
+++ b/candle-examples/examples/mamba2/README.md
@@ -10,3 +10,47 @@ Duality (SSD) framework which unifies structured SSMs and attention variants.
 ```bash
 cargo run --example mamba2 --release -- --prompt "Mamba is the"
 ```
+
+## Supported models
+
+| Model | HuggingFace ID |
+|-------|----------------|
+| Mamba2-130m | `AntonV/mamba2-130m-hf` |
+| Mamba2-370m | `AntonV/mamba2-370m-hf` |
+| Mamba2-780m | `AntonV/mamba2-780m-hf` |
+| Mamba2-1.3b | `AntonV/mamba2-1.3b-hf` |
+| Mamba2-2.7b | `AntonV/mamba2-2.7b-hf` |
+
+## Verification
+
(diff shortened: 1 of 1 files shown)


ivarflakstad (commented):
This is great! 👌 
Everything here looks sane to me, but I was wondering if you had any expected input -> output that we could use to prove correctness?

I usually refer to transformers test code, but they use `Mamba-Codestral` in the integration [test](https://github.com/huggingface/transformers/blob/a7f29523361b2cc12e51c1f5133d95f122f6f45c/tests/models/mamba2/test_modeling_mamba2.py#L350), which isn't in this PR.

ivarflakstad:
Oh btw if you're struggling with getting exactly equal numbers with a reference impl it might be because of small numerical inaccuracies in our normalization kernels. I updated the metal kernels today, so if you happen to be testing on metal then the numbers should be closer now.
Updating the branch with main so you're up to date 🫡

Anri-Lombard:
Added a verification section to the README with expected outputs for mamba2-130m and mamba2-370m. Both match the PyTorch transformers `Mamba2ForCausalLM` reference implementation exactly.

Happy to add more if helpful - would unit tests for the core algorithms (segsum, SSM step) be useful? Or e2e tests? Or do you think documenting in the README is enough?

ivarflakstad:
> Added a verification section to the README with expected outputs for mamba2-130m and mamba2-370m. Both match the PyTorch transformers `Mamba2ForCausalLM` reference implementation exactly.
> 
> Happy to add more if helpful - would unit tests for the core algorithms (segsum, SSM step) be useful? Or e2e tests? Or do you think documenting in the README is enough?

I think the verification against transformers is sufficient 👌
Text is correct (not factually ofc hehe)

Not opposed to adding unit tests to `candle-transformers` later. We don't have enough of them in general.


I think a git op might've gone awry as you have a metal bf16 conv  fix in the PR now.

Anri-Lombard:
@ivarflakstad I've rebased to clean up the git history - removed the merge commit and the stray bf16 conv fix that snuck in.

Also created #3286 to track adding unit tests to candle-transformers for later.

PR #3271, opened 2025-12-28
Sort on cuda fails when tensor size exceeds 1024

slckl:
Stumbled upon an unexpected limitation: `Tensor::sort_last_dim` fails on cuda, if the tensor in question is larger than 1024 elements.

This PR:
- adds a new `asort_big` test, which simply checks sorting on a 2000 element large tensor, it fails on cuda, if you run it on `main`, but will work with `cpu`, no idea about `metal`
- implements fix for cuda by 
  1. capping block_dim to `<= 1024`
  2. modifies the kernel so each thread can handle multiple elements, when `ncols_pad > block_dim`

Opus made the kernel change, it looks ok and all tests pass.  Not entirely sure about performance implications, if any.

### DIFF
--- a/candle-core/tests/tensor_tests.rs
+++ b/candle-core/tests/tensor_tests.rs
@@ -221,6 +221,10 @@ fn asort(device: &Device) -> Result<()> {
 
 /// Test sorting a large tensor that exceeds 1024 elements.
 fn asort_big(device: &Device) -> Result<()> {
+    // Skip on metal for now
+    if device.is_metal() {
+        return Ok(());
+    }
     const SIZE: usize = 2000;
     let data: Vec<f32> = (0..SIZE).map(|x| (SIZE - x) as f32).collect();
     let tensor = Tensor::new(data.as_slice(), device)?;


ivarflakstad:
At a glance this looks pretty good to me.
It's actually a known limitation, and @guoqingbao has a sort impl ([here](https://github.com/guoqingbao/attention.rs/blob/main/src/kernels/src/sort.cu)) which I believe solves the same problem and we were planning on incorporating.

slckl:
Either way sounds good!

I'm currently in the process of porting rfdetr to candle, dirty (but already working version is here): https://github.com/slckl/candle_xps/tree/master/rs/candle_rf_detr
It outputs correct results with this fix:
sample_bike pp

Without this, rfdetr at medium sizes and above runs into this limitation and we have to hack with putting this tensor to cpu then back to gpu, feels wasteful.

P.S. the attention repo seems super cool

ivarflakstad:
> no idea about metal

Can confirm does not work correctly on metal. I'll have a quick look.

guoqingbao:
I believe there are two limitations in the current sort kernel (which is sourced from ggml). Aside from the issue addressed by this PR, the other limitation is related to shared memory: the sort kernel requires all data to reside in shared memory, which constrains the maximum number of input tokens to (maximum shared memory size) / (token data type size). For example, this limits processing to approximately 48K tokens (or elements) on A100, while modern LLMs can support up to 256K context.

ivarflakstad:
Right, and the bitonic sort kernel runs iteratively on subsets of data to be sorted - correct?
I'm still happy to merge this PR, but we should disable the test on mac since the mlx kernel is not set up (in candle at least) to handle sorting large tensors.

ivarflakstad:
We'll switch to your kernel when you're ready @guoqingbao 👌 I might implement a similar kernel for metal actually. Seems easier to maintain than the multiple different sort kernels mlx has.
Merging since it is working on cpu and cuda.

guoqingbao:
> Right, and the bitonic sort kernel runs iteratively on subsets of data to be sorted - correct?

The sort kernel requires a merge pass to be used on limited shared memory. My custom kernel simply use global memory for sorting with thread sync and batch processing, it can in theory, sort any large content but comes with slight performance degradation especially for smaller input sizes (might be neglectable on modern hardware given their fast cache and HBM bandwidth). I mainly use it during prefill stage.

I think sort kernels are mainly used for sampling (the major bottleneck besides attention in LLM inference), I have designed a fused sampling kernel in attention.rs to improve this (approximately 25% overall inference speedup obtained in vllm.rs).

guoqingbao:
Refer to this: https://github.com/guoqingbao/attention.rs/pull/5

May submit a PR to candle if you like @ivarflakstad

PR #3274, opened 2025-12-30
fix: auto-detect GGUF metadata prefix for gemma family

### LINKED ISSUE #3215: cannot find gemma3.attention.head_count in metadata

clocksmith:
Closes #3215  where embeddinggemma-300m-GGUF fails with "cannot find gemma3.attention.head_count in metadata".

As noted by @ivarflakstad , Gemma models use different metadata prefixes: gemma3.* for standard models, gemma-embedding.* for embedding models (ref: unsloth/embeddinggemma-300m-GGUF).

**Solution**: Probe metadata keys to detect actual prefix before lookups, rather than hardcoding gemma3.*. This also future proofs for new variants (gemma4, gemma-vision, etc.)

### DIFF
--- a/candle-transformers/src/models/quantized_gemma3.rs
+++ b/candle-transformers/src/models/quantized_gemma3.rs
@@ -263,34 +263,48 @@ impl ModelWeights {
         reader: &mut R,
         device: &Device,
     ) -> Result<Self> {
-        let md_get = |s: &str| match ct.metadata.get(s) {
-            None => candle::bail!("cannot find {s} in metadata"),
-            Some(v) => Ok(v),
+        // Detect architecture prefix by probing which keys exist in metadata.
+        // This supports gemma3, gemma2, gemma, gemma-embedding, and future variants.
+        let prefix = ["gemma3", "gemma2", "gemma", "gemma-embedding"]
+            .iter()
+            .find(|p| {
+                ct.metadata
+                    .contains_key(&format!("{}.attention.head_count", p))
+            })
+            .copied()
+            .unwrap_or("gemma3");
+
+        let md_get = |s: &str| {
+            let key = format!("{prefix}.{s}");
+            match ct.metadata.get(&key) {
+                None => candle::bail!("cannot find {key} in metadata"),
+                Some(v) => Ok(v),
+            }
         };
 
-        let head_count = md_get("gemma3.attention.head_count")?.to_u32()? as usize;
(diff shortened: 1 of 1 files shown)


clocksmith:
Thanks! Agree on keeping things moving!

Nit: begins with gemma would not work for functiongemma

True it is not 100% automatic, but its now a 1 liner to add support for new models in gemma family! I can also look at the other parsers and see if there is a better pattern to follow.

 We can weight string matching, begins, contains, a filtered list, or any hybrid with fallbacks. What's your preference?

ivarflakstad:
Not sure I have a clear preference.
I like the explicitness of a filtered list, but it's also nice to not have to update it with every release.

ivarflakstad:
How about split at first `.`, check if first item contains `gemma`.
If multiple matches of this exist in the metadata fall back to filtered list?

PR #3279, opened 2026-01-04
Fix BF16 conv_transpose2d using wrong kernel on Metal

amritsingh183:
## Summary

This PR fixes a critical bug where `conv_transpose2d` with BF16 (bfloat16) tensors on Metal was incorrectly using the 1D convolution transpose kernel instead of the 2D kernel.

## The Bug

In `candle-metal-kernels/src/metal_src/conv.metal`, line 606:

```metal
// Before (BUG):
CONVT1D_OP(bfloat, float, conv_transpose2d_bf16)  // Uses 1D kernel!

// After (FIX):  
CONVT2D_OP(bfloat, float, conv_transpose2d_bf16)  // Uses correct 2D kernel
```

This was clearly a copy-paste error. Looking at the surrounding code:

```metal
CONVT2D_OP(float, float, conv_transpose2d_f32)   // F32: ✅ Correct - uses 2D
CONVT2D_OP(half, float, conv_transpose2d_f16)    // F16: ✅ Correct - uses 2D
#if defined(__HAVE_BFLOAT__)
CONVT1D_OP(bfloat, float, conv_transpose2d_bf16) // BF16: ❌ BUG - uses 1D!
#endif
```

## Impact

When running `conv_transpose2d` with BF16 tensors on Metal:
- The operation would run without error
- But produce **completely wrong output** because the 1D kernel processes data along only one spatial dimension
- This would cause subtle numerical errors that are hard to diagnose

## How I Found This

While debugging a SAM3 (Segment Anything Model 3) Rust port, I traced a "spatial collapse" issue where batched mask predictions all became identical. The root cause analysis led me to the `conv_transpose2d` operation in the mask decoder.

During investigation of the Metal kernel source, I spotted this obvious typo.

## Testing

Added a new test `conv_transpose2d_bf16` that:
1. Creates a BF16 tensor and kernel
2. Runs conv_transpose2d
3. Compares output shape and values against F32 reference (should be nearly identical)

## Checklist

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Added test to confirm the fix
- [x] All existing tests pass


candle-core/tests/conv_tests.rs
```diff
@@ -892,3 +892,79 @@ test_device!(
     conv2d_grad_gpu,
     conv2_grad_metal
 );
+
+/// Test that conv_transpose2d works correctly with BF16 tensors.
+/// This test was added to catch a bug where the Metal kernel for BF16 conv_transpose2d
+/// was incorrectly using the 1D convolution kernel (CONVT1D_OP) instead of 2D (CONVT2D_OP).
+/// The bug was in candle-metal-kernels/src/metal_src/conv.metal
+#[allow(dead_code)] // Only used when metal feature is enabled
+fn conv_transpose2d_bf16(dev: &Device) -> Result<()> {
```
ivarflakstad: I understand the sentiment but I'd rather not have tests that are _this_ specific when the cause of the bug is so apparent.

amritsingh183:
let me remove the test

amritsingh183:
i have removed the test.

PR #3285, opened 2026-01-07
feat: simplify metal reduce kernels and standardize on u32 indexing

drbh:
This PR improves the `metal_reduce_.*_strided` kernels

### Benchmark

```bash
cargo bench --features metal -p candle-core --bench bench_main -- "metal_reduce_.*_strided"
```
### Results
```
| Benchmark    | main                | new                      | Speedup      |
|--------------|---------------------|--------------------------|--------------|
| f32_strided  | 5.62ms (711 MiB/s)  | 2.3433 ms (1.6670 GiB/s) | 2.39x faster |
| f16_strided  | 3.98ms (503 MiB/s)  | 2.3631 ms (851.33 MiB/s) | 1.69x faster |
| bf16_strided | 3.95ms (505 MiB/s)  | 2.3545 ms (849.44 MiB/s) | 1.68x faster |
```


ivarflakstad:
Oh yeah big mr huggingface employee coming here showing the people how to vibe code kernel code with proper results, eh?

But seriously this is really good. We've previously talked about how indexing with smaller dtypes is surprisingly efficient, and I see you've also come up with a similar strided index improvement like I did [here](https://github.com/huggingface/candle/pull/3113).
I find this part really interesting (and I suspect is a substantial part of the performance):
```c++
if (is_pow2[k]) {
  r = idx & masks[k];
  idx = idx >> shifts[k];
}
```
and I think it should be unrolled statically like the `get_strided_index_1d` / `get_strided_idx<1>` variants.
(edit: at least if all are pow2)

I couldn't find the branch but at some point I had a kernel impl where I calculated the min required index dtype based on the problem size using function constant values. Maybe worth exploring?

candle-metal-kernels/src/kernels/reduce.rs
```diff
@@ -23,38 +83,37 @@ pub fn call_reduce_contiguous(
...
+            width: out_length,
+            height: 1,
+            depth: 1,
+        },
+        MTLSize {
+            width,
+            height: 1,
+            depth: 1,
+        },
+    );
```
ivarflakstad, line 57: This is functionally the exact same as before, just without named variables.
Let's keep to a minimal amount of changes in this PR.

candle-metal-kernels/src/kernels/reduce.rs
```diff
@@ -3,6 +3,66 @@ use crate::utils::{BufferOffset, EncoderProvider};
...
+            IndexType::U64 => "_u64",
+        }
+    }
+}
+
+struct Pow2Meta {
+    is_pow2: Vec<u8>,
+    masks: Vec<u32>,
+    shifts: Vec<u8>,
+}
```
ivarflakstad: Not required, just fyi you could slap a `#[repr(C)]` on this and create it's equal in metal and we can send in the entire guy instead of it's parts.

candle-metal-kernels/src/kernels/reduce.rs
```diff
@@ -3,6 +3,66 @@ use crate::utils::{BufferOffset, EncoderProvider};
...
+    fn compute(shape: &[usize]) -> Self {
+        let mut is_pow2 = Vec::with_capacity(shape.len());
+        let mut masks = Vec::with_capacity(shape.len());
+        let mut shifts = Vec::with_capacity(shape.len());
+
+        for &dim in shape {
+            let pow2 = dim.is_power_of_two() && dim > 1;
+            is_pow2.push(pow2 as u8);
+            if pow2 {
+                masks.push((dim - 1) as u32);
```
ivarflakstad: I don't see any reason this couldn't be `u16` or `u64`.
If we should store it as `u64` in this calculation and convert to the appropriate dtype on kernel launch we could cover the other cases as well.
This does conflict a bit with the `#[repr(C)]` comment above.

candle-metal-kernels/src/kernels/reduce.rs
```diff
@@ -3,6 +3,66 @@ use crate::utils::{BufferOffset, EncoderProvider};
...
+enum IndexType {
+    U16,
+    U32,
+    U64,
+}
+
+impl IndexType {
+    fn select(shape: &[usize], strides: &[usize]) -> Self {
+        let max_dim = shape.iter().copied().max().unwrap_or(0);
+        let max_stride = strides.iter().copied().max().unwrap_or(0);
```
ivarflakstad: Ok, this is interesting. I was certain you would select the index type based on the maximum _index_ value, aka `shape.iter().product()`.

But looking at the metal fn you're still calculating the actual index using `u64`, but the intermediaries are in the smaller index type. Which has me wondering if this part actually improves performance, or if it is all the "pow2" feature.

Also nit: The stride max will _always_ be larger than shape max, so that's the only one you have to calculate.

I have a weird idea that could allow us to use `u16` for all indexing (even if the elem count of the tensor is way larger). Fallback to `u32`, but probably wouldn't need `u64` (given that the max stride < 4 294 967 295 hehe).
Totally different strided indexing approach. `max_stride` remains as the limiting factor.
Instead of calculating the final index and then indexing we incr/decr the actual data pointers for each dim. I think it could work, but it sounds messy so not sure we want to go down that path 🤔 

Here's the simplest variant. It adjusts the ptr, returns nothing.
```c++
template<typename T, typename IndexT = ushort>
METAL_FUNC void apply_stride_ptr(
    device const T *ptr,
    IndexT idx,
    constant const IndexT &num_dims,
    constant const IndexT *dims,
    constant const IndexT *strides
) {
    for (uint d = 0; d < num_dims; d++) {
        uint dim_idx = num_dims - 1 - d;
        ptr += (idx % dims[dim_idx]) * strides[dim_idx];
        idx /= dims[dim_idx];
    }
}
```
Haven't tested. Probably doesn't with the `const` attribute so some adjustment needed.

A bit tedious if you want to use the same indexing for multiple data sources.

But it would be easy to use `pow2` for all `IndexT`, and simplifies the kernel suffix.

candle-metal-kernels/src/metal_src/reduce.metal
```diff
@@ -46,11 +46,55 @@ METAL_FUNC uint max_shared_mem(uint n) {
...
 }
 
-METAL_FUNC uint get_strided_index(
+METAL_FUNC uint get_strided_index_1d(
+    uint idx,
+    constant const uint *dims,
+    constant const uint *strides
+) {
+    return idx * strides[0];
+}
```
ivarflakstad: As mentioned in my PR where I do the same I'd prefer to wait with this until we have precompilation for the metal backend. I'll add it.
I also like my approach a tiny bit better than claude's hehe.

candle-metal-kernels/src/metal_src/reduce.metal
```diff
@@ -61,6 +105,21 @@ METAL_FUNC uint get_strided_index(
     return strided_i;
 }
 
+METAL_FUNC uint get_strided_index(
```
ivarflakstad: Looking closely this isn't actually used.
You either use `indexer_pow2` or `indexer_u64`.
Which btw could be just `indexer` using templating.

candle-metal-kernels/src/metal_src/reduce.metal
```diff
@@ -631,40 +690,276 @@ kernel void NAME(                                       \
...
+METAL_FUNC ulong get_strided_index_t(
+    uint idx,
+    constant IndexT &num_dims,
+    constant IndexT *dims,
+    constant IndexT *strides,
+    constant uchar *is_pow2,
+    constant uint *masks,
+    constant uchar *shifts
+) {
+    ulong strided_i = 0;
```
ivarflakstad: `ulong`, `unsigned long`, `uint64_t` and `size_t` is the same. I'm happy using `ulong`, just thinking about consistency with for example the function directly below this one.

candle-metal-kernels/src/kernels/reduce.rs
```diff
@@ -73,44 +132,90 @@ pub fn call_reduce_strided(
...
+            width: out_length,
+            height: 1,
+            depth: 1,
+        },
+        MTLSize {
+            width,
+            height: 1,
+            depth: 1,
+        },
+    );
```
ivarflakstad, line 115: Same here

ivarflakstad:
Just checked which kernels are actually being run in the benchmark: `fast_sum_f32_strided`, `fast_sum_f16_strided`, and `fast_sum_bf16_strided`. So we've only hit the `IndexType::U32 => ""` case and only profiled `uint` indexing.

ivarflakstad:
Good news and bad news.

Good news is that I managed to speed this up even more.

Bad news is that I've benchmarked the `pow2` concept in isolation and there is no measurable performance benefit.
I'm assuming it is because using 3 lookups (`is_pow2[k]`, `masks[k]`, `shifts[k]`) counteracts the benefit of mask and shift operations. I even removed the `is_pow2[k]` check and there was still no measurable difference.
It seems all the gains are from using `u32` and static unrolling.

Good news from the bad news - simplifies the improvements considerably!

ivarflakstad:
Just did some quick mafs.
If the tensor is so large that `u32` is not enough to index into it. That means that the underlying buffer has more than 4,294,967,295 elements in it. Even if we were to use `fp8` for this tensor it would still be (4294967295 * 8 / (1024^3) =) 32 GB of data in a single tensor. For metal kernels that's kinda silly.
In other words I'm removing the kernels that use `u64` to index. We'll use a different approach if the need arises.

drbh:
> Just did some quick mafs. If the tensor is so large that `u32` is not enough to index into it. That means that the underlying buffer has more than 4,294,967,295 elements in it. Even if we were to use `fp8` for this tensor it would still be (4294967295 * 8 / (1024^3) =) 32 GB of data in a single tensor. For metal kernels that's kinda silly. In other words I'm removing the kernels that use `u64` to index. We'll use a different approach if the need arises.

thanks for diving into this! and yea I kinda assumed we could get away with just a u32 but didnt want to make too big of an assumption on todays hardware. however that being said, I do think its reasonable to avoid the u64 for now - and if this become a blocker in the future we can revisit

drbh (commented):
would approve the refactor/updates from @ivarflakstad if this wasnt originally opened by me! 

looks great and agree the standardization on u32 is reasonable and realistic

PR #3293, opened 2026-01-09
chore(dep): bump cudarc to 0.18.2

staceymelville:
This is required to support cuda toolkit version 13.1.


EricLBuehler:
Looks like this also brings a grouped gemm api @ivarflakstad!

PR #3298, opened 2026-01-12
add RWKV v7 architecture

nkypy:
Adding the RWKV v7 architecture.


entn-at:
Thanks! I tested with [rwkv7-g1a-0.1b-20250728-ctx4096.pth](https://huggingface.co/BlinkDL/rwkv7-g1/blob/main/rwkv7-g1a-0.1b-20250728-ctx4096.pth) and can confirm that it's working.

PR #2704, opened 2025-01-07
cuda: add opt-in BF16 support for pre-Ampere GPUs

haricot:
tested and works with:
```
nvidia-smi   --query-gpu="compute_cap"  --format=csv
compute_cap
6.1
```
```
cargo run -F "cuda,cudnn" --example llama  --release -- --model-id meta-llama/Llama-3.2-1B-Instruct --temperature 0.1 --which v32-1b-instruct --seed 42 --dtype bf16 --prompt
```

related https://github.com/EricLBuehler/candle/pull/57

PR #3300, opened 2026-01-12
fix candle-kernels build for CC < 700

haricot:
I used cudarc driver API to automatically detect compute capabilities at build time, which seems more practical than relying on the CUDA_COMPUTE_CAP environment variable:

- Works out-of-the-box without user configuration
- Automatically detects multi-GPU setups             
- Falls back to CUDA_COMPUTE_CAP env var if driver init fails (e.g., in CI)

If you prefer a different approach (e.g., nvidia-smi or env var only), I'm happy to adjust.

Currently, the generator's compute_cap method depends on the merging of Narsil/bindgen_cuda#18. And if Narsil/bindgen_cuda#16 is merged, it would be possible to extend CUBIN generation to multiple architectures to accelerate startup and optimization.

### DIFF
--- a/candle-kernels/build.rs
+++ b/candle-kernels/build.rs
@@ -79,7 +79,7 @@ fn main() {
     } else {
         moe_builder = moe_builder.arg("-Xcompiler").arg("-fPIC");
     }
-  
+
     moe_builder.build_lib(out_dir.join("libmoe.a"));
     println!("cargo:rustc-link-search={}", out_dir.display());
     println!("cargo:rustc-link-lib=moe");


ivarflakstad (commented):
Thank you for this!
Looks good to me 👌 

@guoqingbao could you double check the build steps related to the moe kernels? ☺️

guoqingbao:
> Thank you for this! Looks good to me 👌
> 
> @guoqingbao could you double check the build steps related to the moe kernels? ☺️

This also looks good to me.

haricot:
related #3331

DrJesseGlass:
Just want to quickly point out that there are several issue. This won't solve resolves #3331 because BF16 WMMA has a stricter requirement: SM 80+ (#3349 resolves). But this does resolve FP16 WMMA that requires 70+. This and #3349 are complimentary. 

I have a branch which is a subset of this issue created several months back https://github.com/DrJesseGlass/candle/tree/oldgpu/no-changes for Pascal sm_61 but assumed we weren't so interested in backwards compatibility.

This seems to merely need a cargo fmt for this to merge. But wanted to make it apparent that I can readily provide the atomicAdd polyfill for half on Pascal (CC < 70).

haricot:
@DrJesseGlass Thank you for your feedback, I understand your need to Disable BF16 WMMA for pre-Ampere GPUs

My goal is to ensure backward compatibility via https://github.com/huggingface/candle/pull/2704, where I've added `ALLOW_LEGACY_BF16` and `ALLOW_LEGACY_FP8`, as well as [moe_hfma2](https://github.com/haricot/candle/blob/9808d4861d37f6956428ab0ca5a213e144cbf75b/candle-kernels/src/moe/moe_hfma2.cu#L249-L302) (WWMA fallback solution (Tests passed, but testing in real-world conditions is needed)) CC < 70-80. This should therefore work now.
<details>
<summary>The `atomicAdd` function for `__half` is already candle.</summary>

https://github.com/huggingface/candle/blob/3b39794c14b1de8ba4c2d10d49354d557a415547/candle-kernels/src/compatibility.cuh#L38-L59

</details>

I added Polyfill: atomicAdd for bfloat16 for cc < 800 [min/max man](https://github.com/haricot/candle/blob/9808d4861d37f6956428ab0ca5a213e144cbf75b/candle-kernels/src/compatibility.cuh#L18-L25) [atomicAdd bf16 min/max man](https://github.com/haricot/candle/blob/9808d4861d37f6956428ab0ca5a213e144cbf75b/candle-kernels/src/compatibility.cuh#L70C-L92)

The current PR aims for CC < 700 compatibility but primarily offers a way to prepare a step for automatically supporting heterogeneous multi-GPUs.

In the futur using kernel via slang to enable automatic merge operation and backward compatibility via [stensor ](https://github.com/dimforge/stensor) could be a solution, but with less manual optimization setup.

haricot:
Closing in favor of #2704 and   [cudaforge](https://github.com/guoqingbao/cudaforge). And for wmma  either part of #3349 or #2704 (reply hfma2).

PR #3312, opened 2026-01-18
CUDA Tensor::toDevice copies directly via memcopy_dtod

krampenschiesser:
This adds support for dtod copy of tensors.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -44,45 +44,6 @@ jobs:
           export RUSTFLAGS="-C target-feature=neon"
           cargo check --workspace 
 
-  check-cuda:
-    name: Check Cuda
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest]
-        cuda_version: [13.1.0]
-        # cuda_version: [12.9.1, 13.1.0] if we want multiple cuda versions we need to enable concurrency below
-    env:
-      CUDA_HOME: /usr/local/cuda
-      CUDA_LOG_FILE: stderr
-      CUDA_COMPUTE_CAP: 120
-    # concurrency:
-    #   group: ${{ matrix.os }}-cuda
-    steps:
-      - uses: actions/checkout@v6
-
-      - uses: actions/setup-python@v6
-        with:
-          python-version: "3.13"
-
-      - name: Remove cargo config (macOS ring crate fix)
-        if: runner.os == 'macOS'
-        run: rm -f .cargo/config.toml
-      
-      - name: install cuda
-        uses: Jimver/cuda-toolkit@v0.2.30
-        id: cuda-toolkit
-        with:
-          cuda: ${{matrix.cuda_version}}
-          method: 'network'  
-      
-      - run: echo "${{steps.cuda-toolkit.outputs.CUDA_PATH}}" >> "$GITHUB_PATH"
-
-      - uses: dtolnay/rust-toolchain@stable
-
-      - run: cargo check --features cuda
-
   test:
     name: Test Suite
     runs-on: ${{ matrix.os }}


EricLBuehler:
Will review!

krampenschiesser:
sry let me fix the build errors, the logic shouldn't change much though.
still getting used to working on feature gated repositories :(

EricLBuehler:
No worries!

krampenschiesser:
I had to remove the device comparison and cloning of the storage.
When I wrote the test for it I saw the cuda device ptr ids are different and looking into it I realized that a slice clone actually does a d-to-d on the same device.
So I removed the comparison as it's basically a noop with the same logic as the new d-to-d copy.

krampenschiesser:
Will wait for this cudarc PR(https://github.com/chelsea0x3b/cudarc/pull/520) to be released.

.github/workflows/rust-ci.yml
```diff
@@ -22,7 +23,64 @@ jobs:
...
+          export RUSTFLAGS="-C target-feature=avx2"
+          cargo check --workspace 
+
+      - name: Run with arm neon
+        if: matrix.os == 'ubuntu-24.04-arm'
+        run: |
+          export RUSTFLAGS="-C target-feature=neon"
+          cargo check --workspace 
+
+  check-cuda:
```
krampenschiesser: I added this since nothing worked locally but the check passed in ci (3 different cudarc versions)

.github/workflows/rust-ci.yml
```diff
@@ -22,7 +23,64 @@ jobs:
...
+        run: cargo check --workspace --features metal
+
+      - name: Run normal cpu
+        if: matrix.os == 'ubuntu-latest' || matrix.os == 'windows-latest'
+        run: cargo check --workspace
+
+      - name: Run with avx2
+        if: matrix.os == 'ubuntu-24.04'
+        run: |
+          export RUSTFLAGS="-C target-feature=avx2"
```
krampenschiesser, line 38: might be overkill, but i think worthwhile to check avx, neon, and metal

Cargo.toml
```diff
@@ -69,13 +69,18 @@ half = { version = "2.5.0", features = [
     "use-intrinsics",
     "rand_distr",
 ] }
-float8 = { version = "0.5.0", features = ["num-traits", "rand_distr"] }
+float8 = { version = "0.5.1", git = "https://github.com/krampenschiesser/float8", rev = "16e49a9", features = [
```
krampenschiesser: i had to patch alot of 3rd party libs to use the newest cudarc version.
Would love to get rid of these github refs and use crates.io.
Let's see when they are merged and released

EricLBuehler:
Hi @krampenschiesser! Reviewing

Cargo.toml
EricLBuehler: I published the new `float8` as `v0.6.0` (as cudarc bumped its minor version). We can update `float8` to be from `crates.io` again.

.github/workflows/rust-ci.yml
EricLBuehler: Sounds good, especially to catch cpu quantized builds. 👍

.github/workflows/rust-ci.yml
EricLBuehler: 👍 Dont need a cuda machine for this

Cargo.toml
```diff
@@ -96,9 +101,12 @@ tokenizers = { version = "0.22.0", default-features = false }
 tracing = "0.1.37"
 tracing-chrome = "0.7.1"
 tracing-subscriber = "0.3.7"
-ug = "0.5.0"
-ug-cuda = "0.5.0"
-ug-metal = "0.5.0"
+# ug = "0.5.0"
+# ug-cuda = "0.5.0"
+# ug-metal = "0.5.0"
+ug = { git = "https://github.com/krampenschiesser/ug.git", version = "0.5.1", rev = "1558a91" }
```
EricLBuehler: I see that your fork just bumps `ug` to 0.19.0 for cudarc 👀

EricLBuehler: Fixed

PR #3317, opened 2026-01-19
Fix flaky GELU test and clippy warnings in candle-onnx

Douglas-Nyberg:
## Changes

- Fix `test_gelu_operation` in `candle-onnx` to use approximate float comparison (was failing due to precision differences).
- Fix clippy warnings in `candle-onnx/src/eval.rs`.
- Add clippy allow attributes for pre-existing test issues (excessive float precision) and generated protobuf code.

## Why

While running tests locally, I found that `test_gelu_operation` was flaky due to using exact float equality. Additionally, `cargo clippy` revealed several issues in `eval.rs` that needed fixing.

(Note: I originally attempted to add `candle-onnx` to the workspace/CI, but reverted that infrastructure change per maintainer feedback to focus solely on these code fixes.)

## Testing

All 64 `candle-onnx` tests pass locally. `cargo clippy` passes.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -50,15 +50,6 @@ jobs:
       - name: Install lld (Linux only)
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y lld
-      - name: Install protobuf (Linux only)
-        if: runner.os == 'Linux'
-        run: sudo apt-get install -y protobuf-compiler
-      - name: Install protobuf (macOS)
-        if: runner.os == 'macOS'
-        run: brew install protobuf
-      - name: Install protobuf (Windows)
-        if: runner.os == 'Windows'
-        run: choco install protoc -y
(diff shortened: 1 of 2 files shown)


ivarflakstad:
Thanks!

I'm happy to merge parts of this PR, but candle-onnx is actually purposefully excluded from workspace / CI.
We'll be making some changes to candle-core (soon ™️ ) which should make it much easier to maintain candle-onnx (and candle-pyo3). At which point we could definitely add it to workspace and CI.

Douglas-Nyberg:
Thanks for the feedback!

I've reverted the changes to Cargo.toml and the CI workflow as requested.
This PR now only contains the code-level fixes:

Fix for the flaky GELU test: Switched to approximate comparison to handle floating point differences.
Clippy fixes: Fixed bugs in candle-onnx/src/eval.rs and added an allow attribute for the generated protobuf code.

Let me know if you would like to see something else.

ivarflakstad (approved):
Lgtm! Thanks :)

PR #3324, opened 2026-01-23
Fix: main failing Rust 1.93 clippy CI

DrJesseGlass:
Rust 1.93 added `clippy::unnecessary_unwrap` which flags `is_some()` followed by `unwrap()`. 

Replaced with `if let Some(...)` patterns in:
- deepseek2.rs
- glm4_new.rs
- quantized_glm4.rs
- quantized_qwen3_moe.rs


DrJesseGlass:
This was solved by another recent PR

PR #3347, opened 2026-01-30
Use flash attention for qwen3 models

guoqingbao:


DrJesseGlass:
I'll review this PR to make sure that I built along the same lines however I had integrated GPU flash attention and CPU flash attention into Qwen3 (but not MoE) but as I was doing it found substantial reason to upgrade the CPU flash attention and made an early simple submission #3254 to get an improved CPU flash structure before I started serious optimizations.

However, someone else was simultaneously working on #3250 for integrating varlen and the agreement at the time was once varlen was integrated that I would tweak mine to incorporate but it's been sitting unfinished.

guoqingbao:
> I'll review this PR to make sure that I built along the same lines however I had integrated GPU flash attention and CPU flash attention into Qwen3 (but not MoE) but as I was doing it found substantial reason to upgrade the CPU flash attention and made an early simple submission #3254 to get an improved CPU flash structure before I started serious optimizations.

This uses the previously defined flash attention interface and remains compatible with existing implementations.

DrJesseGlass:
Yes. It makes sense to get this integrated sooner if possible. I just meant to call out the redundancy.

guoqingbao:
> Yes. It makes sense to get this integrated sooner if possible. I just meant to call out the redundancy.

It would be better to have a unified entry point for varlen attention (CPU and GPU). I think I can help with GPU varlen attention, but it seems we are missing corresponding use cases in candle. Varlen attention requires parallel requests (though it can work with a single request) to meaningfully demonstrate performance.

ivarflakstad:
#3254 has been merged, so feel free to continue your work @guoqingbao  👌 

I've mentioned it before, but I want to flip the attention logic in candle a bit.
1. Get rid of `args.use_flash_attn`. Perhaps add a thread local static (env var) which can be used to disable flash attn for debugging purposes.
2. Call an model level abstraction like `attention.forward(...)` which does model specific prep and calls `candle_nn::attention(...)` which then automatically choses the optimal path. If `#[cfg(feature = "flash-attn")]` then it calls flash-attn (unless env var debugging), else call manual matmul -> softmax -> matmul attention (again often model specific softcap / masking may exist in which case the fallback must either be in the model code or we need some really good generic abstractions).

guoqingbao:
> #3254 has been merged, so feel free to continue your work @guoqingbao 👌
> 
> I've mentioned it before, but I want to flip the attention logic in candle a bit.
> 
> 1. Get rid of `args.use_flash_attn`. Perhaps add a thread local static (env var) which can be used to disable flash attn for debugging purposes.
> 2. Call an model level abstraction like `attention.forward(...)` which does model specific prep and calls `candle_nn::attention(...)` which then automatically choses the optimal path. If `#[cfg(feature = "flash-attn")]` then it calls flash-attn (unless env var debugging), else call manual matmul -> softmax -> matmul attention (again often model specific softcap / masking may exist in which case the fallback must either be in the model code or we need some really good generic abstractions).

Hi Ivar,

I've added a unified attention entry point (`candle_nn::attention::attention`) and migrated the Qwen3 models to use it. If this approach looks good, I can migrate other models to use the unified attention as well.

Dispatch pattern:

attention(q, k, v, scale, mask)
├── CANDLE_DISABLE_FLASH_ATTN=1? → standard_attn (matmul)
├── #[cfg(feature = "flash-attn")] + GPU? → candle_flash_attn (CUDA flash)
├── CPU + supported dtype? → cpu_flash (optimized kernel)
└── fallback → standard_attn (matmul)

PR #3373, opened 2026-02-17
feat: dynamic loading CUDA dylibs

mayocream:
This PR switches `cudarc`'s feature from `dynamic-linking` to `dynamic-loading`.

Dynamic linking means the Rust executable needs CUDA at startup; it will fail silently on Windows if the CUDA dylibs are missing, which makes debugging harder and provides the worst user experience. However, we could use `cudarc`'s `dynamic-loading` feature, which searches for CUDA dylibs in a few defined paths; it's more flexible. The software I developed (https://github.com/mayocream/koharu) uses this method, but I have to fork candle to achieve that.

candle and its dependencies that depend on cudarc need to switch to `dynamic-loading` at the same time; otherwise, cudarc will panic for both dynamic-linking and dynamic-loading used. That's why I forked `ug` to switch to the dynamic-loading feature.

Also, the recently added MoE implementation is incompatible with dynamic loading; it hard-codes the CUDA APIs and links CUDA at compile time despite cudarc. I modified it to use CUDA APIs at runtime via cudarc instead of ffi.

I locally tested qwen3-moe models with cuda with this change, it works perfectly.

```bash
$ cargo run --features cuda --example quantized-qwen3-moe --release -- --which 32b_q4k --prompt "A train is travelling at 120mph, how far does it travel in 3 minutes 30 seconds?"
    Finished `release` profile [optimized] target(s) in 0.44s
     Running `target\release\examples\quantized-qwen3-moe.exe --which 32b_q4k --prompt "A train is travelling at 120mph, how far does it travel in 3 minutes 30 seconds?"`
loaded 579 tensors (18.55GB) in 0.93s
model built
formatted prompt: <|im_start|>user
A train is travelling at 120mph, how far does it travel in 3 minutes 30 seconds?<|im_end|>
<|im_start|>assistant
We are given:

- Speed of the train = **120 miles per hour (mph)**
- Time = **3 minutes 30 seconds**

### Step 1: Convert time to hours
Since speed is in miles per hour, we need to convert the time into **hours**.
```

@ivarflakstad, could you help review this PR since you are involved in the original MoE PR?


mayocream:
cc @guoqingbao, could you help review this PR? Thanks!

PR #3376, opened 2026-02-18
feat: add Qwen3 ASR example

lumosimmo:
This PR adds a new `qwen3-asr` example to Candle by implementing a core inference path.

The example currently supports:
- Offline transcription
- Streaming transcription
- Loading model/tokenizer from Hugging Face or local path
- Automatic resampling of non-16kHz input audio to 16kHz

This PR intentionally keeps a narrow example scope:
- No forced-aligner timestamp path in this example
- No server/API integration from the standalone project
- No benchmark/regression tooling in this initial merge

## Validation

I ran:

- `cargo check --workspace`
- `cargo test -p candle-examples --example qwen3-asr --features symphonia`
- `cargo clippy -p candle-examples --example qwen3-asr --features symphonia -- -D warnings`
- `cargo clippy -p candle-examples --example qwen3-asr --features "symphonia forced-aligner timing" -- -D warnings`

Smoke-tested example commands:

- Offline:
  - `cargo run -p candle-examples --example qwen3-asr --features symphonia --release -- --model-id Qwen/Qwen3-ASR-0.6B --input sample:jfk --cpu`
  - `cargo run -p candle-examples --example qwen3-asr --features "symphonia metal" --release -- --model-id Qwen/Qwen3-ASR-0.6B --input sample:jfk`
- Streaming:
  - `cargo run -p candle-examples --example qwen3-asr --features symphonia --release -- --model-id Qwen/Qwen3-ASR-0.6B --input sample:jfk --cpu --stream --chunk-size-sec 0.5 --print-intermediate`
  - `cargo run -p candle-examples --example qwen3-asr --features "symphonia metal" --release -- --model-id Qwen/Qwen3-ASR-0.6B --input sample:jfk --stream --chunk-size-sec 0.5 --print-intermediate`

---

Originally I implemented this as a [standalone project](https://github.com/lumosimmo/qwen3-asr-rs). That one has a wider scope that's skipped here, including forced-aligner, benchmark tooling etc. I think this PR is enough for a minimal example, but happy to continue migrating everything over here.


airvzxf:
Nice work on this! I've opened #3509 which takes the same model but places it in `candle-transformers/src/models/qwen3_asr/` following the pattern used by Whisper, LLaMA, and other first-class models — this way other crates and downstream projects can depend on it directly. The example in `candle-examples/` imports from there. Curious to hear your thoughts — would be great to align both efforts.

PR #3382, opened 2026-02-20
Remove unwrap()s from candle-metal-kernels/src/metal/device.rs

jacobgorm:
Replace unwrap() calls with proper error propagation, to avoid crashing the caller if Metal fails to initialize

### DIFF
--- a/candle-metal-kernels/src/metal/device.rs
+++ b/candle-metal-kernels/src/metal/device.rs
@@ -77,7 +77,8 @@ impl Device {
         length: usize,
         options: MTLResourceOptions,
     ) -> Result<Buffer, MetalKernelError> {
-        let pointer = ptr::NonNull::new(pointer as *mut c_void).unwrap();
+        let pointer = ptr::NonNull::new(pointer as *mut c_void)
+            .ok_or_else(|| MetalKernelError::InvalidInput("Null pointer".to_string()))?;
         unsafe {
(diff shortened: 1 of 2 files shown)


candle-metal-kernels/src/metal/device.rs
```diff
@@ -77,7 +77,8 @@ impl Device {
         length: usize,
         options: MTLResourceOptions,
     ) -> Result<Buffer, MetalKernelError> {
-        let pointer = ptr::NonNull::new(pointer as *mut c_void).unwrap();
+        let pointer = ptr::NonNull::new(pointer as *mut c_void)
+            .ok_or_else(|| MetalKernelError::FailedToCreateResource("Null pointer".to_string()))?;
```
ivarflakstad: I don't think `std::ptr::NonNull` failing should result in a "failed to create metal resource".

This would only happen if we sent in a null ptr from our end into this fn.
Maybe we should add a "invalid input" kind of error?

ivarflakstad (commented):
Thanks! Good improvement!

I only had one comment. Can address it in this PR or later - up to you :)

candle-metal-kernels/src/metal/device.rs
jacobgorm: I've amended my change

ivarflakstad (approved):
👍

PR #3384, opened 2026-02-22
fix: make Qwen3 Model::clear_kv_cache public

Marenz:
Makes `clear_kv_cache` public on the Qwen3 `Model` struct, consistent with other model implementations (e.g. Llama, Phi, Mistral) that expose this method publicly.

Without this, downstream users cannot clear the KV cache between generations.


Marenz:
Duplicate of #3381

PR #3387, opened 2026-02-25
feat: add eps() and remove_mean() getters to LayerNorm and RmsNorm

### LINKED ISSUE #3017: candle_nn: LayerNorm does not expose `remove_mean` and `eps` parameter
I am working on serializing a neural net and I noticed that [`LayerNorm`](https://github.com/huggingface/candle/blob/461e8c1685e003bdddfd1e7d1aa5092786ca9df5/candle-nn/src/layer_norm.rs#L64) does not expose the `remove_mean` and `eps` parameters, so serializing this struct isn't possible.

Is hiding these fields on purpose?

jgremes:
Fixes #3017

Exposes `eps` and `remove_mean` as public getters on `LayerNorm`, following the same pattern as `weight()` and `bias()` that already existed.

Also adds `weight()` and `eps()` getters to `RmsNorm` since it has the same serialization limitation.


jgremes:
Hi, just checking in — happy to make any adjustments if needed!

jgremes:
@ivarflakstad - you suggested this in #3017, would love your feedback when you get a chance!

PR #3397, opened 2026-03-07
Fused conv2d + bias on cpu & cudnn backends

slckl:
Conv2d + bias impl has been a bit silly in that it computes conv2d and then uses potentially very expensive strided broadcast add op to add bias. 

This PR implements fused conv2d + bias kernel for cpu and uses existing cudnn kernel on cuda/cudnn backend. Other backends use the historic approach of conv2d + bias as separate ops.

Unfortunately, the best way to impl this that I could think of was to add a new Op for conv2d+bias and expand `Tensor` API, so that's the tradeoff...

Bulk of the work done by Opus 4.6, but under strict human supervision.

# Performance

Performance gains are, uhh, interesting. Below you'll find best median times of 3 runs on i7-12700h + RTX NVIDIA RTX A2000 8GB lappy, only reported for benches with +bias, as the non-bias version uses the old code without changes.

## CPU performance 
Benchmark command: `cargo bench conv2d -p candle-nn`

| Benchmark | main (38e7202) | fused-conv2d-bias | Speedup |
|-----------|------|-------------------|---------|
| F32 i128 k3x3 | 5.532 ms | 136.07 µs | **40.7×** |
| F32 i256 k5x5 | 56.811 ms | 802.12 µs | **70.8×** |
| F16 i128 k3x3 | 8.611 ms | 154.01 µs | **55.9×** |
| F16 i256 k5x5 | 67.679 ms | 807.88 µs | **83.8×** |

## Cudnn performance
Benchmark command: `cargo bench conv2d -p candle-nn --features cuda,cudnn`

| Benchmark | main (38e7202) | fused-conv2d-bias | Speedup |
|-----------|------|-------------------|---------|
| F32 i128 k3x3 | 259.69 µs | 39.609 µs | **6.6×** |
| F32 i256 k5x5 | 2144.6 µs | 40.532 µs | **52.9×** |
| F16 i128 k3x3 | 255.52 µs | 38.055 µs | **6.7×** |
| F16 i256 k5x5 | 2090.1 µs | 38.590 µs | **54.2×** |


slckl:
Sorry, I see that I've screwed up git history somehow on this branch, but the diff seems correct, and a squashed merge will probably put things right >_>

If we merge this, ofc...

Edit: I've since then unscrewed git history for this PR, carry on...

candle-core/src/conv.rs
```diff
@@ -305,6 +345,7 @@ impl Tensor {
         dilation: usize,
         groups: usize,
         cudnn_fwd_algo: Option<CudnnFwdAlgo>,
+        bias: &Option<Tensor>,
```
ivarflakstad: Nit: `Option<&Tensor>` is a tiny bit nicer, though it doesn't matter that much in any case as tensors are Arc'ed and cheap to clone.

candle-core/src/cpu_backend/conv2d.rs
```diff
@@ -3,13 +3,15 @@ use std::borrow::Cow;
...
 use crate::{
+    backend::BackendStorage,
     conv::ParamsConv2D,
     cpu_backend::{copy_strided_src_, Im2Col, Map1, Map2, MatMul},
     shape::dims4,
     Layout, Result, WithDType,
 };
 
 pub(super) struct Conv2D<'a>(pub(super) &'a crate::conv::ParamsConv2D);
+pub(super) struct Conv2DWithBias<'a>(pub(super) &'a crate::conv::ParamsConv2D);
```
ivarflakstad, line 14: Did you try putting `bias: Option<Tensor>` inside `crate::conv::ParamsConv2D`?

I think that might be the smoothest way to handle this. It's semver breaking, but we'll just bump v.

Then instead of `Conv2D.f()` we would use something like your `Conv2DWithBias.map()`
```rust
impl Conv2D {
  pub(super) fn map(
    &self,
    inp: &super::CpuStorage,
    inp_l: &Layout,
    kernel: &super::CpuStorage,
    kernel_l: &Layout,
    bias: Option<(&super::CpuStorage, &Layout)>,
  ) -> Result<super::CpuStorage> {
    match (inp, kernel) {
      (C::U8(inp), C::U8(k)) => {
        let mut out = Conv2D(self.0).f(inp, inp_l, k, kernel_l)?;
        if let Some((C::U8(b), l)) = bias { // Not sure if this syntax is actually correct, but you know what I'm saying
          apply_bias_inplace(&mut out, b, bias_l, self.0);
          Ok(C::U8(out))
        }
      }
      ...
    }
  }
}
```
Alternatively we check for `bias` in params before calling the cpu fn and dispatch to the correct variant.

What do you think? :)

ivarflakstad (commented):
This is great 👌

candle-core/src/conv.rs
slckl: changed to `Option<Tensor>`

candle-core/src/cpu_backend/conv2d.rs
slckl: Ok, so I've been trying to make this work, but I don't think it's gonna happen, unfortunately... At least not with just `ParamsConv2D`.

2 main reasons:
1. The existing `ParamsConv2D` consists of scalars and an enum, is generic for all devices and it derives `PartialEq` and `Eq` without problems. `Tensor` does not have `PartialEq` and `Eq`, so I can't add a `bias: Option<Tensor>` field.
```rust
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct ParamsConv2D {
    pub(crate) b_size: usize,
    pub(crate) i_h: usize,
    pub(crate) i_w: usize,
    pub(crate) k_h: usize,
    pub(crate) k_w: usize,
    pub(crate) c_out: usize,
    pub(crate) c_in: usize,
    pub(crate) padding: usize,
    pub(crate) stride: usize,
    pub(crate) dilation: usize,
    pub cudnn_fwd_algo: Option<CudnnFwdAlgo>,
    // Tensor is not Debug, PartialEq or Eq
    pub(crate) bias: Option<Tensor>,
}
```
2. To handle backwards I need to have the `Tensor` machinery so I can add the appropriate `BackropOp`. Which means I need to know whether I'm doing fused vs non-fused at the `Tensor`-api level:
```rust
    fn conv2d_single_group_with_bias(
        ..
    ) -> Result<Self> {
        let use_fused = match self.device() {
            crate::Device::Cpu => true,
            crate::Device::Cuda(_) => cfg!(feature = "cudnn") && kernel.layout().is_contiguous(),
            crate::Device::Metal(_) => false,
        };
        if use_fused {
            let storage = .. ;
            // Fused has a backprop3
            let op = BackpropOp::new3(self, kernel, bias, |arg, kernel, bias| Op::Conv2DBias {
                arg,
                kernel,
                bias,
                padding: params.padding,
                stride: params.stride,
                dilation: params.dilation,
            });
            let out_dims = params.out_dims();
            Ok(crate::tensor::from_storage(storage, out_dims, op, false))
        } else {
            // non-fused has backprop2 followed by another backprop2
            let out = self.conv2d_single_group(kernel, params)?;
            let b = bias.dims1()?;
            let bias = bias.reshape((1, b, 1, 1))?;
            out.broadcast_add(&bias)
        }
    }
```

I'm open to suggestions on how to better deal with the backprop situation...

ivarflakstad: 1. Those derives aren't actually used. Just remove them? 🤷 
2. I'm probably misunderstanding something here. The backwards pass doesn't care wether you did the fused on unfused variant. The output should be numerically equal.

ivarflakstad: Ah, I see now. One tensor adding two backprops. I'll have a think.

ivarflakstad: Actually I don't think there is a major issue here.
1 is solved, and I don't really see what the problem with 2 is.

Additionally, and note that I haven't fully thought this through, it could be that you could avoid the new backprop type as well and instead add two backprops manually. The backprop chain should be identical to what the `use_fused = false` path produces in that instance, except the storage behind conv2d and broadcast_add backprops is the same. Does that lead to issues? Maybe.

slckl: I think we should be able to do the same backprop ops for fused as the non-fused path. I can try.

So, are you ok with Tensor api extension? I have not found a better solution that would avoid breaking changes to `Tensor` api, that's the only tradeoff here.

No problems otherwise, imo.

ivarflakstad: I'm ok with adding `conv2d_single_group_with_bias`, as it is a private fn, but I'm also ok with changing `conv2d_single_group`. Something like
```rust
fn conv2d_single_group(&self, kernel: &Self, params: &ParamsConv2D) -> Result<Self> {
    if let Some(bias) = params.bias {
        let use_fused = match self.device() {
            crate::Device::Cpu => true,
            crate::Device::Cuda(_) => cfg!(feature = "cudnn") && kernel.layout().is_contiguous(),
            _ => false,
        };
        if use_fused {
            let storage = .. ;
            ...
            Ok(crate::tensor::from_storage(storage, out_dims, op, false))
        } else {
            ...
        }
    } else {
        let storage =
            self.storage()
                .conv2d(self.layout(), &kernel.storage(), kernel.layout(), params)?;
        ...
        Ok(crate::tensor::from_storage(storage, out_dims, op, false))
    }
}
```

PR #3405, opened 2026-03-16
fix: conv2d_tiled produces wrong results when C == H == W

### LINKED ISSUE #3404: conv2d_tiled produces incorrect results compared to conv2d_im2col_gemm (CPU backend)
## Bug Description

`conv2d_tiled` (the default CPU conv2d implementation) produces **incorrect results** that differ significantly from both PyTorch and candle's own `conv2d_im2col_gemm` implementation. The error is not a small floating-point precision difference — output values are **completely wrong** (max diff ~2.4 on outputs with magnitude ~0.27).

## Minimal Reproduction

```rust
use candle_core::{DType, Device, Module, Result, Tensor, Var};
use candle_nn::{Conv2dConfig, VarBuilder, VarMap};

fn main() -> Result<()> {
    let device = &Device::Cpu;

    // Create input [1, 64, 64, 64] and Conv2d(64 -> 384, 3x3, pad=1) with fixed weights
    let input = Tensor::randn(0f32, 1.0, (1, 64, 64, 64), device)?;
    
    // Create conv with random weights
    let varmap = VarMap::new();
    let vb = VarBuilder::from_varmap(&varmap, DType::F32, device);
    let conv = candle_nn::conv2d(64, 384, 3, Conv2dConfig { padding: 1, ..Default::default() }, vb)?;
    
    // Run forward
    let output = conv.forward(&input)?;
    
    // Save input, weight, output as safetensors
    // Then load in PyTorch and compare - they will differ significantly
    println!("Output shape: {:?}", output.dims());
    println!("Output abs_mean: {:.6}", output.to_dtype(DType::F64)?.abs()?.mean_all()?.to_scalar::<f64>()?);
    Ok(())
}
```

### Step 1: Generate reference data from PyTorch
```python
import torch
from safetensors.torch import save_file

torch.manual_seed(42)
x = torch.randn(1, 64, 64, 64)
conv = torch.nn.Conv2d(64, 384, 3, stride=1, padding=1)
with torch.no_grad():
    y = conv(x)

save_file({
    'input': x,
    'weight': conv.weight.contiguous(),
    'bias': conv.bias.contiguous(),
    'output': y,
}, 'conv2d_test.safetensors')
print(f'PyTorch output first 5: {y.flatten()[:5].tolist()}')
```

### Step 2: Load in Candle and compare
```rust
// Load safetensors, create Conv2d with same weights, run forward
// Candle output will differ from PyTorch output by max_diff ~2.4
```

## Root Cause

The issue is in `conv2d_tiled` at `candle-core/src/cpu_backend/conv2d.rs`.

### Proof: Switching to `FullIm2Col` fixes the issue

Changing line 21:
```rust
// BEFORE (broken):
const DEFAULT_CONV2D_IMPL: Conv2dImpl = Conv2dImpl::TiledIm2Col;

// AFTER (correct):
const DEFAULT_CONV2D_IMPL: Conv2dImpl = Conv2dImpl::FullIm2Col;
```

With `FullIm2Col`, the same test produces max_diff=**1.6e-6** (correct fp32 precision).

### Analysis

The `conv2d_tiled` implementation converts the input to NHWC layout and uses a `(c_in, kh, kw)` ordering for im2col patches and kernel flattening (lines 172-181, 217-243). While this should be mathematically equivalent to the standard `(kh, kw, c_in)` ordering, the tiled implementation appears to have a correctness bug in how it interacts with the matmul.

Specifically, the kernel is flattened as `[c_out, kh * kw * c_in]` with `c_in` as the innermost dimension (line 172-181), and the im2col column is built with `c_in` as the innermost dimension too (line 218). This pairing should be correct, but the resulting matmul gives wrong answers for certain channel/spatial configurations.

## Impact

This affects any Conv2d where:
- kernel size > 1x1 (1x1 convs use a separate fast path)
- The combination of in_channels, out_channels, and spatial size triggers the tiled path

In practice, this breaks inference for models with decoder stages that use 3x3 convolutions at medium-to-large spatial resolutions. We discovered this while implementing [BiRefNet](https://github.com/ZhengPeng7/BiRefNet) inference — the model produces correct output at 384x384 (where errors don't accumulate enough to matter) but completely wrong output at 1024x1024.

## Measured Results

| Implementation | max_diff vs PyTorch |
|---------------|-------------------|
| `TiledIm2Col` (default) | **2.396** |
| `FullIm2Col` | **1.6e-6** |
| `Direct` | (not tested) |

## Environment

- candle-core 0.9.2 (also verified on git main — same code)
- CPU backend (macOS ARM64)
- Rust 1.93.1

developer0hye:
## Bug

`conv2d_tiled` produces **completely wrong output** when `C_in == H == W` (e.g., input shape `[1, 64, 64, 64]`).

Fixes #3404

## Root Cause

`conv2d_tiled` converts input from NCHW to NHWC internally. It uses a stride comparison to skip this conversion if the input appears to already be NHWC:

```rust
let layout_is_valid = inp_l.stride() == [H*W*C, W*C, C, 1];
```

**When `C == H == W`, NCHW and NHWC strides are numerically identical:**

| Layout | Strides |
|--------|---------|
| NCHW `[B, C, H, W]` | `[C*H*W, H*W, W, 1]` |
| NHWC `[B, H, W, C]` | `[H*W*C, W*C, C, 1]` |

When `C == H == W = 64`: both become `[262144, 4096, 64, 1]`. The check passes, conversion is skipped, and NCHW data is read as NHWC during im2col — producing completely wrong results.

## Fix

```rust
// Before:
let layout_is_valid = inp_l.stride() == [cont_s0, cont_s1, cont_s2, 1];

// After:
let layout_is_valid = false;
```

Candle tensors are always NCHW. The "already NHWC" fast path can never legitimately fire, so we disable it entirely.

This same fix is applied to both `conv2d_tiled` and `conv2d_direct` which share the same pattern.

## Correctness

Added regression test `conv2d_c_eq_h_eq_w` that:
- Creates `Conv2d(4, 2, 3x3, padding=1)` with input `[1, 4, 4, 4]` (C==H==W==4)
- Verifies output channels are identical for all-ones kernel (sanity check)
- Verifies specific output value matches hand-calculated expected result

**Without fix:** test fails (`got 4.232, expected 1.431, diff=2.80`)
**With fix:** test passes

All 8 existing conv tests continue to pass.

## Performance

The `layout_is_valid` optimization only fired when `C == H == W` — which is the case that produced wrong results. For all other inputs (the vast majority of real models), the conversion was already running.

| Case | Before | After | Impact |
|------|--------|-------|--------|
| `[1,64,64,64]→384` (bug case) | 8,716 us | 10,215 us | +17% (was producing wrong output) |
| `[1,128,56,56]→256` (general) | 8,487 us | 7,746 us | noise |
| `[1,256,28,28]→512` (general) | 8,379 us | 8,097 us | noise |
| `[1,512,14,14]→1024` (general) | 11,821 us | 11,204 us | noise |

**No correct computation becomes slower.** The bug case is slower because it now performs the (necessary) conversion, but the previous "fast" result was incorrect.

## Real-world Impact

This bug prevented correct inference of [BiRefNet](https://github.com/ZhengPeng7/BiRefNet) at 1024x1024 resolution. The model's decoder produces intermediate tensors of shape `[1, 64, 64, 64]`, triggering the bug. With this fix, BiRefNet inference matches PyTorch within fp32 precision at all resolutions.

- [candle-birefnet](https://github.com/developer0hye/candle-birefnet) — BiRefNet implementation that discovered this bug


candle-core/src/cpu_backend/conv2d.rs
```diff
@@ -293,29 +288,24 @@ fn conv2d_direct<T: WithDType + num_traits::Num + Copy + 'static>(
     // Output shape: [b_size, c_out, out_h, out_w].
     let dst = vec![T::zero(); p.b_size * p.c_out * out_h * out_w];
 
-    // Make contiguous input copy if needed.
+    // Convert NCHW input to NHWC layout for direct convolution.
+    // See: https://github.com/huggingface/candle/issues/3404
```
slckl: Same as above, I would remove the issue link to keep comments leaner.

candle-core/src/cpu_backend/conv2d.rs
```diff
@@ -139,29 +139,24 @@ fn conv2d_tiled<T: WithDType + num_traits::Num + Copy + 'static>(
     // Output shape: [b_size, c_out, out_h, out_w].
     let dst = vec![T::zero(); p.b_size * p.c_out * out_h * out_w];
 
-    // Make contiguous input copy if needed.
+    // Convert NCHW input to NHWC layout for tiled im2col.
+    // See: https://github.com/huggingface/candle/issues/3404
```
slckl: Since you're removing a hack instead of adding something extra, I would remove the link to issue here.

candle-core/src/cpu_backend/conv2d.rs
slckl: same as above, I would remove the link to the issue here.

candle-core/src/quantized/mod.rs
```diff
@@ -14,6 +14,7 @@ pub mod imatrix_file;
 pub mod k_quants;
 #[cfg(feature = "metal")]
 pub mod metal;
+#[cfg(feature = "tokenizers")]
```
slckl: unrelated change, pls submit a separate PR for this.

candle-core/tests/conv_tests.rs
```diff
@@ -860,6 +860,86 @@ fn conv2d_grad(dev: &Device) -> Result<()> {
...
+    // total = 1.4310
+    let actual_corner: f32 = res.i((0, 0, 0, 0))?.to_scalar()?;
+    let corner_diff = (actual_corner - 1.4310).abs();
+    assert!(
+        corner_diff < 1e-3,
+        "conv2d with C==H==W: top-left corner mismatch: got {actual_corner}, expected 1.4310, diff={corner_diff}"
+    );
+
+    Ok(())
+}
```
slckl, line 934: Can confirm this test fails on main, and succeeds with your changes, good job!

candle-core/tests/conv_tests.rs
```diff
@@ -860,6 +860,86 @@ fn conv2d_grad(dev: &Device) -> Result<()> {
...
+
+test_device!(
+    conv2d_c_eq_h_eq_w,
+    conv2d_c_eq_h_eq_w_cpu,
+    conv2d_c_eq_h_eq_w_gpu,
+    conv2d_c_eq_h_eq_w_metal
+);
```
slckl: I would move this section below the other `test_device!`, to the very bottom, since these macros follow the same order as test definitions.

candle-core/Cargo.toml
```diff
@@ -33,7 +33,7 @@ safetensors = { workspace = true }
 thiserror = { workspace = true }
 yoke = { workspace = true }
 zip = { workspace = true }
-tokenizers = { workspace = true, features = ["onig"] }
+tokenizers = { workspace = true, features = ["onig"], optional = true }
```
slckl: unrelated change, pls submit a separate PR for this

slckl:
Good find!

I rerun benchmarks since the whole point of tiled conv was to improve the performance of conv2d on cpu. Fortunately, this change does not seem to make the tiled conv meaningfully slower, changes seem within noise levels.

## Benchmarks:
Bo3, on i7-12700h:

### main
```
cpu_conv2d_F32_i128_k1x1_nb/iter
                        time:   [29.254 µs 29.419 µs 29.666 µs]
cpu_conv2d_F32_i128_k5x5_nb/iter
                        time:   [223.53 µs 228.09 µs 233.73 µs]
cpu_conv2d_F32_i512_k3x3_nb/iter
                        time:   [1.8525 ms 1.9264 ms 2.0002 ms]
cpu_conv2d_F16_i128_k1x1_nb/iter
                        time:   [35.495 µs 35.804 µs 36.186 µs]
cpu_conv2d_F16_i128_k5x5_nb/iter
                        time:   [211.55 µs 214.27 µs 217.70 µs]
cpu_conv2d_F16_i512_k5x5_nb/iter
                        time:   [2.4261 ms 2.4615 ms 2.5095 ms]
```

### this branch
```
cpu_conv2d_F32_i128_k1x1_nb/iter
                        time:   [29.710 µs 29.951 µs 30.348 µs]
cpu_conv2d_F32_i128_k5x5_nb/iter
                        time:   [221.32 µs 225.11 µs 229.71 µs]
cpu_conv2d_F32_i512_k3x3_nb/iter
                        time:   [2.0539 ms 2.1074 ms 2.1568 ms]
cpu_conv2d_F16_i128_k1x1_nb/iter
                        time:   [34.808 µs 35.043 µs 35.342 µs]
cpu_conv2d_F16_i128_k5x5_nb/iter
                        time:   [212.97 µs 215.21 µs 217.97 µs]
cpu_conv2d_F16_i512_k5x5_nb/iter
                        time:   [2.5030 ms 2.5259 ms 2.5538 ms]
```

slckl (commented):
A few minor issues, but the fix looks ok.

candle-core/src/cpu_backend/conv2d.rs
developer0hye: Done, removed the issue link.

candle-core/src/cpu_backend/conv2d.rs
developer0hye: Done, removed here as well.

candle-core/src/quantized/mod.rs
developer0hye: Reverted. Will submit as a separate PR.

candle-core/tests/conv_tests.rs
developer0hye: Thanks for verifying!

candle-core/tests/conv_tests.rs
developer0hye: Done, moved it to the bottom after the other test_device! macros.

candle-core/Cargo.toml
developer0hye: Reverted. Will submit as a separate PR.

slckl:
LGTM!
Due to github permission issues I, unfortunately, can't resolve the previous review comments...

developer0hye:
@slckl I resolved them!

developer0hye:
Pushed a formatting fix commit (71a9d69) — ran `cargo fmt --all` to resolve the rustfmt CI failure.

The CI workflows are currently in `action_required` status, awaiting maintainer approval to run (expected for outside collaborator PRs). Once approved, the rustfmt check should pass.

PR #3409, opened 2026-03-19
fix(metal): GPU-native dequantization for quantized matmul precision

jamesbrink:
## Problem

Metal's fused quantized matmul kernels (`kernel_mul_mm_*`) accumulate in half precision (F16). For single-pass inference (LLMs), this is fine. For iterative workloads like diffusion model denoising (20-50 steps), the precision errors compound across iterations and produce incorrect results (washed-out or garbage images).

CUDA doesn't have this problem because for large matrices it dequantizes on GPU using dedicated kernels, then uses cuBLAS GEMM at full precision.

## Fix

Metal already has `kernel_get_rows_*` GPU dequantize kernels in `quantized.metal` — they just had no Rust dispatch function. This PR:

1. **Adds `call_dequantize_f32()`** in `candle-metal-kernels` — dispatches `kernel_get_rows_*` to dequantize a full quantized tensor to F32 on the GPU
2. **Adds `dequantize_on_gpu()`** in `candle-core` — uses the new dispatch to dequantize without a CPU roundtrip
3. **Updates `QMetalStorage::fwd()`** — dequantizes weight to F32 on GPU, uses standard Metal GEMM, frees the temporary copy
4. **Fixes BF16 mapping bug** — `GgmlDType::BF16` was incorrectly mapped to `candle_metal_kernels::GgmlDType::F16`
5. **Falls back to CPU** for unsupported dtypes (Q8_1, Q8K)

### Memory impact

Per-matmul overhead on Apple Silicon:
- Largest typical layer (e.g. FLUX [3072, 18432]): ~225 MB temporary F32 buffer
- Freed immediately after GEMM + sync via `wait_until_completed()`
- GGUF model stays in quantized form — no additional persistent memory

### Tradeoffs

This trades some throughput for correctness on Metal. The fused kernels are faster (no temporary allocation) but produce incorrect results for iterative inference. The approach matches what CUDA already does for large matrices.

## Files changed

| File | Change |
|------|--------|
| `candle-metal-kernels/src/kernels/quantized.rs` | Add `call_dequantize_f32()` + dtype helpers |
| `candle-metal-kernels/src/kernels/mod.rs` | Export new function |
| `candle-core/src/quantized/metal.rs` | `dequantize_on_gpu()`, updated `fwd()`, `dequantize()`, BF16 fix |

## Testing

- Verified GGUF FLUX inference on Apple M3 Max produces correct, clear images (previously garbage)
- Same seed produces identical results across runs
- `cargo test` passes

### DIFF
--- a/candle-core/src/quantized/metal.rs
+++ b/candle-core/src/quantized/metal.rs
@@ -33,76 +33,125 @@ impl QMetalStorage {
         &self.buffer
     }
 
-    pub fn dequantize(&self, elem_count: usize) -> Result<MetalStorage> {
+    /// Dequantize a quantized tensor to F32 using GPU kernels.
+    ///
+    /// Uses Metal `kernel_get_rows_*` to dequantize directly on GPU, avoiding
+    /// the CPU roundtrip. Falls back to CPU for unsupported dtypes (Q8_1, Q8K).
+    fn dequantize_on_gpu(&self, n_rows: usize, n_cols: usize) -> Result<MetalStorage> {
+        let elem_count = n_rows * n_cols;
+
+        // Check if this dtype is supported by GPU dequantize kernels
+        let kernel_dtype: candle_metal_kernels::GgmlDType = self.dtype.into();
+        let gpu_supported = !matches!(
+            self.dtype,
+            GgmlDType::Q8_1 | GgmlDType::Q8K
+        );
+
+        if !gpu_supported {
+            // Fall back to CPU dequantize for unsupported types
+            return self.dequantize_cpu(elem_count);
+        }
+
+        // Allocate F32 output buffer on GPU
+        let output_size = elem_count * 4; // f32 = 4 bytes
+        let dst_buffer = self.device.allocate_zeros(output_size)?;
+
+        // Create sequential index buffer [0, 1, ..., n_rows-1] as i32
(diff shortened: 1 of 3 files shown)


ivarflakstad:
My AI assistance detector is beeping. Which isn't necessarily bad, but they aren't always fully correct.

> Metal's fused quantized matmul kernels (kernel_mul_mm_*) accumulate in half precision (F16).

They accumulate in f32. See `mc` [here](https://github.com/huggingface/candle/blob/main/candle-metal-kernels/src/metal_src/quantized.metal#L7036) and accum fn call [here](https://github.com/huggingface/candle/blob/main/candle-metal-kernels/src/metal_src/quantized.metal#L7099).


> CUDA doesn't have this problem because for large matrices it dequantizes on GPU using dedicated kernels, then uses cuBLAS GEMM at full precision.

Do you mean when `FORCE_DMMV` is set to true? It defaults to false, and uses kernels that have the same approach (but not identical) to the metal variants.

AI often end up making changes than are needed. Have you checked that the entire issue is not fixed by just this? :)
>4. Fixes BF16 mapping bug — GgmlDType::BF16 was incorrectly mapped to candle_metal_kernels::GgmlDType::F16

It's a good fix and I'd happily merge a PR only that.

PR #3439, opened 2026-04-01
feat(auto): AutoModelForCausalLM — 21 float models, GGUF support, impl_causal_lm! macro

### LINKED ISSUE #3120: AutoModel / PreTrainedModel equivalent magic ?
Hello all, first, thanks a lot for this wonderful crate. 

I was wondering if it's on the roadmap or if there is a solution to have the same magic as in python with a `AutoModel.from_pretrained("the_model_name_string")`

As I'm protoyping and am often changing models... which requires to change the architecture everytime and having this "auto load" would save time. 

Alternatives : https://github.com/lucasjinreal/Crane or https://docs.rs/kalosm/latest/kalosm/

Thanks in advance, 
Have a nice day.

ArthurZucker:
## What this adds

A complete `AutoModelForCausalLM` factory with GGUF support, 21 float models, and a macro-driven design that makes adding new models a one-liner.

Refactor examples to use it 😉 

## Key design

**Three macros replace all manual boilerplate:**

```rust
// Every model: wire forward + clear_kv_cache into the CausalLM trait
crate::impl_causal_lm!(mistral::Model, "mistral");

crate::make_auto_map!(config, vb, use_flash_attn, {
    "llama"   => (llama::Config,   llama::LlamaForCausalLM),
    "mistral" => (mistral::Config, mistral::Model),
    // ... 19 more
});

crate::make_gguf_map!(arch, content, reader, device, {
    "llama" => quantized_llama::ModelWeights,
    // ...
});
```

**Standardisation applied across the entire model zoo:**
- `clear_kv_cache()` added to the 3 float models that were missing it
- All models now accept `seqlen_offset: usize` in `forward` — uniform signature everywhere
- `use_flash_attn: bool` with `#[serde(default)]` added to every model config
- `LlamaConfig`/`GraniteConfig` two-struct splits collapsed: `Config` is now directly deserializable, `into_config()` removed, backward-compat type aliases kept

## Model coverage

**Float (safetensors):** llama, mistral, mixtral, phi, phi3, qwen2, qwen2_moe, qwen3, qwen3_moe, gemma, gemma2, gemma3, starcoder2, deepseek_v2, olmo, olmo2, stablelm, yi, helium, granite, chatglm

**GGUF:** llama, phi, phi3, qwen2, qwen3, gemma3

Original PR by @Elvis339 — commits preserved at the base of this branch. closes #3120

### DIFF
--- a/candle-examples/examples/auto/README.md
+++ b/candle-examples/examples/auto/README.md
@@ -1,0 +1,20 @@
+# AutoModelForCausalLM
+
+Load any supported causal language model by name, without manually specifying the architecture.
+
+## Supported Models
+
+- `llama` (Llama 2, Llama 3, etc.)
+- `mistral` (Mistral 7B, etc.)
+- `phi3` (Phi-3)
+- `qwen2` (Qwen2, Qwen2.5)
+- `gemma` (Gemma)
+
+## Usage
+
+```bash
+# Qwen2
+cargo run --example auto --release -- \
+    --model "Qwen/Qwen2-0.5B-Instruct" \
+    --prompt "Hello, I am"
+```
--- a/candle-examples/examples/chatglm/main.rs
+++ b/candle-examples/examples/chatglm/main.rs
@@ -76,7 +76,8 @@ impl TextGeneration {
             let context_size = if index > 0 { 1 } else { tokens.len() };
             let ctxt = &tokens[tokens.len().saturating_sub(context_size)..];
             let input = Tensor::new(ctxt, &self.device)?.unsqueeze(0)?;
-            let logits = self.model.forward(&input)?;
+            let seqlen_offset = tokens.len().saturating_sub(context_size);
+            let logits = self.model.forward(&input, seqlen_offset)?;
             let logits = logits.squeeze(0)?.to_dtype(DType::F32)?;
             let logits = if self.repeat_penalty == 1. {
                 logits
--- a/candle-transformers/src/models/gemma.rs
+++ b/candle-transformers/src/models/gemma.rs
@@ -446,3 +446,4 @@ impl Model {
         }
     }
 }
+crate::impl_causal_lm!(Model, "gemma");
--- a/.github/workflows/ci_cuda.yaml
+++ b/.github/workflows/ci_cuda.yaml
@@ -25,11 +25,11 @@ jobs:
       CUDA_COMPUTE_CAP: 86
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
       - name: Install dependencies
         run: apt update && apt install curl build-essential libssl-dev protobuf-compiler pkg-config -y
       - name: Install Rust Stable
-        uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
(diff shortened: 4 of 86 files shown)


Elvis339:
This is great, thanks for taking this another mile!

ArthurZucker:
SOrry for the delay in reviewing, repo changed a bit since so I thought I'd push this through!

candle-examples/examples/auto/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Load any supported causal language model by name, without manually specifying the architecture.
+
+## Supported Models
+
+- `llama` (Llama 2, Llama 3, etc.)
+- `mistral` (Mistral 7B, etc.)
+- `phi3` (Phi-3)
+- `qwen2` (Qwen2, Qwen2.5)
+- `gemma` (Gemma)
```
ivarflakstad, line 11: Quite a bit more by now!

Maybe we should keep a more formal list in the `auto` module?
This can then only mention models informally and refer to the `auto` itself, so that we don't have to keep this example README updated.

candle-examples/examples/chatglm/main.rs
ivarflakstad, line 1: Not worried about the auto model logic working correctly, more so special case logic like this:
```rust
let tokenizer_filename = match args.tokenizer {
  Some(file) => std::path::PathBuf::from(file),
  None => api
    .model("lmz/candle-chatglm".to_string())
    .get("chatglm-tokenizer.json")?,
};
```
Because this block implies that the tokenizer in `THUDM/chatglm3-6b` is incompatible with candle. And indeed when I check the [repo](https://huggingface.co/zai-org/chatglm3-6b/tree/main) it uses a `SentencePiece` tokenizer. Previously what we have done is convert the SP tokenizer to a `tokenizers` compatible format and upload to HF, or in the case of the `marian-mt` example there's a script that converts it locally (I prefer first alternative).

We use the `tokenizers` python package to convert. I don't think the rust crate exposes `SentencePiece` logic.

I wasn't planning on running every single example you've updated here to verify that they all work correctly. At least not all in one go. Would it be better (slower and boring) to break it up into a couple PRs?

Also: if the auto example runs a model perfectly and the existing model example displays no additional functionality we don't have to update the example - we can remove it 🤷

candle-transformers/src/models/gemma.rs
```diff
@@ -15,6 +15,8 @@ fn default_max_position_embeddings() -> usize {
 
 #[derive(serde::Deserialize, Debug, Clone)]
 pub struct Config {
+    #[serde(default)]
+    pub use_flash_attn: bool,
```
ivarflakstad: Just noting down:

One goal I have for the future is to:

- Add `Attention` custom ops in `candle-nn` , which has fallback to good ol' matmul -> softmax -> etc.
  -  If defined custom ops should use fallback if op is not supported on backend
- Remove all the `#[cfg(feature = "flash-attn")]` toggled `flash_attn` fns. It should simply be best effort based on hardware, installed software, and feature flags.
- Remove `use_flash_attn` from all examples. Again, should just be best effort.
- Consider adding some kind of toggled debug logging that let's users easily inspect what functionality is currently enabled on their system (since it no longer crashes when `use_flash_attn=true` and `flash-attn` is not enabled)

ivarflakstad:
Just tested auto example with
`cargo run --example auto --release --features=metal -- --model "meta-llama/Llama-3.2-1B" --prompt "Hello there"`
Output is sane, but completely different from the llama example. It is also significantly slower (37 tok/s vs 60 tok/s). Any ideas why?

PR #3444, opened 2026-04-02
Metal: fused MoE kernels, QTensor buffer access, encoder reuse, Q4K tuning

emanueleDiVizio:
## Summary

This PR enables fused execution of quantized Mixture-of-Experts (MoE) layers on Metal by exposing direct access to QTensor GPU storage and introducing fused kernels for expert dispatch.

Instead of the standard path:

    dequantize → gather → matmul → activation

this enables:

    direct quantized access → fused gather + matmul + activation (single kernel)

This significantly reduces memory traffic and GPU dispatch overhead for MoE models on Apple Silicon.

### QTensor GPU access
- Expose `QTensor::metal_storage()` for direct access to underlying Metal buffers
- Enables fused kernels to operate directly on quantized weights (no intermediate dequantization)
- Export `Source` enum for external integration

### Fused MoE kernels (Metal)
- `call_gather_mm`: fused token routing + quantized matmul
- Fused gate + up + SiLU kernel
- Eliminates intermediate gather buffers

### Runtime optimizations
- Optional compute encoder reuse (`CANDLE_METAL_REUSE_ENCODER=1`)
  - Reduces command encoder overhead for workloads with many small kernels (e.g. MoE)
- Enable 2 simdgroups for Q4K matmul (aligned with llama.cpp)

## Motivation

MoE models dispatch to multiple experts per token, resulting in many small matmuls per layer. On Metal, this leads to:

- excessive kernel launches
- command encoder overhead
- unnecessary memory movement

This PR addresses these by:
- fusing operations into single kernels
- operating directly on quantized data
- reducing dispatch overhead

## Testing

- Verified on Apple Silicon (M-series) with Mixtral Q4K via mistral.rs
- Confirmed no regression for non-MoE inference
- Tested encoder reuse enabled/disabled


ivarflakstad:
Due to an increase in vibe coded PRs it would be nice to have a simple way to verify the implementation.
Could be model usage via `candle-examples`, with good output and some numbers of what you yourself have seen as the performance gain.

Also, minimal amount of changes per PR is preferred. This PR is at least 3 PRs in one.
1. Compute encoder reusage
   We've explored this before, with no measurable performance gain - but I'm open to run some new benchmarks.
2. Enable 2 simdgroups for Q4K matmul (aligned with llama.cpp)
   This would be merged very quickly if it includes a link to the exact piece of code in llama.cpp that "aligned with" refers to, and preferrably some proof that it improves performance somewhat.
3. Metal fused MoE dispatch for Q4K.
   I'm interested in seeing how it performs, but requiring pre-baked kernels ala `kernel_mul_mv_id_swiglu_q4_K_f32` for each variation of quantization and MoE is outside the scope of candle as a library. It would be better suited as a `CustomOp`.

PR #3456, opened 2026-04-09
Handle Flash Attn V2 SMEM conditions better

EricLBuehler:
For head_dim > 256, the kernel requires `kBlockN >= 64` which needs more shared memory than some GPUs have. This PR adds handling for the following "tiers" and a function to dynamically check compatability.

- Tier 1: >= 192 KB (H100, Blackwell) => 64x64 tile size
- Tier 2: >= 160 KB (A100) => 32x64 tile size
- Tier 3: < 160 KB (sm86/89) => cannot run correctly, caller must check supports_head_dim() and use eager attention.


EricLBuehler:
https://github.com/Dao-AILab/flash-attention/issues/2427

PR #3459, opened 2026-04-11
Refactor read functions to take ownership of reader

Qu1ntev:


ivarflakstad:
Hey!

Could you give some description as to why this is an improvement?

Thanks! :)

Qu1ntev:
Hey, @ivarflakstad !

```rust
use std::io::{Cursor, Read, Seek};

fn test_func<R: Read + Seek>(mut reader: R) { // &mut is not needed (trait satisfied)
    let mut buf = String::new();
    reader.read_to_string(&mut buf).unwrap();
}

fn main() {
    let mut cursor = Cursor::new(b"hello world");
    test_func(&mut cursor); // mut ref
    // test_func(&cursor); // error! not mut
    test_func(cursor); // or move
}
```

but I forgot to write mut var: R in the commit, now there is var: R

edit:
I looked at the code in more detail, it needs a lot of fixes that won’t affect anything, I think the PR is closed :)

PR #3463, opened 2026-04-13
Add fast CUDA MMVQ GGUF kernels

EricLBuehler:
Adds `mmvq_gguf.cu` and surrounding infra enabling a fast-path for GGUF decode on CUDA.

Specifically, it adds:    
- Native BF16 input/output support (no F32 round-trip!)                                                                                                                             
- On-the-fly activation quantization to `Q8_1` with dedicated `BF16`/`F16`/`F32` quantize kernels                                                                                                            
- Per-device `Q8_1` scratch workspace that is lazily allocated and reused across calls                                                                                                                 
- Batch sizes 1 to 8 with compile-time specialized kernel variants per batch size                                                                                                                       
- Automatic fast-path dispatch, falls back to existing PTX-based path for unsupported types or larger batches.

### DIFF
--- a/candle-core/src/quantized/fast_mmvq.rs
+++ b/candle-core/src/quantized/fast_mmvq.rs
@@ -1,0 +1,279 @@
+//! CUDA fast path for GGUF matmul with BF16/F32 activations.
+
+use std::collections::HashMap;
+use std::sync::{Mutex, OnceLock};
+
+use super::cuda::{QCudaStorage, MATRIX_ROW_PADDING};
+use super::GgmlDType;
+use crate::cuda_backend::DeviceId;
+use crate::{backend::BackendStorage, CudaDevice, CudaStorage, DType, Result, Shape};
+
+use cudarc::driver::{CudaSlice, DevicePtr};
+
+const Q8_1_BLOCK_SIZE: usize = 32;
+const Q8_1_TYPE_SIZE: usize = 36; // 2 halves (4 bytes) + QK8_1 int8 = 4 + 32 = 36
+
+#[inline]
+fn pad(p: usize, q: usize) -> usize {
+    p.div_ceil(q) * q
+}
+
+/// Quant types supported by the fast MMVQ kernels.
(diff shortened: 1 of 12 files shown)


candle-core/src/quantized/fast_mmvq.rs
```diff
@@ -0,0 +1,279 @@
...
+    u64,
+    std::sync::MutexGuard<'static, HashMap<DeviceId, WorkspaceSlot>>,
+)> {
+    let map = WORKSPACE.get_or_init(|| Mutex::new(HashMap::new()));
+    let device_key = dev.id();
+    let mut guard = map.lock().unwrap();
+    let slot = match guard.get_mut(&device_key) {
+        Some(slot) => slot,
+        None => {
+            let slice = unsafe { dev.alloc::<u8>(bytes.max(1))? };
```
ivarflakstad: If none this always allocates 1 byte, then immediately overwrites it directly below.

I think this section would be easier to implement using the `HashMap::entry` api.

candle-core/src/quantized/fast_mmvq.rs
```diff
@@ -0,0 +1,279 @@
...
+    self_shape: &Shape,
+    rhs: &CudaStorage,
+    rhs_l: &crate::Layout,
+) -> Result<Option<(CudaStorage, Shape)>> {
+    use candle_kernels::ffi;
+
+    // Gate checks.
+    if super::cuda::FORCE_DMMV.load(std::sync::atomic::Ordering::Relaxed) {
+        return Ok(None);
+    }
```
ivarflakstad: I think it would be enough to check this in the code that calls this fn (especially since we already do anyways)

EricLBuehler: 👍

candle-core/src/quantized/fast_mmvq.rs
EricLBuehler: Done! Thanks

ivarflakstad (approved):
lgtm! 👌

PR #3465, opened 2026-04-15
Add fast CUDA MMQ GGUF kernels

EricLBuehler:
Adds infra enabling a fast-path for GGUF prefill on CUDA, resulting in a +30% prefill T/s improvement for models & sizes I tested (e.g., Gemma 4 @ q8_0 went from 1545.3t/s to 2014.3t/s)


ivarflakstad (approved):
Lgtm!

PR #3467, opened 2026-04-15
Lazy backend

ivarflakstad:
Experimental lazy backend.

The first design used a graph library to capture the ops, but it was a bit slower than I liked, so I wanted to try something that I haven't seen a lot of: Self referential nodes with interior mutability.

Seems illegal in rust, but by wrapping the `Op` in an `AtomicPtr` we get all we need.

Here's a rough outline of how the "graph" is wired together.
```rust
pub enum Op {
  Unary(Unary),
  Binary(Binary),
  // ... etc
}
pub struct Unary {
    src: LazyStorage,
    op: &'static str,
}
pub struct Binary {
    lhs: LazyStorage,
    rhs: LazyStorage,
    op: &'static str,
}
pub struct LazyStorage {
    op: Arc<AtomicOp>,
}

type Graph = &[&LazyStorage];
```

Currently supports metal and enough features to run llama.
Example usage `CANDLE_LAZY=1 cargo run --example llama --release --features=metal -- --prompt "Hey there" --sample-len 200 --which v32-1b --dtype=bf16 --temperature=0 --repeat-penalty=1.0`.

Currently there are no optimizations, so all you get is the CPU overhead that comes with lazy execution. However that overhead is only 4-5% which is pretty good for a first stab (see pytorch team very happy about the <5% overhead of TorchDynamo [here](https://docs.pytorch.org/assets/pytorch2-2.pdf)).

Next step is (obviously) automatic optimizations.
Elementwise fusion, auto-detect special ops (attention), etc.


candle-core/src/device.rs
```diff
@@ -288,7 +302,8 @@ impl Device {
         match self {
             Self::Cpu => DeviceLocation::Cpu,
             Self::Cuda(device) => device.location(),
-            Device::Metal(device) => device.location(),
+            Self::Metal(device) => device.location(),
+            Self::Lazy(_) => DeviceLocation::Lazy,
```
EricLBuehler, line 306: Should all lazy device locations be lazy? Maybe we need a notion of logical vs physical device location

candle-core/src/lazy/mod.rs
```diff
@@ -0,0 +1,1968 @@
...
+static CUSTOM_OP_REGISTRY: LazyLock<Mutex<HashMap<String, CustomOp>>> =
+    LazyLock::new(|| Mutex::new(HashMap::<String, CustomOp>::new()));
+
+#[derive(Debug)]
+pub struct AtomicOp {
+    ptr: AtomicPtr<Op>,
+}
+
+impl AtomicOp {
+    fn get_ptr(op: Op) -> *mut Op {
```
EricLBuehler, line 80: This seems to have an unsound `AtomicOp` ownership model I think, due to provenance + lifetime issues.

It stores `Arc::into_raw`, then `swap()` moves the `Op` out with `value.read()` at without reconstructing/dropping the Arc, leaking allocations.`load_ref()` also returns an unguarded `&Op` from the atomic pointer, so a concurrent `store()`/`swap()` can move/drop the value while references exist.

This likely needs an owning primitive such as `ArcSwap`, `RwLock<Arc<Op>>`, or careful `Arc::from_raw` handling plus reader lifetime protection.

candle-core/src/storage.rs
```diff
@@ -79,6 +112,10 @@ impl Storage {
             Storage::Cpu(storage) => storage.const_set(v, l),
             Storage::Cuda(storage) => storage.const_set(v, l),
             Storage::Metal(storage) => storage.const_set(v, l),
+            Storage::Lazy(storage) => {
+                let _: () = storage.const_set_lazy(v);
```
EricLBuehler, line 116: I think this ignores the layout unlike the other branches?

PR #3477, opened 2026-04-19
metal: fix RMSNorm NaN on large-magnitude F32 inputs

tomsanbear:
## Summary

The fused Metal `rmsnorm_f32` kernel produced NaN when fed F32 activations around ~1e9 or larger, silently breaking any model that pushes activations that high — e.g. BitNet b1.58 where the squared-ReLU × up_proj outer product in the FFN lands around 7e9 and the inner `ffn_sub_norm` collapses to all-NaN.

## Root cause

`RMSReduceOp::operator()` in `candle-metal-kernels/src/metal_src/reduce.metal` evaluated a Welford-shaped correction term:

```cpp
uint new_count = a.count + b.count;
uint nb_over_n = b.count / new_count;
T delta = b.mean - a.mean;
a.mean += b.mean + delta * delta * a.count * nb_over_n;
```

This formula doesn't apply to the `RMS<T>` accumulator. `RMS<T>` only carries `{count, mean}` — no `m2` field — and `RMSLoadOp` accumulates `Σx²` straight into the `mean` field (the name is inherited from the LayerNorm Welford struct and is misleading; it's a sum of squares, not a mean). Welford's parallel-merge correction `delta² * n_a * n_b / new_count` is the m2 update for tracking `Σ(x − M)²`, which is a different quantity from `Σx²`. Compare `LayerNormValue` further down the same file: it has the full `{count, mean, m2}` triple and uses Welford correctly.

The merge of two partial sums-of-squares is therefore plain addition. No correction term should be there at all.

The original code happened to land on the right answer because `nb_over_n = b.count / (a.count + b.count)` is uint integer division and floors to 0 whenever `a.count >= 1` (and when `a.count == 0` the `a.count` factor zeros the term anyway). So the correction was a no-op in exact arithmetic.

The NaN appears because evaluation order is `((delta * delta) * a.count) * nb_over_n` — the F32 squaring happens *before* the multiply by zero. With BitNet magnitudes (~7e9, ~1728 elements per thread) the per-thread `Σx²` is ~8.5e22, so `delta = b.mean - a.mean` is ~1e22, `delta * delta` is ~1e44 → overflows F32 to `+Inf`, and `Inf * 0 = NaN` poisons the reduction.

## Fix

Drop the correction entirely and merge partials by `(sum_sq, count)` addition. Also rename `RMS::mean` → `RMS::sum_sq` so the field name reflects what it actually stores, and tighten the surrounding comments to call out that Welford doesn't apply here.

## Test plan

- [x] Added `rms_norm_large_magnitude` regression in `candle-nn/tests/ops.rs` at realistic BitNet FFN magnitudes. Fails on the old kernel on Metal, passes with the fix on both CPU and Metal.
- [x] Existing `candle-nn` test suite passes on Metal + CPU.
- [x] Downstream BitNet b1.58-2B-4T checkpoint now produces coherent text on Metal (was `"!!!!!!!!!!"` pre-fix).

Reproducing pre-fix on `main`:
```
cd candle-nn && cargo test --features metal rms_norm_large_magnitude
```

Diff: 1 kernel template + 1 regression test + rename pass.


ivarflakstad (commented):
Thanks a lot!

I don't think these comments add anything here. If you remove them I'll merge 👍 

Sorry for the wait, I wanted to see if an proper online implementation (which this never was) would be better, but I have too many things to address so can't prioritize it.

tomsanbear:
> Thanks a lot!
> 
> I don't think these comments add anything here. If you remove them I'll merge 👍
> 
> Sorry for the wait, I wanted to see if an proper online implementation (which this never was) would be better, but I have too many things to address so can't prioritize it.

All fixed, will keep this in mind for the other (and future) prs 😄

ivarflakstad (approved):
👍

PR #3478, opened 2026-04-20
metal: add copy2d kernels for I16 / I32

tomsanbear:
## Summary

The copy2d dispatcher in `candle-core/src/metal_backend/mod.rs` bails with `Metal copy2d DTYPE not implemented` for `DType::I16` and `DType::I32`. The underlying kernel template is dtype-agnostic, so instantiating two more specialisations covers both — no separate kernel code needed.

## Motivation

[bitnet-rs](https://github.com/tomsanbear/bitnet-rs) fuses q/k/v projections by concatenating their packed weights along the output axis. The concatenation touches an `I32` row-sum tensor, which triggers `copy2d` on `I32` — a path not previously reachable from any in-tree model.

## Changes

- `candle-metal-kernels/src/metal_src/unary.metal`: add `COPY2D(copy2d_i16, int16_t)` and `COPY2D(copy2d_i32, int32_t)` instantiations.
- `candle-metal-kernels/src/kernels/unary.rs`: expose the two new kernel names.
- `candle-core/src/metal_backend/mod.rs`: route `DType::I16`/`I32` through the new kernels.

## Test plan

- [x] `cargo build -p candle-core -p candle-metal-kernels`
- [x] Exercised via bitnet-rs FFN path (concat on I32 row_sum tensor) — previously panicked, now succeeds.


ivarflakstad (approved):
👌

PR #3479, opened 2026-04-20
metal: route q_seq > 1 through full SDPA kernel, not vector

tomsanbear:
## Summary

The vector kernel `sdpa_vector` in `scaled_dot_product_attention.metal` dispatches one threadgroup per `(batch, qhead)` pair and writes a single output position per head — `head_idx * D + ...` has no q-axis component and `tid.y` only indexes heads.

When the dispatcher routed `q_seq ∈ [2, 8]` to this kernel, positions `1..q_seq-1` of the output buffer were left uninitialised. Metal buffer pools reuse allocations, so callers saw stale data as logits. Broke speculative-decoding verify batches (`q_seq = K+1`) and in theory any short prefill that reads per-position outputs.

## Fix

Tighten `supports_sdpa_vector` to `q_seq == 1` and let `q_seq > 1` fall through to the full kernel.

## Test plan

- [x] Adds two regression guards in `candle-nn/tests/sdpa.rs` covering `q_seq ∈ [2, 8]` both with and without `do_causal`.
- [x] Before the fix the new tests report relative error 2.7×–411× vs a manual matmul→softmax→matmul reference.
- [x] After the fix all 7 sdpa tests pass.

## Note on performance

Routing `q_seq ∈ [2, 8]` through the full (BQ=32) kernel is correct but leaves ~75% of the Q threads padded out. A follow-up PR adds BQ=8 tile variants to close that gap — correctness-first landing here.


ivarflakstad (commented):
I like the fix (haven't verified it but seems reasonable 👍 ). 
Let's get rid of the added doc and test though.
If we had docs and tests like this for every minor fix/improvement the project would be unreadable.

tomsanbear:
> I like the fix (haven't verified it but seems reasonable 👍 ). Let's get rid of the added doc and test though. If we had docs and tests like this for every minor fix/improvement the project would be unreadable.

Thanks, updated!

ivarflakstad (approved):
Thanks! 👍

PR #3481, opened 2026-04-20
fix clippy lints surfaced by Rust 1.95

tomsanbear:
## Summary

Stable rustc 1.95.0 ships four new clippy lints that fire on pre-existing code in the workspace and block the CI clippy job (`cargo clippy --workspace --tests --examples --benches -- -D warnings`). This PR fixes each one in-place; no functional changes.

## Changes

| File | Lint | Fix |
|------|------|-----|
| `candle-core/src/npy.rs` | `collapsible_match` | Rewrite `',' => { if cnt == 0 { ... } }` as match guard `',' if cnt == 0 => { ... }` |
| `candle-transformers/src/models/fastvit.rs` | `manual_checked_division` | `in_channels.checked_div(group_size).unwrap_or(1)` |
| `candle-transformers/src/models/stella_en_v5.rs` | `manual_checked_division` | `num_heads.checked_div(num_kv_heads).unwrap_or(0)` |
| `candle-transformers/src/models/paddleocr_vl/mod.rs` (×5) | `explicit_counter_loop` | `for (seqlen_offset, _) in (start..).zip(1..N)` — loop-var pattern, manual `+= 1` dropped |
| `candle-examples/examples/quantized-lfm2/main.rs` | `explicit_counter_loop` | Same zip pattern with `index_pos` |

### Correctness notes

- **`manual_checked_division`**: semantically identical for `usize` by construction — `checked_div(d)` returns `None` iff `d == 0`, matching the original `if d == 0 { default }` branch exactly.
- **`explicit_counter_loop`**: mechanical transform. The `RangeFrom::next()` on `(start..)` yields `start, start+1, start+2, ...` via the iterator's internal `+= 1`, so each loop iteration observes the same `offset` values as before. `break` exits at the same point. `seqlen_offset` remains a CPU `usize` — no device-side changes.

### Validation

Automated tests don't exercise these code paths, so correctness of the counter-loop refactors was verified by running the example binaries before and after the refactor with fixed seeds / inputs:

- **quantized-lfm2**: `--which lfm2-350m-q4_k_m --prompt "The quick brown fox" --sample-len 40 --temperature 0 --seed 42` — 39 tokens generated, byte-identical output.
- **paddleocr-vl**: `--image test_ocr.png --task ocr --max-length 128` (single-image path, hits the first of the five loops) — 97 tokens, byte-identical OCR output.

[...]

PR #3483, opened 2026-04-20
Fix clippy warnings for Rust 1.95.

humesze-llc:
## Summary

Fixes clippy warnings introduced by Rust 1.95 (released 2026-04-14). Restores `cargo clippy --workspace --tests --examples --benches -- -D warnings` to a clean state, which is currently the CI clippy command in `.github/workflows/rust-ci.yml`.

The `-D warnings` CI job aborts at the first failing crate (`candle-core`), which was hiding 8 additional warnings in `candle-transformers` and `candle-examples`. All 9 sites are addressed here.

### `clippy::collapsible_match` (1 site)
- `candle-core/src/npy.rs:121`: inner `if` inside a match arm folded into a match guard. Behavior is identical because the original `',' => { if cnt_parenthesis == 0 { ... } }` already fell through to the following iteration when the guard was false.

### `clippy::manual_checked_ops` (2 sites)
- `candle-transformers/src/models/fastvit.rs:200`: `if group_size == 0 { 1 } else { in_channels / group_size }` becomes `in_channels.checked_div(group_size).unwrap_or(1)`.
- `candle-transformers/src/models/stella_en_v5.rs:291`: `if num_kv_heads > 0 { num_heads / num_kv_heads } else { 0 }` becomes `num_heads.checked_div(num_kv_heads).unwrap_or(0)`. Note the fallback is `0`, not `1`, matching the original else branch.

### `clippy::explicit_counter_loop` (6 sites)
- `candle-transformers/src/models/paddleocr_vl/mod.rs:541, 603, 668, 874, 1050`
- `candle-examples/examples/quantized-lfm2/main.rs:300`

Manual `seqlen_offset += 1` and `index_pos += 1` counters replaced with `(start..).zip(range)` iteration.

### Note on paddleocr_vl: deviating from clippy's suggestion

Clippy suggests `for (seqlen_offset, _) in (current_ids.dim(1)?..).zip(1..max_new_tokens)` at the loop site. For four of the five paddleocr_vl sites (lines 541, 603, 668, 874), this suggestion is semantically incorrect because `current_ids` is reassigned on the line immediately before the loop:

```rust
let mut seqlen_offset = current_ids.dim(1)?;            // captures prompt length
current_ids = Tensor::new(&[next_token], ...)?.unsqueeze(0)?;  // reassigns to 1-token tensor
for _ in 1..max_new_tokens { ... }
```

Applying clippy's inline suggestion would evaluate `current_ids.dim(1)?` after the reassignment, producing `1` instead of the prompt length. To preserve semantics, these four sites bind the original dim before the reassignment:

```rust
let initial_offset = current_ids.dim(1)?;
current_ids = Tensor::new(&[next_token], ...)?.unsqueeze(0)?;
for (seqlen_offset, _) in (initial_offset..).zip(1..max_new_tokens) { ... }
```

The fifth site (line 1050) uses `input_ids` for the range, which is not reassigned before the loop, so clippy's suggestion is applied verbatim.

## Environment

- OS: Ubuntu 22.04 on WSL2, x86_64
- `rustc 1.95.0 (59807616e 2026-04-14)`
- `clippy 0.1.95 (59807616e1 2026-04-14)`

## Verification

Locally on the environment above, all of the following pass cleanly:

- `cargo test --workspace`

## Test plan

- [x] Clippy clean under CI command (`-D warnings`, workspace, tests, examples, benches)
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo test --workspace` passes (no regressions)
- [ ] Runtime inference through the modified paddleocr_vl generation loops and the quantized-lfm2 example was not exercised (neither has coverage in the existing test suite, and running them requires downloading model weights). The refactor preserves iteration count and per-iteration counter values by construction, but a runtime sanity check from a reviewer with access to the models would be welcome.
- [ ] Only tested on Linux x86_64. CI covers Ubuntu, Windows, macOS, and Ubuntu ARM; the changes are pure Rust with no platform-specific code, so all should pass, but I have not verified cross-platform locally.

### DIFF
--- a/candle-core/src/npy.rs
+++ b/candle-core/src/npy.rs
@@ -117,11 +117,9 @@ impl Header {
             match c {
                 '(' => cnt_parenthesis += 1,
                 ')' => cnt_parenthesis -= 1,
-                ',' => {
-                    if cnt_parenthesis == 0 {
-                        parts.push(header[start_index..index].to_owned());
-                        start_index = index + 1;
-                    }
+                ',' if cnt_parenthesis == 0 => {
+                    parts.push(header[start_index..index].to_owned());
+                    start_index = index + 1;
                 }
                 _ => {}
             }
--- a/candle-examples/examples/quantized-lfm2/main.rs
+++ b/candle-examples/examples/quantized-lfm2/main.rs
@@ -285,7 +285,6 @@ fn main() -> Result<()> {
         next_token
     };
 
-    let mut index_pos = tokens.len();
     let prompt_dt = start_prompt_processing.elapsed();
 
     all_tokens.push(next_token);
@@ -297,7 +296,7 @@ fn main() -> Result<()> {
     let eos_token = guess_eos_id(tos.tokenizer());
     let mut sampled = 0;
     let start_post_prompt = std::time::Instant::now();
-    for _ in 0..to_sample {
+    for (index_pos, _) in (tokens.len()..).zip(0..to_sample) {
         if let Some(max_ctx) = context_length {
             if index_pos + 1 > max_ctx {
                 println!("\n\ncontext window of {max_ctx} reached, stopping generation");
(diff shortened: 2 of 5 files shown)


throgsoft:
Closing in favor of #3481, same set of lints and same approach.

PR #3484, opened 2026-04-21
Fix causal mask construction for batch size > 1 in Qwen3/GLM4/SmolLM3/Z-Image

agene0001:
## Summary

The `causal_mask` helpers in these models declare the output tensor shape as `(b, 1, tgt, tgt + offset)` but only populate `tgt * (tgt + offset)` mask values. `Tensor::from_slice`'s `ShapeWithOneHole` impl for concrete shapes [silently accepts the mismatch](https://github.com/huggingface/candle/blob/main/candle-core/src/shape.rs#L474-L478) (it ignores the `_el_count` parameter for already-concrete shapes), so the resulting tensor reads **uninitialized storage** for rows `1..b`.

Downstream, those rows frequently evaluate to NaN once added to attention scores, producing NaN/Inf hidden states for every item beyond index 0 in any batched forward pass.

For `b=1` the bug is invisible because only the populated row is ever read, which is why it has gone unnoticed. It reliably reproduces with batched Qwen3 inference — I hit it on `microsoft/harrier-oss-v1-0.6b` (a Qwen3 embedding model) at `batch_size >= 2` on CUDA fp32, which is what motivated this patch.

## Repro (before the fix)

Embedding any batch of ≥2 distinct texts through a Qwen3 model:
- item 0 embedding: correct
- items 1..N: all NaN

Switching to `batch_size = 1` (or calling the forward per-item) produces correct embeddings for the same texts, confirming the issue is specific to the batched mask. Dumping `causal_mask(b=N, ...)` shows valid values only in the first `tgt * (tgt+offset)` slots; the rest is whatever happened to be in allocator-returned memory.

## Fix

Build the mask at `(1, 1, tgt, tgt + offset)` — which matches the `Vec`'s actual length — then `.broadcast_as((b, 1, tgt, tgt + offset))`. Broadcasting is a zero-copy metadata operation, so there is no performance cost, and it eliminates the uninitialized read.

```rust
// Before
Tensor::from_slice(&mask, (b, 1, tgt, tgt + offset), &self.device)?
    .to_dtype(self.dtype)

// After
Tensor::from_slice(&mask, (1, 1, tgt, tgt + offset), &self.device)?
    .broadcast_as((b, 1, tgt, tgt + offset))?
    .to_dtype(self.dtype)
```

Applied to the same pattern in all 8 files that currently carry it:

- `candle-transformers/src/models/qwen3.rs`
- `candle-transformers/src/models/qwen3_moe.rs`
- `candle-transformers/src/models/quantized_qwen3.rs`
- `candle-transformers/src/models/quantized_qwen3_moe.rs`
- `candle-transformers/src/models/glm4_new.rs`
- `candle-transformers/src/models/quantized_glm4.rs`
- `candle-transformers/src/models/smol/smollm3.rs`
- `candle-transformers/src/models/z_image/text_encoder.rs`

## Possible follow-up (not part of this PR)

The underlying footgun is that `ShapeWithOneHole::into_shape` for a fully-concrete shape discards `el_count` without any check that `shape.elem_count() == el_count`. Adding even a debug-only assertion there would have surfaced this class of bug at the `from_slice` call site instead of showing up as silent NaNs layers downstream. Happy to send that as a separate PR if it's something you'd want.

## Disclosure

This patch was generated by Claude (Anthropic) while I was debugging batched-inference NaNs in a downstream embedding pipeline, and pushed as-is once I had verified the fix on my own workload. Flagging this in the spirit of transparency about AI-assisted contributions — happy to adjust anything that doesn't match the project's contribution expectations.

## Test plan

- [x] Patched fork rebuilt and used as a `[patch.crates-io]` override in a downstream app running `microsoft/harrier-oss-v1-0.6b` on CUDA fp32 — previously NaN batches now produce valid L2-normalized embeddings at `batch_size` up to the token budget of a 12 GB GPU.
- [ ] Would be good to add a targeted unit test that calls each patched model's `causal_mask`/forward at `b >= 2` and asserts the output is finite; happy to add one if you'd like it in this PR.


ivarflakstad (approved):
Thanks, nice catch!

And yes the debug assert is a good idea 👍

agene0001:
hi @ivarflakstad would you like me to add the debug assert before the commit is merged?

PR #3487, opened 2026-04-21
Binary broadcast scalar support

ivarflakstad:
Sometimes one (or more) side of a binary op, presented as a strided tensor, is actually a scalar in disguise - typically following a reduce.

The reason for this is that we map the (typically rhs) scalar tensor to the same shape as the (lhs) other tensor by setting the stride to be `lhs.dims().map(0).collect()`.
So given lhs shape `[1, 128, 256]` the rhs stride would be `[0, 0, 0]`. This means that strided indexing will always correctly give you the index `0` - but it's slow. Not mainly because of the index calculation, but because the compiler can't vectorize loads.

Now we can simply check if the tensor `is_scalar_broadcast`, select the appropriate kernel, and to index the scalar we use this marvel:
```cpp
struct scalar_indexer {
    METAL_FUNC uint operator()(
        uint idx,
        constant size_t &num_dims,
        constant size_t *dims,
        constant size_t *strides
    ) {
        return 0;
    }
};
```

I observed around a ~860% benchmark speedup on my machine. Only implemented for metal here, but both cuda and cpu should benefit from the same approach.


candle-core/src/metal_backend/mod.rs
```diff
@@ -1858,11 +1858,24 @@ impl MetalStorage {
             "eq" | "ne" | "le" | "lt" | "ge" | "gt" => DType::U8,
             _ => self.dtype,
         };
+        let lhs_is_scalar = lhs_l.is_scalar_broadcast();
+        let rhs_is_scalar = rhs_l.is_scalar_broadcast();
         let lhs_contiguous = lhs_l.is_contiguous();
         let rhs_contiguous = rhs_l.is_contiguous();
 
+        let kernel = match (lhs_is_scalar, rhs_is_scalar, lhs_contiguous, rhs_contiguous) {
```
EricLBuehler: Once we fix `is_scalar_broadcast`, the dispatch relies on the implicit invariant `is_scalar_broadcast` implies that `!is_contiguous`. Maybe we should note this since we are assuming that?

candle-core/src/metal_backend/mod.rs
```diff
@@ -1858,11 +1858,24 @@ impl MetalStorage {
...
+            (true, false, _, false) => kernel_name(op, &self.dtype, "_rss"),
+            (false, true, true, _) => kernel_name(op, &self.dtype, "_cs"),
+            (false, true, false, _) => kernel_name(op, &self.dtype, "_lss"),
+            (false, false, true, true) => kernel_name(op, &self.dtype, ""),
+            (false, false, true, false) => kernel_name(op, &self.dtype, "_rstrided"),
+            (false, false, false, true) => kernel_name(op, &self.dtype, "_lstrided"),
+            (false, false, false, false) => kernel_name(op, &self.dtype, "_strided"),
+        };
+
         let buffer = if lhs_contiguous && rhs_contiguous {
```
EricLBuehler: Maybe it would be best to treat the kernel name as the source of truth for which ABI it expects? Maybe we can have a small enum alongside the name?

This avoids cases like shapes [] and [1, 1, 1] (and other all-size-1 shapes) where the disconnect between kernel selection and ABI (as `is_scalar_broadcast` will be true for these and cause incorrect ABI selection I believe).

EricLBuehler (changes requested):
👍 One comment with a corner case.

candle-core/src/metal_backend/mod.rs
ivarflakstad: Good catch.

My thinking was that `is_scalar_broadcast` will only be used were appropriate, perhaps only for this specific case.

But we should safeguard against other shapes.
What behaviour would you expect from binary when one shape is `[]`? Probably a noop?
I do agree that `[1, 1, 1]` is a scalar, but it's not a broadcast. Perhaps instead we should have `is_scalar_like()` which is just the union of `is_scalar` and `is_scalar_broadcast`?

EricLBuehler: I think for a `[]` shape we should treat it as a scalar as it does have 1 element. I agree that `is_scalar_like` would be good for clarity & as a general purpose API, but I think that ultimately relying on the kernel name for the correct dispatch would be best as it makes the dispatch system safer in the future.

In addition, the current design has a bug where the kernel name is different from the way it is dispatched (using `lhs_contiguous && rhs_contiguous`). Dispatching based on the name would fix this :)

EricLBuehler (approved):
👍

PR #3490, opened 2026-04-23
fix: make tokenizers dependency optional (left on by default though) in candle-core

butterflysky:
## Problem

candle-core 0.10 added an unconditional (non-wasm) dependency on `tokenizers` with the `onig` feature to support the new `quantized::tokenizer` module (GGUF tokenizer loading). This pulls the Oniguruma C regex library into every consumer of candle-core, even those that only use it for tensor operations or safetensors-based model loading.

Because candle-nn and candle-transformers depend on candle-core with default features, and Cargo unifies features across the dependency graph, downstream consumers have **no way to opt out** — even setting `default-features = false` on their own candle-core dependency doesn't help.

## Fix

Makes the `tokenizers` dependency **opt-out** (default on, so no breaking change):

**candle-core:**
- `tokenizers` is now `optional = true`
- New `tokenizers` feature flag, included in `default`
- `quantized::tokenizer` module gated behind `#[cfg(feature = "tokenizers")]`

**candle-nn / candle-transformers:**
- Forward the `tokenizers` feature through their own feature flags, following the existing pattern used for `cuda`, `metal`, `accelerate`, etc.
- Depend on candle-core (and candle-nn, for transformers) with `default-features = false`, re-enabling defaults via their own `default` feature

**Workspace:**
- Root `Cargo.toml` declares `candle` and `candle-nn` with `default-features = false`
- All other workspace members (examples, book, datasets, pyo3, wasm-tests, tensor-tools) explicitly enable `features = ["default"]`

### For existing consumers

**No change needed.** Default features include `tokenizers`, so the existing API surface is identical.

### To opt out

```toml
candle-core = { version = "0.10", default-features = false }
candle-nn = { version = "0.10", default-features = false }
candle-transformers = { version = "0.10", default-features = false }
```

## Testing

- `cargo check -p candle-transformers` passes with defaults (tokenizers included)
- `cargo check -p candle-transformers --no-default-features` passes without tokenizers
- Downstream project ([memory-mcp](https://github.com/butterflyskies/memory-mcp)) passes all 109 tests with the opt-out, confirmed `onig` is fully eliminated from the dependency tree

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ candle-flash-attn = { path = "./candle-flash-attn", version = "0.10.2" }
 candle-flash-attn-v3 = { path = "./candle-flash-attn-v3", version = "0.10.2" }
 candle-kernels = { path = "./candle-kernels", version = "0.10.2" }
 candle-metal-kernels = { path = "./candle-metal-kernels", version = "0.10.2" }
-candle-nn = { path = "./candle-nn", version = "0.10.2" }
+candle-nn = { path = "./candle-nn", version = "0.10.2", default-features = false }
 candle-onnx = { path = "./candle-onnx", version = "0.10.2" }
 candle-transformers = { path = "./candle-transformers", version = "0.10.2" }
 candle-ug = { path = "./candle-ug", version = "0.10.2" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ ab_glyph = "0.2.23"
 accelerate-src = { version = "0.3.2" }
 anyhow = { version = "1", features = ["backtrace"] }
 byteorder = "1.4.3"
-candle = { path = "./candle-core", package = "candle-core", version = "0.10.2" }
+candle = { path = "./candle-core", package = "candle-core", version = "0.10.2", default-features = false }
 candle-datasets = { path = "./candle-datasets", version = "0.10.2" }
 candle-flash-attn = { path = "./candle-flash-attn", version = "0.10.2" }
 candle-flash-attn-v3 = { path = "./candle-flash-attn-v3", version = "0.10.2" }
--- a/candle-book/Cargo.toml
+++ b/candle-book/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 accelerate-src = { workspace = true, optional = true }
 candle = { workspace = true, features = ["default"] }
 candle-datasets = { workspace = true }
-candle-nn = { workspace = true }
+candle-nn = { workspace = true, features = ["default"] }
 candle-transformers = { workspace = true }
 candle-flash-attn = { workspace = true, optional = true }
 safetensors = { workspace = true }
--- a/candle-book/Cargo.toml
+++ b/candle-book/Cargo.toml
@@ -11,7 +11,7 @@ readme = "README.md"
 
 [dependencies]
 accelerate-src = { workspace = true, optional = true }
-candle = { workspace = true }
+candle = { workspace = true, features = ["default"] }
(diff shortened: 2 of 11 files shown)


ivarflakstad:
Looks sensible to me.

Just curious - is the reasoning that you'd prefer to not depend on oniguruma as it is no longer actively developed / archived?

Or is it bloat in general?

I haven't checked but it could be that tokenizers is open to removing the dependency (I saw it was added for accurate GPT-2 parity, not sure if it is used/needed for more modern models).

butterflysky:
> Just curious - is the reasoning that you'd prefer to not depend on oniguruma as it is no longer actively developed / archived?
> 
> Or is it bloat in general?

Both really - though more the former than the latter. But these aren't hard requirements, just philosophical preferences.

> I haven't checked but it could be that tokenizers is open to removing the dependency (I saw it was added for accurate GPT-2 parity, not sure if it is used/needed for more modern models).

Oh interesting, I haven't explored this either.

Jerboas86:
the `onig` transitive feature flag seems also to hurt Windows support [Tokenizers issue](https://github.com/huggingface/tokenizers/issues/1581)

fbbbkaya:
Hit the same onig_sys error on Windows. It's specific to the x86_64-pc-windows-gnu (MinGW) target — the generated config.h redefines uid_t/gid_t and clashes with MinGW headers. Building on the MSVC toolchain (x86_64-pc-windows-msvc) compiles onig_sys fine, so switching the toolchain is a working workaround.

Worth noting the root isn't tokenizers itself — candle-core pulls tokenizers with features = ["onig"], and Cargo's feature unification means downstream crates can't turn it off. The proper fix probably belongs on the candle side (gating tokenizers/onig behind a feature flag), so this might be better tracked there.

fbbbkaya:
The onig_sys failure here is specific to the x86_64-pc-windows-gnu (MinGW) target: the generated config.h redefines uid_t/gid_t and clashes with MinGW's headers. Building on the MSVC toolchain compiles onig_sys cleanly, so as an immediate workaround:


rustup default stable-x86_64-pc-windows-msvc
cargo clean && cargo build --release
But the real issue isn't tokenizers itself — it's that candle-core pulls tokenizers with features = ["onig"] as a non-optional dependency, and Cargo's feature unification means downstream crates can't turn it off (setting default-features = false on your own tokenizers/candle-core doesn't help). So onig is forced on everyone who uses candle-core, even for tensor-only or safetensors workloads.

candle-core only uses tokenizers in a single module (quantized::tokenizer). Gating it behind a feature fixes this at the root. On candle-core 0.11.0 the change is two edits:

candle-core/Cargo.toml


[target.'cfg(not(target_arch = "wasm32"))'.dependencies.tokenizers]
version = "..."
features = ["onig"]
default-features = false
optional = true            # <-- add

[features]
default = ["tokenizers"]           # <-- keep tokenizers on by default (non-breaking)
tokenizers = ["dep:tokenizers"]    # <-- new feature
candle-core/src/quantized/mod.rs


#[cfg(all(not(target_arch = "wasm32"), feature = "tokenizers"))]
pub mod tokenizer;
I verified locally: with this patch, default builds are unchanged, but a consumer using candle-core = { default-features = false } drops tokenizers → onig → onig_sys from the dependency tree entirely, and candle-core still compiles cleanly without the module. So the proper fix belongs on the candle side (gating tokenizers behind a feature); this issue might be better tracked there.

ivarflakstad:
Ah. I misremembered this was an issue, not a PR.
Sorry for the auto-closing. I think switching to `fancy-regex` solved what you were trying to fix here though 🤞 

Let me know if it's still a problem and I'll reopen.

PR #3493, opened 2026-04-25
quantized: fix UAF in `as_t_slice` when called with `Cow::Owned`

tomsanbear:
## What's broken

`as_t_slice` (`candle-core/src/quantized/mod.rs:43`) takes `Cow<'_, [u8]>` by value and returns a `&[T]` whose elided lifetime is tied to that input. With `Cow::Owned`, the inner `Vec` drops at the helper's exit and the returned slice dangles. Both `pub fn from_data` paths route through it (`QStorage::from_data`, `GgmlDType::from_data`), so they UAF on every backend — CPU's `.to_vec()` and Metal/CUDA's GPU upload both read the dangling slice.

The borrow checker doesn't catch it because `Cow<'a, T>`'s lifetime parameter only describes the borrow inside the `Borrowed` variant; for `Owned` it's unconstrained.

No in-tree caller currently triggers this (the GGUF loaders bypass these APIs entirely), but they're `pub`, so any downstream user reaching for `Cow::Owned` silently corrupts their model weights.

## How I ran into this

While extending Metal kernel test coverage post-#3477, I wrote a CPU-vs-Metal dequant equivalence test that quantized on CPU, copied the packed bytes into a `Vec<u8>`, wrapped them in `Cow::Owned`, and reloaded onto Metal via `QStorage::from_data`. The two dequant outputs disagreed by `max |Δ| = 1.0625`. Tracing showed the GPU buffer held freed memory by the time of the upload — switching to `Cow::Borrowed(&bytes)` made the test pass bit-equivalently. Same shape on the CPU branch's `.to_vec()`.

## Fix

`as_t_slice` now takes `&[u8]`. Both `pub fn from_data` keep their `Cow<'_, [u8]>` parameter — they bind `let data: &[u8] = &data;` and pass the borrow through, which keeps the inner `Vec` alive in the public function's frame for the whole call.

No public signature changes. No allocs added, no memcpys added. `Cow::Borrowed` paths compile identically.

## Tests

Two new tests in `candle-core/tests/quantized_tests.rs`:

1. **`from_data_dequant_matches_canonical_when_caller_passes_cow_owned`** — dispatched via `test_device!` to CPU + CUDA + Metal. Quantizes on CPU, round-trips packed bytes through `QStorage::from_data(Cow::Owned(bytes), device, …)`, asserts dequant matches the canonical CPU dequant. Pre-fix `_cpu` and `_metal` fail with `max |Δ|` ranging from `1.0` to `~5e5` across runs (depends on what the allocator hands the freed page). Post-fix: green on both. CUDA path is fixed by the same change but I couldn't run it locally.

2. **`from_data_with_cow_owned_must_not_read_freed_memory_under_miri`** — `#[cfg(miri)]`-gated. Run with `cargo +nightly miri test -p candle-core --test quantized_tests`. On unpatched candle Miri reports:
   ```
   error: Undefined Behavior: constructing invalid value:
   encountered a dangling reference (use-after-free)
       --> candle-core/src/quantized/mod.rs:348:36
        |
    348 |   Self::Q4_0 => Box::new(as_t_slice::<BlockQ4_0>(data).to_vec()),
        |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
   ```
   With the fix: passes.

`cargo test --features metal -p candle-core` is fully green (294 tests).

## Notes

- A cleaner alternative is to drop `Cow` from `from_data`'s public signature entirely (`fn from_data(data: &[u8], …)`). Same correctness, semver-breaking. Happy to do that as a follow-up if preferred.
- The behavioral test is reliable on macOS Apple Silicon (100% repro across many runs) but could be flaky on other allocators. The Miri test is the deterministic backstop.


ivarflakstad (commented):
Great catch!

If we could make the comments more concise / keep only the what we need that would be excellent 👍

tomsanbear:
> Great catch!
> 
> If we could make the comments more concise / keep only the what we need that would be excellent 👍

Thanks! Updated!

ivarflakstad (approved):
LGTM! 👌

PR #3495, opened 2026-04-25
Uniform blocks

ivarflakstad:
New `StridedBlocks` variant.

`UniformBlocks` represents blocks of memory where the block itself is contiguous, but the striding between blocks is not.
If the blocks were not internally contiguous then it would have been fully strided. If the striding between blocks were contiguous then it would just be one contiguous block of memory.


EricLBuehler (approved):
👍

PR #3503, opened 2026-04-27
Add BF16 CPU matmul support via F32 cast

### LINKED ISSUE #920: falcon example error on m1 macbook pro
# run output
```bash
cargo run --example falcon   --release -- --prompt "how to learn css quickly"
    Finished release [optimized] target(s) in 0.61s
     Running `target/release/examples/falcon --prompt 'how to learn css quickly'`
Running on CPU, to run on GPU, build this example with `--features cuda`
retrieved the files in 1.320542ms
loaded the model in 35.508071542s
starting the inference loop
Error: unsupported dtype BF16 for op matmul
```

# code version
git version 
commit 871efc0307e39441236327643abb397b3ab200d0 (HEAD -> main, origin/main, origin/HEAD)

# os info
sys info as below:
  model：	MacBook Pro
  model id：	MacBookPro17,1
  type：	MYD92CH/A
  cpu：	Apple M1
  kernels：	8
  memory：	8 GB
  firmware：	8422.141.2
  bios：	8422.141.2

# toolchain info
> rust toolchain info:
> rustup show -v
> Default host: aarch64-apple-darwin
> 
> installed targets for active toolchain
> 
> aarch64-apple-darwin
> wasm32-unknown-unknown
> 
> active toolchain
> 
> stable-aarch64-apple-darwin (default)
> rustc 1.70.0 (90c541806 2023-05-31)

ninthhousestudios:
## Summary

- Adds BF16 CPU matmul support by casting operands to F32, running the existing F32 gemm/BLAS path, and casting back.
- Uses `half::slice::HalfFloatSliceExt` for bulk bf16<->f32 conversion (SIMD-optimized in the `half` crate).
- Covers all three CPU backends: default (gemm crate), MKL, and Accelerate.
- Expands the `matmul_bf16` test to run on CPU (previously skipped), with accuracy validation against F32 reference and batched matmul coverage.

## Motivation

BF16 models (e.g. Qwen3-VL) cannot run on CPU because `MatMul::f` rejects `DType::BF16`. The underlying `gemm` crate (v0.19) has no native BF16 support. This has been open since #920.

The cast-to-F32 approach adds negligible overhead — the O(n) conversion is dominated by the O(n³) matmul for any non-trivial matrix. Temporary F32 buffers hold only the matmul operands, not the full model.

## Design decisions

- **`supports_bf16()` unchanged** — kept returning `false` for CPU. That function is used by `bf16_default_to_f32()` and examples as a "bf16 is efficient here" signal. We don't want to change downstream behavior; we just want matmul to *work* when users explicitly request BF16 on CPU.
- **No gemm crate changes** — this is purely within candle-core.
- **No new dependencies** — the `half` crate is already present.

## Test plan

- [x] `matmul_bf16_cpu` — exact 2×2 result, accuracy vs F32 reference on 8×16 × 16×4, batched 2×4×3 × 2×3×5
- [x] Full `candle-core` test suite passes with no regressions
- [ ] MKL/Accelerate paths follow the same pattern but are untested (no MKL/Accelerate on test machine)

Closes #920


ivarflakstad:
Both approaches are acceptable. Claude kinda borked the implementation though.

In the original commit you could have used the slice conversion functions that `half` already provides. Less noise. Probably more efficient.

In the second commit you're referencing your own local fork of `gemm`, making the PR unusable. You would have to get the approach you've implemented locally accepted upstream to the gemm repo - not really sure if it would be.

ninthhousestudios:
Thanks for the review! Good catches on both points.

I've force-pushed a cleaned up version that:

1. **Uses `half::slice::HalfFloatSliceExt`** for bulk bf16↔f32 conversion (`to_f32_vec()` and `convert_from_f32_slice()`) instead of manual element-wise loops. Cleaner and potentially SIMD-optimized.
2. **Dropped the second commit** that referenced the local gemm fork. This PR is now a single commit with the cast-to-F32 approach only — no upstream gemm changes needed.

ivarflakstad:
Thanks. Would be nice if you could apply a bit of the DRY principle (don't repeat yourself) to `MatMul::f`.
I can test the accelerate path if you'd like.

ninthhousestudios:
Thanks for the feedback! I've pushed two new commits:

1. **DRY up BF16 matmul conversion with shared helpers** — Extracts `bf16_to_f32()` and `f32_as_bf16_dst()` helpers so the repeated unsafe bf16<->f32 pointer casting is centralized in one place across the gemm, accelerate, and mkl paths.

2. **Add debug assertions to BF16 helpers and test edge cases** — Adds `debug_assert_eq!(T::DTYPE, DType::BF16)` to both helpers so the type-punning precondition is enforced at the function boundary. Also adds BF16 matmul tests for narrowed views (nonzero start_offset) and transposed (non-contiguous) inputs.

I tested the accelerate path on a MacBook Air — all matmul tests pass with `--features accelerate`.

candle-core/src/cpu_backend/mod.rs
```diff
@@ -1365,6 +1382,71 @@ impl Map2 for MatMul {
         use gemm::{gemm, Parallelism};
 
         match T::DTYPE {
+            DType::BF16 => {
```
ivarflakstad: Wrt DRY I was actually referring to this block.
Other than the casting before and after it's identical to the f32/f16 path.

ninthhousestudios: Good call — I misread which duplication you meant. Replaced the whole BF16 gemm/sgemm block in each backend with a 7-line preamble that converts bf16→f32, recursively calls `self.f::<f32>()` to reuse the existing matmul path, then converts back. Net -103 lines.

PR #3513, opened 2026-05-06
refactor: converge CPU flash attn kernels on shared primitives

DrJesseGlass:
Converges the three CPU flash attention kernels (causal, standard, varlen) onto                                                                                                                           
two shared primitives: a single dot-product trait and a single online-softmax
helper. Reduces duplication and aligns kernel structure across the module.                                                                                                                                
                                                                                                                                                                                                          
## New: candle-nn/src/attention/cpu_flash/dot.rs                                                                                                                                                          
                                                                                                                                                                                                          
`pub trait DotF32 { fn dot_f32(q: &[f32], k: &[Self]) -> f32; }` with impls for                                                                                                                           
f32, f16, bf16, f64. All accumulate in f32; non-f32 dtypes use an 8-wide
unrolled loop with per-element conversion.                                                                                                                                                                
                                                                                                                                                                                                          
### Removed (4 duplicate dot impls)
                                                                                                                                                                                                          
- `dot_f32` in causal.rs and varlen.rs                                                                                                                                                                    
- `dot_qf32_kf16` in varlen.rs (folded into `f16::dot_f32`)
- `vec_dot<T>` in standard.rs                                                                                                                                                                             
                                                                                                                                                                                                          
### Migrated call sites
                                                                                                                                                                                                          
- f32 paths call `f32::dot_f32(q, k)` directly (no precision change).
- Generic paths (causal generic + standard) now dequantize Q to a thread-local
  f32 buffer once per row, then call `T::dot_f32(q_f32, k_row)`. Correctness                                                                                                                              
  improvement for f16/bf16: the old `vec_dot::<T>` accumulated in T (lossy);
  the new path accumulates in f32 throughout.                                                                                                                                                             
                                                                
## standard.rs: lean dispatch                                                                                                                                                                             
                                                                
New `flash_attn_decode_lean` and `flash_attn_prefill_lean` mirror causal's lean                                                                                                                           
structure. When mask is None, max_bias is 0, and softcap is 0, we skip the
per-step mask check and softcap branches entirely. General-purpose decode and                                                                                                                             
prefill paths still handle masks, ALiBi, and softcap as before.
                                                                                                                                                                                                          
## New: candle-nn/src/attention/cpu_flash/online_softmax.rs     
                                                                                                                                                                                                          
One `pub(crate)` `#[inline(always)]` helper. Takes a closure for V                                                                                                                                        
accumulation so each kernel keeps its own loop shape (contiguous vs strided,
f32 native vs `to_f32()` conversion).                                                                                                                                                                     
                                                                                                                                                                                                          
### 13 call sites collapsed
                                                                                                                                                                                                          
- causal.rs: 7 sites (5 f32 paths plus 2 generic paths)         
- standard.rs: 4 sites (decode, decode_lean, prefill, prefill_lean)
- varlen.rs: 2 sites (f32, f16)                                                                                                                                                                           
                                                                                                                                                                                                          
## Dropped: f64 path in qwen3 CPU flash dispatch                                                                                                                                                          
                                                                                                                                                                                                          
qwen3.rs had a `DType::F64 => flash_attn::<f64>(...)` arm as a fallback. No                                                                                                                               
test exercised it and no real model uses f64 weights for attention. Removed
the arm so f64 hits the existing `bail!("Unsupported dtype...")`. Also                                                                                                                                    
removed the `DotF32 for f64` impl since it no longer has a caller.


DrJesseGlass:
CPU flash attention Qwen3-0.6B-Q4_K_M performance

|             | Prompt (t/s) | Generation (t/s) |                                                                                                                                                         
|-------------|--------------|------------------|
| Main (warm) | 42.77 avg    | 26.74 avg        |                                                                                                                                                         
| PR (warm)   | 53.16 avg    | 28.25 avg        |                                                                                                                                                         
| Delta       | +24.3%       | +5.6%            |
 
Improvements may be noise, but plausibly come from compiler wins enabled by the refactor:
- Uniform V (KV-cache value) accumulation. Before, the new-max branch did acc[t] += v_row[t] and the else branch did acc[t] += v_row[t] * w - two different SIMD shapes. The shared helper routes both branches through one closure with weight w (1.0 on new-max), so LLVM sees one canonical inner loop and picks one vectorization strategy. The * 1.0 is free on finite f32 and the compiler folds it.
- Less divergent control flow at the V (KV-cache value) load site. Two for t in 0..d loops became one shared loop body, lowering register pressure and giving the scheduler less to track.

ivarflakstad:
Great work!
I'll merge #3570 asap.
Since removing `#[inline(always)]` made the intrinsic based dot product significantly faster I think we can get rid of `DotF32` etc here, and instead just rely on good old `WithDType` (which implies `crate::cpu::kernels::VecOps`)

DrJesseGlass:
Merged main (incl. #3570 "Improved neon vec dot") and converged the CPU flash-attention kernels onto that shared primitive. ~15% prompt and generation speed improvements from #3570

EricLBuehler (approved):
Great, thank you @DrJesseGlass!

PR #3516, opened 2026-05-06
Simplify vec_dot_f32 in candle-core::cpu (compiler-vectorized over intrinsics)

DrJesseGlass:
Brings the compiler-vectorized f32 dot loop into `candle-core::cpu::vec_dot_f32`, where it can serve conv1d / conv2d / convtranspose2d on the f32 CPU path.

The simple compiler-vectorized form was integrated into the cpu flash attention kernels (#3254, refined in #3513) because it was found faster than the hand-written `CurrentCpu::vec_dot_f32` intrinsics in that context. @ivarflakstad observed the same result more fundamentally https://github.com/huggingface/candle/pull/3254#discussion_r3133873546.

The best design across hardware is not fully settled (he also noted he had a tweaked intrinsics version that was slightly faster, "let's look into it later"). This PR makes the bet that the simpler version is at least as good in the cases that matter today, and trades the cfg-gated intrinsics complexity for portability and less unsafe.

### What changes                                                                                                                                                                                                                                                           
`candle-core/src/cpu/mod.rs::vec_dot_f32` body is replaced. The cfg-gated SIMD intrinsics path (NEON / AVX2 / SIMD128) and the scalar fallback are both replaced by one portable compiler-vectorized 4-wide loop. Same I/O contract: writes the dot product to `*c` (overwriting).
- Drops the unsafe pointer-arithmetic codepath through the `CurrentCpu` trait for f32.                                                                                                                    
- Drops the four cfg gates.                                                                                                                                                                               
- One function, ~17 lines, safe slice access via `from_raw_parts`.
- The intrinsics paths for `vec_dot_f16`, `vec_dot_bf16`, and `vec_sum`, plus the `CurrentCpu` trait itself, are unchanged.                                                                               
                                                                                                                                                                                                          
### Benchmark
Added a conv2d benchmark at `candle-core/benches/benchmarks/conv2d.rs`. Workload is a ResNet-style block: input `(1, 128, 28, 28)`, kernel `(256, 128, 3, 3)`, stride 1, padding 1. This is the codepath  that calls `T::vec_dot` for f32.                                
                                                                                                                                                                                                          
Hardware: MacBook Air, Apple M1 (NEON). N=4 runs for compiler-vectorized, N=2 for intrinsics, alternated to control for thermal state. 
| Version             | Run 1   | Run 2   | Run 3   | Run 4   | Median  |
|---------------------|---------|---------|---------|---------|---------|
| Compiler-vectorized | 3.60 ms | 4.90 ms | 4.13 ms | 4.74 ms | 4.43 ms |
| Intrinsics (main)   | 4.30 ms | 4.89 ms | --      | --      | 4.60 ms |
                                                                                         
Compiler-vectorized is at least as fast as intrinsics on M1 NEON, plausibly slightly faster (~4 percent on median, well within run-to-run noise). The clear win here is maintenance, not raw throughput.  AVX hardware not yet measured. 
               
### Out of scope
`vec_dot_f16`, `vec_dot_bf16`, and `vec_sum` still use the intrinsics path. Those involve narrow-type-to-f32 conversion in the inner loop where the autovectorizer is less reliably effective. Worth revisiting in a follow-up with measurements.

### DIFF
--- a/candle-core/benches/bench_main.rs
+++ b/candle-core/benches/bench_main.rs
@@ -7,6 +7,7 @@ criterion_main!(
     benchmarks::binary::benches,
     benchmarks::broadcast::benches,
     benchmarks::copy::benches,
+    benchmarks::conv2d::benches,
     benchmarks::conv_transpose2d::benches,
     benchmarks::matmul::benches,
     benchmarks::qmatmul::benches,
--- a/candle-core/benches/benchmarks/conv2d.rs
+++ b/candle-core/benches/benchmarks/conv2d.rs
@@ -9,7 +9,7 @@ fn run(x: &Tensor, k: &Tensor, padding: usize, stride: usize, dilation: usize, g
 }
 
 fn run_benchmark(c: &mut Criterion, device: &Device, dtype: DType, name: &str) {
-    let t = Tensor::arange(0.0f32, (1 * 128 * 28 * 28) as f32, device)
+    let t = Tensor::arange(0.0f32, (128 * 28 * 28) as f32, device)
         .unwrap()
         .reshape((1, 128, 28, 28))
         .unwrap()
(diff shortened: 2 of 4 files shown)


ivarflakstad:
Figured out what the issue was. #3570 is several times faster than both main and the autovectorized variant introduced here.

Closing :)

PR #3518, opened 2026-05-07
fix(qwen3_vl): build causal mask for prefill, not single-token decode (#3505)

### LINKED ISSUE #3505: Qwen3 VL forward pass doesn't construct attention mask properly
Current `forward()` implementation in Qwen3 VL skips constructing the attention mask if `seqlen > 1` (typo?).

0xSoftBoi:
## Summary

In `Qwen3VLModel::forward`, the gate that decides whether to build a causal attention mask is inverted:

```rust
let attention_mask = if seqlen <= 1 {
    Some(self.prepare_decoder_attention_mask(...))   // mask built when *not* needed
} else {
    None                                             // mask skipped when needed
};
```

The decoder causal mask is only useful when **more than one token** is being processed at once (prefill / multi-token prompt). For single-token autoregressive decode steps the KV cache already constrains attention and the mask can be skipped. The current code does the opposite.

Every other candle text model (mistral, starcoder2, helium, olmo2, phi3, deepseek2, csm, gemma4, …) follows the canonical pattern, e.g. mistral.rs:

```rust
let attention_mask = if seq_len <= 1 {
    None
} else {
    let mask = self.prepare_decoder_attention_mask(seq_len, seqlen_offset)?;
    Some(mask)
};
```

This PR aligns Qwen3 VL with that shape and adds a comment explaining the invariant so the inversion isn't reintroduced.

## Test
- `cargo build -p candle-transformers` — clean
- `cargo clippy -p candle-transformers --lib -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean

Fixes #3505

### DIFF
--- a/candle-transformers/src/models/qwen3_vl/mod.rs
+++ b/candle-transformers/src/models/qwen3_vl/mod.rs
@@ -67,19 +67,14 @@ impl Qwen3VLModel {
         seqlen_offsets: &[usize],
     ) -> Result<Tensor> {
         let (bs, seqlen) = input_ids.dims2()?;
-        // The causal mask is only needed when processing more than one token at once
-        // (prefill / multi-token prompt). For single-token autoregressive decode steps
-        // we can rely on the KV cache and skip building the mask. The previous logic
-        // had this condition inverted (issue #3505), which built the mask only for
-        // single-token steps and skipped it during prefill — leaving the model with
-        // no attention mask when it actually needed one.
         let attention_mask = if seqlen <= 1 {
             None
         } else {
+            let seqlen_offset = seqlen_offsets.first().copied().unwrap_or(0);
             Some(self.prepare_decoder_attention_mask(
                 bs,
                 seqlen,
-                seqlen_offsets[0],
+                seqlen_offset,
                 self.text.dtype,
                 input_ids.device(),
             )?)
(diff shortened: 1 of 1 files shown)


0xSoftBoi:
Gentle nudge here. This inverts the causal-mask gate in Qwen3VLModel::forward to match the canonical pattern used across the other text models: build the mask for prefill (multi-token) and skip it for single-token decode. The change is small and idiomatic, and build/clippy/fmt are clean. Happy to rebase. Thanks!

0xSoftBoi:
Rebased onto current `main` (66 commits); clean, and `qwen3_vl` was untouched upstream in the meantime, so the change applies as-is. `cargo check -p candle-transformers` passes.

Restating the bug briefly, since the diff is small enough to be easy to misread. The mask gate in `Qwen3VLModel::forward` is inverted:

```rust
let attention_mask = if seqlen <= 1 {
    Some(self.prepare_decoder_attention_mask(...))   // built for single-token decode
} else {
    None                                             // skipped for multi-token prefill
};
```

A causal mask is needed for the multi-token prefill and is meaningless for a single-token decode step, so this is backwards relative to every other text model in the repo. The patch swaps the branches to match.

It also replaces `seqlen_offsets[0]` with `seqlen_offsets.first().copied().unwrap_or(0)`, since the index panics on an empty slice — previously unreachable only because the branch never ran during prefill.

Issue #3515, opened 2026-05-06
candle-flash-attn vendored kernels ~13 months stale; long-context bf16 divergence vs PyTorch flash_attn 2.8.3


toddwbucy:
# Summary

`candle-flash-attn`'s vendored Tri Dao Flash Attention 2 kernels were last bumped in December 2024 (#2688/#2689/#2690). Tri Dao's upstream has since shipped v2.7.0, 2.7.1, 2.7.2, 2.7.4, 2.8.0, 2.8.1, 2.8.2, and 2.8.3. The Python `flash_attn` package, which HuggingFace `transformers` uses, is at v2.8.3.

We've measured a significant correctness divergence between `candle_flash_attn::flash_attn[_windowed]` and Python `flash_attn` 2.8.3 on the same inputs at long context. The divergence is bounded below ~500 tokens and grows with sequence length.

## Repro / measurement

Tested against [Jina V4](https://huggingface.co/jinaai/jina-embeddings-v4) (Qwen2.5-VL base, 36 layers, hidden 2048, 16 heads / 2 KV heads, head_dim 64, sliding_window 32768) on a 1000-sample stratified fixture. bf16 inference on a single A6000 (sm86), comparing pooled final-layer embeddings.

Reference: HuggingFace `transformers.AutoModel` with `attn_implementation='flash_attention_2'` explicitly set, also bf16. So **both sides use FA2; both sides use bf16; the only difference is the FA2 kernel implementation.**

| Token band | n | P95 cosine | mean | min |
|---|---:|---:|---:|---:|
| 0–200 | 830 | **0.9997** | 0.9999 | 0.9986 |
| 200–500 | 65 | **0.9921** | 0.9979 | 0.9851 |
| 500–1000 | 5 | 0.9867 | 0.9918 | 0.9867 |
| 4000–8000 | 1 | 0.9472 | 0.9472 | — |
| 8000–12000 | 20 | 0.8667 | 0.9154 | 0.8588 |
| 12000–16000 | 16 | 0.8265 | 0.8711 | 0.8265 |
| 16000–20000 | 14 | 0.7018 | 0.8328 | 0.7018 |
| 20000–25000 | 27 | 0.7614 | 0.8191 | 0.7591 |
| 25000–32000 | 19 | 0.7200 | 0.7928 | 0.7200 |
| 32000+ | 3 | 0.7422 | 0.7729 | 0.7422 |

[...]

PR #2690, opened 2024-12-30
Flash-Attn upgrade / SoftCap Candle-FlashAttn [3/n]

michaelfeil:
This is a series of PR's to pull in flash-attn-2 kernels and support:
- [x] pull in latest FA2:
   - [x] launch of different grid size for sm86 / sm89
- [x] softcap FA2
- [x] unpadded_lse

Most likely I will implement this as a series of 3 stacked PRs. *This is PR 3/3.*

Closes #2687 
This PR is stacked on #2689 #2690

```
arning: `candle-core` (lib) generated 1 warning
   Compiling candle-flash-attn v0.8.1 (/workspace/model-performance/michaelfeil/candle/candle-flash-attn)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 1m 13s
     Running unittests src/lib.rs (target/debug/deps/candle_flash_attn-58541d909d490f26)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/flash_attn_tests.rs (target/debug/deps/flash_attn_tests-a64401a027b99c85)

running 3 tests
test flash_attn_varlen ... ok
test flash_attn_acausal ... ok
test flash_attn_acausal_softcap ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 36.33s

   Doc-tests candle_flash_attn

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

```

PR #2689, opened 2024-12-30
Flash-Attn upgrade / SoftCap Candle-FlashAttn [2/n]

michaelfeil:
This is a series of PR's to pull in flash-attn-2 kernels and support:
- [x] pull in latest FA2:
   - [x] launch of different grid size for sm86 / sm89
- [x] softcap FA2
- [ ] unpadded_lse

Most likely I will implement this as a series of 3 stacked PRs. *This is PR 2.*

Closes #2687 
This PR is stacked on #2689

```
running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/flash_attn_tests.rs (target/debug/deps/flash_attn_tests-a64401a027b99c85)

running 3 tests
test flash_attn_varlen ... ok
test flash_attn_acausal_softcap ... ok
test flash_attn_acausal ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 34.96s

   Doc-tests candle_flash_attn

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

```

PR #2688, opened 2024-12-29
Flash-Attn upgrade / SoftCap Candle-FlashAttn [1/n]

michaelfeil:
This is a series of PR's to pull in flash-attn-2 kernels and support:
- [x] pull in latest FA2:
   - [x] launch of different grid size for sm86 / sm89
- [ ] softcap FA2
- [ ] unpadded_lse

Most likely I will implement this as a series of 3 stacked PRs. *This is PR 1.*

Closes #2687

PR #3521, opened 2026-05-07
candle-flash-attn: update vendored Tri Dao FA kernels to v2.8.3

toddwbucy:
Closes the kernel-vendoring portion of #3515.

## Summary

Bumps the vendored Flash Attention 2 kernel sources from the
post-Dec-2024 state (~v2.6.x with #2688 / #2689 / #2690 cherry-picks)
to upstream **v2.8.3** (Tri Dao commit 060c918, 2025-08-14).

Single squashed commit. ~700 net lines including 24 new split-KV
kernel files; the bulk is mechanical kernel updates from upstream
v2.8.3.

## What's in the bump

* All `kernels/*.h` and forward sm80 `*.cu` files replaced with
  v2.8.3 sources, path-remapped from upstream
  `csrc/flash_attn/src/` to candle's flat `kernels/`.
* 24 new split-KV forward sm80 kernels added (`flash_fwd_split_hdim*_*_sm80.cu`)
  — {fp16, bf16} × {dense, causal} × {hdim 32, 64, 96, 128, 192, 256}.
* `flash_api.cu` dispatcher mirrors v2.8.3's
  `run_mha_fwd(params, stream, force_split_kernel)` shape: branches
  between `run_mha_fwd_<>` (dense) and `run_mha_fwd_splitkv_dispatch<>`
  based on `params.num_splits` and `force_split_kernel`.
* `extern "C" run_mha` FFI gains four params (`num_splits`,
  `softmax_lseaccum_ptr`, `oaccum_ptr`, `force_split_kernel`).
  `src/ffi.rs` mirrors. Existing call sites pass safe defaults so
  pre-bump behavior is preserved exactly.
* Rust-side ports of upstream's `num_splits_heuristic` (efficiency
  search, ≥85% of peak occupancy, capped at 128) and the
  `set_params_splitkv` accumulator-allocation logic.
* Dropped legacy hdim 160 / 224 / 512 forward kernels — Tri Dao
  removed their launch-template helpers in v2.x and the previously
  vendored .cu files for those head dims no longer compile against
  v2.8.3's launch template.
* CUTLASS pin retained at `7d49e6c7` (v2.8.3's pinned `dc481792`
  is not required for the kernels we compile).

## PyTorch dependency stub

v2.8.3's dropout codepath references `at::PhiloxCudaState` and
`at::cuda::philox::unpack`. Provide a candle-side stub in
`kernels/philox_unpack.cuh` so the path compiles without dragging
the torch C++ runtime in. Tripwire at the dispatch site asserts
`params.p_dropout == 1.f` to prevent any future caller from
silently triggering the (broken-by-stub) dropout codepath.

## Compile-time fix: extern template declarations

The naive splitkv-aware `flash_api.cu` produced a runaway nvcc
compile (~30+ minutes, ~17 GB cicc RSS) during development. Root
cause: `flash_fwd_launch_template.h` defines the primary template
`run_mha_fwd_splitkv_dispatch<>`, whose body has 7 nested
`*_SWITCH` chains (128 kernel specialisations per
(dtype, hdim, causal) tuple, plus a 14-way combine-kernel chain).
Implicitly instantiating all 24 tuples in the dispatcher TU
expanded ~3,400 kernel specialisations through CUTLASS GEMM
templates in a single nvcc invocation.

Fixed by adding 24 `extern template` declarations in `flash_api.cu`
after the include. nvcc no longer instantiates the splitkv
specialisations in this TU; linker resolves to the explicit
instantiations in the per-hdim `flash_fwd_split_hdim*_*_sm80.cu`
files (which compile in parallel as 24 ~30s TUs). The dense-path
counterpart `run_mha_fwd_<>` was already forward-declared in
`flash.h` without a primary template definition, so it never had
this problem.

## Testing

cuda:0, sm_86, RTX A6000:

```text
$ cargo build           # 4m36s clean (49 kernels)
$ cargo test
  flash_attn_acausal          ok
  flash_attn_acausal_softcap  ok
  flash_attn_varlen           ok
  flash_attn_acausal_splitkv  ok   # NEW
```

The new `flash_attn_acausal_splitkv` test uses a shape
(batch=1, heads=2, seqlen_q=8, seqlen_k=512, head_dim=64) chosen
so `num_splits_heuristic` returns ≥ 2 on any modern sm80+ GPU,
and asserts max-abs < 5e-3 against an fp32 attention reference.

## Public Rust API

`flash_attn`, `flash_attn_windowed`, `flash_attn_alibi`,
`flash_attn_varlen`, etc. — unchanged. New FFI params are internal.

## Note on framing

Originally filed under the hypothesis "FA kernels are stale,
bumping fixes long-context divergence." The bump is structurally
fine and worth landing on its own merits (newer kernels + splitkv
support + no regressions), but it is **correctness-neutral on
long-context** by itself. The actual long-context fix turned out
to be in `candle-transformers/src/models/qwen2.rs::RotaryEmbedding::new`
(filed as a separate PR, link to follow if maintainers want to
review them together). See the [diagnostic update on #3515](https://github.com/huggingface/candle/issues/3515#issuecomment-4399608908)
for the full chain.

This PR is the "newer kernels + splitkv" half of the original ask.


toddwbucy:
Cross-link: #3520 is the actual long-context correctness fix (5-line dtype change in qwen2.rs RoPE). Originally framed in #3515 as one issue; the diagnostic chain split them into two independent contributions. Either review order works.

PR #3520, opened 2026-05-07
qwen2: build RoPE cos/sin tables in fp32 (fixes long-context divergence)

toddwbucy:
## Summary

`RotaryEmbedding::new` in `candle-transformers/src/models/qwen2.rs`
casts `inv_freq` and the position vector `t` to `dtype` (typically
bf16) *before* the matmul, then computes `.sin()` / `.cos()` on the
bf16 result. At long-context positions, bf16's 7-bit mantissa
cannot represent integer positions above ~256 exactly, so the
trig outputs become noise rather than trigonometric values.

Concretely: position 15962 rounds to neighbouring bf16 values 15936
or 15968 (16-step granularity at this magnitude). For low-frequency
dims where `inv_freq[0] = 1.0`, the angle becomes the rounded
position, and `cos(15962) ≈ -0.547` vs `cos(15968) ≈ -0.882` are
essentially uncorrelated.

## Fix

Keep `inv_freq` and `t` in fp32, do the matmul + trig in fp32,
cast to `dtype` only at the output. This mirrors:

- HuggingFace `transformers`' `Qwen2_5_VLRotaryEmbedding.forward`
  (uses `torch.autocast(enabled=False)` + `.float()` to force fp32);
- candle's own `candle-transformers/src/models/qwen3_vl/text.rs::RotaryEmbedding::new`
  (already does this correctly).

5-line diff.

## Evidence

Discovered via layer-by-layer bisection on Jina Embeddings V4
(Qwen2.5-VL backbone). Using a 1000-sample stratified cosine
harness against `transformers` + `flash_attn 2.8.3` on cuda:0 bf16:

| Stratum (~tokens) | P95 before | P95 after |
|---|---:|---:|
| retrieval.query (~50–500B) | 0.9996 | 0.9996 |
| retrieval.passage (~200–2kB) | 0.9970 | 0.9997 |
| text-matching | 0.9995 | 0.9995 |
| code | 0.9959 | 0.9990 |
| matryoshka | 0.9995 | 0.9996 |
| **long-context (~16k tokens)** | **0.7705** | **0.9999** |

Long-context min cosine: 0.7032 → 0.9994. Per-position max-abs
diff at post-RoPE Q on a 15962-token input: 38.0 → 0.25.

## Refs

- Closes the long-context divergence portion of #3515 (the FA bump
  there is structurally fine but is correctness-neutral on
  long-context — see [diagnostic update](https://github.com/huggingface/candle/issues/3515#issuecomment-4399608908)).
[...]

PR #3525, opened 2026-05-08
candle-core: add Device::enable_peer_access for cross-CUDA P2P transfers

### LINKED ISSUE #3524: Cross-CUDA Tensor::to_device fails with CUDA_ERROR_INVALID_CONTEXT (no peer-access enable)
## Bug

`Tensor::to_device(&Device::new_cuda(M))` from a tensor on `Device::new_cuda(N)` (where `N != M`) fails with `CUDA_ERROR_INVALID_CONTEXT` on first use.

## Reproduction

```rust

fn main() -> anyhow::Result<()> {
    let dev0 = Device::new_cuda(0)?;
    let dev1 = Device::new_cuda(1)?;
    let t = Tensor::randn(0f32, 1., (1024, 1024), &dev0)?;
    // ↓ errors with CUDA_ERROR_INVALID_CONTEXT
    let _t_on_dev1 = t.to_device(&dev1)?;
    Ok(())
}
```

Tested on 2× A6000 with NVLink (also reproduces over PCIe-only). cudarc 0.19.4, CUDA 13. The error path is deterministic — happens on the very first cross-card transfer in a fresh process.

## Root cause

`Tensor::to_device` → `CudaStorage::transfer_to_device` (`candle-core/src/cuda_backend/mod.rs:1261`) → `cudarc::CudaStream::clone_dtod`. cudarc's `memcpy_dtod` correctly detects that source and destination contexts differ and dispatches to `result::memcpy_peer_async`:

```rust
// cudarc-0.19.4 driver/safe/core.rs:1591
pub fn memcpy_dtod<T, ...>(...) -> Result<...> {
    self.ctx.bind_to_thread()?;
    let src_ctx = src.stream().context();
    let dst_ctx = self.context();
    ...
    if src_ctx == dst_ctx {
        unsafe { result::memcpy_dtod_async(dst, src, num_bytes, self.cu_stream) }
    } else {
        unsafe {
            result::memcpy_peer_async(
                dst_ctx.cu_ctx, dst,
                src_ctx.cu_ctx, src,
                num_bytes, self.cu_stream,
            )
        }
    }
}
```

`cuMemcpyPeerAsync` requires `cuCtxEnablePeerAccess` to have been called between the two contexts beforehand. Candle never calls it, so the driver rejects the very first cross-context transfer with `CUDA_ERROR_INVALID_CONTEXT`.

## Suggested fix

Two possible API shapes:

**(A) Explicit opt-in** — new public method:
```rust
pub fn enable_peer_access(&self, other: &Device) -> Result<()>
```
Idempotent (same-ordinal pairs no-op; `PEER_ACCESS_ALREADY_ENABLED` mapped to `Ok(())`); enables both directions in one call. Operators call it once between every pair they intend to do cross-card transfers across.

**(B) Auto-enable in `BackendDevice::new`** — opportunistically enable peer access between the new `CudaDevice` and all already-constructed `CudaDevice`s. Requires global state tracking in the cuda_backend. Larger surgery, surprising side effects (every `Device::new_cuda(N)` mutates global driver state), but zero caller burden.

I'd lean (A) for its explicitness — peer access has real costs (UVA address-space coordination, no-op on hardware that doesn't support it) and surfacing the opt-in keeps those costs visible. But (B) matches the principle of least surprise more closely; the call should "just work."

## Workaround until fixed

Host-staged transfer via CPU works at worst-case bandwidth:
```rust
let mid = t.to_device(&Device::Cpu)?;
let t_on_dev1 = mid.to_device(&dev1)?;
```

## Adjacent context

PR #1388 (Feb 2024, still open) noted that multi-GPU is "not well covered at the moment" per @LaurentMazare. This is one of the load-bearing primitives that needs to work for any multi-GPU model-parallel or unified-encoder/decoder pattern; happy to send a PR for whichever of (A) or (B) maintainers prefer.

I have a working implementation of option (A) on a fork branch and can open it as a PR right after this issue if (A) is acceptable.

toddwbucy:
Fixes #3524.

Adds explicit \`Device::enable_peer_access(&other)\` so GPU-direct cross-card tensor operations (\`Tensor::to_device(&other_cuda_device)\`) succeed instead of erroring with \`CUDA_ERROR_INVALID_CONTEXT\` on first use.

## What

- \`CudaDevice::enable_peer_access(&self, other: &Self) -> Result<()>\` — calls \`cuCtxEnablePeerAccess\` in both directions (\`self ←→ other\`), bound to the appropriate context for each direction.
- \`Device::enable_peer_access(&self, other: &Self) -> Result<()>\` — public surface, cuda-only (\`#[cfg(feature = "cuda")]\`); errors clearly if either side isn't a CUDA device.

## Idempotent

- Same-ordinal pairs no-op, returning \`Ok(())\` — same-context "peer" access is meaningless.
- Repeat calls between the same context pair are safe — the driver's \`CUDA_ERROR_PEER_ACCESS_ALREADY_ENABLED\` (704) is folded into \`Ok(())\` inside the helper. So callers can put \`enable_peer_access\` on a hot path without worrying about state tracking.

## Why explicit, not auto-enable in \`BackendDevice::new\`

Per #3524 I considered (A) explicit opt-in vs (B) opportunistic auto-enable in \`BackendDevice::new\`. Going with (A) here:

- Peer access is a real driver-state mutation with costs (UVA coordination on platforms that have it, no-op rejection on heterogeneous configs). Surfacing the opt-in keeps those visible.
- (B) requires global state — a registry of all already-constructed \`CudaDevice\`s — which is a much larger surgery and creates surprising side effects (every \`Device::new_cuda(N)\` would mutate driver state for every other extant CUDA device).
- (A) is forward-compatible with (B) — if you later want auto-enable, \`BackendDevice::new\` can just call this same helper.

Happy to switch to (B) instead if maintainers prefer that direction; the underlying \`enable_peer_access_one_way\` helper is the same either way.

## Errors

Returns the underlying \`DriverError\` if the device pair doesn't support peer access (some heterogeneous or IOMMU-isolated configurations) or if either context is in a terminal state. Callers wanting to probe support before attempting can use \`cuDeviceCanAccessPeer\` separately; not wrapped here.

## Validation

- \`cargo build -p candle-core --features cuda\` — clean
- Tested in a downstream project's topology bench (2× A6000 + NVLink, also tested over PCIe with NVLink bridge physically removed) where \`Tensor::to_device(&other_cuda_device)\` previously errored with \`INVALID_CONTEXT\`. With this patch + a one-time \`encoder_device.enable_peer_access(&decoder_device)?\` call before the first transfer, the cross-card path succeeds and routes through NVLink (or PCIe P2P).

## Adjacent

This is the third small fix from our Jina V4 + multi-GPU work. The other two are #3520 (qwen2 RoPE fp32 cos/sin tables) and #3521 (FA v2.8.3 vendored kernel bump). All three were independently discovered while bringing a Jina V4 embedder up under candle on a 2-A6000 rig; happy to provide more context on any of them.


candle-core/src/device.rs
```diff
@@ -255,6 +255,31 @@ impl Device {
...
+    /// Idempotent: same-ordinal pairs and already-enabled pairs both
+    /// return `Ok(())`. See
+    /// [`crate::CudaDevice::enable_peer_access`] for the underlying
+    /// semantics.
+    ///
+    /// Returns an error if either device is not CUDA (e.g. `Cpu` or
+    /// `Metal`), or if the underlying driver call rejects the request
+    /// (peer access unsupported on this hardware pair, etc.).
+    #[cfg(feature = "cuda")]
+    pub fn enable_peer_access(&self, other: &Self) -> Result<()> {
```
EricLBuehler, line 263: This gate (`#[cfg(feature = "cuda")]`) breaks the file's convention where all methods are always exposed. They compile unconditionally and bail at runtime (see the methods on `Device`).

candle-core/src/cuda_backend/device.rs
```diff
@@ -173,6 +192,58 @@ impl CudaDevice {
...
+        // the call is a no-op (already enabled); fold that into Ok(())
+        // so the helper is safely idempotent across repeat calls.
+        self.context.bind_to_thread().w()?;
+        enable_peer_access_one_way(other_ctx)?;
+        other.context.bind_to_thread().w()?;
+        enable_peer_access_one_way(self_ctx)?;
+        // Restore self as current — bind_to_thread above pushed `other`
+        // onto this OS thread; callers that proceed with `self` work
+        // immediately after this method shouldn't have to re-bind.
+        self.context.bind_to_thread().w()?;
```
EricLBuehler, line 345: If we bail out with `enable_peer_access_one_way?`, then the final self restoration does not occur. Can you please verify correctness of this implementation & edge case?

It seems that the happy path is correct, and the only behavioral gap is that on a failure in the second direction the thread is left bound to other instead of self. It's not a soundness bug, and it's more of a robustness/consistency cleanup.

candle-core/src/cuda_backend/device.rs
```diff
@@ -173,6 +192,58 @@ impl CudaDevice {
...
+    /// driver rejects with `CUDA_ERROR_INVALID_CONTEXT`.
+    ///
+    /// # Errors
+    ///
+    /// Returns the underlying CUDA error if the device pair doesn't
+    /// support peer access (e.g. some heterogeneous or
+    /// IOMMU-isolated configurations) or if either context is in a
+    /// terminal state. Check `cuDeviceCanAccessPeer` separately if you
+    /// need to probe support before attempting to enable it.
+    pub fn enable_peer_access(&self, other: &Self) -> Result<()> {
```
EricLBuehler, line 328: The method mutates the calling thread's current CUDA context (ends bound to self, regardless of what was bound before). The doc comment explains the internal rebind but doesn't warn the caller that their thread's current context changes as a side effect. I think it may be worth one line in the # Errors/notes section, since it's a surprising side effect for a method named `enable_*`.

candle-core/src/cuda_backend/device.rs
```diff
@@ -173,6 +192,58 @@ impl CudaDevice {
...
+        let self_ord = self.context.ordinal();
+        let other_ord = other.context.ordinal();
+        if self_ord == other_ord {
```
EricLBuehler: This compares `ordinal()`, not the context pointer. Two distinct `CudaDevice `instances on the same physical GPU but with different contexts would short-circuit without enabling anything, yet a transfer between them still has `src_ctx != dst_ctx` and would route through memcpy_peer_async.

This is of course a very niche configuration (candle normally has one context per ordinal anyway) and `cuCtxEnablePeerAccess` between two contexts on the same device would error anyway, so the early return is defensible.

However, the doc says "same physical device … no peer to enable," which glosses over it. Perhaps we should note the assumption?

EricLBuehler (changes requested):
Hey @toddwbucy!

Thanks for the PR - it solves a real bug introduced in #3312 where context was not properly configured. In fact, once we get this PR merged, we should enhance the current `to_device` path in candle to try peer-enable (new) + peer copy (as current); on failure, fall back to host staging (the old path). However, that should be a later PR.

This PR looks nice, but I have a few review comments.

One thing I did not mention in the specific comments is the added documentation. The code added is paired with plenty of documentation, but I think that in this case it is a bit too much. If you look at the codebase, you will notice that we have more self-explanatory code with the goal of avoiding such long comment blocks. Of course, this PR does introduce methods that bring with them complex semantics, so documenting that is fine, but can you please correct the comments where they are overly detailed (or forced to wrap at ~80 columns although the codebase does not use that convention) or use special characters?

toddwbucy:
Thanks @EricLBuehler — addressed all of it and rebased onto current main:

- ordinal → context: now compares context pointers, so same-GPU/different-context handles aren't wrongly short-circuited.
- error-path restore: the two-way enable is captured and self is unconditionally rebound afterward, so a failure in either direction can't leave the thread on other (verified all four exit paths).
- side effect: documented under # Side effects (thread ends bound to self).
- cfg gate: dropped; added a runtime-bail stub in the dummy backend so it compiles unconditionally like the other Device methods.
- docs: trimmed throughout, dropped the 80-col wrapping and the unicode arrow.
- Verified with a GPU0/1 P2P round-trip (to_device 0→1→0).

Issue #2977, opened 2025-06-01
No backward pass for `LayerNorm` if tensor is contiguous and remove_mean is true


toolness:
This is similar to #2168, but the root cause appears to be different:

* In #2212, a new `candle_nn::ops::layer_norm` function was added, which uses `apply_op3_no_bwd`.

* In #2213, `LayerNorm`'s forward pass was modified to call `candle_nn::ops::layer_norm` when the tensor is contiguous and `remove_mean` is true.

izderadicka:
I hit this one also -  learning candle on simple transformer example -  it hit me hard -  if probably with get me stopped - if Fable will not help.
I found quite bad -  so significant change of default behavior  of `Module` and not even mentioned in docs.  At least it deserves proper documenting.

Issue #2168, opened 2024-05-07
No backward pass for `RmsNorm` if tensor is contiguous


agerasev:
`RmsNorm` switches to faster implementation if tensor is contiguous:

https://github.com/huggingface/candle/blob/82b641fd2752e3b14db6a9c91faef70e3329f3b5/candle-nn/src/layer_norm.rs#L174-L175

But it does not support backward pass:

https://github.com/huggingface/candle/blob/82b641fd2752e3b14db6a9c91faef70e3329f3b5/candle-nn/src/ops.rs#L640

Maybe it's better to implement `ModuleT` rather than `Module` for `RmsNorm` and use faster implementation only if `train == false`?

computer-whisperer:
Just ran into this, had to make my own copy of LayerNorm as a workaround

toolness:
Same here, I was very confused as to why LayerNorm wasn't working, then I started logging my gradients and noticed that everything before the norm had no gradient.  Very unexpected!

Actually, it looks like the root cause of `LayerNorm`'s problem is different (fixing `RmsNorm` won't automatically fix `LayerNorm`) so I filed #2977 separately.

getupforone:
I also experienced same problem. I spent almost three days to figure out the problem.I just noticed by logging every gradients. The training was not changing weight matrix before layernorm in the Vamap.The root cause of gradient was blocked by layer norm. There were no documentaion about this problem.

nxrobins:
Still reproduces today (2026-05-28) on candle main HEAD \`1d7e9274591ef7eb101d965b71243a41994214d2\`. PR #3526 has been awaiting review since 2026-05-08 with a fix that both removes the contiguous fast-path AND adds a proper \`bwd\` implementation to the \`RmsNorm\` CustomOp2 — including the gradient derivation in the diff. Tested by that PR's author on Qwen2.5-0.5B; independently corroborated by us on the same model.

Minimal standalone reproducer (CPU, <60s, no external assets):
https://github.com/nxrobins/candle-bug-3-rmsnorm-no-bwd

Sibling \`apply_op*_no_bwd\` issues filed this week for the same pattern in other functions:
- #3568 — \`rotary_emb::rope\` / \`rope_i\` / \`rope_thd\`
- #3569 — \`ops::softmax_last_dim\`

Could a maintainer take a look at #3526? It's been ~3 weeks and has a clean fix ready to land.

PCfVW:
[...]

PR #3526, opened 2026-05-08
Fix RmsNorm backward pass for contiguous tensors (issue #2168)

Gravirus:
﻿## Fix RmsNorm backward pass for contiguous tensors (Issue #2168)

### Problem

In candle-nn 0.10.2, RmsNorm::forward has a fast path for contiguous tensors that uses apply_op2_no_bwd, which does not register a backward pass in the autograd graph. Since most tensors in a typical model are contiguous in memory, this means gradients silently stop flowing through RmsNorm during training.

The bug affects all models using RmsNorm from candle-nn, including LLaMA, Mistral, Qwen, Phi, Gemma, and others.

### Root Cause

Two interconnected issues:

1. layer_norm.rs: RmsNorm::forward checks xs.is_contiguous() and routes to ops::rms_norm() for contiguous tensors
2. ops.rs: ops::rms_norm() uses apply_op2_no_bwd and the RmsNorm CustomOp2 has no bwd method

### Fix

Two changes in two files:

**candle-nn/src/layer_norm.rs**: Removed the is_contiguous() fast path. RmsNorm::forward now always delegates to LayerNorm::forward (with remove_mean=false), which uses elementary operations (sqr, sum_keepdim, broadcast_div, broadcast_mul) that automatically build the autograd graph.

**candle-nn/src/ops.rs**: Added a bwd method to the RmsNorm CustomOp2 implementation with proper gradient formulas, and changed apply_op2_no_bwd to apply_op2 so the fast path can be used with gradient tracking in the future.

### Changes

`diff
// layer_norm.rs
impl Module for RmsNorm {
-       if xs.is_contiguous() {
-           crate::ops::rms_norm(xs, &self.0.weight, self.0.eps as f32)
-       } else {
-           self.0.forward(xs)
-       }
+       self.0.forward(xs)
    }
}
`

`diff
// ops.rs
impl candle::CustomOp2 for RmsNorm {
    // ... existing cpu_fwd, cuda_fwd, metal_fwd ...
    
+   fn bwd(
+       &self,
+       arg1: &Tensor,
+       arg2: &Tensor,
+       _res: &Tensor,
+       grad_res: &Tensor,
+   ) -> Result<(Option<Tensor>, Option<Tensor>)> {
+       let n = arg1.dim(D::Minus1)? as f32;
+       let rms = (arg1.sqr()?.sum_keepdim(D::Minus1)? / n + self.eps)?.sqrt()?;
+       let alpha_over_rms = arg2.broadcast_div(&rms)?;
+       let x_dot_grad_alpha = (arg1 * grad_res * arg2)?.sum_keepdim(D::Minus1)?;
+       let correction = arg1 * &x_dot_grad_alpha / (rms.sqr() * n)?;
+       let grad_x = alpha_over_rms.broadcast_mul(&(grad_res - correction)?)?;
+       let grad_alpha = (arg1 * grad_res / rms)?.sum_keepdim(D::Minus1)?;
+       Ok((Some(grad_x), Some(grad_alpha)))
+   }
 }

 pub fn rms_norm(xs: &Tensor, alpha: &Tensor, eps: f32) -> Result<Tensor> {
     // ... validation ...
-    xs.apply_op2_no_bwd(alpha, &RmsNorm { eps })
+    xs.apply_op2(alpha, RmsNorm { eps })
 }
`

### Testing

Tested on a Qwen2.5-0.5B model with DynamicLoRA adapters:

| Metric | Before Fix | After Fix |
|--------|-----------|-----------|
| Variables with gradients | 0/6 | 6/6 |
| Training loss (epoch 0) | 21.23 | 21.23 |
| Training loss (epoch 39) | N/A (no learning) | 2.78 |
| Generation quality | Random tokens | Coherent output |

### Related Issues

- Issue #2168: No backward pass for RmsNorm if tensor is contiguous (OPEN since May 2024)
- Issue #2977: No backward pass for LayerNorm if tensor is contiguous
- Issue #3011: LayerNorm Gradient Flow Issue in candle-nn

### DIFF
--- a/candle-nn/src/layer_norm.rs
+++ b/candle-nn/src/layer_norm.rs
@@ -201,11 +201,7 @@ impl RmsNorm {
 
 impl Module for RmsNorm {
     fn forward(&self, xs: &Tensor) -> Result<Tensor> {
-        if xs.is_contiguous() {
-            crate::ops::rms_norm(xs, &self.0.weight, self.0.eps as f32)
-        } else {
-            self.0.forward(xs)
-        }
+        self.0.forward(xs)
     }
 }
 
--- a/candle-nn/src/ops.rs
+++ b/candle-nn/src/ops.rs
@@ -617,6 +617,45 @@ impl candle::CustomOp2 for RmsNorm {
         let newstorage = candle::MetalStorage::new(output, device.clone(), elem_count, s1.dtype());
         Ok((newstorage, l1.shape().clone()))
     }
+
+    fn bwd(
+        &self,
+        arg1: &Tensor,
+        arg2: &Tensor,
+        _res: &Tensor,
+        grad_res: &Tensor,
+    ) -> Result<(Option<Tensor>, Option<Tensor>)> {
+        // y = x * alpha / rms, where rms = sqrt(sum(x^2) / n + eps)
+        // n = last dimension size
+        let n = arg1.dim(D::Minus1)? as f64;
+        let n_inv = 1.0 / n;
+        // rms = sqrt(sum(x^2) / n + eps)
+        let sum_x2 = arg1.sqr()?.sum_keepdim(D::Minus1)?;
+        let rms = ((sum_x2 * n_inv)? + self.eps as f64)?.sqrt()?;
+        // rms^2 = sum(x^2) / n + eps
+        let rms_sq = rms.sqr()?;
+
+        // grad_x:
+        // dy_i/dx_j = alpha_i / rms      if i == j
+        //           = -x_j * alpha_i * x_i / (rms^3 * n)   if i != j (through rms)
+        //
+        // Simplification:
+        // grad_x_i = (alpha_i / rms) * grad_i - (x_i * alpha_i) / (rms^3 * n) * sum_j(x_j * grad_j)
+        //
+        // More compactly:
+        // grad_x = alpha / rms * (grad - x * sum(x * grad * alpha) / (rms^2 * n))
+        let alpha_over_rms = arg2.broadcast_div(&rms)?;
+        let x_dot_grad_alpha = (arg1 * grad_res * arg2)?.sum_keepdim(D::Minus1)?;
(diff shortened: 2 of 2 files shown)


nxrobins:
Independent corroboration: this bug also surfaced during Phase 0.5b validation of an external project against a Python reference implementation, on the same model you tested (Qwen2.5-0.5B). Confirming both the symptom and the fix.

**Reproducer (standalone, no external assets, <60s):** https://github.com/nxrobins/candle-bug-3-rmsnorm-no-bwd

\`\`\`bash
git clone https://github.com/nxrobins/candle-bug-3-rmsnorm-no-bwd && cd candle-bug-3-rmsnorm-no-bwd && cargo run --release
# exit 1 at candle main HEAD 1d7e9274: BUG PRESENT
# exit 0: bug fixed
\`\`\`

The reproducer pins to commit \`7c7a8c570e\` (the 0.10.2 release commit), builds a tiny \`RmsNorm\` via the standard \`candle_nn::rms_norm\` constructor, passes a contiguous \`Var\` through \`RmsNorm::forward\`, and asserts that \`grads.get(var)\` returns \`None\`. Positive control: \`forward_diff\` propagates gradient (\`max(|g|) = 0.24\` on this fixture).

This pattern is also present in two sibling functions that PR #3526 doesn't address:
- \`candle-nn/src/rotary_emb.rs:537\` (\`rope\` and friends) — filed as #3568
- \`candle-nn/src/ops.rs:430\` (\`softmax_last_dim\`) — filed as #3569

Both filed independently with the same \`bwd\`-on-CustomOp shape as this PR as the suggested supremum fix; they reference this PR as precedent for the approach. If the maintainers prefer one bundled PR addressing all four (RmsNorm + rope + softmax + an open one for Qwen2 causal mask, #3567), happy to help coordinate.

Full validation context with all four bugs that surfaced together: https://github.com/nxrobins/wave-agent/blob/main/docs/upstream-candle-bugs.md

PR #3532, opened 2026-05-14
[Metal] Improved inter-encoder sync and gemv

ivarflakstad:
Changes include
* Use mutex guard wrapper instead of semaphore the guard access to compute/blit encoders.
* Switch to untracked mlt buffers and manual synchronization between encoders.
* Track dependencies between commands on buffers using `Input` and `Output`, which we can use to only wait for the minimum amount of required fences. If two encoders have no depencies between them they run in parallel. This is inter-encoder parallellism, while #3511 addressed intra-encoder parallelism.
* Add residency set for buffers, so that we no longer have to remember `useResource` at (for example) kernel call sites.
* Add mlx's `gemv` and `gemv_t` kernels.

With this and the previous PR I'm seing ~2-2.5x faster tok/s for several non-quantized models, while quantized models get ~20% boost to tok/s.

| model  | old | new |
|-------|-----|------------|
| Llama-3.2-1B | 55 tok/s  | 131 tok/s   |
| Qwen3-0.6B-Q4_K_M    | 170.5 tok/s  | 211.96 tok/s |

### DIFF
--- a/candle-core/src/metal_backend/device.rs
+++ b/candle-core/src/metal_backend/device.rs
@@ -172,8 +172,6 @@ impl MetalDevice {
             .wait_until_completed()
             .map_err(MetalError::from)?;
 
-        // Drop unused buffers only after all in-flight CBs are confirmed complete.
-        // With commandBufferWithUnretainedReferences, buffers must stay alive until GPU is done.
         self.drop_unused_buffers()?;
         Ok(())
     }
--- a/candle-core/src/metal_backend/device.rs
+++ b/candle-core/src/metal_backend/device.rs
@@ -4,14 +4,13 @@ use crate::{DType, Result};
 use candle_metal_kernels::metal::ComputePipeline;
 use candle_metal_kernels::{
     metal::{
-        BlitCommandEncoder, Buffer, BufferMap, Commands, ComputeCommandEncoder, Device,
-        MTLResourceOptions,
+        BlitCommandsGuard, Buffer, BufferMap, Commands, CommandsGuard, Device, MTLResourceOptions,
+        ResidencySet,
     },
     Kernels,
 };
 use objc2_foundation::NSURL;
 use objc2_metal::{MTLCaptureDescriptor, MTLCaptureDestination, MTLCaptureManager};
-
 use std::path::Path;
 use std::sync::{Arc, Mutex, RwLock};
 
@@ -39,7 +38,7 @@ pub struct MetalDevice {
     /// Raw metal device: <https://developer.apple.com/documentation/metal/mtldevice?language=objc>
     pub(crate) device: Device,
 
-    pub(crate) commands: Arc<RwLock<Commands>>,
+    pub(crate) commands: Arc<Commands>,
 
     /// Simple allocator struct.
     /// The buffers are stored in size buckets since ML tends to use similar shapes over and over.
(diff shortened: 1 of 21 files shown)


candle-core/src/quantized/metal.rs
```diff
@@ -331,17 +333,18 @@ impl QMetalStorage {
...
 
     pub fn data(&self) -> Result<Vec<u8>> {
         let buffer = self.device.allocate_buffer(self.buffer.length())?;
         {
-            let blit = self.device.blit_command_encoder()?;
+            let mut blit = self.device.blit_command_encoder()?;
             blit.set_label("blit_to_cpu");
             blit.copy_from_buffer(&self.buffer, 0, &buffer, 0, self.buffer.length());
-            blit.end_encoding();
+            blit.as_ref().end_encoding();
```
EricLBuehler: It seems that `end_encoding` may be called twice here, once manually, and once via drop? It seems that `BlitCommandsGuard` doens't check if encoding was ended already.

candle-metal-kernels/src/metal/commands.rs
```diff
@@ -299,11 +310,67 @@ impl Commands {
...
+        };
+
+        let mut prev_ce_outputs = self.prev_ce_outputs.lock().unwrap();
+
+        // Wait for any prior encoder that wrote to buffers we're reading (RAW protection).
+        let mut waiting_on = HashMap::new();
+        for input in all_inputs.iter() {
+            if let Some(fence) = prev_ce_outputs.get(input) {
+                if !waiting_on.contains_key(input) {
+                    encoder.wait_for_fence(fence);
```
EricLBuehler: I think these calls to `wait_for_fence` are dead as they are at the end of the encoder after everything has been encoded, since [`waitForFence`](https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/waitforfence(_:)) only blocks subsequent commands and previously encoded dispatches are not affected.

candle-metal-kernels/src/kernels/sort.rs
```diff
@@ -92,7 +92,7 @@ fn multi_block_sort(
                 /* stride_sorted_axis */ 1i32,
                 /* nc_dim */ 1i32,
                 /* nc_shape */ nrows as i32,
-                /* nc_str */ ncols as i32
+                /* nc_str */ ncols as i64
```
EricLBuehler, line 95: 👍

candle-core/src/metal_backend/device.rs
```diff
@@ -67,19 +66,21 @@ pub struct MetalDevice {
...
+    pub(crate) residency_set: Arc<ResidencySet>,
 }
 
 // Resource options used for creating buffers. Shared storage mode allows both CPU and GPU to access the buffer.
-pub const RESOURCE_OPTIONS: MTLResourceOptions =
-    objc2_metal::MTLResourceOptions(MTLResourceOptions::StorageModeShared.bits());
-//| MTLResourceOptions::HazardTrackingModeUntracked.bits(),
-//);
-
+pub const RESOURCE_OPTIONS: MTLResourceOptions = objc2_metal::MTLResourceOptions(
```
EricLBuehler, line 74: We now have `RESOURCE_OPTIONS` as defined in candle-metal-kernels which is Shared, and here we are defining it as Shared and Untracked. Is this correct / should we unify it somehow?

candle-metal-kernels/src/metal/residency_set.rs
```diff
@@ -0,0 +1,56 @@
+use super::{Buffer, Device};
+use objc2::{rc::Retained, runtime::ProtocolObject};
+use objc2_metal::{MTLAllocation, MTLDevice as _, MTLResidencySet, MTLResidencySetDescriptor};
+
+/// Keeps Metal buffers resident in GPU memory, removing the need for `useResource` calls.
+///
+/// Add the set to the command queue once via `MTLCommandQueue::addResidencySet`. Then
+/// register every buffer at allocation time and unregister at free time.
+pub struct ResidencySet {
```
EricLBuehler, line 9: Since [`MTLResidencySet`](https://developer.apple.com/documentation/metal/mtlresidencyset) requires macos 15+, I think we are effectively constraining support? This is fine I think, but should we have a fallback?

EricLBuehler (changes requested):

candle-metal-kernels/src/metal/commands.rs
ivarflakstad: I think you're right. Which also means we don't need to extract `all_inputs`.

candle-core/src/metal_backend/device.rs
ivarflakstad: Right. I'll make the one in candle-metal-kernels Shared | Untracked as well.
We should remove it, but currently it is being used to create buffers in `multi_block_sort`. Imo those should be passed in so they are part of the buffer pool.

The only other usages are in tests, so we can just make it pub(crate).
The other alternative of course is to let candle-metal-kernels define the options for candle. Not sure what makes most sense.

candle-metal-kernels/src/metal/residency_set.rs
ivarflakstad: Hmm released september 2024.
Looking at [Supported hardware](https://en.wikipedia.org/wiki/MacOS_Sequoia#Supported_hardware) on wikipedia I think we're pretty safe , same goes for [iOS](https://en.wikipedia.org/wiki/IOS_18#Supported_devices).

EricLBuehler (approved):
🚀 Awesome!

PR #3537, opened 2026-05-16
feat: Add CPU Conv3d support

SorenDreano:
Add a real 5D Conv3d path for Candle instead of relying on the model-specific split-Conv2d workaround used by Qwen3-VL. That workaround only covered the Qwen3-VL temporal-patch case by slicing a depth-2 kernel into two Conv2d calls; it did not provide general Conv3d semantics, arbitrary output depth, bias handling, or a backend Conv3d op. The tensor op uses NCDHW inputs and OI DHW weights, supports per-dimension padding, stride and dilation, and handles grouped convolution. CPU now implements forward plus autograd support for input and weight gradients. candle-nn also gets a Conv3d module; bias is applied with the usual broadcast-add path, so bias gradients come from regular autograd.

The CPU implementation follows the existing Conv2d approach where that makes sense. Forward and weight-gradient use tiled im2col buffers feeding MatMul for f16, f32 and f64, with a 1x1 fast path. bf16 and other dtypes fall back to direct kernels because the generic CPU MatMul path does not support bf16 here. Input-gradient is still direct; optimizing it would need a separate col2im/scatter-add path and is better left for another change.

CUDA and Metal are intentionally left unsupported in this patch. The backend trait has Conv3d hooks, but non-CPU backends return the normal unsupported-op error.
I have access to Nvidia and Apple hardware, so if the maintainers are happy with this MR I can implement CUDA and Metal Conv3d support in a follow-up patch.

Tests cover forward, grouped convolution, gradients, non-contiguous inputs, invalid arguments, and f16/bf16 CPU behavior. The checked-in Rust fixtures follow the existing Conv2d style: each PyTorch reference snippet lives above the Rust test, and the expected values are hard-coded with rounded test_utils assertions.

During development this was also validated with a separate PyTorch comparison harness kept outside this Candle repository, in another repo on my GitHub profile.
https://github.com/SorenDreano/test_conv3d_candle
That harness was used for thorough parity testing: it runs Candle and torch.nn.functional.conv3d on the same deterministic fixtures and compares forward output, input gradient, weight gradient, and bias gradient across a broader set of Conv3d parameter combinations.

Validation run:

- ./test_conv3d_candle/run_conv3d_validation.sh
- cargo test -p candle-core conv3d
- cargo test -p candle-nn conv3d
- cargo fmt --all -- --check
- cargo test -p candle-core
- cargo test -p candle-nn
- cargo clippy -p candle-core -p candle-nn --all-targets -- -D warnings

### DIFF
--- a/candle-core/src/backend.rs
+++ b/candle-core/src/backend.rs
@@ -57,6 +57,36 @@ pub trait BackendStorage: Sized {
         _params: &crate::conv::ParamsConv2D,
     ) -> Result<Self>;
 
+    fn conv3d(
+        &self,
+        _l: &Layout,
+        _kernel: &Self,
+        _kernel_l: &Layout,
+        _params: &crate::conv::ParamsConv3D,
+    ) -> Result<Self> {
+        crate::bail!("conv3d is not implemented for this backend")
+    }
+
+    fn conv3d_backward_input(
+        &self,
+        _l: &Layout,
+        _kernel: &Self,
+        _kernel_l: &Layout,
+        _params: &crate::conv::ParamsConv3D,
+    ) -> Result<Self> {
+        crate::bail!("conv3d backward input is not implemented for this backend")
+    }
+
+    fn conv3d_backward_weight(
+        &self,
+        _l: &Layout,
+        _grad: &Self,
+        _grad_l: &Layout,
+        _params: &crate::conv::ParamsConv3D,
+    ) -> Result<Self> {
+        crate::bail!("conv3d backward weight is not implemented for this backend")
+    }
+
     fn conv_transpose2d(
         &self,
         _l: &Layout,
--- a/candle-core/src/backprop.rs
+++ b/candle-core/src/backprop.rs
@@ -80,6 +80,11 @@ impl Tensor {
                         kernel: rhs,
                         ..
                     }
+                    | Op::Conv3D {
+                        arg: lhs,
+                        kernel: rhs,
+                        ..
+                    }
                     | Op::ConvTranspose2D {
                         arg: lhs,
                         kernel: rhs,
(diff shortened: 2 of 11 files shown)


SorenDreano:
I have rebased my branch.

To demonstrate that this op can replace the model-specific split-Conv2d workaround mentioned in the description, I've put together a branch that ports Qwen3-VL to the generic `Conv3d`: [`conv3d-qwen3vl-demo`](https://github.com/SorenDreano/candle/tree/conv3d-qwen3vl-demo) 

([diff vs main](https://github.com/SorenDreano/candle/compare/conv3d...SorenDreano:candle:conv3d-qwen3vl-demo)).

# Changes
- `PatchEmbed` uses `candle_nn::conv3d` (this branch) instead of the custom `Conv3dNoBias`
- `conv3d_temporal_2.rs` is deleted
- Checkpoint compatibility is preserved: the same `proj.weight` (5D) and `proj.bias` tensors are loaded under the same names.
- The stride now maps directly to the HF PyTorch definition, `Conv3d(kernel_size=(t, p, p), stride=(t, p, p))`, instead of handling the temporal dimension implicitly by slicing.

# Validation
The branch includes a unit test that builds `PatchEmbed` from a `VarBuilder` and checks its output against the previous two-Conv2d decomposition on the same weights: max abs difference is below 1e-5 in f32 (the results are equivalent within float tolerance rather than bit-identical, since the accumulation order differs).

# Tests
- `cargo clippy -p candle-transformers --tests -- -D warnings`
- `cargo test -p candle-transformers qwen3_vl`

# No PR for this change for now
The split-Conv2d workaround currently runs on all backends, while this PR's Conv3d is CPU-only, so swapping it now would regress Qwen3-VL on CUDA/Metal. The intended sequence is: this PR first, CUDA/Metal kernels as the follow-up mentioned above, then the Qwen3-VL swap as a small final PR. Happy to open the swap as a draft PR instead if that's easier to review.

SorenDreano:
Hi! Just checking in on this PR, is there anything blocking it from being reviewed or merged?

I'd be happy to make any changes needed to get it accepted, including adjusting the API/implementation, adding tests or benchmarks, or changing the scope. If CPU-only Conv3d isn't the preferred direction, I'd also appreciate guidance on what you'd like to see before this can move forward.

I have asked around a bit on the dedicated Discord channel but did not any response there.

Regards

PR #3542, opened 2026-05-18
[Metal] add support for debug labels and refactor to forwards compatible buffer builders on metal

tomsanbear:
Opt-in Metal GPU-capture instrumentation behind new, dependency-free, default-off feature flags: `debug-labels` (candle-metal-kernels) and `metal-debug-labels` (candle-core, implies `metal`), with passthrough from candle-nn / candle-transformers / candle-examples so it is reachable when running a real model or example.

Infrastructure: DebugGroupGuard + ComputeCommandEncoder::debug_group (RAII pushDebugGroup/popDebugGroup, CommandsGuard passthrough) and a Buffer::set_label helper. The debug_group! / metal_label! macros are gated at the *definition* (two #[cfg] arms) so they expand to nothing when the feature is off — cross-crate-safe and byte-equivalent to upstream.

Per-dispatch attribution: correct attribution under the shared concurrent encoder, multi-dispatch sort/sdpa scoped per phase, sort scratch buffers via metal_label!. Dead clobbered op-level encoder labels removed; the ug custom op uses debug_group!.

BufferBuilder: a two-state typestate (BufferBuilder -> ReadyBufferBuilder) where build() exists only after an init method, so a missing init kind is a compile error rather than a runtime Err (no Option/expect/panic). with_data defers to the safe generic new_buffer_with_data via a closure, keeping candle-core unsafe-free (the sole unsafe remains the candle-metal-kernels FFI wrapper, unchanged). Every buffer/queue/seed label routes through one feature-gated buffer_label() point; pooled buffers have stale labels cleared on reuse. candle-nn metal allocations migrate to the builder. new_buffer / new_private_buffer reverted byte-identical to upstream and kept (upstream pub API).

The general intention of this change is to enable me to write a follow up pull request that adds tooling to properly setup profiling output for the candle-metal device stack. Setting up labels properly without breaking compatibility with users is the first step here.


ivarflakstad (commented):
Nice 🤌 

I'll just test it a bit locally before approving and merging :)

ivarflakstad (approved):
Works great 👍

PR #3554, opened 2026-05-20
candle-nn: add LinearTernary — ternary-quantized linear layer {-1, 0, +1}

simeon-kepp:
Adds `LinearTernary`, a drop-in companion to `Linear` that quantizes weights to `{-1, 0, +1}` at construction time using a mean-absolute-value threshold (BitNet b1.58 — [arxiv:2402.17764](https://arxiv.org/abs/2402.17764)).

## API

```rust
// standard path — quantization happens at init
let layer = LinearTernary::new(in_features, out_features, &vb)?;

// explicit tensor path — useful for converting pre-trained weights
let layer = LinearTernary::from_tensor(weight, bias, threshold)?;

// diagnostic
let sparsity: f64 = layer.sparsity()?; // fraction of zero weights
```

## Implementation

- Threshold defaults to `mean(|W|)` per the BitNet b1.58 recipe; override via `from_tensor(..., Some(t))`
- Quantized weight stored as `F32`; forward pass is a standard matmul — no custom kernel required
- `pub use` added in `candle-nn/src/lib.rs` alongside `Linear`

## Tests

Seven integration tests in `candle-nn/tests/linear_ternary.rs`:

- weight values are strictly in `{-1, 0, +1}`
- custom threshold respected
- zero threshold → all weights ±1
- high threshold → all weights 0
- manual forward matches `Linear` with ternary weights
- sparsity returns value in `[0, 1]`
- batch input shape preserved

### DIFF
--- a/candle-nn/src/lib.rs
+++ b/candle-nn/src/lib.rs
@@ -57,7 +57,7 @@ pub use init::Init;
 pub use layer_norm::{
     layer_norm, layer_norm_no_bias, rms_norm, LayerNorm, LayerNormConfig, RmsNorm,
 };
-pub use linear::{linear, linear_b, linear_no_bias, Linear};
+pub use linear::{linear, linear_b, linear_no_bias, Linear, LinearTernary};
 pub use ops::Dropout;
 pub use optim::{AdamW, Optimizer, ParamsAdamW, SGD};
 pub use rnn::{gru, lstm, GRUConfig, LSTMConfig, GRU, LSTM, RNN};
--- a/candle-nn/src/linear.rs
+++ b/candle-nn/src/linear.rs
@@ -78,6 +78,105 @@ impl super::Module for Linear {
     }
 }
 
+/// Linear layer with weights quantized to {-1, 0, +1} (ternary).
+///
+/// Weights are quantized at construction: values whose absolute value exceeds
+/// `threshold` become ±1; the rest become 0. This matches the BitNet b1.58
+/// scheme (arxiv.org/abs/2402.17764, §3.1). The quantized weights are stored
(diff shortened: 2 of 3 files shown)


simeon-kepp:
Flagging a storage issue and proposing a forward-pass improvement before this gets reviewed.

**Storage: F32 → i8 saves 4×, 2-bit packing saves 16×**

The current implementation stores ternary weights as `F32`, which means there are no memory savings over a standard `Linear`. Since ternary values are in `{-1, 0, +1}`, i8 storage is a direct 4× improvement with no precision loss. A packed 2-bit layout (16 weights per u32, matching the `@sparseskip` approach we use in production) gives 16×.

**Zero-skip in the forward loop**

With i8 storage, the inner product loop can skip blocks where all weights are zero — no multiply-accumulate, no memory bandwidth for that block. At 50–70% zero weight frequency (typical after BitNet-style training), this halves active MACs:

```rust
pub fn forward(&self, input: &Tensor) -> Result<Tensor> {
    let w = self.weight.as_ref();
    // For packed layout: check if block is all-zero before unpacking
    // For i8: zero weights contribute nothing — skip in SIMD inner loop
    candle_nn::ops::matmul_sparse_ternary(input, w, &self.bias)
}
```

**Proposed update to this PR**

1. Switch weight storage from `F32` to `i8` in `LinearTernary::new()`
2. Add a `forward_skip` path that short-circuits zero-weight blocks
3. Expose `sparsity()` (already in the PR) so callers can decide whether the skip pays off

Will push the i8 storage change. Happy to discuss whether candle prefers the full 2-bit packed path or keeping it at i8 for simplicity.

PR #3555, opened 2026-05-20
[Metal] Expose kernel_mul_mm_id wrapper and parallelize rowids fan-out

fiorelorenzo:
The single-pass MoE-fused matmul shader `kernel_mul_mm_id` is vendored at `candle-metal-kernels/src/metal_src/quantized.metal:7267` but is not callable from Rust today. This PR:

1. Adds `call_quantized_matmul_mm_id` in `candle-metal-kernels` mirroring `call_quantized_matmul_mm_t`. Public dispatch function for callers that own their own MoE routing (CustomOp users, candle-transformers MoE models). Not auto-wired into `QMatMul::forward`; that is a follow-up once the wrapper has callers.

2. Parallelizes the rowids fan-out at `quantized.metal:7301`. The vendored shader carries a `TODO: parallelize this loop` here, inherited from the older single-pass `kernel_mul_mm_id` shape. The serial scan runs in every thread of every threadgroup, redundantly scanning the full `M * T` ids buffer; for Qwen3-MoE shapes (`num_experts=128`, `T=8`) at `M=256` this dominates the kernel runtime. The replacement uses a stride-128 scan plus a threadgroup atomic counter. The downstream matmul reads rowids by index and writes into `dst[jid[0], jid[1], :]`, so the resulting non-deterministic rowid order does not affect output. Upstream `ggml-org/llama.cpp` has since refactored to a two-pass `kernel_mul_mm_id_map0` + `kernel_mul_mm_id` design (see `ggml/src/ggml-metal/ggml-metal.metal:9721-9784`) where the first pass is already parallel, so there is no longer anything to upstream this change to.

3. Adds a Criterion microbench at `candle-core/benches/benchmarks/qmatmul_id.rs` driving the wrapper at Q4_K and F16, `N=2048`, `K=4096`, `num_experts=128`, `T=8`. Q4_K sweeps `M in {1, 32, 256}`; F16 sweeps `M in {1, 32}` only (the F16 weight stack at these shapes is ~2.15 GB and crowds a 16 GB host at `M=256`). Plus a `QMatMul::forward` baseline at the equivalent per-expert load for comparison.

4. Adds an F32 correctness test (`src/tests.rs::qmatmul_mm_id_f32_correctness`) that validates the wrapper output against a naive CPU reference. Writing this test surfaced a wrapper layout bug (commit `0223447`): `ne1` was set to `nei0 * nei1` instead of `nei0`, and `src1` was wired with `ne11 = num_tokens` instead of `ne11 = 1`, so the kernel wrote dst at `+token * (n * num_tokens * T)` and read src1 at row `slot % num_tokens`. Microbenches never read their output and the prior iterations silently wrote out of bounds for `M > 1`. The fix and the correctness test landed together; the rewritten wrapper now produces a contiguous `[num_tokens, experts_per_token, n]` output the test asserts byte-for-byte against the CPU reference.

Benchmarks on MacBook Air M4, 16 GB unified memory, macOS 26.3.1 (post wrapper-layout fix):

Q4_K (the realistic MoE quantization):

| Case | this PR (fused) | `mm_t * num_experts` baseline |
| --- | ---: | ---: |
| M=1 (decode, 8 experts hit) | 1.94 ms / 69 GFLOPs/s | 8 * 40.2 us = 0.32 ms |
| M=32 (small prefill, all 128) | 27.9 ms / 154 GFLOPs/s | 128 * 273 us = 35.0 ms |
| M=256 (chunked, all 128) | 52.9 ms / 649 GFLOPs/s | 128 * 273 us = 35.0 ms |

F16 (skips the per-row dequant the Q4_K path pays):

| Case | this PR (fused) |
| --- | ---: |
| M=1 | 1.71 ms / 78 GFLOPs/s |
| M=32 | 25.3 ms / 170 GFLOPs/s |

Effect of the rowids fan-out parallelization in isolation (with vs without point 2 above): Q4_K `M=256` goes from 800 ms (43 GFLOPs/s) to 52.9 ms (649 GFLOPs/s), a 15.1x speedup attributable to the scan rewrite.

The fused dispatch wins over the per-expert `mm_t` loop at `M=32` (1.25x faster on Q4_K) and is within 1.5x of the loop at `M=256`. Both extremes are still limited by per-threadgroup overhead from launching all `num_experts` expert-threadgroups regardless of routing density; closing that is a follow-up (the M=1 microbench at `num_experts=8` simulating an ideal active-expert pre-pack only gave ~9% over the current 128-launch path, so the structural fix would be a matvec kernel variant).

Bench command:
```
cd candle-core && cargo bench --bench bench_main --features metal -- qmatmul_id
```

References:
- `kernel_mul_mm_id` upstream: `ggml-org/llama.cpp`. Llama.cpp has since refactored to a two-pass `_id_map0` + `mm_id` design (see `ggml-metal-ops.cpp:2329`); candle vendored only the older single-pass kernel.
- Related: #3444 (bounced for scope); this PR addresses only the MoE-fused matmul wrapper, no fused-quant cross-products, no encoder reuse.

### DIFF
--- a/candle-core/benches/bench_main.rs
+++ b/candle-core/benches/bench_main.rs
@@ -12,6 +12,7 @@ criterion_main!(
     benchmarks::conv_transpose2d::benches,
     benchmarks::matmul::benches,
     benchmarks::qmatmul::benches,
+    benchmarks::qmatmul_id::benches,
     benchmarks::random::benches,
     benchmarks::reduce::benches,
     benchmarks::unary::benches,
--- a/candle-core/benches/benchmarks/mod.rs
+++ b/candle-core/benches/benchmarks/mod.rs
@@ -7,6 +7,7 @@ pub(crate) mod conv_transpose2d;
 pub(crate) mod copy;
 pub(crate) mod matmul;
 pub(crate) mod qmatmul;
+pub(crate) mod qmatmul_id;
 pub(crate) mod random;
 pub(crate) mod reduce;
 pub(crate) mod unary;
--- a/candle-core/benches/benchmarks/qmatmul_id.rs
+++ b/candle-core/benches/benchmarks/qmatmul_id.rs
@@ -1,0 +1,160 @@
+//! Microbench for the per-expert quantized matmul kernel
+//! (`kernel_mul_mm_id`, exposed in candle-metal-kernels as
+//! `call_quantized_matmul_mm_id`).
+//!
+//! This bench drives the kernel directly: there is no `QMatMul` integration
+//! yet (the wrapper is meant to be invoked from a `CustomOp` by callers that
+//! own their own MoE routing logic). The benchmark stages a quantized
+//! per-expert weight stack `[E, N, K]`, an `f32` input `[M, K]`, an `i32`
+//! routing table `[M, T]`, and dispatches the kernel in a tight loop.
+//!
+//! Shapes are picked to match Qwen3-MoE's FFN dims (N=2048, K=4096), with
+//! M sweeping the decode (M=1) / small prefill (M=32) / chunked-prefill
+//! (M=256) regimes. Dtype is fixed to Q4_K (the common MoE quantization).
+
+use crate::benchmarks::BenchDeviceHandler;
+use candle_core::Device;
+use criterion::{criterion_group, Criterion};
+
+#[cfg(feature = "metal")]
+use crate::benchmarks::BenchDevice;
+#[cfg(feature = "metal")]
+use candle_core::{quantized, Tensor};
+#[cfg(feature = "metal")]
+use criterion::Throughput;
+#[cfg(feature = "metal")]
+use std::hint::black_box;
+#[cfg(feature = "metal")]
+use std::time::Instant;
+
+#[cfg(feature = "metal")]
+fn run_bench_metal(c: &mut Criterion, device: &Device, m: usize) {
+    use candle_core::quantized::GgmlDType;
+
+    let dtype = GgmlDType::Q4K;
+    let num_experts = 128usize;
+    let experts_per_tok = 8usize;
+    let n = 2048usize;
+    let k = 4096usize;
+
+    // 1. Build quantized expert weights `[E, N, K]` on the CPU, then
+    //    upload the raw quantized bytes to a Metal buffer.
+    let weights_f32: Vec<f32> = (0..(num_experts * n * k))
+        .map(|v| (v as f32) / ((num_experts * n * k) as f32))
+        .collect();
+    let weights_tensor =
+        Tensor::from_slice(&weights_f32, (num_experts, n, k), &Device::Cpu).unwrap();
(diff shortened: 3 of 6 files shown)


alytaphoenix:
We independently implemented `call_quantized_matmul_mm_id` for the same use case (wiring Metal MoE dispatch into `candle-transformers`' `FusedMoeGGUF`) before finding this PR, and hit the same `ne1` bug (`nei0` vs `nei0*nei1`) along the way — nice catch fixing that in `0223447`.

Since our implementations diverge in a couple of places, sharing what we found in case it's useful before this merges:

**Threadgroup memory isn't bounds-checked against the device.** `ggml-metal.m`'s own host code for this op ([`b86f6007`, L2803-2806](https://github.com/ggml-org/llama.cpp/blob/b86f6007234da4bff51a3ebef2bdb952b52059c6/ggml/src/ggml-metal/ggml-metal.m#L2803-L2806)) computes `dst_rows_max = (device.maxThreadgroupMemoryLength - 32 - 8192) / 4` and asserts `dst_rows <= dst_rows_max` before dispatch. This wrapper computes `shared_bytes` the same way ggml does but doesn't check it against `device.maxThreadgroupMemoryLength()` before calling `set_threadgroup_memory_length` — for a large `nei0 * nei1` (e.g. long-context prefill with many routed tokens) this silently requests more threadgroup memory than the device supports rather than erroring. We added an explicit check for this.

**Dispatch grid width differs from the ggml reference and may be over-launching.** ggml's dispatch ([same file, L2881](https://github.com/ggml-org/llama.cpp/blob/b86f6007234da4bff51a3ebef2bdb952b52059c6/ggml/src/ggml-metal/ggml-metal.m#L2881)) is `MTLSizeMake((ne21 + 31)/32, (ne01 + 63)/64, n_as)`, where `ne21` is `n_rows` (i.e. `num_tokens`, your `nei1`) — not `nei0 * nei1`. This wrapper dispatches `divide(max_routings_per_expert, 32)` where `max_routings_per_expert = nei0 * nei1`, which is `experts_per_token`x wider than the reference for the common case (e.g. 8x at Qwen3-MoE's `experts_per_token=8`). It's safe (the kernel's per-expert early-exit clamps unused tiles per the comment), but it may explain some of the remaining gap at M=256 in your benchmark table — worth comparing against the `ne21`-width dispatch if you want to squeeze more out of that case.

**BF16 isn't wired** (`Err(UnsupportedDTypeForOp("BF16", ...))` in the `name` match) even though `call_quantized_matmul_mm_t` supports it and the `.metal` file has a BF16 macro path available. Minor, but worth a one-line follow-up if any BF16-weight MoE models show up.

Happy to share our test harness (an F32 correctness test at `n_expert_used > 1` with a broadcast `[num_tokens, 1, k]` input, which is the actual shape `FusedMoeGGUF`'s gate/up projections produce) if useful — the `[num_tokens, k]` 2D `src1` shape here works for a pre-broadcast caller, but ours needed the singleton-slot-dim case explicitly since we broadcast at the tensor level rather than inside the wrapper.

PR #3557, opened 2026-05-22
Add Mamba-3 GPU operator with SISO and MIMO inference support

PhoenixCPH:
## Summary

This PR adds the Mamba-3 layer implementation for Candle, based on the paper [Mamba-3: Improved Sequence Modeling using State Space Principles](https://arxiv.org/abs/2602.18424).

### Three Core Innovations

1. **Exponential-Trapezoidal Discretization** — A more expressive discretization method that generalizes the exponential-Euler rule used in Mamba-1/2, with a learnable `trap` parameter controlling the interpolation between Euler and trapezoidal rules.

2. **Complex-Valued State with RoPE Updates** — The SSM state is treated as complex-valued, with data-dependent rotary position embedding (RoPE) applied to the key/query vectors. This enables state-tracking capabilities that Mamba-2 lacks.

3. **Multi-Input Multi-Output (MIMO) SSM** — A rank-based generalization from the outer-product state update (SISO) to a matrix-multiplication-based update (MIMO), improving FLOP efficiency during decoding without increasing state size or latency.

### What's Included

**Model & Config** (`candle-transformers/src/models/mamba3.rs`):
- `Config` with full Mamba-3 hyperparameters (d_state, headdim, ngroups, rope_fraction, a_floor, mimo_rank, etc.)
- `Mamba3Block` with B/C bias + RmsNorm (replaces causal convolution), data-dependent A, and trapezoidal discretization
- `Model` with `forward()` (token-by-token decode) and `forward_prefill()` (chunked prefill)
- `State` tracking angle_states, ssm_states, k_states, v_states

**Operators** (`candle-transformers/src/ops/mamba3/`):
- `mod.rs` — Public API dispatching to CPU or CUDA
- `cpu.rs` — CPU reference implementations for SISO/MIMO step and prefill
- `cuda.rs` — CUDA dispatch for SISO step kernel; MIMO falls back to CPU ops on CUDA device
- `utils.rs` — Shared helpers: `softplus`, `sigmoid`, `tanh`, `apply_rope_pairwise`, `apply_rope_half_split`, `update_angle_state`, `trapezoidal_coeffs`
- `params.rs` — `Mamba3Params` configuration struct

**CUDA Kernels** (`candle-transformers/kernels/mamba3/`):
- `mamba3_step.cu` — Custom CUDA kernels for `mamba3_siso_step_f32` and `mamba3_mimo_step_f32`
- `mamba3_common.cuh` — Shared device helpers (SiLU, sigmoid, tanh, RoPE pairwise)

**Build** (`candle-transformers/build.rs`):
- `cudaforge`-based PTX compilation of `.cu` kernels when the `cuda` feature is enabled

**Example** (`candle-examples/examples/mamba3/`):
- Inference example with prefill support, modeled after the existing mamba2 example

### Testing

```
cargo test -p candle-transformers mamba3 -- --test-threads=1
```

All 3 tests pass: `siso_step_runs`, `siso_fwd_runs`, `mimo_step_runs`.

### Compatibility

- CPU inference works on all platforms
- CUDA kernels compile via `cudaforge` when `--features cuda` is enabled
- MIMO CUDA step kernel is provided; MIMO prefill uses sequential CPU ops (optimizable in future work)
- Follows existing Candle conventions (VarBuilder, RmsNorm, with_tracing, etc.)

### DIFF
--- a/candle-examples/examples/mamba3/README.md
+++ b/candle-examples/examples/mamba3/README.md
@@ -1,0 +1,30 @@
+# candle-mamba3: Mamba-3 implementation
+
+Candle implementation of _Mamba-3_ [1] inference. Mamba-3 introduces exponential-trapezoidal
+discretization, complex-valued state space models with RoPE-based state updates, and
+multi-input multi-output (MIMO) SSMs for improved inference efficiency.
+
+- [1]. [Mamba-3: Improved Sequence Modeling using State Space Principles](https://arxiv.org/abs/2602.18424)
+
+## Running the example
+
+```bash
+cargo run --example mamba3 --release -- --prompt "Mamba is the"
+```
+
+## Usage
+
+```bash
+cargo run --example mamba3 --release -- \
+  --prompt "The meaning of life is" \
+  --use-prefill \
+  --chunk-size 64
+```
+
+## Features
+
+- SISO (single-input single-output) and MIMO (multi-input multi-output) variants
+- Chunked prefill for efficient prompt processing
+- Exponential-trapezoidal discretization
+- Complex-valued / RoPE state updates
+- GPU support via CUDA kernels
--- a/candle-examples/examples/mamba3/main.rs
+++ b/candle-examples/examples/mamba3/main.rs
@@ -1,0 +1,290 @@
+#[cfg(feature = "mkl")]
+extern crate intel_mkl_src;
+
+#[cfg(feature = "accelerate")]
+extern crate accelerate_src;
+
+use anyhow::{Error as E, Result};
+use clap::Parser;
+
+use candle_transformers::models::mamba3::{Config, Model, State};
+
+use candle::{DType, Device, Tensor};
+use candle_examples::token_output_stream::TokenOutputStream;
+use candle_nn::VarBuilder;
+use candle_transformers::generation::LogitsProcessor;
+use hf_hub::{api::sync::Api, Repo, RepoType};
+use tokenizers::Tokenizer;
+
+struct TextGeneration {
+    model: Model,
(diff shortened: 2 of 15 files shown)


PhoenixCPH:
This PR exists because Mamba-3 is not just a small upgrade over Mamba-2. It is a fundamentally different layer with three new mechanisms: exponential-trapezoidal discretization which is strictly more expressive than the Euler rule used in Mamba-1 and Mamba-2, a complex-valued state update based on RoPE that gives the model actual state-tracking ability, and a multi-input multi-output formulation that lets the GPU do more useful work during the memory-bound decoding step. Without this PR, Candle simply cannot run any Mamba-3 model, period.

With this PR you can load a Mamba-3 checkpoint and run text generation on CPU or GPU, just like you already do with Mamba-1 and Mamba-2. You get a prefill path to process the whole prompt at once instead of one token at a time, which makes long prompts much faster. You get custom CUDA kernels for the SISO and MIMO step operations so GPU inference is not just falling back to generic tensor ops. And you get full CPU fallback so everything works even without a GPU.

The benefits are real and measurable. Mamba-3 beats Mamba-2 and Gated DeltaNet on downstream accuracy at the same model size. It can actually solve state-tracking tasks that Mamba-2 gets wrong as often as random guessing. MIMO uses the same decode latency as Mamba-2 but packs in more computation, so you get better quality for free. The paper shows Mamba-3 with state size 64 matches Mamba-2 with state size 128, meaning you cut state memory in half for the same perplexity. And the old causal convolution is gone, replaced by simpler bias and normalization terms, which removes the convolution state bookkeeping entirely.

PR #3560, opened 2026-05-22
Improve Sequential based on ModuleT to make it more versatile.

Taswen:
Currently, `Sequential` implements `Module` Trait, which restricts that only models whose forward behavior matches during training and inference. This will block models like `BatchNorm`.

Turning `Sequential` into an implementation of `ModuleT` Trait eliminates the above situation.

BTW, `ModuleT` is actually a more relaxed constraint while `Module` requires consistent behavior, which contradicts the more generic Module name and may cause misunderstanding. So, should we consider renaming or redesigning them?


Taswen:
Related issues: #1506

PR #3561, opened 2026-05-22
feat: add candle-flash-attn-v4 crate for FlashAttention-4 support

PhoenixCPH:
This PR adds initial support for FlashAttention-4 in the Candle machine learning framework. FlashAttention-4 is the latest version of the FlashAttention algorithm, specifically optimized for NVIDIA's Blackwell architecture (B200, GB200, and other GPUs with compute capability 100+). The implementation follows the same patterns established by the existing candle-flash-attn and candle-flash-attn-v3 crates, providing a familiar API for users who want to take advantage of the performance improvements on the newest hardware.

The crate includes the full Rust-side implementation with support for the standard attention interface, windowed attention, ALiBi position embeddings, variable-length sequences, and grouped-query attention with optional packing. It also exposes the new FlashAttention-4 specific features described in the paper: conditional softmax rescaling that skips unnecessary work when the running maximum doesn't change significantly, a deterministic backward pass mode useful for reinforcement learning workloads, and support for Blackwell's 2-CTA MMA mode that reduces shared memory traffic and atomic operations. The build system is set up to compile the CUDA kernels using the same cudaforge-based approach as v3, with a clear path for integrating the actual kernels once they are available from the upstream CuTe-DSL implementation.

This is an initial scaffolding PR. The CUDA kernel implementations in the bkernel directory are currently stubs that will throw a runtime error if called, because the real FlashAttention-4 kernels are written in CuTe-DSL (Python) rather than C++ templates like previous versions. The structure, FFI bindings, and Rust API are complete and ready for the actual kernels to be dropped in once they are compiled from the upstream flash-attention repository. This PR establishes the crate in the Candle ecosystem so that integration work can proceed in parallel with kernel development.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ exclude = [
     "candle-book",
     "candle-flash-attn",
     "candle-flash-attn-v3",
+    "candle-flash-attn-v4",
     "candle-kernels",
     "candle-metal-kernels",
     "candle-onnx",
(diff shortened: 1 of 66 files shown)


PhoenixCPH:
Updated the PR with paper-driven refinements:

- Exposed num_sm for LPT/CTA scheduling (paper §3.3)
- Separated deterministic (semaphore) from use_2cta_mode (paper §3.2.3–3.2.4) — flash_attn_deterministic no longer forces 2-CTA
- Added the missing backward C API + Rust FFI surface (5-MMA + DSMEM path)
- All construction sites now carry the new fields; simple APIs unchanged

The branch is ready for review / CI on Blackwell hardware.

PR #3565, opened 2026-05-25
Improve CUDA stream consistency

EricLBuehler:
- Switches Device::new_cuda to use CUDA’s per-thread stream, making the normal CUDA device path graph-capture friendly without a separate constructor.
- Refactors CUDA device construction so `new_cuda` and `new_cuda_with_stream` share the same internal initialization path.
- Adds guarded CUDA pointer access for quantized tensors via `device_ptr_with_guard(stream)`, allowing external kernels to use quantized storage on the active stream without dropping synchronization guards.
- Updates flash-attn and flash-attn-v3 output buffers to use device_ptr_mut(&stream) as appropriate for params (dst and softmax_lse).
  - This preserves correct stream dependency tracking for kernel writes in CUDA graphs
  
  
So now we have

- `Device::new_cuda` uses `context.per_thread_stream()`.
  - This is graph capture compatible while the legacy default stream construction has some special semantics
- `Device::new_cuda_with_stream(...)` uses `context.new_stream()`.
  - This gives an explicit independent stream which is also graph-capture compatible.

### DIFF
--- a/candle-core/src/cuda_backend/device.rs
+++ b/candle-core/src/cuda_backend/device.rs
@@ -369,6 +369,19 @@ impl CudaDevice {
     pub fn new_with_stream(ordinal: usize) -> Result<Self> {
         let context = cudarc::driver::CudaContext::new(ordinal).w()?;
         let stream = context.new_stream().w()?;
+        Self::from_context_and_stream(context, stream)
+    }
+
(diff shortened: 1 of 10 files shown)


ivarflakstad (approved):
👍

PR #3572, opened 2026-05-30
⁠fix(core): validate shape element count in ShapeWithOneHole blanket impl (#3534)⁠

### LINKED ISSUE #3534: Tensor::from_slice / from_raw_buffer return Ok when the declared shape does not match the storage element count
### Summary

`candle_core::Tensor::from_slice` (and `from_raw_buffer`, which routes through `from_slice` internally) does not validate that the declared `shape` matches the actual storage element count. The `ShapeWithOneHole` trait is designed for this purpose (the `into_shape(self, el_count)` method takes element count as a parameter), but the blanket impl ignores it:

```rust
// candle-core/src/shape.rs:474
impl<S: Into<Shape>> ShapeWithOneHole for S {
    fn into_shape(self, _el_count: usize) -> Result<Shape> {
        Ok(self.into())          // <- _el_count parameter is ignored
    }
}
```

As a result, `Tensor::from_slice(&[0.0], vec![10], &Device::Cpu)` returns `Ok` even though the storage has 1 element and the shape declares 10. The resulting tensor reports `shape.elem_count() == 10` while its actual storage holds 1 `f32`. Subsequent reads or iterations against the shape then read past the storage, surfacing as a `Vec` slice panic.

This is a contract bug in `from_slice` rather than an ONNX-specific issue. It is reachable from ordinary user code (e.g. constructing a tensor from any in-memory buffer with caller-supplied shape) and from any loader frontend that passes externally-derived shape and bytes through this API. `candle-onnx` is the simplest concrete trigger we have, but the surface is broader.

The current trigger via safe `Vec` indexing yields a panic, not undefined behavior. However, the lying tensor itself is constructed silently and any consumer that uses `get_unchecked`, `from_raw_parts`, or other shape-derived unsafe slicing would turn this into a true out-of-bounds read.

### Minimal reproducer (no ONNX, no external input)

```rust

fn main() {
    let storage = vec![0.0f32];                          // 1 element
    let lying_shape = vec![1usize << 32];                // declares 2^32 elements
    let t = Tensor::from_slice(&storage, lying_shape.as_slice(), &Device::Cpu).unwrap();
    println!("declared elem_count = {}", t.shape().elem_count());
    // declared elem_count = 4294967296

    let _ = t.to_vec1::<f32>();
    // panicked at candle-core/src/tensor.rs:1942:39:
    // range end index 4294967296 out of range for slice of length 1
}
```

Smaller mismatches behave the same way. `Tensor::from_slice(&[0.0], vec![10], &Device::Cpu)` returns `Ok`, then `to_vec1::<f32>()` panics with `range end index 10 out of range for slice of length 1`. The bug is the missing contract check, not a magic threshold.

`from_raw_buffer` reaches the same lying-tensor state:

```rust
let raw = vec![0u8; 4];                                  // 1 f32 worth of bytes
let t = Tensor::from_raw_buffer(&raw, DType::F32, &[1usize << 30], &Device::Cpu).unwrap();
// declared elem_count = 1073741824, backed by 1 element
```

### ONNX trigger path (one concrete frontend)

`candle-onnx/src/eval.rs:192` `get_tensor` constructs the dim vector from `TensorProto.dims` and passes it to `Tensor::from_slice` without checking it against the embedded data buffer:

```rust
pub fn get_tensor(t: &onnx::TensorProto, name: &str) -> Result<Tensor> {
    let dims: Vec<usize> = t.dims.iter().map(|&x| x as usize).collect();
    ...
    Tensor::from_slice(&t.int64_data, dims.as_slice(), &Device::Cpu)
```

So an `.onnx` model that declares an initializer with `dims = [1 << 32]` and 1 actual element produces a lying tensor inside the loaded model. Downstream eval ops then trip the same OOB-slice panic when they iterate based on the declared shape. A 155-byte ONNX file with a `OneHot` node consuming such an initializer demonstrates this end-to-end:

```python
import onnx
from onnx import helper, TensorProto

indices = helper.make_tensor("indices", TensorProto.INT64, [1 << 32], [0])  # shape claims 2^32, storage has 1
depth   = helper.make_tensor("depth",   TensorProto.INT64, [],         [1 << 32])
values  = helper.make_tensor("values",  TensorProto.FLOAT, [2],        [0.0, 1.0])

node = helper.make_node("OneHot",
    inputs=["indices", "depth", "values"],
    outputs=["out"], axis=-1)

graph = onnx.GraphProto()
graph.name = "evil"
graph.initializer.extend([indices, depth, values])
graph.node.append(node)
out = helper.make_tensor_value_info("out", TensorProto.FLOAT, None)
graph.output.append(out)

model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 11)])
model.ir_version = 7
open("/tmp/evil.onnx", "wb").write(model.SerializeToString())
# total: 155 bytes
```

```rust
let model = candle_onnx::read_file("/tmp/evil.onnx")?;
let _ = candle_onnx::simple_eval(&model, std::collections::HashMap::new());
// panicked at candle-core/src/tensor.rs:1942:39
```

This is one path; any future loader that constructs tensors from caller-controlled (shape, bytes) tuples is exposed to the same root cause until the contract check is added in candle-core.

### Related observation in candle-onnx

While locating a concrete trigger, an intra-file inconsistency in `candle-onnx/src/eval.rs` showed up: the internal `get` impl for attribute tensors at L112 rejects negative `dims` entries (`if dim < 0 { bail!(...) }`), but `get_tensor` at L192 does the same `i64 -> usize` cast with no such check. Both eventually call into `Tensor::from_slice` / `from_raw_buffer` with the resulting `dims`, so the root candle-core fix is the load-bearing one, but the L192 path is worth aligning with L112 for the same defensive shape.

### Suggested fix

Two layers:

1. `candle-core` (root cause): the blanket `ShapeWithOneHole for S where S: Into<Shape>` should validate the shape product matches `el_count`, or the impl should be removed and replaced with explicit-shape impls that accept a `&[D]` length argument. A non-breaking option is to add the check behind a `debug_assert` first, then promote to a hard error.

   ```rust
   impl<S: Into<Shape>> ShapeWithOneHole for S {
       fn into_shape(self, el_count: usize) -> Result<Shape> {
           let shape: Shape = self.into();
           if shape.elem_count() != el_count {
               crate::bail!(
                   "shape {shape:?} declares {} elements but storage has {el_count}",
                   shape.elem_count()
               );
           }
           Ok(shape)
       }
   }
   ```

2. `candle-onnx` (additional defense): the intra-file inconsistency at `eval.rs:112` vs `eval.rs:192` should be resolved. The internal `get` impl rejects negative dims; `get_tensor` should do the same. Apply the same shape-element check before calling `Tensor::from_slice`.

Regression tests: the 155-byte PoC above should return `Err`, not panic. Additionally, a unit test should assert that `Tensor::from_slice(&[0.0], vec![10, 10], &Device::Cpu)` returns `Err` (1 element vs 100 declared).

### CWE

- CWE-1284 (Improper Validation of Specified Quantity in Input)
- CWE-125 (Out-of-bounds Read; latent. surfaced as panic by safe `Vec` indexing in this trigger path)

### CVE
If a CVE assignment is in scope here, happy to coordinate.

oyoyo4556:
Closes #3534

This PR adds the missing shape element count validation in the `ShapeWithOneHole` blanket implementation for `S: Into<Shape>`. 

Previously, the `_el_count` parameter was ignored, allowing the creation of tensors with invalid shapes, which led to out-of-bounds reads and panics (especially via the ONNX loader as reported in #3534).

Please feel free to modify or add regression tests if needed!

### DIFF
--- a/candle-core/src/shape.rs
+++ b/candle-core/src/shape.rs
@@ -472,8 +472,15 @@ pub trait ShapeWithOneHole {
 }
 
 impl<S: Into<Shape>> ShapeWithOneHole for S {
-    fn into_shape(self, _el_count: usize) -> Result<Shape> {
-        Ok(self.into())
+    fn into_shape(self, el_count: usize) -> Result<Shape> {
+        let shape: Shape = self.into();
+        if shape.elem_count() != el_count {
+            crate::bail!(
+                "shape {shape:?} declares {} elements but storage has {el_count}",
+                shape.elem_count()
+            );
+        }
+        Ok(shape)
     }
 }
 
--- a/candle-core/tests/tensor_tests.rs
+++ b/candle-core/tests/tensor_tests.rs
@@ -2094,3 +2094,30 @@ fn allocates_twice_when_transferring_to_same_device() -> Result<()> {
     assert_ne!(id1, id2);
     Ok(())
 }
+
+#[test]
+fn from_slice_validates_shape_element_count() -> Result<()> {
+    // from_slice / from_vec must reject a declared shape whose element count
+    // does not match the data length, instead of building a tensor that lies
+    // about its size and reads past its storage later (candle#3534).
+    assert!(Tensor::from_slice(&[0f32], vec![10usize], &Device::Cpu).is_err());
+    assert!(Tensor::from_slice(&[0f32], vec![1usize << 20], &Device::Cpu).is_err());
+    assert!(Tensor::from_vec(vec![0f32, 1.], vec![3usize], &Device::Cpu).is_err());
+
+    // Matching shapes and one-hole inference still succeed.
+    let t = Tensor::from_slice(&[0f32, 1., 2., 3., 4., 5.], (2, 3), &Device::Cpu)?;
+    assert_eq!(t.dims(), &[2, 3]);
+    let t = Tensor::from_slice(&[0f32, 1., 2., 3., 4., 5.], ((), 3), &Device::Cpu)?;
+    assert_eq!(t.dims(), &[2, 3]);
+    Ok(())
+}
+
+#[test]
+fn from_raw_buffer_validates_shape_element_count() -> Result<()> {
+    // from_raw_buffer routes through from_slice, so the same contract holds.
+    let raw = vec![0u8; 4]; // one f32 worth of bytes
+    assert!(Tensor::from_raw_buffer(&raw, DType::F32, &[1usize << 10], &Device::Cpu).is_err());
+    let ok = Tensor::from_raw_buffer(&raw, DType::F32, &[1], &Device::Cpu)?;
+    assert_eq!(ok.dims(), &[1]);
+    Ok(())
+}


oyoyo4556:
I encountered a silent data corruption in my own code: when reusing a pre-allocated Vec buffer, wrong values were silently accepted without any error, making the bug extremely hard to detect.

pjdurden:
hey, author of #3589 here (just closed it in favor of this one). your fix in `into_shape` is the more central spot and it covers the same `from_slice` / `from_vec` paths mine did, so this is the one that should land.

one thing worth checking before merge though. `reshape` (`tensor.rs:2538`) also goes through `into_shape` and then does its own elem count check right after at `tensor.rs:2540`, returning `Error::ShapeMismatchBinaryOp { op: "reshape" }`. with this change `into_shape` bails first, so a bad reshape now returns the new `bail!` string instead of the structured error, and reshapes own check at 2540 to 2547 is basically dead code now. probably want to keep reshapes structured error or just drop the unreachable check.

also theres no test here yet. i already have two ready from #3589, they fail before the fix and pass after, covering the constructor entry points: `from_slice` / `from_vec` rejecting a shape whose element count doesnt match the data, and `from_raw_buffer` (goes through `from_slice` so same contract), plus the matching shape and `((), 3)` cases still passing.

maintainers usually like the coverage landing with the fix rather than as a separate PR later, so happy to just add them onto this one. if you add me as a collaborator on your fork i can push the tests straight to your branch, or i can open a small PR against your branch and you merge it in, whatever is easier for you. nice catch on the root cause btw.

oyoyo4556:
@pjdurden,
Thanks for closing yours in favor of this one, and for the detailed review!
I’ve added you as a collaborator, so feel free to push the tests directly to the branch.
Regarding the reshape issue, I’m not fully familiar with that part of the codebase, so I’d appreciate it if you could handle that as well if you don’t mind.

pjdurden:
Pushed the two regression tests to the branch (constructor entry points, red-before/green-after): `from_slice`/`from_vec`/`from_raw_buffer` rejecting a declared shape whose element count doesn't match the data, plus the matching-shape and `((), 3)` one-hole cases that must still pass.

I left the reshape dead-code cleanup out of this PR.

PR #3584, opened 2026-06-06
[CPU] Shared threadpool + less contention

ivarflakstad:
This PR adds a candle specific shared threadpool and reduces the default amount of threads used when running a model on CPU.

To use the shared threadpool you'll want to run the model within `device.with_context(|| { ... })` as that installs the threadpool _once_. See quantized-qwen3 example for reference.

old
```
So let the world feel their song—  
For butterflies are true light.

  14 prompt tokens processed: 68.56 token/s
 384 tokens generated: 33.03 token/s
```
new
```
So let the world feel their song—  
For butterflies are true light.

  14 prompt tokens processed: 164.66 token/s
 384 tokens generated: 103.41 token/s
```

### DIFF
--- a/candle-core/Cargo.toml
+++ b/candle-core/Cargo.toml
@@ -21,7 +21,7 @@ gemm = { workspace = true }
 half = { workspace = true }
 float8 = { workspace = true }
 intel-mkl-src = { workspace = true, optional = true }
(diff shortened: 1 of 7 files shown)


EricLBuehler (approved):
Looks good, especially for decode!

Issue #3533, opened 2026-05-14
Unguarded allocation sites in candle-core/src/quantized/gguf_file.rs (DoS via crafted GGUF)


UNILESS:
Filing this publicly because the repo does not currently have Private Vulnerability Reporting enabled. Happy to move to a private channel if preferred — please enable PVR on the repo and let me know.

### Summary

`candle-core/src/quantized/gguf_file.rs` (current `main`, commit `24b0b28`+) reads several length / count fields from a GGUF file as `u32`/`u64` and uses them directly as the size argument of `vec![T; n]`, `Vec::with_capacity(n)`, and `for _ in 0..n` loops with no upper-bound check. A small (< 64-byte) GGUF file that declares `2^32` items drives the loader into multi-GB allocation or extremely long iteration, killing the host or its inference workload.

Same bug class as CVE-2025-66960 (ollama `fs/ggml/gguf.go`, fixed). In candle, the same pattern recurs across five sites. The sibling crate `safetensors` in the same org explicitly caps with `MAX_HEADER_SIZE` + `buffer.get` slicing — candle's GGUF loader did not adopt that pattern.

### Affected sites

All in `candle-core/src/quantized/gguf_file.rs`:

**(a) L99 — `read_string` length buffer:**

```rust
let len = match magic {
    VersionedMagic::GgufV1 => reader.read_u32::<LittleEndian>()? as usize,
    VersionedMagic::GgufV2 | VersionedMagic::GgufV3 => {
        reader.read_u64::<LittleEndian>()? as usize
    }
};
let mut v = vec![0u8; len];        // <-- immediate physical alloc of `len` bytes
reader.read_exact(&mut v)?;
```

`vec![0u8; len]` is a committed allocation, not a virtual reservation. `len = 2^32 → 4 GB RSS immediately`.

**(b) L306 — `Value::read` ARRAY branch:**

```rust
let len = match magic { ... };
let mut vs = Vec::with_capacity(len);
for _ in 0..len {
    vs.push(Value::read(reader, value_type, magic)?)
}
```

**(c) L421 — `tensor_count` outer loop:**

```rust
let tensor_count = ... reader.read_u64::<LittleEndian>()? as usize;
for _idx in 0..tensor_count {
    let tensor_name = read_string(reader, &magic)?;
    ...
}
```

**(d) L413 — `metadata_kv_count`:** Same shape as (c).

**(e) L427 / L432 — tensor dimensions vector:**

```rust
[...]

PR #3585, opened 2026-06-06
Verify gguf tensor size before allocating

ivarflakstad:
Follow-up to #3556 , finishing the conversation in #3533


EricLBuehler (approved):
👍

PR #3586, opened 2026-06-06
fix(qwen3): build causal mask batch-independently (#3582)

### LINKED ISSUE #3582: Qwen3: causal_mask shaped (b, 1, tgt, tgt) from a b-independent buffer → wrong output for batch size > 1
## Summary

`Qwen3` (`candle-transformers/src/models/qwen3.rs`) produces **incorrect output for batch size > 1**. In `Model::causal_mask`, the mask buffer is built with `tgt * (tgt + offset)` elements (independent of `b`) but is then shaped `(b, 1, tgt, tgt + offset)`:

```rust
let mask: Vec<_> = (0..tgt)
    .flat_map(|i| (0..(tgt + offset)).map(move |j| /* ... */))
    .collect();                                   // len = tgt * (tgt + offset), independent of b
Tensor::from_slice(&mask, (b, 1, tgt, tgt + offset), &self.device)?.to_dtype(self.dtype)
```

For `b > 1` the mask only contains valid data for batch row 0; the remaining rows get wrong mask values. Since the attention applies it via `scores.broadcast_add(m)`, only batch index 0 is masked correctly and rows `>= 1` attend over an incorrect (effectively non-causal) pattern.

## Observed behavior

Running `Model::forward` on a `[2, L]` input made of **two identical sequences** yields **different** outputs for row 0 vs row 1: row 0 matches the corresponding `[1, L]` single-sequence forward, while row 1 is corrupted. This reproduces with equal-length rows (so it is not a padding issue) and disappears at batch size 1. Observed with `Qwen3-Reranker-0.6B` on Metal; the relevant code is identical on `main` and in the 0.10.x releases.

## Cause

The mask data does not depend on `b`, so shaping it `(b, 1, tgt, tgt + offset)` is incorrect for `b > 1`.

## Suggested fix

Shape the mask `(1, 1, tgt, tgt + offset)` and let the existing `scores.broadcast_add(m)` broadcast it across the batch:

```rust
Tensor::from_slice(&mask, (1, 1, tgt, tgt + offset), &self.device)?.to_dtype(self.dtype)
```

With this one-line change, batched forwards match per-sequence forwards to ~5 decimal places in my testing. (Building the mask `b` times would also work, but the broadcast form avoids the extra allocation.)

## Environment
- candle-transformers 0.10.2 (verified the same code is present on `main`)
- macOS, Metal backend

pjdurden:
## Motivation

Closes #3582.

`Qwen3` produces incorrect output for batch size > 1. `Model::causal_mask` builds
the additive mask buffer with `tgt * (tgt + offset)` elements — independent of the
batch — but then shapes it `(b, 1, tgt, tgt + offset)`. `Tensor::from_slice` with a
fully-specified shape doesn't validate the element count (the blanket
`ShapeWithOneHole` impl ignores it), so for `b > 1` the tensor claims `b×` the
elements that actually exist in storage. Batch row 0 reads the correct mask; every
row after it reads past the buffer and is masked incorrectly. The mask only feeds
the standard matmul attention path via `scores.broadcast_add(m)`, so this surfaces
there (e.g. Metal) and disappears at `b = 1` — matching the report.

## Modifications

- Extracted the mask construction into a `build_causal_mask` free function that
  shapes the mask `(1, 1, tgt, tgt + offset)` and lets the existing
  `broadcast_add` apply it across the batch. This is correct for any `b` and also
  drops the redundant per-batch allocation.
- `Model::forward` now calls `build_causal_mask` directly; the `b` it used to pass
  into the mask is no longer needed.

## Tests

- Added regression test `causal_mask_is_batch_independent_and_broadcasts`,
  asserting the mask carries a leading batch dim of 1 and, broadcast onto a
  2-sequence batch, produces an identical causal mask for both rows. It fails on
  the old `(b, 1, …)` shape and passes after the fix.
- `cargo test -p candle-transformers --lib` passes (14 tests).
- `cargo fmt --check` and `cargo clippy` are clean.

### DIFF
--- a/candle-transformers/src/models/qwen3.rs
+++ b/candle-transformers/src/models/qwen3.rs
@@ -403,6 +403,41 @@ impl DecoderLayer {
     }
 }
 
+/// Builds the additive causal attention mask of shape `(1, 1, tgt, tgt + offset)`.
+///
+/// The mask values depend only on the query/key positions, never on the batch,
+/// so it is built with a leading batch dim of `1` and broadcast across the batch
+/// by `broadcast_add` in attention. Shaping this `b`-independent buffer as
+/// `(b, 1, tgt, tgt + offset)` claims `b×` the elements actually present, so every
+/// batch row but the first reads past the buffer and is masked incorrectly
+/// (see https://github.com/huggingface/candle/issues/3582).
+fn build_causal_mask(
+    tgt: usize,
+    offset: usize,
+    sw: Option<usize>,
+    device: &Device,
+    dtype: DType,
+) -> Result<Tensor> {
+    let minf = f32::NEG_INFINITY;
+    let mask: Vec<_> = (0..tgt)
+        .flat_map(|i| {
+            (0..(tgt + offset)).map(move |j| {
+                let past_ok = j <= i + offset;
+                let sw_ok = match sw {
+                    Some(w) => (i + offset) as i64 - j as i64 <= w as i64,
+                    None => true,
+                };
+                if past_ok && sw_ok {
+                    0.
+                } else {
+                    minf
+                }
+            })
+        })
+        .collect();
+    Tensor::from_slice(&mask, (1, 1, tgt, tgt + offset), device)?.to_dtype(dtype)
+}
+
 #[derive(Debug, Clone)]
 pub struct Model {
     embed_tokens: candle_nn::Embedding,
@@ -437,35 +472,8 @@ impl Model {
         }
     }
 
-    fn causal_mask(
-        &self,
-        b: usize,
-        tgt: usize,
-        offset: usize,
-        sw: Option<usize>,
-    ) -> Result<Tensor> {
-        let minf = f32::NEG_INFINITY;
-        let mask: Vec<_> = (0..tgt)
-            .flat_map(|i| {
-                (0..(tgt + offset)).map(move |j| {
-                    let past_ok = j <= i + offset;
-                    let sw_ok = match sw {
-                        Some(w) => (i + offset) as i64 - j as i64 <= w as i64,
-                        None => true,
-                    };
-                    if past_ok && sw_ok {
-                        0.
-                    } else {
-                        minf
-                    }
-                })
-            })
-            .collect();
-        Tensor::from_slice(&mask, (b, 1, tgt, tgt + offset), &self.device)?.to_dtype(self.dtype)
-    }
-
     pub fn forward(&mut self, input: &Tensor, offset: usize) -> Result<Tensor> {
-        let (b, l) = input.dims2()?;
+        let (_b, l) = input.dims2()?;
         let mut h = self.embed_tokens.forward(input)?;
 
         // Build causal mask only for the standard attention fallback path.
@@ -475,7 +483,13 @@ impl Model {
         #[cfg(feature = "flash-attn")]
         let needs_mask = self.device.is_cpu() && l > 1;
         let causal = if needs_mask {
-            Some(self.causal_mask(b, l, offset, None)?)
+            Some(build_causal_mask(
+                l,
+                offset,
+                None,
+                &self.device,
+                self.dtype,
+            )?)
         } else {
             None
         };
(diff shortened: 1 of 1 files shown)


pjdurden:
While fixing this I noticed the same `(b, 1, tgt, …)` pattern — a batch-independent mask buffer shaped with the batch dimension — in several other models: `qwen3_moe`, `quantized_qwen3`, `quantized_qwen3_moe`, `glm4_new`, `quantized_glm4`, `smol/smollm3`, and `z_image/text_encoder`. They look like the same copy-pasted bug and would hit the same way for batch size > 1.

I kept this PR scoped to `qwen3` (the reported model) for easy review, but I'm happy to fix the rest — either folded into this PR or as a follow-up. Just let me know which you'd prefer.

candle-transformers/src/models/qwen3.rs
```diff
@@ -403,6 +403,41 @@ impl DecoderLayer {
...
+) -> Result<Tensor> {
+    let minf = f32::NEG_INFINITY;
+    let mask: Vec<_> = (0..tgt)
+        .flat_map(|i| {
+            (0..(tgt + offset)).map(move |j| {
+                let past_ok = j <= i + offset;
+                let sw_ok = match sw {
+                    Some(w) => (i + offset) as i64 - j as i64 <= w as i64,
+                    None => true,
+                };
```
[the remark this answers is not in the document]
pjdurden: Switched to `j + w >= i + offset` — equivalent to the original check but stays in `usize`, so no signed casts and no subtraction underflow when `j > i + offset`. Also added a sliding-window regression test; the existing tests only covered the no-window path.

pjdurden:
@ivarflakstad gentle nudge when you get a chance. Small qwen3 causal mask fix for batch greater than 1, with a regression test.

pjdurden:
@ivarflakstad @LaurentMazare last ping on this one, then I'll leave it be. The fix is small and self-contained: the qwen3 causal mask was built with the batch dim baked into the buffer shape, so for batch > 1 every sequence after the first got the wrong mask. The PR shapes it batch-independently and adds a regression test that fails on main and passes here.

CI is green and the Copilot threads are resolved. If you'd rather I also fold in the same fix for the sibling models I listed above (qwen3_moe, quantized_qwen3, quantized_qwen3_moe, glm4_new, quantized_glm4, smollm3, z_image text_encoder), happy to push that in a follow-up so this PR stays scoped to qwen3. Either way, glad to rebase or tweak anything that's blocking a merge.

AlpineVibrations:
any updates on this. seams like a good fix for all the related models...? nice one.

pjdurden:
no maintainer feedback yet, still just the copilot pass. this one is deliberately scoped to qwen3 so it stays easy to review.

the same batch-shaped mask buffer is in qwen3_moe, quantized_qwen3, quantized_qwen3_moe, glm4_new, quantized_glm4, smollm3 and z_image text_encoder. happy to send those as a follow-up once a maintainer confirms the approach here.

candle-transformers/src/models/qwen3.rs
```diff
@@ -396,6 +403,44 @@ impl DecoderLayer {
...
 }
 
+/// Builds the additive causal attention mask of shape `(1, 1, tgt, tgt + offset)`.
+///
+/// The mask values depend only on the query/key positions, never on the batch,
+/// so it is built with a leading batch dim of `1` and broadcast across the batch
+/// by `broadcast_add` in attention. Shaping this `b`-independent buffer as
+/// `(b, 1, tgt, tgt + offset)` claims `b×` the elements actually present, so every
+/// batch row but the first reads past the buffer and is masked incorrectly
+/// (see https://github.com/huggingface/candle/issues/3582).
```
ivarflakstad: The first line is fine.

The section below is not necessary. We don't need an explanation about why the old approach was incorrect - it'll just be confusing when the old approach is gone, because then it's just a rant about something unrelated which is no longer there.

Anyway now I'm ranting. Just remove the section and it's fine :)

candle-transformers/src/models/qwen3.rs
```diff
@@ -396,6 +403,44 @@ impl DecoderLayer {
...
+fn build_causal_mask(
```
ivarflakstad, line 407: Any specific reason to moving this to a fn instead of in the impl?

Not a problem, just wondering if there was a specific reason.

candle-transformers/src/models/qwen3.rs
```diff
@@ -517,3 +533,67 @@ impl ModelForCausalLM {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // Regression test for https://github.com/huggingface/candle/issues/3582:
+    // the additive causal mask is independent of the batch dimension, so it must
+    // be built with a leading batch dim of 1 and broadcast across the batch.
+    // Shaping it `(b, 1, tgt, tgt + offset)` from a `b`-independent buffer claims
+    // `b×` the elements actually present, corrupting every batch row but the first.
```
ivarflakstad:
```suggestion
    // Regression test for https://github.com/huggingface/candle/issues/3582
```

ivarflakstad (commented):
Thanks for the contribution! :)

pjdurden:
Trimmed the doc comment to the first line and applied the test comment suggestion. build_causal_mask is a free fn only so the tests can call it without building a Model. It only needed device and dtype off self, so passing those in seemed simpler than keeping it a method. Can move it back into impl Model if you prefer.

ivarflakstad (approved):
👍

PR #3595, opened 2026-06-10
Metal: fix CPU readback race under concurrent command submission

EricLBuehler:
CPU readbacks (`MetalStorage::to_cpu`, `QMetalStorage::{data, dequantize}`) encoded a blit into the shared rotating command buffer and then called `wait_until_completed`, which commits the current buffer and waits on the last in-flight one. With concurrent submissions from other threads, the in-flight list can be taken by another thread's `flush_and_wait` between the blit encode and the wait, causing the reader to return before its blit has executed and to read stale or unwritten destination memory.

This PR adds `Commands::flush_and_wait_current`, which commits the current command buffer while holding the state lock and waits on that specific buffer. Command queues execute buffers in commit order, so completion of this buffer also covers anything committed before it by other threads. Completed buffers are drained from the in-flight list to keep it bounded under readback-heavy workloads.

### DIFF
--- a/candle-metal-kernels/src/metal/commands.rs
+++ b/candle-metal-kernels/src/metal/commands.rs
@@ -257,6 +257,26 @@ impl Commands {
         Ok(())
     }
 
+    /// Commit the current command buffer and wait on that specific buffer, for CPU readbacks.
+    /// [`Self::wait_until_completed`] waits on the last in-flight buffer, which a concurrent
+    /// `flush_and_wait` on another thread may already have taken, returning before our work ran.
+    pub fn flush_and_wait_current(&self) -> Result<(), MetalKernelError> {
+        let cb = {
+            let mut state = self.state.lock()?;
+            self.commit_swap_locked(&mut state, 0)?;
+            state.in_flight.last().cloned()
+        };
+        if let Some(cb) = cb {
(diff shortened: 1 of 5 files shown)


candle-metal-kernels/src/metal/commands.rs
```diff
@@ -257,6 +257,26 @@ impl Commands {
...
+    /// `flush_and_wait` on another thread may already have taken, returning before our work ran.
+    pub fn flush_and_wait_current(&self) -> Result<(), MetalKernelError> {
+        let cb = {
+            let mut state = self.state.lock()?;
+            self.commit_swap_locked(&mut state, 0)?;
+            state.in_flight.last().cloned()
+        };
+        if let Some(cb) = cb {
+            Self::ensure_completed(&cb)?;
+            // queue is FIFO: everything committed before cb is done too; keep errored for reporting
```
ivarflakstad: Not sure what "keep errored for reporting" means in this context

ivarflakstad (approved):
👍

candle-metal-kernels/src/metal/commands.rs
EricLBuehler: Yep, my mistake! It was from an earlier path I did

PR #3598, opened 2026-06-11
Parameter cache for CUDA kernel launch

guoqingbao:
**Summary**

Candle uses `htod` to copy kernel launch parameters to the device, which makes CUDA graph capture inefficient because it captures tens of host-to-device operations in each forward pass. This PR addresses the kernel launch bottleneck by caching kernel parameters, delivering ~10% performance speedup across **ALL** models, as validated in vLLM.rs (xinfer): https://github.com/guoqingbao/xinfer/pull/371.

**Usage:**

Before CUDA graph capture, enable the parameter cache and run a warmup step. This caches the kernel launch parameters:

```rust
let _guard = CudaDevice::enable_cuda_graph_htod_cache();
for b in 0..bs {
    let _ = self.model.forward(&input_ids_bs, ...)?;
}
```

Then perform the actual CUDA graph capture:

```rust

for b in 0..bs {
   self.model.start_capture(b)?;
    let out = self.model.forward(&input_ids_bs, ...)?;
    // Save out.
    self.model.end_capture()?;
}

```

**Results** from xinfer:

| Model | Quantization | Size | GPU | Conventional CUDA graph | CUDA graph + parameter cache | Speedup |
|---|---|---:|---|---:|---:|---:|
| Qwen3-30B-A3B | NVFP4 | 30B MoE | RTX 5090 | 175.30 tokens/s | 181.59 tokens/s | 3.59% |
| Gemma4-26B-A4B | NVFP4 | 26B MoE | RTX 5090 | 131.00 tokens/s | 137.23 tokens/s | 4.76% |
| Ministral-3-3B (Multimodal) | ISQ (BF16→Q4K) | 3B | A100 | 171.92 tokens/s | 193.67 tokens/s | 12.65% |
| DeepSeek-R1-0528-Qwen3-8B | Q4_K_M | 8B | A100 | 124.87 tokens/s | 139.25 tokens/s | 11.52% |
| Llama-3.1-8B | ISQ (BF16→Q4K) | 8B | A100 | 120.74 tokens/s | 133.10 tokens/s | 10.24% |
| Qwen3-VL-8B-Instruct (Multimodal) | Q8_0 | 8B | A100 | 105.31 tokens/s | 112.51 tokens/s | 6.84% |
| Qwen3.6-35B-A3B (Multimodal) | FP8 | 35B MoE | Hopper | 102.00 tokens/s | 110.00 tokens/s | 7.84% |
| Qwen3-30B-A3B | NVFP4 | 30B MoE | V100 | 67.10 tokens/s | 72.86 tokens/s | 8.58% |
| GLM-4-9B-0414 | Q4_K_M | 9B | A100 | 70.38 tokens/s | 77.48 tokens/s | 10.09% |
| MiniMax-M2.5 | NVFP4 | 229B MoE | Hopper ×2 | 62.00 tokens/s | 64.50 tokens/s | 4.03% |
| Qwen3.5-27B (Multimodal) | Q4_K_M | 27B Dense | Hopper | 45.20 tokens/s | 49.33 tokens/s | 9.14% |
| Qwen3.5-27B/Qwen3.6-27B | FP8 | 27B Dense | Hopper | 42.00 tokens/s | 45.00 tokens/s | 7.14% |
| QwQ-32B | Q4_K_M | 32B | A100 | 41.36 tokens/s | 46.02 tokens/s | 11.27% |
| Gemma4-31B | ISQ (BF16→Q4K) | 31B Dense | Hopper | 41.00 tokens/s | 47.00 tokens/s | 14.63% |

Note: This PR does not benefit `candle-examples` by default. It is intended for downstream projects that rely on Candle and support CUDA graphs.

### DIFF
--- a/candle-core/src/cuda_backend/mod.rs
+++ b/candle-core/src/cuda_backend/mod.rs
@@ -10,6 +10,9 @@ use cudarc::driver::{
     CudaSlice, DevicePtr, DeviceRepr, LaunchConfig, PushKernelArg, ValidAsZeroBits,
 };
 use half::{bf16, f16};
+use std::cell::Cell;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex, OnceLock};
 
 #[cfg(feature = "cudnn")]
 pub mod cudnn;
@@ -20,8 +23,35 @@ pub use device::{CudaDevice, DeviceId};
 pub use error::{CudaError, WrapErr};
 pub use utils::{Map1, Map1Any, Map2, Map2Any, Map2InPlace, Map3, S};
 
+static CUDA_PARAM_CACHE: OnceLock<Mutex<HashMap<(DeviceId, Vec<usize>), Arc<CudaSlice<usize>>>>> =
+    OnceLock::new();
+
+thread_local! {
+    static CUDA_PARAM_CACHE_ENABLED: Cell<bool> = const { Cell::new(false) };
+}
+
+pub struct CudaParamCacheGuard {
+    previous: bool,
+}
+
+impl Drop for CudaParamCacheGuard {
+    fn drop(&mut self) {
+        CUDA_PARAM_CACHE_ENABLED.with(|enabled| enabled.set(self.previous));
+    }
+}
+
+pub fn cuda_param_cache_scope(enabled: bool) -> CudaParamCacheGuard {
+    let previous = CUDA_PARAM_CACHE_ENABLED.with(|cache_enabled| {
+        let previous = cache_enabled.get();
+        cache_enabled.set(enabled);
+        previous
+    });
+    CudaParamCacheGuard { previous }
+}
+
 pub enum SlicePtrOrNull<T> {
     Ptr(CudaSlice<T>),
+    Cached(Arc<CudaSlice<T>>),
     Null,
 }
 
@@ -29,6 +59,7 @@ impl<T: DeviceRepr> SlicePtrOrNull<T> {
     pub fn builder_arg<'a, 'b: 'a>(&'b self, builder: &mut cudarc::driver::LaunchArgs<'a>) {
         match self {
             SlicePtrOrNull::Ptr(slice) => builder.arg(slice),
+            SlicePtrOrNull::Cached(slice) => builder.arg(slice.as_ref()),
             SlicePtrOrNull::Null => builder.arg(&0usize),
         };
     }
@@ -53,13 +84,33 @@ impl crate::scalar::Scalar {
 }
 
 impl SlicePtrOrNull<usize> {
+    pub fn params_from_vec(dev: &CudaDevice, params: Vec<usize>) -> Result<Self> {
+        if CUDA_PARAM_CACHE_ENABLED.with(|enabled| enabled.get()) {
+            let key = (dev.id(), params.clone());
+            let cache = CUDA_PARAM_CACHE.get_or_init(|| Mutex::new(HashMap::new()));
+            if let Some(slice) = cache.lock().unwrap().get(&key).cloned() {
+                return Ok(SlicePtrOrNull::Cached(slice));
+            }
+
+            let is_capturing = dev.cuda_stream().capture_status()
+                == Ok(cudarc::driver::sys::CUstreamCaptureStatus::CU_STREAM_CAPTURE_STATUS_ACTIVE);
+            if !is_capturing {
+                let slice = Arc::new(dev.clone_htod(&params)?);
+                let mut cache = cache.lock().unwrap();
+                let slice = cache.entry(key).or_insert_with(|| slice).clone();
+                return Ok(SlicePtrOrNull::Cached(slice));
+            }
+        }
+
+        Ok(SlicePtrOrNull::Ptr(dev.clone_htod(&params)?))
+    }
+
     pub fn params_from_layout(dev: &CudaDevice, l: &Layout) -> Result<Self> {
-        let ds = if l.is_contiguous() {
-            SlicePtrOrNull::Null
+        if l.is_contiguous() {
+            Ok(SlicePtrOrNull::Null)
         } else {
-            SlicePtrOrNull::Ptr(dev.clone_htod(&[l.dims(), l.stride()].concat())?)
-        };
-        Ok(ds)
+            Self::params_from_vec(dev, [l.dims(), l.stride()].concat())
+        }
     }
 }
 
@@ -187,7 +238,7 @@ impl Map1 for Im2Col1D {
         let l_out = self.l_out(dims[2]);
         let threads = dims[0] * l_out * dims[1];
         let cfg = LaunchConfig::for_num_elems(threads as u32);
-        let ds = dev.clone_htod(&[dims, layout.stride()].concat())?;
+        let ds = SlicePtrOrNull::params_from_vec(dev, [dims, layout.stride()].concat())?;
         let src = &src.slice(layout.start_offset()..);
         let func = dev.get_or_load_func(&kernel_name::<T>("im2col1d"), &kernels::CONV)?;
         // SAFETY: Set later by running the kernel.
@@ -199,7 +250,7 @@ impl Map1 for Im2Col1D {
         barg!(builder, self.stride);
         barg!(builder, self.padding);
         barg!(builder, self.dilation);
-        builder.arg(&ds);
+        ds.builder_arg(&mut builder);
         builder.arg(src);
         builder.arg(&dst);
         // SAFETY: ffi.
@@ -238,7 +289,7 @@ impl Map1 for Im2Col {
         let (h_out, w_out) = self.hw_out(dims[2], dims[3]);
         let dst_el = dims[0] * h_out * w_out * dims[1] * self.h_k * self.w_k;
         let cfg = LaunchConfig::for_num_elems(dst_el as u32);
-        let ds = dev.clone_htod(&[dims, layout.stride()].concat())?;
+        let ds = SlicePtrOrNull::params_from_vec(dev, [dims, layout.stride()].concat())?;
         let src = &src.slice(layout.start_offset()..);
         let func = dev.get_or_load_func(&kernel_name::<T>("im2col"), &kernels::CONV)?;
         // SAFETY: Set later by running the kernel.
@@ -252,7 +303,7 @@ impl Map1 for Im2Col {
         barg!(builder, self.stride);
         barg!(builder, self.padding);
         barg!(builder, self.dilation);
-        builder.arg(&ds);
+        ds.builder_arg(&mut builder);
         builder.arg(src);
         builder.arg(&dst);
         // SAFETY: ffi.
@@ -330,7 +381,8 @@ impl Map1Any for FastReduce<'_> {
             block_dim: (block_dim as u32, 1, 1),
             shared_mem_bytes: 0,
         };
-        let ds = dev.clone_htod(&[dims.as_slice(), stride.as_slice()].concat())?;
+        let ds =
+            SlicePtrOrNull::params_from_vec(dev, [dims.as_slice(), stride.as_slice()].concat())?;
         let src = &src.slice(layout.start_offset()..);
         let (name, check_empty, return_index) = match self.1 {
             ReduceOp::Sum => ("fast_sum", false, false),
@@ -350,7 +402,7 @@ impl Map1Any for FastReduce<'_> {
             barg!(builder, src_el);
             barg!(builder, el_to_sum_per_block);
             barg!(builder, src_dims.len());
-            builder.arg(&ds);
+            ds.builder_arg(&mut builder);
             builder.arg(src);
             builder.arg(&out);
             // SAFETY: ffi.
@@ -363,7 +415,7 @@ impl Map1Any for FastReduce<'_> {
             barg!(builder, src_el);
             barg!(builder, el_to_sum_per_block);
             barg!(builder, src_dims.len());
-            builder.arg(&ds);
+            ds.builder_arg(&mut builder);
             builder.arg(src);
             builder.arg(&out);
             // SAFETY: ffi.
@@ -429,7 +481,7 @@ impl Map1 for IndexSelect<'_> {
         };
         let ids_shape = ids_l.shape();
         let ids_dims = ids_shape.dims();
-        let ds = dev.clone_htod(&[ids_dims, ids_l.stride()].concat())?;
+        let ds = SlicePtrOrNull::params_from_vec(dev, [ids_dims, ids_l.stride()].concat())?;
         let src = match src_l.contiguous_offsets() {
             Some((o1, o2)) => src.slice(o1..o2),
             None => Err(crate::Error::RequiresContiguous { op: "index-select" }.bt())?,
@@ -446,7 +498,7 @@ impl Map1 for IndexSelect<'_> {
         let mut builder = func.builder();
         barg!(builder, dst_el);
         barg!(builder, ids_dims.len());
-        builder.arg(&ds);
+        ds.builder_arg(&mut builder);
         barg!(builder, ids);
         builder.arg(&src);
         builder.arg(&out);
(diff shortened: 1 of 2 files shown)


guoqingbao:
@EricLBuehler Do you have time to try this? I think you've supported cuda graph in mistral.rs recently?

candle-core/src/cuda_backend/mod.rs
```diff
@@ -20,15 +23,43 @@ pub use device::{CudaDevice, DeviceId};
 pub use error::{CudaError, WrapErr};
 pub use utils::{Map1, Map1Any, Map2, Map2Any, Map2InPlace, Map3, S};
 
+static CUDA_PARAM_CACHE: OnceLock<Mutex<HashMap<(DeviceId, Vec<usize>), Arc<CudaSlice<usize>>>>> =
```
EricLBuehler: This is a process-global state, maybe it should be at least thread-local or somewhat bounded with lifecycle, or some way to clear or manage it?

candle-core/src/cuda_backend/mod.rs
```diff
@@ -53,13 +84,33 @@ impl crate::scalar::Scalar {
...
 impl SlicePtrOrNull<usize> {
+    pub fn params_from_vec(dev: &CudaDevice, params: Vec<usize>) -> Result<Self> {
+        if CUDA_PARAM_CACHE_ENABLED.with(|enabled| enabled.get()) {
+            let key = (dev.id(), params.clone());
+            let cache = CUDA_PARAM_CACHE.get_or_init(|| Mutex::new(HashMap::new()));
+            if let Some(slice) = cache.lock().unwrap().get(&key).cloned() {
+                return Ok(SlicePtrOrNull::Cached(slice));
+            }
+
+            let is_capturing = dev.cuda_stream().capture_status()
```
EricLBuehler, line 72: This falls back to `dev.clone_htod(&params)` during capture if a param vector was not warmed up and cached. That means a missed warmup can still capture the host-to-device copy the PR is trying to eliminate.

Current main’s H2D cache catches this class of mistake by erroring on missing cached host data during capture, so maybe this path can do the same or reuse that mechanism?

EricLBuehler (changes requested):
Hey @guoqingbao! Thanks for the PR; I think it's a cool approach.

However, I merged #3564 a few weeks ago which seems to implement the same thing (cuda kernel param vec h2d copies), just differently. With my approach, we have a cache for small htod copies (up to 4096 bytes) during uploads, whereas this seems to be stricter and explicitly a cache for that case.

I think your approach is probably better and more clear, espeically since it forces us to be more correct.

However, I have some general points:
- I think the current method of `enable_cuda_graph_htod_cache` is better than setting `CUDA_PARAM_CACHE_ENABLED` manually. Internally, this is probably fine, but I think the API should be a bit cleaner.
- Additionally, since we have the chance, I think we should make `h2d` and `d2h` copies hard-error if we are in capture model. 

What do you think about these points?

guoqingbao:
> However, I have some general points:
> 
> * I think the current method of `enable_cuda_graph_htod_cache` is better than setting `CUDA_PARAM_CACHE_ENABLED` manually. Internally, this is probably fine, but I think the API should be a bit cleaner.
> * Additionally, since we have the chance, I think we should make `h2d` and `d2h` copies hard-error if we are in capture model.
> 
> What do you think about these points?

Hi Eric,

I think these are good suggestions. We can reuse the enable_cuda_graph_htod_cache mechanism instead of adding a new interface. And yes, we should make it a hard error if the parameter cache is enabled while we are in the capturing state. It may be better to start with a warning and progressively turn it into a hard error, which would help detect any violations gradually.

candle-core/src/cuda_backend/mod.rs
guoqingbao: We can try thread level cache or bounded with cuda stream.

candle-core/src/cuda_backend/mod.rs
guoqingbao: This is the intended design. It has no side effects if no warm-up is triggered. A warm-up is needed to trace all kernel launch parameters before capturing, but we can add a warning if warm-up is not executed before capturing.

EricLBuehler: I think warning is too weak during capture. A missed warmup means we either capture the H2D copy or silently lose the guarantee this PR is adding. Can we make capture-time cache misses return an error, like the existing H2D cache does?

candle-core/src/cuda_backend/mod.rs
EricLBuehler: Sounds good!

EricLBuehler:
Thanks, that sounds good to me.

One small clarification: I think we should only hard-error during capture when a host-to-device copy would miss the parameter cache and fall back to an actual h2d upload. I’d also prefer making it a hard error immediately rather than warning first. If warmup missed a param vector, the capture is already not doing what this PR is meant to guarantee, and warnings are easy to miss. This also matches the existing `enable_cuda_graph_htod_cache` behavior from #3564, where a cache miss during capture returns an error. For d2h, I think any copy during capture should hard-error, since there is no analogous cached-device-data path like the h2d parameter cache.

For the global state, thread-local sounds like the right direction, maybe reusing or integrating with the existing CUDA graph h2d cache mechanism. If this remains a separate param cache, it should have a bounded lifecycle/clear semantics, probably tied to the guard or device/stream.

guoqingbao:
> Thanks, that sounds good to me.
> 
> One small clarification: I think we should only hard-error during capture when a host-to-device copy would miss the parameter cache and fall back to an actual h2d upload. I’d also prefer making it a hard error immediately rather than warning first. If warmup missed a param vector, the capture is already not doing what this PR is meant to guarantee, and warnings are easy to miss. This also matches the existing `enable_cuda_graph_htod_cache` behavior from #3564, where a cache miss during capture returns an error. For d2h, I think any copy during capture should hard-error, since there is no analogous cached-device-data path like the h2d parameter cache.
> 
> For the global state, thread-local sounds like the right direction, maybe reusing or integrating with the existing CUDA graph h2d cache mechanism. If this remains a separate param cache, it should have a bounded lifecycle/clear semantics, probably tied to the guard or device/stream.

Thanks for the suggestion, will revise that accordingly.

guoqingbao:
> Thanks, that sounds good to me.
> 
> One small clarification: I think we should only hard-error during capture when a host-to-device copy would miss the parameter cache and fall back to an actual h2d upload. I’d also prefer making it a hard error immediately rather than warning first. If warmup missed a param vector, the capture is already not doing what this PR is meant to guarantee, and warnings are easy to miss. This also matches the existing `enable_cuda_graph_htod_cache` behavior from #3564, where a cache miss during capture returns an error. For d2h, I think any copy during capture should hard-error, since there is no analogous cached-device-data path like the h2d parameter cache.
> 
> For the global state, thread-local sounds like the right direction, maybe reusing or integrating with the existing CUDA graph h2d cache mechanism. If this remains a separate param cache, it should have a bounded lifecycle/clear semantics, probably tied to the guard or device/stream.

@EricLBuehler  Hi Eric, I've revised the implementation:

1) H2D cache miss during capture → hard error: If params_from_vec is called during active graph capture and the requested parameter vector is not already in the cache, it now returns an immediate error rather than falling through to an actual clone_htod. Cache hits during capture continue to work normally (returning the cached device pointer).

2) D2H during capture → hard error: Added a check at the top of to_cpu_storage() that errors immediately if a device-to-host copy is attempted while the stream is in capture mode (gated by the param cache being enabled, which signals we're in a CUDA graph workflow).

3) Bounded cache lifecycle: The cache entries are Arc-ref-counted, so device memory remains alive as long as any captured graph holds a reference via SlicePtrOrNull::Cached. An explicit clear_cuda_param_cache() function is provided for cleanup when rebuilding graphs or shutting down — clearing the map is safe even with live graphs since Arc refcounts keep the underlying device allocations alive. The guard's Drop restores the thread-local enabled flag (supporting nesting), ensuring non-graph code paths are unaffected.

candle-core/src/cuda_backend/mod.rs
```diff
@@ -1693,6 +1759,16 @@ impl BackendStorage for CudaStorage {
     }
 
     fn to_cpu_storage(&self) -> Result<CpuStorage> {
```
EricLBuehler, line 1731: This only prevents D2H through `to_cpu_storage`. However, other D2H paths like `CudaDevice::clone_dtoh` and memcpy_dtoh remain unchecked, and some CUDA code may use them directly.

I think it may be best to move this check and one for H2D to the `CudaDevice::clone_dtoh` layer? The H2D check in this PR is for just the params vec, while I think it should be at the `CudaDevice::clone_htod` level too, to catch any copies in either direction.

candle-core/src/cuda_backend/mod.rs
```diff
@@ -20,15 +23,50 @@ pub use device::{CudaDevice, DeviceId};
...
+    previous: bool,
+}
+
+impl Drop for CudaParamCacheGuard {
+    fn drop(&mut self) {
+        CUDA_PARAM_CACHE_ENABLED.with(|enabled| enabled.set(self.previous));
+    }
+}
+
+pub fn clear_cuda_param_cache() {
```
EricLBuehler: I think this is not safe with live captured graphs. `clear_cuda_param_cache` drops the map’s `Arc<CudaSlice<usize>>` entries, but CUDA graph capture only records the raw device pointer passed through `LaunchArgs`, not the Rust Arc.

So in the case where the cache is cleared while a graph that used those parameter slices is still executable, the graph can retain dangling device pointers. This is more of a lifetime issue than a direct soundness issue.

I think the fix should make the lifetime of cached parameter slices at least as long as any captured graph that references them. For example, instead of exposing a global `clear_cuda_param_cache()`, the cache ownership could be tied to a graph-capture/cache guard, and the captured graph wrapper would keep the relevant `Arc<CudaSlice<usize>>` values alive until the graph is dropped. Alternatively, if we cannot tie the slices to the graph object, we should not expose a clear API that can invalidate live graph pointers.

EricLBuehler (changes requested):
Thanks for the changes @guoqingbao! I think we're very close now. I added a few comments.

candle-core/src/cuda_backend/mod.rs
guoqingbao: Thanks, that makes sense. One thing this reminded me of is that the current change is mainly around parameter caching, which is host-to-device. With that scope in mind, I’m not sure we should block D2H at this layer.

The cache and H2D validation in this PR are currently tied to entries that go through `params_from_vec`. I agree there may be a broader question around whether copy validation should live lower down, e.g. in `CudaDevice::clone_dtoh` / `clone_htod`, so that all copy paths are covered. I’m just wondering if that should be handled separately from this PR.

Another reason I’m hesitant to move this directly to `clone_htod` is that the parameters handled here are usually small and reused frequently for kernel launches, which makes them a good fit for this cache. In contrast, `clone_htod` may also be used for larger transfers that are not necessarily suitable for this caching behavior.

candle-core/src/cuda_backend/mod.rs
guoqingbao: Thanks, that’s a good point. My original intention was for `clear_cuda_param_cache()` to be an explicit/manual cache-management API, where users are responsible for ensuring that any captured graphs using cached parameter slices are released before calling it.

That said, I agree this requirement is easy to miss and potentially unsafe. Since CUDA graph capture only records the raw device pointer and this layer does not currently have access to the graph objects, we cannot reliably tie the cached `Arc<CudaSlice<usize>>` lifetime to the captured graph lifetime here.

Given that, I think the safer fix for this PR is to remove the public `clear_cuda_param_cache()` API for now, instead of exposing an API that depends on users manually guaranteeing there are no live captured graphs.

EricLBuehler: Thanks, agreed. Removing `clear_cuda_param_cache()` addresses my concern here.

Longer term, if we want an explicit clear API, I think it needs to be tied to graph/cache ownership so the cached parameter slices live at least as long as any captured graph that references them. But removing the public clear API is the safer fix here, and if we need that API we can add it later 👍

candle-core/src/cuda_backend/mod.rs
EricLBuehler: Yeah, I think that distinction makes sense. I agree we should not cache arbitrary large `clone_htod` transfers in this parameter cache.

My concern is slightly different though: validation during graph capture does not have to mean caching every transfer. If we are actively capturing and a path would perform a real H2D/D2H copy, I think that should error rather than silently become part of the captured graph.

So I think we should either move/reuse the capture checks at the `CudaDevice::{clone_htod,memcpy_htod,clone_dtoh,memcpy_dtoh}` level. What do you think?

guoqingbao: > My concern is slightly different though: validation during graph capture does not have to mean caching every transfer. If we are actively capturing and a path would perform a real H2D/D2H copy, I think that should error rather than silently become part of the captured graph.
> 
> So I think we should either move/reuse the capture checks at the `CudaDevice::{clone_htod,memcpy_htod,clone_dtoh,memcpy_dtoh}` level. What do you think?

From the scope of the current PR, I think it may be safer to allow those outlier copies to remain captured in the graph for now. I did some profiling test in Candle-vLLM and found that the current strategy removes all H2D copies (they are all kernel-launch parameter copies). Ideally, there shouldn't be a D2H copy captured in the graph (this happen after forward pass, logits output using D2H, for sampling), as shown in the test statistics below:

**Qwen3.5 35B FP8**

```
[CUDA Graph Node Statistics]
========================================================================
**Before:**

BS=  1: Total=5067, Kernel=1975, Memcpy[H2D=241, D2H=0, D2D=0, other=0], Memset=251, MemAlloc=2437, MemFree=41, WaitEvent=1, EventRecord=121
========================================================================

**After:** (this PR)
BS=  1: Total=4585, Kernel=1975, Memcpy[H2D=0, D2H=0, D2D=0, other=0], Memset=251, MemAlloc=2196, MemFree=41, WaitEvent=1, EventRecord=121
```

I also tried to eliminate MemAlloc/MemFree graph nodes by caching those allocations, but saw no obvious benefit since CUDA already handles them efficiently through its memory pool during graph replay.

EricLBuehler: Thanks, that data is helpful. I agree CUDA graphs can technically contain memcpy nodes, so this is more about the invariant we want for Candle’s graph-capture mode than about CUDA graph validity in general.

My preference is that when the CUDA graph/cache guard is enabled, any real H2D/D2H copy during active capture should fail fast unless it goes through an explicit supported path, like the warmed H2D cache for params upload. Otherwise, a missed warmup or an unexpected host/device transfer can silently become part of the captured graph.

So I’m fine not caching arbitrarily large `clone_htod` transfers, but I still think validation belongs at the lower copy APIs. `clone_htod` / `memcpy_htod` can use the existing cached path when appropriate and error on uncached capture-time copies; `clone_dtoh` / `memcpy_dtoh` should probably hard-error during capture since there is no equivalent cached D2H path.

If we later want to support intentional memcpy nodes in captured graphs, I think that should be an explicit opt-in API with a different contract rather than the default behavior.

What do you think?

guoqingbao: Right, I've moved the validation down to CudaDevice::{clone_htod, memcpy_htod, clone_dtoh, memcpy_dtoh} in the new commit.

EricLBuehler: Thanks, moving the validation down to `CudaDevice` is the right direction.

  I think there is still one issue though: the new D2H checks are gated on `CUDA_PARAM_CACHE_ENABLED`, while the H2D checks still come from the separate
  `enable_cuda_graph_htod_cache()` state. So the capture-safety behavior is split across two independent guards.

  Could we unify this so there is one graph/cache guard controlling both:
  - param cache usage / H2D cache-miss behavior
  - D2H hard errors during active capture

  My preference would be to reuse `CudaDevice::enable_cuda_graph_htod_cache()` as that single guard, and have the new param cache and D2H checks key off the same state.

guoqingbao: Unified — removed the separate CudaParamCacheGuard and made both the param cache and D2H capture checks key off the existing enable_cuda_graph_htod_cache() guard, so there's now a single state controlling all capture-safety behavior.

EricLBuehler (approved):
Thank you @guoqingbao! Looks good 🙌

PR #3600, opened 2026-06-11
Optimize CUDA kernels

guoqingbao:
## Summary

This PR adds another round of CUDA kernel optimizations for binary ops, cast ops, reductions, and selected unary ops.

The main focus is improving hot contiguous BF16/F32 paths by optimizing cuda kernels with specialized vectorized implementations, reducing indexing overhead, and using more efficient CUDA primitives where appropriate.

Combined, these CUDA kernel optimizations deliver up to **15% end-to-end speedup** in various models.

### Binary Ops

Updated `binary.cu` and `binary_op_macros.cuh`.

- Optimized BF16 add/mul:
  - `badd_bf16`
  - `bmul_bf16`
- Replaced generic implementations with vectorized kernels for contiguous inputs.
- Contiguous BF16 kernels now use `float4` loads, processing 8 BF16 elements per load.
- BF16 add/mul use `__hadd2` and `__hmul2` intrinsics.
- Added semi-contiguous non-contiguous paths for LHS-only or RHS-only strided inputs.
- Optimized F32 add/div/mul:
  - `badd_f32`
  - `bdiv_f32`
  - `bmul_f32`
- F32 contiguous kernels now use `float4` loads, processing 4 F32 elements per load.
- Added `BINARY_OP_BF16_VEC` in `binary_op_macros.cuh` for reusable vectorized BF16 binary ops with float-promoted per-element computation.

### Cast Ops

Updated `cast.cu`.

- Optimized `cast_bf16_f32` with vectorized contiguous loads.
- BF16 to F32 conversion now uses `float4` loads, processing 8 BF16 values per load.
- Conversion is unrolled via `__bfloat162float`.
- Optimized `cast_f32_bf16` with vectorized contiguous loads.
- F32 to BF16 conversion now uses `float4` loads, processing 4 F32 values per load.
- Conversion uses `__float2bfloat16_rn`.

### Reduce Ops

Updated `reduce.cu` and `mod.rs`.

- Reworked the `fast_sum` template.
- Replaced shared-memory tree reduction with a warp-shuffle two-phase reduction using `__shfl_xor_sync`.
- Added a contiguous fast path that skips `get_strided_index`.
- Accumulation now uses `float` for better precision with half-precision input types.
- Added `fast_sum_bf16_vec`, a specialized BF16 vectorized reduce kernel using:
  - `float4` loads
  - 8 BF16 elements per load
  - warp-shuffle reduction
- Added an explicit optimized `fast_sum_f32` path using:
  - `float4` loads
  - 4 F32 elements per load
  - warp-shuffle reduction
- Replaced the previous generic `FAST_OP` macro path for F32 sum.
- Added `fast_sum_small` kernels for small reductions where `el_to_sum <= 32`.
- Added specialized small-reduction variants for:
  - BF16
  - F16
  - F32
  - F64
- Updated Rust dispatch in `mod.rs` with `use_small_reduce` logic.
- Sum reductions with `el_to_sum <= 32` now dispatch to `fast_sum_small` using 256-thread blocks instead of launching one block per output element.

### Unary Ops

Updated `unary.cu`.

- Optimized `ucopy_bf16` with vectorized contiguous copies using `float4` loads.
- Optimized `usilu_bf16` with vectorized contiguous processing:
  - 8 BF16 elements per thread iteration
  - float-promoted computation
  - computes `x / (1 + exp(-x))`
- Optimized `usigmoid_bf16` with vectorized contiguous processing:
  - 8 BF16 elements per thread iteration
  - float-promoted computation
  - computes `1 / (1 + exp(-x))`

## Expected Impact

These changes should improve throughput for common CUDA workloads, especially contiguous BF16 and F32 operations.

The largest expected gains are in:

- BF16 add/mul
- F32 add/div/mul
- BF16/F32 cast operations
- Contiguous sum reductions
- Small sum reductions with `el_to_sum <= 32`
- BF16 copy, SiLU, and sigmoid

The optimizations reduce scalar memory access, avoid unnecessary stride indexing on contiguous paths, reduce shared-memory synchronization overhead in reductions, and improve utilization through vectorized loads.


candle-kernels/src/reduce.cu
```diff
@@ -647,13 +724,145 @@ fast_argmax(const size_t src_numel, const size_t el_to_sum_per_block,
...
 
+// Small-reduce kernel: one thread per output element, for el_to_sum <= 32.
+// Eliminates massive block-count overhead when reducing over tiny dimensions (e.g., MoE topk=8).
+template <typename T>
+__device__ void
+fast_sum_small_impl(const size_t src_numel, const size_t el_to_sum_per_block,
+                    const size_t num_dims, const size_t *info, const T *src, T *dst) {
+  const size_t *dims = info;
+  const size_t *strides = info + num_dims;
+  const size_t dst_el = src_numel / el_to_sum_per_block;
```
EricLBuehler, line 735: I think here we should handle empty reductions before dividing by the reduction length

A valid empty view such as `Tensor::new(&[[1f32, 2.]; 4], &device)?.narrow(1, 0, 0)?.sum(1)?` reaches this kernel with both `src_numel` and `el_to_sum_per_block` equal to zero. This calculation therefore divides by zero instead of producing the zero outputs that the previous kernel returned. The undefined output bound also risks OOB writes. The BF16 specialization has the same issue I think.

candle-kernels/src/reduce.cu
```diff
@@ -647,13 +724,145 @@ fast_argmax(const size_t src_numel, const size_t el_to_sum_per_block,
...
+  size_t gid = blockIdx.x * blockDim.x + threadIdx.x;
+  if (gid >= dst_el) return;
+
+  T sum = 0;
+
+  if (is_contiguous(num_dims, dims, strides)) {
+    size_t start = gid * el_to_sum_per_block;
+    for (size_t i = 0; i < el_to_sum_per_block; ++i) {
+      sum += src[start + i];
```
EricLBuehler, line 744: I think here it would be best to use a float accumulator for FP16 and convert once when storing the result, as the BF16 specialization already does, for numerical stability.

candle-kernels/src/unary.cu
```diff
@@ -97,8 +97,93 @@ __device__ T sign_(T t) {
...
+                for (int j = 0; j < 8; j++) { \
+                    float xf = __bfloat162float(vp[j]); \
+                    vp[j] = __float2bfloat16(FLOAT_FUNC); \
+                } \
+                out4[i] = v; \
+            } \
+            const size_t tail_start = vec_numel * 8; \
+            for (unsigned int i = tail_start + blockIdx.x * blockDim.x + threadIdx.x; i < numel; i += blockDim.x * gridDim.x) { \
+                __nv_bfloat16 x = inp ? inp[i] : out[i]; \
+                out[i] = SCALAR_FUNC; \
```
EricLBuehler, line 132: The vector loop evaluates the activation in float, but this tail and the scalar fallbacks still use BF16 intermediate arithmetic. I think it would be best to keep things in float and only convert when necessary, as possible for numerical stability.

EricLBuehler (commented):
Hey @guoqingbao! Thanks for the PR. I've attached some comments, but I dont see anything critically wrong.

Aside from this and the merge conflicts I think it looks great.

guoqingbao:
Hi @EricLBuehler, I’ve addressed the review comments in the latest two commits. Empty reductions are now handled safely, FP16 reductions accumulate in float, and BF16 unary fallback paths compute in float before conversion.

candle-kernels/src/reduce.cu
guoqingbao: Addressed in fast_sum_small_f16 specification.

EricLBuehler (approved):
Looks great. Thank you @guoqingbao!

PR #3627, opened 2026-06-17
[CPU] Persistent barrier based threadpool and improved qgemm

ivarflakstad:
This PR adds 3 things:

1. `GgmlType::vec_dot_4` and `::vec_dot_2` with default impls. Let's us load RHS (LHS activation) once per 4 vec dot calls.
2. LHS scratch buffer so we do not have to heap allocate a new vec each call.
3. Persistent threadpool, which is used in quantized matmul instead over rayon.

(`../poem.txt` is just 23 lines of previous output)
```
cargo run --example quantized-qwen3 --release -- --prompt "$(cat ../poem.txt)" --which 0.6b --sample-len 350 --temperature=0 --repeat-penalty=1.0

main
 331 prompt tokens processed: 177.43 token/s
 296 tokens generated: 86.40 token/s

4 x vec_dot per LHS
 331 prompt tokens processed: 231.06 token/s
 296 tokens generated: 97.38 token/s

BarrierPool (instead of rayon)
 331 prompt tokens processed: 285.40 token/s
 296 tokens generated: 110.25 token/s

This + recently merged improvements
 331 prompt tokens processed: 361.25 token/s
 296 tokens generated: 120.00 token/s
```


candle-core/src/utils.rs
```diff
@@ -1,19 +1,250 @@
...
+                    .name(format!("candle-bp-{tid}"))
+                    .spawn(move || {
+                        set_thread_affinity();
+                        let inner = unsafe { &*(inner_ptr as *const BarrierPoolInner) };
+                        let slot = &inner.slots[tid];
+                        let mut gen = 0usize;
+
+                        loop {
+                            // Continuously check `slot.go` for next work item.
+                            loop {
```
EricLBuehler, line 140: I think this may allow idle worker threads to spin forever. Since `barrier_pool()` stores the pool in a static `OnceLock`, after the first quantized matmul calls it, the worker threads sit in the `slot.go` loop calling `spin_loop()` until the next job. Since the static pool is never dropped, stop is never set in normal use.

That means one quantized matmul can leave `get_num_threads() - 1` CPU threads burning cores for the rest of the process. This likely needs parking/backoff, a condvar, or maybe reuse of Rayon-style sleeping workers.

candle-core/src/utils.rs
```diff
@@ -1,19 +1,250 @@
...
+            call_lock: Mutex::new(0),
+        }
+    }
+
+    #[inline]
+    pub fn n_workers(&self) -> usize {
+        self.threads.len()
+    }
+
+    pub fn execute<F: Fn(usize) + Sync>(&self, f: F) {
```
EricLBuehler, line 246: I think this is not panic-safe. If a worker closure panics, that worker exits before `slot.done.store()`, and the caller spins forever. There is also a main-thread unwind case, as `execute` stores a raw pointer to the borrowed closure in `work`, then runs workers concurrently. If `f(n)` panics on the caller thread, it can unwind while workers are still using that borrowed closure/scratch state.

Essentially, I think we should catch/propagate panics or otherwise guarantee the barrier is completed before unwinding.

EricLBuehler (changes requested):

candle-core/src/utils.rs
ivarflakstad: Fair point. I'll try adding a spin limit and use `std::thread::park` if the limit is reached

candle-core/src/utils.rs
ivarflakstad: Good catch!
I'll use `std::panic::catch_unwind` around the closure. We can save the panic for after we signal `done`, and then propagate it to the main thread.

candle-core/src/utils.rs
```diff
@@ -1,19 +1,320 @@
...
+
+                            // Reduce: wait for all children before signalling parent.
+                            for &child in &children {
+                                while inner.slots[child].done.load(Ordering::Acquire) != gen {
+                                    spin_loop();
+                                }
+                            }
+
+                            // Signal done unconditionally so main/parent never deadlocks.
+                            slot.done.store(gen, Ordering::Release);
```
EricLBuehler, line 190: I think we may have a race condition here w/ the panic handling. `BarrierGuard::drop` waits for done, then immediately takes `panic_payload`. So main can observe done before the worker stores the panic payload. That can miss the panic or leave it to be propagated on a later execute. I think we should signal done after the panic payload is taken?

EricLBuehler (changes requested):
Just one small thing, looks good otherwise!

EricLBuehler (approved):
Looks good! I think CI is failing though

PR #3629, opened 2026-06-18
[CPU] Add neon optimized path to `BlockQ8K::from_float`

ivarflakstad:
```
before
  14 prompt tokens processed: 166.36 token/s
 495 tokens generated: 100.58 token/s

after
  14 prompt tokens processed: 175.61 token/s
 361 tokens generated: 112.47 token/s
```


candle-core/src/quantized/neon.rs
```diff
@@ -601,6 +601,74 @@ pub(crate) fn vec_dot_q2k_q8k(n: usize, xs: &[BlockQ2K], ys: &[BlockQ8K]) -> f32
...
+            let mut out = y.qs.as_mut_ptr();
+            let mut p = chunk.as_ptr();
+            for _ in 0..QK_K / 16 {
+                let f0 = vmulq_f32(vld1q_f32(p), vscale);
+                let f1 = vmulq_f32(vld1q_f32(p.add(4)), vscale);
+                let f2 = vmulq_f32(vld1q_f32(p.add(8)), vscale);
+                let f3 = vmulq_f32(vld1q_f32(p.add(12)), vscale);
+                p = p.add(16);
+                let s01 = vcombine_s16(
+                    vqmovn_s32(vcvtnq_s32_f32(f0)),
```
EricLBuehler: `vcvtnq_s32_f32` rounds ties to even, while the scalar path uses `f32::round` which rounds away from 0. This small divergence may introduce some error which would be platform-dependent.

I think using `vcvtaq_s32_f32` would match the rounding behavior of `f32_round`

Rounding behavior:
- https://doc.rust-lang.org/core/arch/aarch64/fn.vcvtnq_s32_f32.html
- https://doc.rust-lang.org/core/arch/aarch64/fn.vcvtaq_s32_f32.html
- https://doc.rust-lang.org/std/primitive.f32.html#method.round

candle-core/src/quantized/neon.rs
```diff
@@ -601,6 +601,74 @@ pub(crate) fn vec_dot_q2k_q8k(n: usize, xs: &[BlockQ2K], ys: &[BlockQ8K]) -> f32
...
+            let mut vmax3 = vdupq_n_f32(0.0f32);
+            let mut p = chunk.as_ptr();
+            for _ in 0..QK_K / 16 {
+                vmax0 = vmaxq_f32(vmax0, vabsq_f32(vld1q_f32(p)));
+                vmax1 = vmaxq_f32(vmax1, vabsq_f32(vld1q_f32(p.add(4))));
+                vmax2 = vmaxq_f32(vmax2, vabsq_f32(vld1q_f32(p.add(8))));
+                vmax3 = vmaxq_f32(vmax3, vabsq_f32(vld1q_f32(p.add(12))));
+                p = p.add(16);
+            }
+            let amax = vmaxvq_f32(vmaxq_f32(vmaxq_f32(vmax0, vmax1), vmaxq_f32(vmax2, vmax3)));
```
EricLBuehler: I think this will lose the sign of the max-magnitude value. Scalar Q8K uses the signed max for `iscale` (see `let iscale = -128f32 / max;` in `BlockQ8K::from_float`), and if the largest magnitude value is negative, this computes `iscale = -128 / abs(max)`, so that element quantizes to +128 and saturates to +127 (scalar is correct with -128).

I think a fix would be to track the signed value associated with the maximum absolute value, like the scalar loop does, and use that signed `max` for `iscale`. For example, keep both `vabs` and original-value vectors and select the original lane when its abs value wins.

EricLBuehler (changes requested):
Looks mostly good, just 2 comments

candle-core/src/quantized/neon.rs
ivarflakstad: Hmm. I made both `d = 1/iscale` and the quantized values have flipped sign relative to the scalar impl when amax's origin is negative, and was fairly certain that was correct.
Now I'm not certain we're fully covered though, I'll have a look.

ivarflakstad: Wouldn't this have led to fairly significant errors?

EricLBuehler: Yeah, I think you're right that this would not be a huge sign-inversion error for most lanes: if both `d` and `qs` flip sign, the dequantized values mostly cancel out.

The case I’m worried about is the asymmetric `i8` range at the very edges. The scalar path uses the signed `max`, so the max-magnitude lane maps to `-128`. With the abs-only scale, if the max-magnitude lane is negative, that lane maps to `+128` and then saturates to `+127`.

For example `[-1.0, 0.5]` gives roughly:
- scalar: `d = 1/128`, `qs = [-128, 64]`, dequant `[-1.0, 0.5]`
- abs-only: `d = -1/128`, `qs = [127, -64]`, dequant `[-0.9921875, 0.5]`

So I agree this is probably small, but it still means the NEON path is not bit/quantization-equivalent to scalar, and `bsums` inherit the same off-by-one.

Could we either track the signed max like the scalar path? Alternatively, maybe we can include edge cases like this in our tests so that we can make sure we capture them in the future? But maybe I missed something of course!

EricLBuehler (approved):

PR #3632, opened 2026-06-19
Metal SDPA: support combining an additive mask with causal masking

astorise:
call_sdpa_full previously forced do_causal=false whenever a mask buffer was present, to avoid a 0/0 = NaN when a fully-masked query row reached the final softmax divide. That silently dropped causal masking for callers passing a padding/bias mask together with do_causal=true, letting queries attend to future keys.

Fix the NaN at its source instead: clamp a zero softmax normalizer to 1 before the final divide in the steel attention kernel, so a fully-masked row yields a zeroed output row. With that guard in place, the additive mask and do_causal can be applied together, so the wrapper no longer needs to drop causal masking.

Add regression tests covering (1) a non-causal padding mask combined with do_causal, and (2) a fully-masked query row staying finite (zero).

Claude-Session: https://claude.ai/code/session_01BB5dWNeZGmPzWtyRpBEdMM

### DIFF
--- a/candle-metal-kernels/src/kernels/sdpa.rs
+++ b/candle-metal-kernels/src/kernels/sdpa.rs
@@ -109,11 +109,12 @@ pub fn call_sdpa_full(
     let align_k = (kl % bk) == 0;
     let has_mask = mask_buffer.is_some();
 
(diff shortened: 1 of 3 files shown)


astorise:

PR #3634, opened 2026-06-20
[CPU] Remove flash attn specific threadpool

ivarflakstad:


EricLBuehler (approved):
LGTM!

EricLBuehler:
Do we want to make it use the Candle threadpool instead of the Rayon global threadpool? Maybe through `Device::with_context`, or internally, `candle_pool().install(f)`?

ivarflakstad:
`Device::with_context` should wrap the entire inference pipeline so that there is as little rayon overhead as possible. Installing within an op causes fairly significant performance degradation unfortunately.

PR #3635, opened 2026-06-20
Render chat templates via hf-chat-template for transformers parity

GregoryBolshakov:
### The decision

`candle-examples` renders chat templates itself, and for many real models the output does not match `transformers.apply_chat_template`. This PR fixes that by rendering through a small crate, [`hf-chat-template`](https://crates.io/crates/hf-chat-template), instead of the hand-rolled renderer. **The one call to make is whether you are OK taking that dependency in `candle-examples`** (it replaces the existing `minijinja` dep there, which nothing else uses). If you would rather vendor the logic, say so and I will adapt the PR.

The crate is MIT/Apache-2.0; its deps are `minijinja` (+ `minijinja-contrib`) and `serde`/`serde_json`. It only renders a string; it does not pull in a tokenizer or runtime.

### Why the current renderer needs fixing

It builds a bare `minijinja::Environment` and registers only `raise_exception`, so it errors or silently diverges from `transformers` on:

- Python string methods on values (`strip`, `split`, `startswith`, `title`): Qwen3, QwQ, Command-R, Falcon, OpenChat all error with `unknown method`.
- `strftime_now`: Granite errors with `unknown function`.
- `trim_blocks` / `lstrip_blocks`: `transformers` sets both; without them, templates that put tags on their own lines (Zephyr) gain stray newlines. The current `result.trim_start()` only hides the leading case.
- Named sub-templates (`tool_use`, `rag`) and the `tools` argument have no path through the API.

### Result

Same models through both renderers, diffed against `transformers`, over the cases this module's API can express:

| | byte-identical | wrong (error or silent) |
|---|---|---|
| before | 35 | 19 |
| after | 52 | 0* |

\* The two non-matches left are `strftime_now` date stamps, which differ only by wall-clock instant.

### What changes

- Rendering is delegated to `hf-chat-template`. The module's public API (`Message`, `ChatTemplate`, `ChatTemplateOptions`, `Conversation`, the presets) is unchanged, so `smollm3` and the other examples need no edits. The existing `chat_template` tests still pass and a byte-exact Qwen2.5 test is added.
- Second commit: boxes a large enum variant in the `smollm3` example, which `clippy` 1.96 now flags with `large_enum_variant` (unrelated to the above, folded in so CI is green; happy to split out).

I could not run the full `candle-examples` build locally (an unrelated `openssl-sys` system dep), so the module and its tests were validated against the real crate in isolation. A natural follow-up is to widen the surface to pass `tools`/documents and select named sub-templates, which the backing crate already supports.

### DIFF
--- a/candle-examples/Cargo.toml
+++ b/candle-examples/Cargo.toml
@@ -26,7 +26,7 @@ hf-hub = { workspace = true, features = ["tokio"] }
 image = { workspace = true }
 intel-mkl-src = { workspace = true, optional = true }
 num-traits = { workspace = true }
-minijinja = { version = "2", features = ["loader"] }
+hf-chat-template = "0.2"
 palette = { version = "0.7.6", optional = true }
 enterpolation = { version = "0.3.0", optional = true }
 pyo3 = { version = "0.27", features = [
--- a/candle-examples/Cargo.toml
+++ b/candle-examples/Cargo.toml
@@ -26,7 +26,7 @@ hf-hub = { workspace = true, features = ["tokio"] }
 image = { workspace = true }
 intel-mkl-src = { workspace = true, optional = true }
 num-traits = { workspace = true }
-hf-chat-template = "0.2"
+hf-chat-template = "1.0"
 palette = { version = "0.7.6", optional = true }
 enterpolation = { version = "0.3.0", optional = true }
 pyo3 = { version = "0.27", features = [
--- a/candle-examples/examples/smollm3/main.rs
+++ b/candle-examples/examples/smollm3/main.rs
@@ -27,7 +27,7 @@ const DEFAULT_PROMPT: &str = "Write a Rust function to calculate the factorial o
 
 enum SmolLM3Model {
     Quantized(QuantizedModelForCausalLM),
-    Full(ModelForCausalLM, Config), // Store config alongside model
+    Full(ModelForCausalLM, Box<Config>), // Store config alongside model
 }
 
 impl SmolLM3Model {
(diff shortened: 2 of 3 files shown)


GregoryBolshakov:
@EricLBuehler would you be open to this? The change is byte-for-byte chat-template parity with `transformers` (the current renderer errors or silently diverges on Qwen3/QwQ/Command-R/Granite/Zephyr), and the module's public API is unchanged. The one real decision is taking the `hf-chat-template` dependency in `candle-examples` in place of the direct `minijinja` dep.

GregoryBolshakov:
I opened another PR for the failing Clippy CI at https://github.com/huggingface/candle/pull/3636. It's unrelated to the chat template change.

GregoryBolshakov:
Tidied this up: folded the clippy fix (previously #3636) in as a second commit and rebased on main, so it is now a single self-contained change. Rewrote the description to lead with the actual decision: whether you are open to taking the `hf-chat-template` dependency in `candle-examples`, or would prefer the logic vendored. CI looks like it is waiting on the first-time-contributor approval to run.

GregoryBolshakov:
CI here is red only on the Clippy job — Rust 1.97 promoted a few lints to errors across the whole workspace (unrelated to this PR). Fix is in #3754; once that lands I'll rebase and this goes green.

GregoryBolshakov:
@EricLBuehler #3754 has landed and I've merged main in, so the clippy failure from last time is gone, CI just needs a workflow approval to run again. Ready for a look whenever you have a chance.

GregoryBolshakov:
@ivarflakstad could you take a look at this one, or point me to who should? You merged my clippy fix in #3754, so I hope it is fine to ask you here.

It is a bug fix in candle-examples. The current renderer errors or silently diverges from transformers on Qwen3, QwQ, Command-R, Granite and Zephyr.

Two things changed since I opened it. The crate is now 1.0, so its API is frozen and no release of it can break candle before a 2.0. It also has no third-party types in its public API at all, so it cannot force a version of anything into your tree. I just pushed a commit moving the pin from 0.2 to 1.0, and no code change was needed. The same rendering bug also turned up in kalosm and a fix landed there last week, so it is not only a candle problem.

The last full CI run on this branch passed every job. The bump I just pushed needs a workflow approval before the checks can run again.

The only real decision is the dependency. It is a swap, not an addition. candle-examples drops the direct minijinja dep and takes hf-chat-template instead, which brings minijinja anyway.

PR #3636, opened 2026-06-20
Box the Config field in smollm3's model enum

GregoryBolshakov:
clippy 1.96 (current `stable`) flags `SmolLM3Model` with `large_enum_variant`:

```
error: large size difference between variants
  --> candle-examples/examples/smollm3/main.rs:28:1
   = the largest variant contains at least 456 bytes
   = the second-largest variant contains at least 216 bytes
```

This fails `cargo clippy --workspace --tests --examples --benches -- -D warnings` on any PR. Box the `Config` field in the `Full` variant, as clippy's machine-applicable suggestion proposes; the only construction site and the `cfg` field accesses are unaffected.


GregoryBolshakov:
Folded into #3635 as a second commit so that PR has green CI in a single change. Closing this one.

PR #3643, opened 2026-06-23
[CPU] Quantized perf improvements

ivarflakstad:
On my M4 this actually brings us to decode performance parity (or better) with llama.cpp on Qwen3-0.6B-Q4_K_M and Qwen3-0.6B-Q8_0.

This actually fixes a longstanding discrepancy between candle and llama.cpp regarding how we use `iscale = -128 / max` in q8k quantization when llama.cpp has used `-127` for quite a while [ref](https://github.com/ggml-org/llama.cpp/commit/7ef6389694d188e2ec4ef103b69bc07359d718ad).

### DIFF
--- a/candle-core/src/quantized/k_quants.rs
+++ b/candle-core/src/quantized/k_quants.rs
@@ -189,6 +189,20 @@ pub struct BlockQ8K {
 }
 const _: () = assert!(4 + QK_K + QK_K / 16 * 2 == std::mem::size_of::<BlockQ8K>());
 
+/// 8 Q4K blocks packed in interleaved format facilitating 8-column GEMV.
+/// Currently only compiled on AArch64 (with dotprod enabled).
+#[cfg(all(target_arch = "aarch64", target_feature = "dotprod"))]
+#[derive(Clone)]
+#[repr(C)]
+pub(crate) struct BlockQ4Kx8 {
+    pub(crate) d: [f16; 8],
+    pub(crate) dmin: [f16; 8],
+    pub(crate) scales: [u8; 96],
+    pub(crate) qs: [u8; 1024],
+}
+#[cfg(all(target_arch = "aarch64", target_feature = "dotprod"))]
+const _: () = assert!(std::mem::size_of::<BlockQ4Kx8>() == 1152);
+
 impl GgmlType for BlockQ4_0 {
     const DTYPE: GgmlDType = GgmlDType::Q4_0;
     const BLCK_SIZE: usize = QK4_0;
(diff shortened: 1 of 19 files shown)


candle-core/src/quantized/k_quants.rs
```diff
@@ -2408,6 +2471,158 @@ pub fn matmul<T: GgmlType>(
...
+/// Pack Q4K blocks into the 8-column interleaved format for 8 x GEMV
+#[cfg(all(target_arch = "aarch64", target_feature = "dotprod"))]
+pub(crate) fn pack_to_q4kx8(blocks: &[BlockQ4K], n: usize) -> Vec<u8> {
+    debug_assert!(n.is_multiple_of(8));
+    debug_assert_eq!(blocks.len() % n, 0);
+    let k_blocks = blocks.len() / n;
+    let n_groups = n / 8;
+    let block_bytes = std::mem::size_of::<BlockQ4Kx8>();
+    let mut out = vec![0u8; n_groups * k_blocks * block_bytes];
+    let out_ptr = out.as_mut_ptr() as *mut BlockQ4Kx8;
```
EricLBuehler: Could this be `Vec<BlockQ4Kx8>` instead? `Vec<u8>` does not guarantee `BlockQ4Kx8` alignment, so creating `BlockQ4Kx8` references/slices from this buffer is technically relying on allocator behavior.

I think this function should return `Vec<BlockQ4Kx8>` to make this sound?

ivarflakstad: Very good point.
I kind of want the output to be `BlockQ4Kx8` as well, just to make it completely obvious what is going on. Then transmute to `Vec<u8>` when we store it in `repacked_qs`

ivarflakstad: Think I managed a fairly sleek solution

candle-core/src/quantized/k_quants.rs
```diff
@@ -2408,6 +2471,182 @@ pub fn matmul<T: GgmlType>(
...
+    debug_assert!(n.is_multiple_of(8));
+    debug_assert_eq!(blocks.len() % n, 0);
+    let k_blocks = blocks.len() / n;
+    let n_groups = n / 8;
+    let count = n_groups * k_blocks;
+    let mut out: Vec<BlockQ4Kx8> = Vec::with_capacity(count);
+    let out_ptr = out.as_mut_ptr();
+    for g in 0..n_groups {
+        for b in 0..k_blocks {
+            let dst = unsafe { &mut *out_ptr.add(g * k_blocks + b) };
```
EricLBuehler: (nit) This forms a `&mut BlockQ4Kx8` before the element is initialized.

However, given the loop invariants this is easy to make fully sound by using a raw pointer write pattern here instead of forming a reference first.

candle-core/src/quantized/k_quants.rs
```diff
@@ -2408,6 +2471,182 @@ pub fn matmul<T: GgmlType>(
...
+    const { assert!(std::mem::align_of::<T>() <= 4) };
+
+    let ptr = v.as_mut_ptr() as *mut u8;
+    let len = v.len() * std::mem::size_of::<T>();
+    let cap = v.capacity() * std::mem::size_of::<T>();
+
+    // Prevent the original Vec from freeing the memory buffer
+    std::mem::forget(v);
+
+    unsafe { Vec::from_raw_parts(ptr, len, cap) }
```
EricLBuehler: I think this is unsound, subtly. This changes the owner from `Vec<BlockQ4Kx8>` to `Vec<u8>`. Since the `Vec<u8>` will eventually deallocate the buffer with `u8` layout/alignment, this violates `Vec::from_raw_parts`’ same-layout requirement.

Can we keep `repacked_qs` typed as `Vec<BlockQ4Kx8>` instead of converting it back to bytes, and then do casts as necessary to convert?

ivarflakstad: Yeah `Vec::from_raw_parts` says layout must be the same - the docs even states 
> `T` having a less strict alignment is not sufficient,

I made it copy the data to get rid of the UB.

> Can we keep repacked_qs typed as Vec<BlockQ4Kx8>

That would mean that `repacked_qs` can only ever be used with `Q4K`. Repacking is afaik useful for several quantizations.

candle-core/src/quantized/k_quants.rs
ivarflakstad: Good point. Updated to normal struct and vec initialization.

candle-core/src/quantized/k_quants.rs
```diff
@@ -2408,6 +2471,172 @@ pub fn matmul<T: GgmlType>(
     })
 }
 
+/// Copy a `Vec<T>` to `Vec<u8>`, consuming the input.
+#[allow(dead_code)]
+pub(crate) fn vec_to_bytes<T: Copy>(v: Vec<T>) -> Vec<u8> {
+    let len = v.len() * std::mem::size_of::<T>();
+    unsafe { std::slice::from_raw_parts(v.as_ptr() as *const u8, len) }.to_vec()
```
EricLBuehler: I think this still has the original alignment issue. `vec_to_bytes` now fixes the `Vec::from_raw_parts` deallocation UB, but `repacked_bytes` is still owned as `Vec<u8>`, so this cast to `&[BlockQ4Kx8]` is not guaranteed aligned.

I think the simples solution is to store store an aligned typed buffer in `repacked_qs`, e.g. an enum with a `Q4Kx8(Vec<BlockQ4Kx8>)` variant?

ivarflakstad: > I think this still has the original alignment issue.

Yes, solved now with `zerocopy` (was a transitive dependency anyway).

> e.g. an enum with a Q4Kx8(Vec<BlockQ4Kx8>) variant?

Imo a simple chunk of bytes (now with correct conversions via `zerocopy`) is easier to maintain.

EricLBuehler (approved):
🔥

EricLBuehler (approved):
🔥

PR #3647, opened 2026-06-25
Upstream onnx device candle-onnx: propagate device through simple_eval instead of hard-coding CPU

astorise:
Fixes #3491

## Problem

`candle_onnx::simple_eval` ignored the device of its input tensors and always
built initializers, `Constant`/`ConstantOfShape` values, `RandomUniform`/
`RandomNormal` outputs, and `Gemm`'s `alpha`/`beta` scalars on `Device::Cpu`.
Mixing those CPU-resident tensors with GPU inputs raised a device-mismatch
error, making the ONNX backend effectively unusable on CUDA/Metal.

## Fix

- `simple_eval`, `simple_eval_`, `get_tensor`, and the `AttrOwned` trait now
  take an explicit `&Device` that is threaded through every previously
  hard-coded call site.
- Ops that already had a tensor to read a device from (`Range`, `Gemm`) reuse
  that tensor's device instead of taking a new parameter.
- `candle-pyo3`'s `Onnx.run()` infers the device from the inputs, preferring
  any non-CPU device among them (HashMap iteration order is unspecified, so
  naively picking "the first" input could non-deterministically select a CPU
  metadata tensor over a GPU data tensor and reintroduce this same bug).
- The bundled `onnx`, `onnx-llm`, `onnx_basics`, and `silero-vad` examples are
  updated for the new parameter.
- Adds device-propagation regression tests for initializers, `Constant`,
  `ConstantOfShape`, `RandomUniform`, `RandomNormal`, `Range`, and `Gemm`, run
  on CPU plus cuda/metal-gated variants (mirroring `candle-core`'s
  `test_device!` convention), and a `cuda`/`metal` feature section in
  `candle-onnx`'s `Cargo.toml` so those variants can be gated.
- Relaxes the GELU ONNX test's float comparison from exact equality to
  6-decimal rounding (`to_vec2_round`), since CPU and GPU execution can
  legitimately differ at the float32 ULP level (e.g. `0.8413447` vs
  `0.8413448`), which surfaced once the new cuda-gated tests started
  exercising this path on real hardware.

## Commits

1. `candle-onnx: propagate device through simple_eval instead of hard-coding CPU`
   — the core fix + regression tests.
2. `candle-pyo3: pick a non-CPU device deterministically in Onnx.run()`
   — fixes a non-determinism edge case in the pyo3 binding's device inference.
3. `Relax GELU ONNX test tolerance`
   — pre-existing float-tolerance issue surfaced while running the new
[...]

PR #3648, opened 2026-06-25
candle-onnx: add pipeline-parallel evaluation via simple_eval_with_placement

astorise:
Depends on #3647 (must merge first; this branch is stacked on top of it).

## Motivation

`simple_eval` runs a whole ONNX graph on a single device. For models that
don't fit on one GPU, this adds an additive, non-breaking API to spread a
single graph across several devices (pipeline parallelism) — e.g. running
the first transformer blocks on `cuda:0` and the rest on `cuda:1`.

## API

- New `Placement` type: resolves each node's device from the longest
  matching node-name prefix rule, otherwise the device of the node's first
  already-computed input (so a stage's device flows along the graph and
  only the first node of each stage needs an explicit rule), otherwise a
  fallback device.
- New `simple_eval_with_placement`; `simple_eval` now delegates to it with
  a uniform placement, preserving its exact existing behavior.
- Initializers are materialized lazily on the device of the node that first
  consumes them (instead of all up front on a single device), which is what
  makes real memory splitting possible. Graph outputs that are unconsumed
  initializers are still materialized before returning.
- Inputs are pre-staged onto each node's device, so cross-device copies are
  inserted automatically at stage boundaries via `Tensor::to_device`.

Cross-device transfers currently require CUDA (CPU<->CUDA and CUDA<->CUDA);
Metal-to-Metal copies are not implemented in candle-core. Placement is
expected to be monotone along the topological order, which a by-depth
pipeline split satisfies.

## Fixes folded in from review

- `Placement::uniform` (used by plain `simple_eval`) now always resolves a
  node's device to the fallback, instead of inheriting an already-present
  input's device when no prefix rule matches. With zero rules,
  input-device inheritance was taking priority over the requested device,
  letting `simple_eval` silently run on the wrong device.
- `If` subgraphs now inherit their enclosing graph's initializers instead
  of only seeing their own. Lazy initializer materialization only
  populates `values` when a node directly consumes an initializer, so a
  parent-scope initializer referenced solely inside an `If` branch was
  never visible to that branch and evaluation failed with "cannot find
  ... for op".

## Testing

- CPU tests cover placement resolution, lazy-initializer handling, and
  uniform-placement equivalence with `simple_eval`.
- Feature-gated CUDA tests cover a real CPU<->cuda:0 round-trip (one GPU)
  and a cuda:0 -> cuda:1 split (two GPUs, `--include-ignored`), both
  passing on real dual-GPU hardware (2x RTX 3060).
- A CPU-runnable regression test for the `If`/initializer fix and a
  cuda-gated regression test for the uniform-placement fix are included,
  both verified to fail without their respective fix.
- README documents the new API and its limitations.

### DIFF
--- a/candle-examples/examples/onnx-llm/main.rs
+++ b/candle-examples/examples/onnx-llm/main.rs
@@ -170,7 +170,7 @@ pub fn main() -> Result<()> {
             }
         }
 
-        let outputs = candle_onnx::simple_eval(&model, inputs)?;
+        let outputs = candle_onnx::simple_eval(&model, inputs, &device)?;
 
         let logits = outputs.get("logits").unwrap();
 
--- a/candle-examples/examples/onnx/main.rs
+++ b/candle-examples/examples/onnx/main.rs
@@ -66,9 +66,10 @@ pub fn main() -> anyhow::Result<()> {
 
     let model = candle_onnx::read_file(model)?;
     let graph = model.graph.as_ref().unwrap();
+    let device = image.device().clone();
     let mut inputs = std::collections::HashMap::new();
     inputs.insert(graph.input[0].name.to_string(), image.unsqueeze(0)?);
-    let mut outputs = candle_onnx::simple_eval(&model, inputs)?;
+    let mut outputs = candle_onnx::simple_eval(&model, inputs, &device)?;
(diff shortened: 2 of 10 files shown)


astorise:

PR #3663, opened 2026-06-27
cuda: reject launch configs above u32::MAX

### LINKED ISSUE #3604: candle-core CUDA backend: LaunchConfig::for_num_elems(el as u32) silently truncates; ops fail on tensors with >= 2^32 elements
## Summary

Many launch sites in `candle-core/src/cuda_backend/mod.rs` compute their launch config as

```rust
let cfg = LaunchConfig::for_num_elems(el as u32);
```

`el` is a `usize` element count. For a tensor with exactly 2^32 elements, `el as u32` wraps to 0, giving grid `(0, 1, 1)` and `CUDA_ERROR_INVALID_VALUE`; for counts slightly above 2^32 it silently computes on the wrong (wrapped) element count, which is worse.

## Repro

```rust
// 128 * 8192 * 32 * 128 = 2^32 elements
let x = Tensor::zeros((128, 8192, 32, 128), DType::F32, &dev)?;
let y = x.to_dtype(DType::BF16)?; // DriverError(CUDA_ERROR_INVALID_VALUE)
```

Hit in practice while benchmarking flash-attn-v3 at batch=128, seqlen=8192, 32 heads, headdim=128 — a perfectly realistic H200-sized activation (~8.6 GB in f16).

## Scope

`grep -n "for_num_elems(.* as u32)" candle-core/src/cuda_backend/mod.rs` currently shows 10+ call sites (`to_dtype`, `affine`, unary/binary ops, `where_cond`, etc.), plus similar `as u32` casts in `candle-nn` CUDA kernels. Anything that can see a >= 2^32-element tensor is affected.

## Possible directions

- Bail with a clear error when `el > u32::MAX` (cheap, honest, no silent wrong results), or
- chunk the launch into multiple grids / use 64-bit indexing in the kernels (real fix, larger change).

Happy to send a PR for the bail-with-error variant if that is the preferred direction.

ogarciarevett:
Fixes #3604.

This replaces the `as u32` truncation in the CUDA backend with a checked helper, so oversized tensors fail with a clear error instead of silently wrapping the launch config.

There's also a focused regression test for the `u32::MAX + 1` case.

### DIFF
--- a/candle-core/src/cuda_backend/mod.rs
+++ b/candle-core/src/cuda_backend/mod.rs
@@ -13,6 +13,25 @@ use half::{bf16, f16};
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex, OnceLock};
 
+fn launch_config_for_num_elems(elem_count: usize) -> Result<LaunchConfig> {
+    let elem_count = u32::try_from(elem_count).map_err(|_| {
+        crate::Error::Msg(
+            format!("cuda kernel launch element count {elem_count} exceeds u32::MAX").into(),
+        )
+        .bt()
+    })?;
+    Ok(LaunchConfig::for_num_elems(elem_count))
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn launch_config_rejects_more_than_u32_max_elements() {
+        let err = super::launch_config_for_num_elems(u32::MAX as usize + 1).unwrap_err();
+        assert!(err.to_string().contains("exceeds u32::MAX"), "{err}");
+    }
+}
+
 #[cfg(feature = "cudnn")]
 pub mod cudnn;
 mod device;
@@ -143,7 +162,7 @@ impl Map1 for Affine {
         let shape = layout.shape();
         let dims = shape.dims();
         let el = shape.elem_count();
-        let cfg = LaunchConfig::for_num_elems(el as u32);
+        let cfg = launch_config_for_num_elems(el)?;
         let ds = SlicePtrOrNull::params_from_layout(dev, layout)?;
         let src = &src.slice(layout.start_offset()..);
         let func = dev.get_or_load_func(&kernel_name::<T>("affine"), &kernels::AFFINE)?;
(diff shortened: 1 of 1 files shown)


astorise:
Thanks for this — silently truncating an oversized element count into a bogus, too-small launch grid instead of erroring is exactly the kind of bug that's easy to miss and hard to debug when it happens (wrong results, not a crash).

Two things found while integrating this into a fork, one addition and one regression it introduces:

1. copy_strided_src regression: this function computes the new checked launch_config_for_num_elems(el_count)? unconditionally before branching on src_l.is_contiguous(), but every dtype arm's contiguous branch only calls dev.memcpy_dtod(&src, &mut dst)? and never touches the resulting LaunchConfig — only the non-contiguous else branch calls builder.launch(cfg). So a contiguous CUDA copy with more than u32::MAX elements (e.g. force_contiguous, variable creation) now hard-fails on the launch-count check even though the memcpy path never needed a launch grid and can handle arbitrarily large contiguous copies fine. Moving the launch_config_for_num_elems call into each dtype arm's non-contiguous else branch (right before it's actually used) fixes this without touching the fast path.

2. Same pattern exists in candle-nn, untouched by this PR: [candle-nn/src/ops.rs](https://github.com/astorise/candle/blob/integrate/pr-3663/candle-nn/src/ops.rs) (CUDA sigmoid) and all three rotary-embedding variants in [candle-nn/src/rotary_emb.rs](https://github.com/astorise/candle/blob/integrate/pr-3663/candle-nn/src/rotary_emb.rs) have the identical unchecked LaunchConfig::for_num_elems(x as u32) pattern this PR fixes in candle-core. Since the new helper is private to candle-core's cuda_backend module, I added an equivalent small helper in each candle-nn file and routed those 4 call sites through it.

One review note on the existing copy2d fix in this PR, which I checked carefully and is correct: it changes the old code (which cast d1/d2 to u32 individually before multiplying, so either one alone overflowing u32::MAX silently truncated with no error at all) to check the product d1 * d2 while still usize — that's the right bound to validate against for_num_elems, good catch on the original author's part.

Pushed both fixes on top of this PR in my fork: astorise/candle@f5a2cbb (candle-nn extension) and astorise/candle@9942f5e (copy_strided_src fix). Happy to open as a PR against this branch or main if useful.

ogarciarevett:
Good catch. I confirmed the regression in the source flow: `launch_config_for_num_elems(el_count)` currently runs before the contiguous `memcpy_dtod` branch, so a copy that needs no kernel can now fail the launch-grid bound.

I cannot validate the CUDA path locally because this machine has no `nvcc`, so I will not push an unverified follow-up. Since you already have the focused fix in `9942f5e`, please open it against this branch if convenient and include the CUDA checks you ran. I would keep the `candle-nn` audit from `f5a2cbb` as a separate follow-up so this PR first removes the regression it introduced.

PR #3669, opened 2026-06-27
Add CUDA Graph capture for decode + FlashInfer-style decode-attention backend (CUDA/Metal/CPU) (#3651)

### LINKED ISSUE #3651: [feature] CUDA Graph capture for autoregressive decode + optional FlashInfer backend
### Motivation
During token-by-token decoding, per-kernel launch overhead dominates once the
model is small relative to the GPU. CUDA Graphs capture the steady-state step and
replay it with near-zero launch overhead; FlashInfer provides decode-attention
kernels that often beat flash-attn for batched generation. Both are recurring
needs for any candle-based inference server.

### Gap in candle
- No capture/replay API around a sequence of ops on a CUDA stream.
- No FlashInfer attention backend.

### Proposed implementation
- `candle-core` experimental API: `CudaGraph::capture(&device, || { ... })?` plus
  `.replay()`, wrapping a closure of ops with static shapes on a given stream.
- An optional FlashInfer backend (behind a feature) as an alternative to
  `candle-flash-attn` on the decode path.

### Use-case / example
```rust
// Capture the decode step once, replay it per generated token.
let graph = CudaGraph::capture(&device, || model.decode_step(&tok, &mut cache))?;
for _ in 0..max_new_tokens {
    graph.replay()?; // no per-kernel relaunch
}
```

astorise:
## What this does

Implements both halves of #3651:

1. **CUDA Graph capture/replay for autoregressive decode** — a new
   `candle_core::CudaGraph` API that captures the fixed-shape sequence of
   kernels issued during a decode step once and replays it with a single
   `cuGraphLaunch`, avoiding per-token launch overhead.

2. **Optional FlashInfer-style decode-attention backend** — a new
   `candle-flashinfer-kernels` crate exposing
   `flashinfer_decode_attention(q, k, v, scale)` for the single-new-token
   decode workload (one query token per sequence attending over a KV cache,
   GQA supported). This is a reference, numerically-stable (streaming softmax)
   kernel with the same shape contract FlashInfer's batch-decode kernels
   target — not a port of their tensor-core/split-KV kernels.

## CudaGraph (candle-core)

- `CudaGraph::capture(device, |f|)` / `replay()` built on the raw driver API
  (`cuStreamBeginCapture_v2` / `cuStreamEndCapture` / `cuGraphInstantiateWithFlags`
  / `cuGraphLaunch`).
- Event tracking is paused during capture: candle drives cudarc in multi-stream
  mode, so every launch would otherwise emit `cuStreamWaitEvent`/`cuEventRecord`
  against per-tensor dependency events, which is illegal during capture
  (`CUDA_ERROR_INVALID_VALUE`).
- Documented usage constraints: callers must (a) warm up the ops once before
  capturing so module loads / HtoD uploads don't happen mid-capture, and
  (b) write into **persistent, pre-allocated** buffers — a tensor allocated
  inside the closure becomes a graph-owned allocation node whose memory is only
  valid during a replay.

## FlashInfer decode-attention backend

- **CUDA** kernel (`f32`/`f16`/`bf16`), one block per `(batch, query head)`.
- **Metal** kernel for Apple Silicon (`f32`/`f16`), compiled at runtime; mirrors
  the CUDA path (threadgroup-reduced dot product + streaming softmax).
- **CPU** reference `cpu_fwd` (`f32`/`f16`/`bf16`, GQA, stride/offset aware),
  parallelized across `(batch, head)` with rayon.
- `cuda` and `metal` are **opt-in features**; with neither, the crate builds
  CPU-only (no toolchain needed) and the CPU fallback runs. This also lets the
  workspace build without a CUDA toolchain.

## Testing

- **CUDA**: `cuda_graph_tests` and the decode-attention kernel validated on a
  real GPU runner (`ci_cuda.yaml`).
- **Metal**: decode-attention validated on a GitHub-hosted `macos-15` Apple
  Silicon runner with `CANDLE_METAL_REQUIRED=1` (`ci_metal.yaml`).
- **CPU**: `decode_attention_cpu_matches_reference` runs everywhere, no GPU
  required.

Closes #3651.

### DIFF
--- a/.github/workflows/ci_cuda.yaml
+++ b/.github/workflows/ci_cuda.yaml
@@ -28,10 +28,10 @@ jobs:
       security-events: write
     env:
       CUDA_COMPUTE_CAP: 86
-      # Scoping the test step to GPU-relevant crates (below) wasn't enough on its own: this
-      # runner still OOM-kills (exit 137) while compiling candle-transformers at full rustc
-      # parallelism. Cap parallel rustc jobs to reduce peak memory; the two fixes address
-      # different causes (crate count vs. per-crate codegen parallelism) and are both needed.
+      # `cargo test --features cuda` at the workspace root builds every member, including
+      # candle-pyo3 and the wasm/yew examples, which pulls in pyo3/wasm-bindgen/yew and OOM-kills
+      # (exit 137) this runner. Scope to the GPU-relevant crates instead, and cap parallel rustc
+      # jobs to further reduce peak memory.
       CARGO_BUILD_JOBS: 2
     steps:
       - name: Checkout repository
@@ -44,7 +44,7 @@ jobs:
         uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8  # stable
       - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32  # v2
       - name: Test (cuda)
-        run: cargo test --features cuda
+        run: cargo test --features cuda -p candle-core -p candle-nn -p candle-transformers -p candle-examples
       - name: Build+test (cuda) - CudaGraph capture/replay
         run: cargo test -p candle-core --features cuda --test cuda_graph_tests
       - name: Build+test (cuda) - FlashInfer-style decode attention kernel
--- a/.github/workflows/ci_cuda.yaml
+++ b/.github/workflows/ci_cuda.yaml
@@ -2,6 +2,11 @@ name: CI / cuda
 
 on:
   workflow_dispatch:
+    inputs:
+      test_ref:
+        description: 'Branch/tag/SHA to checkout and test (defaults to the dispatched ref)'
+        required: false
+        type: string
   pull_request:
 
 jobs:
@@ -9,10 +14,10 @@ jobs:
     concurrency:
       group: ${{ github.workflow }}-${{ github.job }}-${{ github.head_ref || github.run_id }}
       cancel-in-progress: true
-    runs-on:
-      group: aws-g5-4xlarge-cache
+    runs-on: arc-gpu-candle
     container:
       image: nvidia/cuda:13.0.2-cudnn-devel-ubuntu24.04
+      options: --gpus all
     if: ${{ github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name }}
     permissions:
       contents: write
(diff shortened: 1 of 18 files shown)


astorise:
Update on this PR: it contains a real bug that was found and fixed downstream, now validated on production hardware — worth landing before merge review continues.

Bug: a second, independent CudaGraph::capture on the same device could fail a later, completely unrelated allocation with CUDA_ERROR_INVALID_VALUE (thrown from the very first op of the second caller, before any of its own capture/replay code ran). Full writeup and history: [astorise/candle#17](https://github.com/astorise/candle/issues/17).

Two distinct, additive causes, both in [candle-core/src/cuda_backend/graph.rs](https://github.com/astorise/candle/blob/main/candle-core/src/cuda_backend/graph.rs):

Event-tracking guard scoped too narrowly. pause_event_tracking() was only held for the capture() call itself, not the graph's replay/lifetime. cudarc unconditionally re-records a tensor's existing read/write dependency event on every access regardless of the tracking flag; a tensor allocated before capture (e.g. a model-level KV cache written into by every capture) already owns such an event, and recording it from inside a capture turns that recording into a node of that graph. With tracking back on as soon as capture() returned, any later access to that tensor — between replays, or by a second independent capture — recorded a plain, non-captured update to an event the graph still owned a captured node for. Fix: hold the guard for the CudaGraph's entire lifetime (capture → every replay → destruction).
Graph memory pool not reconciled on destruction. A tensor allocated inside the captured closure becomes a graph memory-alloc/free node pair, backed by the CUDA driver's per-device graph memory pool — not a plain cuMemAllocAsync/cuMemFreeAsync. By design, that pool retains a destroyed graph's physical memory for reuse by future graph launches instead of releasing it immediately. A capture with many such nodes (e.g. several intermediates per layer of a real multi-layer transformer forward pass) can leave a later independent capture, or any other stream-ordered allocation on the same device, observing a pool the driver hasn't finished reconciling. Fix: call cuDeviceGraphMemTrim on the graph's device in Drop for CudaGraph, right after destroying the graph exec.
Both fixes have regression tests in [candle-core/tests/cuda_graph_tests.rs](https://github.com/astorise/candle/blob/main/candle-core/tests/cuda_graph_tests.rs) (second_independent_capture_after_first_graph_dropped and second_independent_capture_with_many_graph_owned_allocations), and — beyond the candle-core test suite — the combined fix was validated against a real multi-layer Llama decode capture/replay production workload on real GPU hardware (see the last few comments on [astorise/candle#17](https://github.com/astorise/candle/issues/17) for details).

Happy to open this as a separate PR against this branch or rebase it in directly, whichever is easier for review here.

astorise:

PR #3682, opened 2026-06-29
WASM Qwen3-0.6B: 3.5x faster decode (9 -> 32 tok/s) via threads, lazy embed, and relaxed-SIMD Q4_K

DrJesseGlass:
This PR stacks three additions to the WASM32 SIMD path for the quantized Qwen3
example, taking decode from ~9 tok/s to ~32 tok/s (M1, Chrome).

## Changes

**WASM multithreading** 
routes the quantized matmul through rayon (backed by wasm-bindgen-rayon) behind a new `wasm-threads` feature. The native barrier-pool path is unchanged; non-threaded WASM still falls back to the serial path. Requires shared memory and cross-origin isolation (the example's serve.py sets COOP/COEP).

**Lazy embedding dequant** 
the token embedding is kept quantized and only the input-token rows are dequantized each step (via `QTensor::embedding`), instead of materializing the full embedding matrix. The tied lm_head shares the same `QTensor`.

**Relaxed-SIMD Q4_K matmul** 
a `vec_dot` for Q4_K x Q8_K built on `i32x4_relaxed_dot_i8x16_i7x16_add`, gated on `relaxed-simd`, with the existing simd128 kernel as fallback.

Also includes a simd128 WASM build fix: the f16/bf16 vector helpers were gated to NEON/AVX2 so simd128 no longer fails to compile.

## Why all-Q4_K (not Q8_0 or Q4_K_M)

Q8_0 was previously fastest under plain SIMD. Q4_K_M/S/XS were consistently slower because they keep Q6_K in the most critical tensors (the tied lm_head), the dominant matmul (full vocab every decoded token)  and Q6_K underperformed Q8_0 there. Requantizing those down to a uniform all-Q4_K model puts the lm_head on the fast Q4_K relaxed-dot path, which is where the largest win comes from.


candle-core/src/quantized/k_quants.rs
```diff
@@ -2427,28 +2430,75 @@ pub fn matmul<T: GgmlType>(
...
+                            *base.add(2) = d2;
+                            *base.add(3) = d3;
+                        }
+                    }
+                });
+            }
+            #[cfg(all(target_arch = "wasm32", feature = "wasm-threads"))]
+            {
+                use rayon::prelude::*;
+                main.par_chunks_mut(4)
```
ivarflakstad, line 2481: Oh. Yeah this is a bit unfortunate.

I guess what we really need is a generic threading abstraction that chooses the correct path.
We would lose some of the nice rayon helpers like `.par_chunks_mut`, but it would still be an improvement.

DrJesseGlass: That seems like a nice goal. I'm also under the opinion that we should aim to remove rayon as a dependency but perhaps that is shortsighted.

PR #3685, opened 2026-06-29
candle-core: free Q4_K source weights after building the aarch64 matmul repack (~2x -> ~1x weight memory)

DrJesseGlass:
On aarch64+dotprod, the Q4_K matmul fast path repacks each weight into `BlockQ4Kx8` and caches it on the `QTensor`. The original Q4_K blocks were kept alongside that repack for the tensor's lifetime, so every Q4_K matmul weight was held in memory twice. For Qwen3-0.6B that's ~320 MB of pure duplication.

This change builds the repack eagerly in `QMatMul::from_arc` (the moment we commit to the quantized path) and frees the original blocks, so only the repack is retained.

## How

In `QMatMul::from_arc`, in the branch where the tensor stays quantized (after the `CANDLE_DEQUANTIZE_ALL`/`_F16` dequantize paths are ruled out):

```rust
let mut qtensor = qtensor;
if let Some(qt) = Arc::get_mut(&mut qtensor) {
    qt.drop_source_after_repack();
}
Self::QTensor(qtensor)
```

drop_source_after_repack (private) builds the BlockQ4Kx8 repack via the existing path, then swaps storage for an empty same-dtype buffer. No-op off aarch64/dotprod and for non-Q4_K tensors.
Arc::get_mut means we only release when we uniquely own the tensor. A shared tensor (e.g. a tied lm_head reusing the token-embedding Arc) is left intact, so it keeps the blocks it still needs for embedding/dequantize.
Because the release sits after the dequantize branches, models run under CANDLE_DEQUANTIZE_ALL[_F16] are unaffected (they never take this branch).

The matmul cpu_fwd was refactored to fetch the repack via a small q4kx8_repack() helper (build-once, cached) rather than inline, so it no longer reads storage once the repack exists, which is what lets the source be freed.

Effects
Resident memory: ~2x -> ~1x for Q4_K matmul weights. Qwen3-0.6B (M1): 639.5 MB -> 319.7 MB across the 197 matmul weights.
First-token latency is lower -- the repack moved off the first matmul onto load.
Load time is marginally higher (the repack work shifts to load); total work unchanged.
Peak memory is lower -- the transient 2x during first-forward no longer occurs.

Scope
aarch64 + dotprod only (Graviton, Apple Silicon); no-op and zero overhead on x86/wasm/CUDA/Metal.
Lives entirely in candle-core (QMatMul::from_arc), so it benefits every quantized model with no per-model changes.

### DIFF
--- a/candle-core/src/quantized/mod.rs
+++ b/candle-core/src/quantized/mod.rs
@@ -709,6 +709,45 @@ impl QTensor {
         }
     }
 
+    // Build-once cached BlockQ4Kx8 repack of a Q4_K weight with n output rows.
+    // Later calls return it without reading storage, so the original can be freed.
+    #[cfg(all(target_arch = "aarch64", target_feature = "dotprod"))]
+    fn q4kx8_repack(&self, n: usize) -> &Option<Vec<u8>> {
+        use zerocopy::IntoBytes;
+        self.repacked_qs.get_or_init(|| {
+            let s = match &self.storage {
+                QStorage::Cpu(s) => s,
+                _ => return None,
+            };
+            let total_blocks = s.storage_size_in_bytes() / std::mem::size_of::<BlockQ4K>();
+            if total_blocks == 0 {
+                return None;
+            }
+            let blocks =
+                unsafe { std::slice::from_raw_parts(s.as_ptr() as *const BlockQ4K, total_blocks) };
+            Some(k_quants::pack_to_q4kx8(blocks, n).as_bytes().to_vec())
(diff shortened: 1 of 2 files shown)


DrJesseGlass:
Withdrawing: the mechanism moves the Q4_K repack to load time (eager) to free the source, which front-loads work into construction, conflicts with zero-copy/mmap loading, and leaves a used weight non-dequantizable. If pursued, a lazy (first-matmul) drop is the better shape.

PR #3688, opened 2026-06-30
Fix pickle complexity DoS: add memo_get node budget

### LINKED ISSUE #3620: Pickle memo bomb — CPU + memory exhaustion (algorithmic-complexity DoS) in candle-core pickle reader
## Summary

candle-core's pickle parser deep-clones the entire `PyObject` tree on every memo fetch (`BinGet`) and
every memo store (`BinPut` / `Put`). A crafted pickle that fetches the same memo entry twice,
combines the two copies, and writes the result back doubles the node count on each cycle. After N
cycles, parsing requires O(2^N) work / memory. This is a pickle memo bomb (the same amplification
principle as the "Billion Laughs" XML attack) and results in CPU and memory exhaustion
(algorithmic-complexity DoS) on a crafted `.pt` / `.pth` or raw pickle. Two paths share the root
cause: (A) CPU exhaustion via `BinGet` + `Tuple2` + `BinPut`; (B) memory exhaustion via `BinGet` +
`Build`(dict) + `BinPut`. A developer source comment at the fetch site already flags this risk
("Maybe we should use refcounting rather than doing potential large clones here").

## Affected component

- Repository: `huggingface/candle` (`candle-core` crate)
- File: `candle-core/src/pickle.rs:390` — `Stack::memo_get()` returns `obj.clone()` (deep clone of
  the whole `PyObject` tree on every fetch); the dev TODO comment sits at this site. The companion
  store path (`memo_put`, ~line 400) clones the top of stack on every `BinPut`. `PyObject` is a
  recursive enum, so each clone copies the full subtree. No `Rc`/`Arc` sharing, no node-count limit.
- Reachable from candle's standard model/checkpoint loading path on any `.pt` / `.pth` or raw pickle.

## Reproduction

PoC (Path A — CPU exhaustion): `findings/candle/poc-078-candle-memo-bomb.pkl` (144 bytes, N=20;
0.175s at N=20, ~5.7s at N=25 on a release binary).

```python
NEWFALSE = b'\x89'; BINPUT = b'q'; BINGET = b'h'; TUPLE2 = b'\x86'; STOP = b'.'
N = 20
poc = NEWFALSE + BINPUT + b'\x00'
for _ in range(N):
    poc += BINGET + b'\x00' + BINGET + b'\x00' + TUPLE2 + BINPUT + b'\x00'
poc += STOP
open('poc-078-candle-memo-bomb.pkl', 'wb').write(poc)
```

PoC (Path B — memory exhaustion): `findings/candle/poc-078b-dict-build-oom.pkl` (182 bytes; the same
doubling driven through `BUILD` dict-merge → ~92 GB RSS measured on a 128 GB host before
termination).

Load via any `candle_core::pickle` path (`Stack::empty()` + `read_loop()`, or a `.pt` model load).
Measured CPU doubles per cycle; raising N increases the stall / allocation accordingly.

## Root cause

```rust
// pickle.rs — Stack::memo_get() (line ~390)
pub(crate) fn memo_get(&self, idx: u32) -> Result<PyObject> {
    match self.memo.get(&idx) {
        None => crate::bail!("missing memo {idx}"),
        Some(obj) => Ok(obj.clone()),  // deep clone of entire PyObject tree
        // "Maybe we should use refcounting rather than doing potential large clones here"
    }
}

// Stack::memo_put() (line ~400) — called by BinPut / Put
let obj = self.last()?.clone();  // deep clone of top-of-stack on every BinPut
```

Each `BinGet`+`BinGet`+`Tuple2`+`BinPut` cycle (Path A), or `BinGet`+`BinGet`+`Build`+`BinPut` on
dicts (Path B), doubles the node/entry count in the memo slot. After N cycles, work and peak memory
are O(2^N) from an input that grows only linearly in N.

## Suggested fix

Replace owned clones with reference-counted sharing — exactly what the existing developer comment
points at:

```rust
use std::rc::Rc;

pub struct Stack {
    stack: Vec<Rc<PyObject>>,
    memo: HashMap<u32, Rc<PyObject>>,
    // ...
}

fn memo_get(&self, idx: u32) -> Result<Rc<PyObject>> {
    self.memo.get(&idx).cloned().ok_or(/* ... */)  // O(1) refcount bump, no tree copy
}
```

Alternatively, track a cumulative node count during parsing and reject inputs exceeding a configurable
threshold.

## Relationship to a sibling finding

This is an independent occurrence of the same bug class found in tracel-ai/burn (CRUCIBLE-2026-077).
The two codebases share no code; both independently implemented a deep-cloning pickle memo store, and
the same PoC bytes trigger both. This issue covers candle only.

## Revalidation

Source-level revalidation on 2026-06-14 against candle HEAD `65ecb58`: `pickle.rs:390` still returns
`obj.clone()` from `memo_get` with the developer TODO comment intact, and no `Rc`/`Arc` sharing or
node-count limit has been added. The clone-cycle amplification is live.

Status: advisory drafted; re-routing to public issue + VulDB; revalidated 2026-06-14 source-level vs
candle HEAD `65ecb58`.

## Proof-of-concept files (base64)

Decode with `base64 -d > file`.

**poc-078-candle-memo-bomb.pkl** (144 bytes):
```
iXEAaABoAIZxAGgAaACGcQBoAGgAhnEAaABoAIZxAGgAaACGcQBoAGgAhnEAaABoAIZxAGgAaACGcQBoAGgAhnEAaABoAIZxAGgAaACGcQBoAGgAhnEAaABoAIZxAGgAaACGcQBoAGgAhnEAaABoAIZxAGgAaACGcQBoAGgAhnEAaABoAIZxAGgAaACGcQAu
```

AnkitNakhawa:
Fix pickle memo bomb: add cumulative complexity limit to `memo_get` (closes #3620)

## What

Adds a cumulative node-budget guard to `Stack::memo_get` in `candle-core/src/pickle.rs` to prevent the algorithmic-complexity DoS described in #3620.

## Root cause

`memo_get` returned `obj.clone()` unconditionally — a deep clone of the entire `Object` tree on every `BinGet`. A crafted pickle that fetches the same memo slot twice, combines the copies, and writes the result back doubles the node count each cycle. After N cycles, parsing requires O(2^N) work and memory from an input whose size grows only linearly with N (the same "Billion Laughs" amplification principle). Two paths were confirmed:
- **Path A (CPU):** `BinGet` + `BinGet` + `Tuple2` + `BinPut` — N=25 stalls ~5.7s on a release build
- **Path B (memory):** `BinGet` + `BinGet` + `Build` (dict-merge) + `BinPut` — ~92 GB RSS measured before OOM on a 128 GB host

## Fix

`memo_get` now counts the nodes in the object being retrieved, checks the budget, and only clones if the budget is not exceeded. The cost is added to a running `complexity` counter on the `Stack`; if it exceeds `max_complexity` (default: 1,000,000 nodes),parsing is aborted before the expensive object-tree clone occurs. The node walk is iterative rather than recursive so a deeply-nested input cannot stack-overflow before the guard fires.

The 1,000,000-node default is a conservative initial bound — real checkpoint complexity is estimated well below this, but it has not been validated against large production checkpoints and can be adjusted based on maintainer guidance.

`Stack::with_limit(n)` is exposed for callers that need a custom budget. `u64::MAX` disables the check entirely.

## Tests added

- Exact PoC bytes from the issue report (both Path A and Path B) are rejected
- Exponential tuple-doubling and dict-merge payloads exceeding the default limit are rejected
- A deeply nested memo-expansion payload is rejected cleanly — node counting uses iterative traversal rather than recursion, so a deeply-nested input cannot stack-overflow before the guard fires
- Normal small payloads and no-memo pickles continue to parse successfully
- Custom and zero budgets behave as documented

Closes #3620

### DIFF
--- a/candle-core/src/pickle.rs
+++ b/candle-core/src/pickle.rs
@@ -461,7 +461,7 @@ impl Stack {
         self.memo
             .get(&id)
             .cloned()
-            .ok_or_else(|| crate::Error::Msg("missing object in memo".into()))
+            .ok_or_else(|| crate::Error::Msg(format!("missing object in memo {id}")))
     }
 
     fn memo_put(&mut self, id: u32) -> Result<()> {
@@ -1040,18 +1040,12 @@ mod tests {
 
     #[test]
     fn test_deep_chain_does_not_stack_overflow() {
-        // 100_000-level deep Tuple1 chain. With a recursive node_count() this
-        // would overflow the native call stack (~8 MB) before the complexity
-        // guard could fire. The iterative implementation must handle it safely —
-        // either returning an error (complexity exceeded) or succeeding, but
-        // never panicking.
-        //
-        // Each BinGet in iteration k charges k+1 nodes (memo[0] grows by one
-        // node per iteration as the chain deepens). Total complexity after K
-        // iterations = K*(K+1)/2, which exceeds 1_000_000 at K ≈ 1_414.
-        // Parsing therefore returns a complexity error well before depth=100_000;
-        // the important property under test is that it errors cleanly rather
-        // than crashing with a native stack overflow.
+        // Payload containing 100_000 memo-expansion iterations. The complexity
+        // budget fires at roughly iteration 1_414 (cumulative cost K*(K+1)/2
+        // exceeds 1_000_000), so the parser never walks all 100_000 levels.
+        // The payload is large to ensure a recursive node_count() would overflow
+        // the native stack before the guard could fire; the iterative
+        // implementation must reject cleanly without panicking.
         let payload = make_deep_chain_payload(100_000);
         // If parse panics (stack overflow), the test harness catches it and the
         // test fails — that is the "no stack overflow" guarantee. This assertion
--- a/candle-core/src/pickle.rs
+++ b/candle-core/src/pickle.rs
@@ -330,7 +330,7 @@ const DEFAULT_MAX_COMPLEXITY: u64 = 1_000_000;
 pub struct Stack {
     stack: Vec<Object>,
     memo: HashMap<u32, Object>,
-    /// Cumulative count of nodes materialised so far (charged on every memo_get).
+    /// Cumulative count of nodes cloned by memo retrievals so far.
     complexity: u64,
     /// Budget: an error is returned once `complexity` exceeds this value.
     max_complexity: u64,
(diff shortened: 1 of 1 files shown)


astorise:
Heads-up for triage: there's meaningful overlap with #3628 ("Bound the pickle VM's working set and nesting depth"). #3628's byte-based working-set cap also bounds the memo-replay amplification this node budget targets — each memo-expanded clone charges bytes against the cap, and #3628 even ships a rejects_memo_replay_amplification test for exactly that DoS. So the two guard largely the same surface via different metrics (cumulative node count here vs. a heap-bytes ceiling there). Might be worth deciding whether both are wanted or whether they should be unified, to avoid two overlapping limits on the same path.

Issue #3653, opened 2026-06-25
[feature] Add DeepSeek-V3, Llama 4 and Qwen3-MoE model architectures


astorise:
### Motivation
`candle-transformers` already covers a lot (llama, qwen2/qwen3, deepseek2,
gemma2/3/4, glm4, mixtral, plus VLMs), but a few recent dense/MoE families that
users frequently ask for are still missing.

### Gap in candle (please confirm)
- **DeepSeek-V3** — only `deepseek2` (V2) is present; V3 adds aux-loss-free MoE
  routing, Multi-Token Prediction and FP8.
- **Llama 4** — MoE + interleaved/iRoPE; not present.
- **Qwen3-MoE** — only dense `qwen3` appears to be present.

### Proposed implementation
- New modules `candle-transformers/src/models/{deepseek3,llama4,qwen3_moe}.rs`,
  following the existing MoE models (`mixtral`, `deepseek2`).
- HF `config.json` parsing + `VarBuilder` loading + quantized variant where it
  makes sense.
- A minimal `candle-examples/examples/<model>` per family for validation.

### Use-case / example
```rust
// Load Qwen3-MoE like any other candle-transformers model.
let cfg: Qwen3MoeConfig = serde_json::from_slice(&std::fs::read("config.json")?)?;
let model = Qwen3Moe::load(vb, &cfg)?;
let logits = model.forward(&input_ids, /*pos=*/0)?;
```

haydonryan:
Deepseek v4 flash is my target model on dual rtx6000 pro. Currently candle doesn't support this either. 
Interestingly vllm does not support it on Blackwell as of v0.24. 

It'd be great to support it before they did. I can put some cycles and tokens into it.

astorise:
> Deepseek v4 flash is my target model on dual rtx6000 pro. Currently candle doesn't support this either. Interestingly vllm does not support it on Blackwell as of v0.24.
> 
> It'd be great to support it before they did. I can put some cycles and tokens into it.

Thanks for the interest — contributions are very welcome!

[...]

PR #3692, opened 2026-06-30
Add DeepSeek-V3 model architecture

astorise:
## Summary

- Implements DeepSeek-V3, reusing the Multi-head Latent Attention (MLA) and YaRN rotary embeddings already implemented for DeepSeek-V2 (`deepseek2.rs`) rather than duplicating ~700 lines of code. A handful of shared items (`Attention`, `Mlp`, `TopkMethod`, `ScoringFunc`, a couple of config fields, `masked_fill`) are widened to `pub(crate)` so `deepseek3.rs` can build on them via a `From<&DeepSeekV3Config> for DeepSeekV2Config` adapter — the same pattern `qwen3_moe.rs` already uses to reuse `qwen3.rs`.
- Adds DeepSeek-V3's auxiliary-loss-free ("noaux_tc") sigmoid MoE gate, which differs from V2's softmax/group-limited-greedy gate: sigmoid (not softmax) scoring, a learned `e_score_correction_bias` applied only for expert *selection* (the final combination weights use the raw sigmoid scores), group scoring via the sum of the top-2 experts per group (V2 uses max), and an unconditional `routed_scaling_factor`.
- Adds a `candle-examples/examples/deepseekv3` example mirroring the existing `deepseekv2` one (`Which::V3` / `Which::V3Base`, pointing at `deepseek-ai/DeepSeek-V3` / `deepseek-ai/DeepSeek-V3-Base`), with a fallback to a single `model.safetensors` file for checkpoints that aren't sharded.
- Adds a real end-to-end integration test against [`yujiepan/deepseek-v3-tiny-random`](https://huggingface.co/yujiepan/deepseek-v3-tiny-random), a randomly-initialized checkpoint that uses the actual DeepSeek-V3 architecture (256 routed experts, sigmoid noaux_tc gate, MLA, YaRN RoPE) at a download size small enough to run anywhere (~9 MB) — the real 671B model can't be exercised in CI even quantized (even 2-bit would need ~170 GB).
- Multi-token prediction (MTP) and FP8 weights are intentionally out of scope: MTP only matters for training, and FP8 is a storage format that doesn't change the (dequantized, e.g. bf16) inference forward pass.

Closes the DeepSeek-V3 portion of #3653 (Llama 4 and Qwen3-MoE — the latter already implemented separately — are tracked there as well).

## Changes

- `candle-transformers/src/models/deepseek2.rs`: widen visibility of `Attention`, `Mlp`, `TopkMethod`, `ScoringFunc`, and two `DeepSeekV2Config` fields to `pub(crate)` so they can be shared with V3. No functional change.
- `candle-transformers/src/models/deepseek3.rs` (new): `DeepSeekV3Config`, the `noaux_tc` `MoeGate`, `Moe`/`MoeOrMlp`, `DecoderLayer`, and the top-level `DeepSeekV3` model.
- `candle-transformers/src/models/mod.rs`: register the new `deepseek3` module.
- `candle-examples/examples/deepseekv3/{main.rs,README.md}` (new): example mirroring `deepseekv2`, plus a fallback to a single safetensors file when no sharding index is present.
- `candle-examples/tests/deepseekv3_integration.rs` (new): `#[ignore]`d end-to-end test (network required) — downloads the tiny-random checkpoint and runs a real prefill + KV-cache decode step through the production model code.

## Test plan

- [x] `cargo test -p candle-transformers --lib deepseek3` — unit tests with a tiny synthetic config + hand-built weight map covering: forward-pass output shape and KV-cache continuation across a prefill + a follow-up single-token step, and rejection of `topk_method` values other than `noaux_tc` (the only one DeepSeek-V3 uses).
- [x] `cargo test -p candle-examples --release --test deepseekv3_integration -- --ignored` — real end-to-end generation against `yujiepan/deepseek-v3-tiny-random` (real download, real tokenizer, real weights, real forward pass + KV-cache decode step), run on both CPU and CUDA.
- [x] `cargo check -p candle-transformers` / `cargo clippy -p candle-transformers --lib --examples -p candle-examples --example deepseekv3` (default features) — clean.
- [x] `cargo fmt -- --check` — clean.
- [x] CUDA: `cargo test --features cuda` and the integration test above, both run on real GPU hardware (2x RTX 3060, compute capability 8.6) — passing.

### DIFF
--- a/candle-examples/examples/deepseekv3/README.md
+++ b/candle-examples/examples/deepseekv3/README.md
@@ -1,0 +1,26 @@
+# DeepSeek V3
+
+DeepSeek V3 is an MoE model that builds on DeepSeek V2's Multi-head Latent
+Attention (MLA), and switches the MoE router to an auxiliary-loss-free
+("noaux_tc") sigmoid gate: expert and group selection is biased by a learned
+per-expert correction term, while the weights used to combine expert outputs
+come from the un-biased sigmoid scores.
+
+Multi-token prediction and FP8 weights, which the original DeepSeek-V3
+checkpoint also introduces, are not implemented here; this example expects
+dequantized (e.g. bf16) weights and only runs the main next-token-prediction
+path.
+
+- Context length of **128k tokens**
+- 256 routed experts + 1 shared expert per MoE layer, 8 experts active per
+  token
+
+## Running the example
+
+```bash
+$ cargo run --example deepseekv3 --release --features metal -- --prompt "Recursive fibonacci code in Rust:" --sample-len 150
+```
+
+Note that DeepSeek-V3 is a ~671B parameter model, so running it requires a
+machine with a large amount of RAM/VRAM (or a quantized/dequantized subset of
+the weights).
--- a/candle-examples/examples/deepseekv3/main.rs
+++ b/candle-examples/examples/deepseekv3/main.rs
@@ -234,7 +234,11 @@ fn main() -> Result<()> {
         args.revision,
     ));
     let tokenizer_filename = repo.get("tokenizer.json")?;
-    let filenames = candle_examples::hub_load_safetensors(&repo, "model.safetensors.index.json")?;
+    let filenames =
+        match candle_examples::hub_load_safetensors(&repo, "model.safetensors.index.json") {
+            Ok(filenames) => filenames,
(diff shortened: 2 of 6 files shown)


astorise:

PR #3693, opened 2026-06-30
Add Llama 4 (Scout/Maverick) text-decoder model

astorise:
## Summary

Adds a Candle implementation of the Llama 4 text-only decoder (Scout / Maverick), Meta's MoE-based LLM family:

- iRoPE: alternating RoPE/NoPE attention layers, configurable via `no_rope_layers` or `no_rope_layer_interval`
- Parameterless QK-norm (L2 normalization), applied only on RoPE layers
- Attention temperature tuning on NoPE layers
- Interleaved-pair RoPE (`rope_i`) matching the reference implementation
- Top-k sigmoid-gated MoE block with an always-on shared expert, using the real checkpoint's batched expert parameter layout (`experts.gate_up_proj` / `experts.down_proj`, applied with a per-expert matmul rather than per-expert `nn.Linear` sub-modules)
- Dense-vs-MoE layer selection via `interleave_moe_layer_step`

Includes a new `llama4` example (`candle-examples/examples/llama4`) mirroring the existing `deepseekv2`/`llama` examples, plus a README.

## Testing

- Unit test (`forward_smoke_test`) exercising prefill + KV-cache decode with synthetic weights.
- New `#[ignore]`d end-to-end integration test (`candle-examples/tests/llama4_integration.rs`) that downloads `yujiepan/llama-4-tiny-random` — a randomly-initialized checkpoint using the real Llama 4 architecture — and runs a real prefill + decode step through the production model code:

  ```
  cargo test -p candle-examples --release --test llama4_integration -- --ignored
  ```

  This caught three real-checkpoint incompatibilities that synthetic-weight unit tests couldn't:
  - `attn_temperature_tuning` can be serialized as a raw integer rather than a JSON boolean in some configs
  - MoE expert weights are stored as batched parameter tensors, not per-expert named sub-weights
  - the shared expert uses `intermediate_size`, not `intermediate_size_mlp`
- `cargo fmt` / `cargo clippy` clean on `candle-transformers` and `candle-examples`.
- Manually ran `cargo run --example llama4 -- --model-id yujiepan/llama-4-tiny-random --prompt "Hello" --sample-len 5` end-to-end.

The real Scout (109B total / 17B active) and Maverick (400B total / 17B active) checkpoints are gated on the Hub and too large to exercise in CI; the tiny-random checkpoint validates the loading/shape/architecture path against the real Llama4ForConditionalGeneration weight layout instead.

## Test plan

- [x] `cargo test -p candle-transformers --lib llama4`
- [x] `cargo test -p candle-examples --release --test llama4_integration -- --ignored`
- [x] `cargo fmt --check` / `cargo clippy`
- [x] `cargo run --example llama4 -- --prompt "..." --sample-len ...` manually

### DIFF
--- a/candle-examples/examples/llama4/README.md
+++ b/candle-examples/examples/llama4/README.md
@@ -1,0 +1,20 @@
+# Llama 4
+
+Llama 4 is Meta's MoE-based LLM family (Scout / Maverick). This implementation covers
+the text-only decoder: iRoPE (alternating RoPE/NoPE attention layers), QK-norm,
+attention temperature tuning on NoPE layers, and a top-k sigmoid-gated MoE block with
+an always-on shared expert.
+
+- Scout: 17B active / 109B total params, 16 routed experts.
+- Maverick: 17B active / 400B total params, 128 routed experts.
+
+Both are very large checkpoints; running them requires significant RAM/VRAM.
+
+## Running the example
+
+```bash
+$ cargo run --example llama4 --release -- --prompt "Recursive fibonacci code in Rust:" --sample-len 150
+```
+
+By default this targets `meta-llama/Llama-4-Scout-17B-16E-Instruct` (gated on the Hub,
+requires `huggingface-cli login`); pass `--model-id` to point at a different checkpoint.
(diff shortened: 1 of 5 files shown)


astorise:

Issue #2082, opened 2024-04-18
`Op::Unary(arg, UnaryOp::Silu)` branch in Tensor::backward() may cause NaN


Adamska1008:
I tried to fine-tune tinyllama model while I implemented the model structure and the Lora layer independently. During the train process the loss went to NaN, and after 2 thousand years I finally found that the first NaN occurs in the backward process of Silu。Here is related code
```rust
Op::Unary(arg, UnaryOp::Silu) => {
    let sum_grad = grads.or_insert(arg)?;
    // d/dx silu = sigmoid(x) * (1 + x * (1 - sigmoid(x)))
    let mut sigmoid_arg = (*node / arg)?;
    let silu_grad = (&sigmoid_arg * (1. + (arg * (1. - &sigmoid_arg)?)?)?)?;
    *sum_grad = sum_grad.add(&(&grad * silu_grad)?)?;
}
```
`(*node / arg)` here may cause div-zero and result in NaN. I fix it by adding eps, is there any better way to fix it?

LaurentMazare:
Ah thanks for reporting this, that's indeed an issue, I just made #2083 that should fix it in a slightly less efficient way (the tricky bit with a epsilon is that it could still return some nans or some very unsable values if you're unlucky).

Adamska1008:
> Ah thanks for reporting this, that's indeed an issue, I just made #2083 that should fix it in a slightly less efficient way (the tricky bit with a epsilon is that it could still return some nans or some very unsable values if you're unlucky).

Great, I switch to git dependency in cargo.toml and it works well.

singhaki:
@Adamska1008 can your share lora finetuning code? or guide in that direction?

Adamska1008:
> @Adamska1008 can your share lora finetuning code? or guide in that direction?

you can try this candle-lora here https://github.com/EricLBuehler/candle-lora

PR #3694, opened 2026-06-30
candle-kernels: pass /Zc:preprocessor on MSVC for CUDA 13.x

### LINKED ISSUE #3686: candle-kernels 0.10.2 fails to build with CUDA 13.2 on Windows (CCCL requires /Zc:preprocessor)
## Describe the bug

`candle-kernels` fails to compile on Windows with CUDA 13.2 due to a CCCL preprocessor check.

The build aborts with:

```
fatal error C1189:
#error: MSVC/cl.exe with traditional preprocessor is used.
This may lead to unexpected compilation errors.
Please switch to the standard conforming preprocessor by passing /Zc:preprocessor to cl.exe.
You can define CCCL_IGNORE_MSVC_TRADITIONAL_PREPROCESSOR_WARNING to suppress this warning.
```

The error originates from:

```
include/cccl/cuda/std/__cccl/preprocessor.h
```

and is triggered while compiling `candle-kernels`.

---

## Environment

- OS: Windows 11 x64
- Rust: (stable)
- MSVC: 19.44.35225
- CUDA Toolkit: 13.2.78
- NVCC:

```
Cuda compilation tools, release 13.2, V13.2.78
```

- candle-core: 0.10.2
- candle-nn: 0.10.2
- candle-transformers: 0.10.2
- candle-kernels: 0.10.2

---

## Reproduction

Simply build a project that depends on Candle with CUDA enabled:

```bash
cargo build --release
```

The build fails while compiling `candle-kernels`.

---

## Build output

```
warning: candle-kernels@0.10.2: Compiling 11 of 11 PTX kernels

fatal error C1189:
#error: MSVC/cl.exe with traditional preprocessor is used.
Please switch to the standard conforming preprocessor by passing /Zc:preprocessor to cl.exe.

Error: CompilationFailed {
    path: "src\\reduce.cu",
    message: "nvcc error:"
}
```

---

## Additional information

MSVC itself is recent and supports `/Zc:preprocessor`.

```
Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35225 for x64
```

This appears to be caused by `nvcc` invoking `cl.exe` without forwarding `/Zc:preprocessor`.

CUDA 13.2's CCCL headers now require the standard-conforming preprocessor and emit a hard error otherwise.

Would it make sense for `candle-kernels`' build script to pass:

```
-Xcompiler=/Zc:preprocessor
```

when compiling CUDA kernels on Windows, or is CUDA 13.2 currently unsupported?

Thanks!

raskevichai:
Closes #3686.

## Symptom

On Windows with CUDA 13.x, building anything that pulls in `candle-kernels` fails during PTX compilation:

```
fatal error C1189:
#error: MSVC/cl.exe with traditional preprocessor is used.
Please switch to the standard conforming preprocessor by passing
/Zc:preprocessor to cl.exe.

Error: CompilationFailed { path: "src\reduce.cu", message: "nvcc error" }
```

## Root cause

CUDA 13.x ships CCCL headers (`include/cccl/cuda/std/__cccl/preprocessor.h`) that emit a hard `#error` when MSVC's `cl.exe` is invoked in traditional-preprocessor mode. nvcc invokes `cl.exe` as its host compiler, but does not forward `/Zc:preprocessor` by default, so the CCCL include explodes before any kernel source is touched.

The same class of build failure was reported downstream (e.g. mistral.rs #2082) with the same workaround: `NVCC_PREPEND_FLAGS="-Xcompiler /Zc:preprocessor"`. Encoding this in the kernel build script removes the per-user environment dance.

## Fix

`candle-kernels/build.rs` already branches on `TARGET.contains("msvc")` to add `-D_USE_MATH_DEFINES` to the moe builder, and on the non-msvc branch to add `-Xcompiler -fPIC`. The fix follows the same shape:

- Hoist the msvc detection to the top of `main()`.
- On the PTX builder, conditionally append `-Xcompiler=/Zc:preprocessor`.
- On the moe builder, fold the new flag into the existing msvc block alongside `-D_USE_MATH_DEFINES`.

```rust
let is_target_msvc = std::env::var("TARGET")
    .map(|t| t.contains("msvc"))
    .unwrap_or(false);

let mut ptx_builder = KernelBuilder::new()
    .source_dir("src")
    ...;
if is_target_msvc {
    ptx_builder = ptx_builder.arg("-Xcompiler=/Zc:preprocessor");
}
```

`-Xcompiler=...` is the standard nvcc form for forwarding flags to the host compiler. On non-msvc hosts the conditional is false, so the flag is never emitted - no risk of regressing Linux / macOS builds.

## Verification

- `cargo check` in `candle-kernels/` compiles the modified `build.rs` cleanly (failure beyond that is unrelated: cudaforge correctly errors with `NvccNotFound` when no CUDA toolkit is present on the build host).
- The diff is contained to `build.rs`. No `.cu` source changed, no public Rust changed, no downstream crate API change.

End-to-end verification of the Windows + CUDA 13.x build path requires that toolchain, which I do not have available locally. The flag itself is the documented CCCL requirement, called out in the CUDA 13 release notes and in the error message the compiler emits.

### DIFF
--- a/candle-kernels/build.rs
+++ b/candle-kernels/build.rs
@@ -8,16 +8,24 @@ fn main() -> Result<()> {
     println!("cargo::rerun-if-changed=src/cuda_utils.cuh");
     println!("cargo::rerun-if-changed=src/binary_op_macros.cuh");
 
+    let is_target_msvc = std::env::var("TARGET")
+        .map(|t| t.contains("msvc"))
+        .unwrap_or(false);
+
     // Build for PTX
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
     let ptx_path = out_dir.join("ptx.rs");
-    let bindings = KernelBuilder::new()
+    let mut ptx_builder = KernelBuilder::new()
         .source_dir("src") // Scan src/ for .cu files
         .exclude(&["moe_*.cu", "mmvq_gguf.cu", "mmq_*.cu"]) // Exclude statically compiled kernels from ptx build
         .arg("--expt-relaxed-constexpr")
         .arg("-std=c++17")
-        .arg("-O3")
-        .build_ptx()?;
+        .arg("-O3");
+    // CUDA 13.x CCCL headers hard-error unless cl.exe is in /Zc:preprocessor mode.
+    if is_target_msvc {
+        ptx_builder = ptx_builder.arg("-Xcompiler=/Zc:preprocessor");
+    }
+    let bindings = ptx_builder.build_ptx()?;
 
     bindings.write(&ptx_path)?;
 
@@ -52,12 +60,10 @@ fn main() -> Result<()> {
         moe_builder = moe_builder.arg("-DNO_BF16_KERNEL");
     }
 
-    let mut is_target_msvc = false;
-    if let Ok(target) = std::env::var("TARGET") {
-        if target.contains("msvc") {
-            is_target_msvc = true;
-            moe_builder = moe_builder.arg("-D_USE_MATH_DEFINES");
-        }
+    if is_target_msvc {
+        moe_builder = moe_builder
+            .arg("-D_USE_MATH_DEFINES")
+            .arg("-Xcompiler=/Zc:preprocessor");
     }
 
     if !is_target_msvc {


Mahavishnu-K:
Glad you came up with fix!

I wanted it for CUDA 13+ candle-kernel support for my repo `ore-kernel` [https://github.com/Mahavishnu-K/ore-kernel](url)

entergirl:
Hi @raskevichai 鈥?thanks for this fix! I hit the exact same C1189 failure
(CCCL traditional-preprocessor error on quantized.cu / moe_gguf.cu) with
candle-kernels 0.10.2 + CUDA 13.3 + MSVC on Windows.

I applied the same change locally (vendored copy wired via
[patch.crates-io]) and can confirm it works end-to-end:

- Environment: Windows 11, CUDA 13.3, MSVC 2022 (cl 19.44), RTX 4060 (sm_89)
- Command: cargo check -p ramaria-llm --features cuda
- Result: all 11 PTX kernels + 3 moe kernels compile successfully (previously
  the PTX build aborted with atal error C1189)
- Non-CUDA default build is unaffected

This adds the real-machine verification that was missing here. Would be great
to see this merged. Thanks!

---

Note: This is my first time commenting on an upstream proposal, and this
comment was drafted with AI assistance.

PR #3697, opened 2026-07-01
Add aarch64 quant repack kernels

EricLBuehler:


candle-core/src/cpu_backend/mod.rs
```diff
@@ -2981,6 +2982,18 @@ impl BackendStorage for CpuStorage {
         lhs_l: &Layout,
         rhs_l: &Layout,
     ) -> Result<Self> {
+        // no bf16 gemm kernel yet: widen to f32, multiply, narrow back
+        if let (Self::BF16(lhs_v), Self::BF16(rhs_v)) = (self, rhs) {
+            let lhs32 = Self::F32(lhs_v.iter().map(|v| v.to_f32()).collect());
+            let rhs32 = Self::F32(rhs_v.iter().map(|v| v.to_f32()).collect());
```
ivarflakstad: `half` has slice/vec conversion utils through the `HalfFloatSliceExt` trait :)

candle-core/src/cpu_backend/utils.rs
```diff
@@ -361,3 +361,91 @@ pub fn unary_map_vec<T: Copy, U: Copy, F: FnMut(T) -> U, FV: FnMut(&[T], &mut [U
...
+        },
+        crate::StridedBlocks::SingleBlock {
+            start_offset: lo_r,
+            len: len_r,
+        },
+    ) = (lhs_l.strided_blocks(), rhs_l.strided_blocks())
+    {
+        if len_l == len_r && len_l >= PAR_ELEMWISE_MIN {
+            let a = &lhs[lo_l..lo_l + len_l];
+            let b = &rhs[lo_r..lo_r + len_r];
```
ivarflakstad: Could use [NdIter](https://docs.rs/candle-core/latest/candle_core/nditer/struct.NdIter.html) instead of `strided_blocks` to have this work for non-contiguous layouts

candle-core/src/cpu_backend/mod.rs
EricLBuehler: Done!

candle-core/src/cpu_backend/utils.rs
EricLBuehler: Done!

candle-core/src/cpu/neon.rs
```diff
@@ -1,9 +1,10 @@
-use super::Cpu;
+use super::{Cpu, CpuBF16, CpuF16};
 #[cfg(target_arch = "arm")]
 use core::arch::arm::*;
 
 #[cfg(target_arch = "aarch64")]
 use core::arch::aarch64::*;
+use half::{bf16 as f16b, f16};
```
ivarflakstad, line 7: bf16 as f16b is a bit odd

candle-core/src/cpu_backend/mod.rs
```diff
@@ -3053,6 +3054,18 @@ impl BackendStorage for CpuStorage {
...
+        // no bf16 gemm kernel yet: widen to f32, multiply, narrow back
+        if let (Self::BF16(lhs_v), Self::BF16(rhs_v)) = (self, rhs) {
+            let lhs32 = Self::F32(lhs_v.to_f32_vec());
+            let rhs32 = Self::F32(rhs_v.to_f32_vec());
+            let out = MatMul(bmnk).map(&lhs32, lhs_l, &rhs32, rhs_l)?;
+            let Self::F32(out_v) = out else {
+                crate::bail!("matmul dtype mismatch")
+            };
+            return Ok(Self::BF16(
+                out_v.iter().map(|v| half::bf16::from_f32(*v)).collect(),
```
ivarflakstad, line 3066: Could use the slice conversion that half provides here as well

candle-core/src/nditer.rs
```diff
@@ -169,6 +171,30 @@ impl<const N: usize> Iterator for NdIter<N> {
         Some(item)
     }
 
+    #[inline]
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
```
ivarflakstad, line 175: 🤌

candle-core/src/quantized/avx.rs
ivarflakstad, line 1: Could you double check that these are right?
I dont think all have to be marked unsafe, but they may have to be wrapped in an unsafe block at the call site (if the calling fn does not have the same target feature attributes enabled)

ivarflakstad (approved):
Looks great!👍

There are a lot of moving pieces here so let's be a bit careful and make sure we run the different paths on various architectures before we merge (and address the comments from this review ofc)

candle-core/src/quantized/avx.rs
EricLBuehler: Yep, fixed. Some didn't need that!

PR #3721, opened 2026-07-06
Add additive Cache::Paged seam to candle-transformers::models::llama

### LINKED ISSUE #3720: Add an additive Cache::Paged seam to candle-transformers::models::llama for external paged-attention wiring
## Use case

I'm building a downstream inference-serving project on top of `candle` that needs to switch a Llama-family model from the contiguous per-request KV cache to a block-paged one (vLLM-style), so that a block allocator/eviction component can operate at block granularity instead of reallocating a whole-request cache.

`candle-flash-attn` already ships the kernel needed for the paged path — `candle_flash_attn::flash_attn_varlen_paged_windowed(q, k, v, seqlens_q, seqlens_k, block_table, mm_prefix_ranges, max_seqlen_q, max_seqlen_k, softmax_scale, window_size_left, window_size_right, page_block_size, softcap)` — with `k`/`v` laid out as `(num_blocks, page_block_size, num_heads_kv, head_size)` and `block_table` as `(batch_size, max_blocks)` physical block indices per sequence.

## The gap

`candle-transformers::models::llama` has no seam for a downstream crate to use that kernel. `CausalSelfAttention`, `Mlp`, `Block`, and their `forward` methods are private (`struct`/`fn`, not `pub`), and `Cache` hardcodes the contiguous concat-and-narrow KV path (`Cache.kvs[block_idx]`). A caller can drive `Llama::forward`/`Llama::load` end to end, but cannot substitute paged storage into an existing model without either forking this file or reimplementing the whole transformer stack (RoPE, GQA `repeat_kv`, norm placement) to get access to the attention call site.

## Proposed additive API

An additive `Cache` seam, caller-owned paged storage, no behavior change for existing callers:

```rust
/// Physical KV storage and per-sequence block table for paged attention.
/// Constructed and owned by the caller; block_table/seqlens_k are read-only
/// from the model's perspective, key_cache/value_cache are written to.
pub struct PagedKvCache {
    pub key_cache: Tensor,   // (num_blocks, page_block_size, num_kv_heads, head_dim)
    pub value_cache: Tensor, // (num_blocks, page_block_size, num_kv_heads, head_dim)
    pub block_table: Tensor, // (batch_size, max_blocks), physical block ids
    pub seqlens_k: Tensor,   // (batch_size + 1,), cumulative
    pub page_block_size: usize,
}
```

`Cache` gains a per-layer, opt-in paged slot (e.g. `Cache::set_paged_kv(block_idx, PagedKvCache)` / `clear_paged_kv` / `paged_kv`), defaulting to `None` everywhere so `Cache::new` and every existing field/caller is unaffected. When a layer has paged storage attached, `CausalSelfAttention::forward` writes the new K/V into the caller-owned `key_cache`/`value_cache` at the slot `block_table` designates, then calls `flash_attn_varlen_paged_windowed` instead of the dense matmul/softmax or existing `flash_attn` path; otherwise it's the existing contiguous path, byte-for-byte.

Block allocation, eviction, and admission policy are intentionally **not** part of this ask — that stays a caller concern. This is scoped to the minimal seam needed to drive the existing kernel from outside the crate.

Happy to open the PR for this if the shape above looks reasonable.

astorise:
## Summary

Adds an additive `Cache::Paged` seam to `candle-transformers::models::llama` so a downstream crate can drive block-paged KV storage through `Llama::forward` without forking the model or reimplementing RoPE/GQA/norm placement to reach the attention call site (#3720).

- `PagedKvCache`: caller-owned paged KV storage (`key_cache`/`value_cache` shaped `(num_blocks, page_block_size, num_kv_heads, head_dim)`, `block_table`, `seqlens_k`, `page_block_size`).
- `Cache::set_paged_kv(block_idx, PagedKvCache)` / `clear_paged_kv(block_idx)` / `paged_kv(block_idx)`: attach/detach paged storage per transformer layer. Every layer defaults to `None`.
- `CausalSelfAttention::forward` checks the per-layer paged slot first: if set, it writes the new K/V into the caller's `key_cache`/`value_cache` at the slots `block_table` designates, then attends via `candle_flash_attn::flash_attn_varlen_paged_windowed`. If unset, it falls through to today's contiguous concat-and-narrow path, unchanged.

`Cache` stays a struct rather than becoming an enum, so every existing field (`use_kv_cache`, etc.) and every existing call site (`Cache::new`, `candle-examples/examples/llama`, `candle-examples/examples/llava`) is unaffected. The paged path is purely opt-in.

## Why not an enum

An earlier sketch proposed turning `Cache` into `enum Cache { Contiguous(..), Paged(..) }`. That breaks direct field access (`cache.use_kv_cache`) at existing call sites and would require `CausalSelfAttention`/`Block`/`Mlp` to become `pub`. The additive-field version gets the same functional seam — attention already routes through `Llama::forward`, callers never need those internals to be public — with zero behavior change for anyone who doesn't opt in.

## Test plan

- [x] `cargo test -p candle-transformers --lib models::llama::` — 4 new CPU-only unit tests covering: paged storage defaults to `None` (no behavior change), `set_paged_kv`/`clear_paged_kv` lifecycle including out-of-range rejection, slot-index math writing new K/V into the exact physical block/offset the block table designates, and rejecting a block table that doesn't cover the requested position.
- [x] `cargo build -p candle-transformers` / `cargo clippy -p candle-transformers --lib` — clean.
- [x] `cargo build -p candle-examples --example llama --example llava` — confirms zero breakage at the two call sites that use `llama::Cache` directly.
- [x] `cargo test -p candle-transformers --release --features flash-attn --lib models::llama::` on a CUDA GPU — 1 additional test (`paged_attention_matches_dense_causal_attention_on_gpu`, gated behind the `flash-attn` feature since it needs real CUDA hardware) builds a `CausalSelfAttention` by hand and checks that attention through a single-block `PagedKvCache` + `flash_attn_varlen_paged_windowed` matches the existing dense causal path within tolerance.

Fixes #3720

### DIFF
--- a/candle-transformers/src/models/llama.rs
+++ b/candle-transformers/src/models/llama.rs
@@ -142,11 +142,123 @@ impl Config {
     }
 }
 
+/// Physical KV storage and per-sequence block table for paged attention.
+///
+/// Constructed and owned by the caller (e.g. an external block allocator/eviction
+/// engine such as a vLLM-style scheduler). `key_cache`/`value_cache` must be
+/// contiguous: the model writes the newly computed K/V for the current forward
+/// pass into the slots designated by `block_table`, then reads the full paged
+/// history back through
+/// [`candle_flash_attn::flash_attn_varlen_paged_windowed`]. `block_table` and
+/// `seqlens_k` are read-only from the model's perspective — the caller owns
+/// block allocation, eviction, and keeping them in sync with what it wants
+/// written. One `PagedKvCache` covers a single transformer layer; attach one per
+/// layer via [`Cache::set_paged_kv`].
+#[derive(Debug, Clone)]
+pub struct PagedKvCache {
+    /// `(num_blocks, page_block_size, num_kv_heads, head_dim)`, contiguous.
+    pub key_cache: Tensor,
+    /// `(num_blocks, page_block_size, num_kv_heads, head_dim)`, contiguous.
+    pub value_cache: Tensor,
+    /// `(batch_size, max_blocks)`, physical block ids per sequence.
+    pub block_table: Tensor,
+    /// `(batch_size + 1,)`, cumulative sequence lengths (including the tokens
+    /// about to be written by this forward pass), used to drive the varlen kernel.
+    pub seqlens_k: Tensor,
+    pub page_block_size: usize,
+}
+
+impl PagedKvCache {
+    /// Scatters the newly computed K/V (shape `(b_sz, num_kv_heads, seq_len,
+    /// head_dim)`) into `key_cache`/`value_cache` at the physical slots that
+    /// `block_table` designates for absolute positions
+    /// `index_pos..index_pos+seq_len`.
+    fn write_new_kv(&self, k: &Tensor, v: &Tensor, index_pos: usize) -> Result<()> {
+        if !self.key_cache.is_contiguous() || !self.value_cache.is_contiguous() {
+            candle::bail!("PagedKvCache key_cache/value_cache must be contiguous")
+        }
+        let (b_sz, num_kv_heads, seq_len, head_dim) = k.dims4()?;
+        let (num_blocks, page_block_size, kv_heads_cache, head_dim_cache) =
+            self.key_cache.dims4()?;
+        if page_block_size != self.page_block_size {
+            candle::bail!(
+                "PagedKvCache.page_block_size ({}) does not match key_cache shape (block size {page_block_size})",
+                self.page_block_size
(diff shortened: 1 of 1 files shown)


astorise:
The follow-up implementation for #3809 was merged into this PR's head branch via https://github.com/astorise/candle/pull/40.
This PR now includes the BlockMlp factory seam and its bit-for-bit dense-path regression test; merging #3721 will resolve #3809.

astorise:

Issue #3728, opened 2026-07-09
Add a graph-replay-safe decode position seam (`Cache::set_decode_position`) for CUDA graph decode capture


astorise:
## Use case

During autoregressive decoding, the steady-state single-token decode step is identical from one step to the next except for the KV position, which makes it a natural candidate for capture-once/replay-many via a CUDA graph (`cuStreamBeginCapture`/`cuGraphLaunch`): capture the sequence of kernel launches once, then replay it every subsequent step instead of relaunching each kernel individually. This can meaningfully cut per-token launch overhead once the model is small relative to the GPU.

A CUDA graph records the exact kernel launches and memory addresses used during capture — it does not re-derive them. Any buffer whose *contents* need to change between replays must be pre-allocated and updated in place; a value that only exists as a host-side scalar baked into a kernel launch at capture time will not change on replay.

## The gap: rotary position handling is not graph-replay-safe

`candle_transformers::models::llama`'s rotary embedding application computes, per decode step:

```rust
let cos = cache.cos.narrow(0, index_pos, seq_len)?; // index_pos: usize, host-side
let sin = cache.sin.narrow(0, index_pos, seq_len)?;
```

`Tensor::narrow` is a zero-copy view: it clones the same underlying storage and adjusts a host-computed offset into the existing `Layout`. That offset is derived from `index_pos` *at the Rust/host level, before the kernel launch*, and becomes part of the actual device pointer passed to whatever kernel consumes the narrowed tensor. A CUDA graph capturing this sequence bakes in that specific pointer. Since `index_pos` increments every decode step, replaying a graph captured at step *N* at step *N+1* would silently read the rotary embeddings for step *N* again — a correctness bug, not just a missed optimization, if graph capture is layered naively on top of the existing API.

(The existing contiguous KV cache, `Cache.kvs[block_idx]`, has the same class of problem for a different reason: it grows via `Tensor::cat` — a fresh, larger allocation every step — which is likewise disallowed for a captured buffer, since a tensor allocated inside the captured closure becomes a graph-owned allocation whose device memory is only valid while the graph is executing. A pre-allocated, in-place-updated KV cache design solves this half separately; this issue only concerns the rotary position lookup, and is expected to compose with such a cache rather than solve the contiguous-cache case itself.)

## Proposed additive change

A way to drive the rotary embedding lookup from a persistent, in-place-updatable device tensor instead of a host `usize`, e.g.:

```rust
impl Cache {
[...]

PR #3729, opened 2026-07-09
Add CUDA graph capture/replay and a graph-replay-safe decode position seam for Llama

astorise:
## Summary

Addresses #3728.

Adds two related pieces:

1. **`CudaGraph`** (`candle-core/src/cuda_backend/graph.rs`): a thin wrapper around `cuStreamBeginCapture`/`cuStreamEndCapture`/`cuGraphInstantiateWithFlags`/`cuGraphLaunch` that lets a fixed sequence of CUDA operations be captured once and replayed with a single `cuGraphLaunch` instead of relaunching every individual kernel — useful for autoregressive decoding, where the same kernel sequence runs once per generated token and per-launch overhead can dominate for small models. Host-to-device uploads of small constants during capture are served from a cache (`CudaDevice::enable_cuda_graph_htod_cache`), since a fresh upload cannot be recorded as a replayable graph node; the same guard also covers the kernel-launch parameter vectors used for non-contiguous tensor layouts. The warm-up run callers must perform before capturing (to JIT-load kernels and seed that cache) needs to be wrapped in the same guard, which is documented on `CudaGraph::capture`.

2. **`Cache::set_decode_position`** (`candle-transformers/src/models/llama.rs`): an additive seam so a decode step captured with `CudaGraph` doesn't silently go stale. Today, `CausalSelfAttention`'s rotary embedding lookup computes `cache.cos.narrow(0, index_pos, seq_len)` — a host-derived offset baked into the exact device pointer a graph capture would record. Since `index_pos` increments every decode step, replaying a graph captured at step *N* at step *N+1* would silently reuse step *N*'s rotary embeddings, a correctness bug rather than a missed optimization. `Cache::set_decode_position(block_idx, position: Tensor)` attaches a persistent, caller-owned position tensor (shape `(1,)`) per transformer layer; when attached, the rotary lookup uses `Tensor::index_select` against that device tensor instead of `narrow`. The caller updates the tensor's *contents* in place (e.g. via `Tensor::slice_set`) before each `CudaGraph::replay()`; its identity/address stays fixed, so the same captured kernel launches pick up the new position on the next replay.

Both changes are purely additive:
- `CudaGraph` is a new, independent type; nothing existing calls into it.
- `Cache::set_decode_position` defaults every layer to `None`, which preserves the existing `narrow`-based path byte-for-byte. `Cache::new` and every existing caller of `Llama::forward`/`forward_input_embed` are unaffected.

The seam is scoped to the decode step only (`seq_len == 1`, one query token); `CausalSelfAttention::forward` returns an error if a layer with a decode position attached is called with `seq_len != 1`, rather than silently doing something wrong on a misuse. Prefill (`seq_len > 1`) always uses the existing `narrow`-based path, whether or not a decode position happens to be attached.

## Test plan

- [ ] `cargo test -p candle-core --features cuda` — includes a `CudaGraph` capture/replay round-trip test (`candle-core/tests/cuda_graph_tests.rs`): captures an elementwise multiply into a persistent output buffer, checks the buffer is unchanged immediately after capture (before any replay), then that replaying it reproduces the expected result, twice.
- [ ] `cargo test -p candle-transformers` (CPU, no feature flags) — new `Cache::set_decode_position` unit tests: the seam is additive by default, `set_decode_position`/`clear_decode_position` round-trip and reject an out-of-range layer or a wrong-shaped position tensor, the `index_select`-based rotary lookup matches the existing `narrow`-based one for the same position, and a multi-token (`seq_len != 1`) call is rejected once a decode position is attached.
- [ ] `cargo test -p candle-transformers --release --features flash-attn --lib models::llama::` on a CUDA device — a dedicated GPU test captures a decode step's rotary lookup inside an actual `CudaGraph`, replays it once at the captured position, updates the attached position tensor's *contents* in place, replays again, and asserts the output tracks the new position rather than the one live at capture time (i.e. it isn't stuck on the captured value). Verified passing on real GPU hardware.

### DIFF
--- a/candle-core/src/cuda_backend/device.rs
+++ b/candle-core/src/cuda_backend/device.rs
@@ -51,6 +51,24 @@ impl Drop for CudaGraphHtodCacheGuard {
     }
 }
 
+/// Restores cudarc's event-tracking setting when dropped.
+/// Created by [`CudaDevice::pause_event_tracking`].
+#[must_use]
+pub(crate) struct EventTrackingGuard {
+    context: Arc<cudarc::driver::CudaContext>,
+    restore_to_enabled: bool,
+}
+
+impl Drop for EventTrackingGuard {
+    fn drop(&mut self) {
+        if self.restore_to_enabled {
+            // SAFETY: re-enabling the tracking that `pause_event_tracking` disabled,
+            // restoring the context to the state it was in before the guard.
+            unsafe { self.context.enable_event_tracking() };
+        }
+    }
+}
+
 pub struct ModuleStore {
     mdls: [Option<Arc<cudarc::driver::CudaModule>>; kernels::ALL_IDS.len()],
 }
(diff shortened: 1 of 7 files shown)


astorise:
Closing — I opened three overlapping PRs for CUDA graph capture instead of iterating on one branch. Closing all three; this work is moving to a separate crate.

PR #3733, opened 2026-07-10
Add CUDA graph capture/replay support with a graph-safe paged-attention decode path

### LINKED ISSUE #3732: PagedKvCache::write_new_kv's host readback breaks CudaGraph capture for the decode step
## Use case

A downstream inference-serving deployment wants to capture the steady-state
single-token decode step with `candle_core::CudaGraph` and replay it every
subsequent step instead of relaunching each kernel individually, composed
with the paged-attention KV cache. This surfaced after wiring graph-replay-safe
rotary position handling (a device-tensor decode-position seam for RoPE) on
top of the existing paged-attention path.

## The gap: `PagedKvCache::write_new_kv`'s scatter-index computation is not graph-capturable

`CudaGraph::capture`'s own doc comment states the constraint plainly: a
captured closure's operations must be pure device-side kernel launches with
stable buffer addresses — no host/device synchronization, no new allocations.

`write_new_kv` (`candle-transformers/src/models/llama.rs`) is called on
*every* forward pass through the paged path (prefill and decode) and its
very first operation is:

```rust
fn write_new_kv(&self, k: &Tensor, v: &Tensor, index_pos: usize) -> Result<()> {
    ...
    let block_table = self.block_table.to_dtype(DType::U32)?.to_vec2::<u32>()?;
    ...
    let last_pos = index_pos + seq_len - 1;
    let mut slots = Vec::with_capacity(b_sz * seq_len);
    for row in &block_table {
        // host-side arithmetic over the block table's *contents*
        ...
        slots.push((physical_block * self.page_block_size + offset) as u32);
    }
    let indices = Tensor::from_vec(slots, (b_sz * seq_len, 1, 1), device)?...;
    ...
    key_flat.scatter_set(&indices, &k_flat, 0)?;
    value_flat.scatter_set(&indices, &v_flat, 0)?;
    Ok(())
}
```

`Tensor::to_vec2` on a CUDA tensor is a blocking device-to-host copy. Issuing
this (or any synchronizing operation) on a stream that's mid-capture is
invalid in CUDA's stream-capture API and would either fail the capture
outright, or — if some layer tolerated it — silently freeze the *destination*
scatter indices to whatever `index_pos`/block-table contents produced at
capture time, while `k`/`v` *values* keep updating correctly on each replay
via the captured kernels. That would silently misplace every subsequent
decode step's K/V into the position captured at step one. `Tensor::from_vec(slots,
...)` inside the closure is also a fresh allocation, independently disallowed
during capture.

Note this is a real correctness/capturability gap in the *existing*,
already-merged paged-attention path, found by reading the source before
attempting a graph capture — not a regression from the decode-position work.

## Proposed additive change

A way to drive `write_new_kv`'s scatter destinations without a host
round-trip. Two shapes seem reasonable (illustrative, not prescriptive):

1. **Compute scatter indices on-device.** `block_table` (`(batch,
   max_blocks)`, already a device tensor) combined with `index_pos`/`seq_len`
   could derive the flat scatter offsets via `Tensor::gather`/arithmetic ops
   entirely on-device (e.g. `logical_block = pos / page_block_size`, `offset
   = pos % page_block_size`, gather the physical block id from `block_table`
   at `logical_block`, then `physical_block * page_block_size + offset`),
   replacing the `to_vec2` + host loop with device tensor ops. `index_pos`
   itself would need the same device-tensor treatment (not a host `usize`)
   to keep the whole thing graph-safe end to end for the decode step.
2. **Accept caller-precomputed device indices directly**, e.g. a new method
   like `write_new_kv_at(&self, k: &Tensor, v: &Tensor, indices: &Tensor)`
   that skips the internal derivation entirely. The caller already tracks
   each sequence's block assignment on the host before ever uploading
   `block_table` to the device — it doesn't strictly need to re-derive those
   same indices from a device readback of the tensor it just built. For the
   graph-capture case specifically, the caller would maintain a persistent
   device index tensor (written in place before each replay) and pass it
   straight through.

Either way, this should remain scoped to the decode step (`seq_len == 1`) if
that simplifies the derivation — prefill can keep the existing
`write_new_kv` behavior unconditionally.

## Scope

Additive only — default behavior (existing `write_new_kv` call sites,
unchanged) must be unaffected.

astorise:
Title: Add CUDA graph capture/replay support with a graph-safe paged-attention decode path

## Summary

This PR adds three additive building blocks that together let a decode step
be captured once with a CUDA graph and replayed on every subsequent step,
composed with a vLLM-style paged KV cache:

1. **`candle_core::CudaGraph`** (`candle-core/src/cuda_backend/graph.rs`): a
   thin capture/replay wrapper around the CUDA stream-capture API, plus an
   htod-upload cache (`CudaDevice::enable_cuda_graph_htod_cache`) so that
   repeated host-constant uploads inside a captured closure reuse a cached
   allocation instead of failing capture.
2. **`PagedKvCache`** (`candle-transformers/src/models/llama.rs`):
   caller-owned, block-paged KV storage (vLLM-style) attached per
   transformer layer via `Cache::set_paged_kv`. When attached, that layer's
   attention routes through a paged/varlen flash-attention kernel against
   caller-managed `key_cache`/`value_cache`/`block_table` instead of the
   contiguous concat-and-narrow cache. Fully additive: layers default to the
   existing contiguous path.
3. **`Cache::set_decode_position`**: a caller-owned, persistent device
   tensor that a layer's rotary-embedding lookup reads via
   `Tensor::index_select` instead of `narrow(0, index_pos, seq_len)`, so the
   RoPE lookup's device pointer doesn't get baked into a graph capture at a
   position that goes stale on replay.

## The bug this PR also fixes (#3732)

With the above three in place, driving the paged decode path through an
actual `CudaGraph::capture` surfaces a fourth gap: `PagedKvCache`'s
scatter-index computation for writing new K/V is not graph-capturable.

`write_new_kv`'s first operation is a blocking device-to-host readback of
`block_table` (`Tensor::to_vec2`) followed by a host-side loop and a fresh
`Tensor::from_vec` upload of the resulting scatter indices — both of which
are invalid on a stream mid-capture. Left unfixed, a capture either fails
outright or (worse) silently freezes the scatter *destination* to whatever
`block_table` contained at capture time, while the K/V *values* keep
updating correctly on replay — silently misplacing every subsequent decode
step's KV into the position captured at step one.

Fixes this with the same pattern as `set_decode_position`:

```rust
impl Cache {
    /// Attaches a persistent, caller-owned device tensor of flat scatter
    /// indices (shape `(b_sz,)`) for one layer's `PagedKvCache::write_new_kv`
    /// decode step. When attached, `write_new_kv` scatters directly against
    /// it instead of deriving indices from a host readback of `block_table`.
    /// Scoped to the decode step only (`seq_len == 1`).
    pub fn set_paged_kv_decode_slot(&mut self, block_idx: usize, indices: Tensor) -> Result<()>;
    pub fn clear_paged_kv_decode_slot(&mut self, block_idx: usize);
    pub fn paged_kv_decode_slot(&self, block_idx: usize) -> Option<&Tensor>;
}
```

The caller updates the attached tensor's *contents* in place (e.g. via
`Tensor::slice_set`) before each `CudaGraph::replay()`; its identity/address
stays fixed across replays, which is what makes the capture valid.

## Scope / compatibility

Everything here is additive:
- `Cache::new` and every existing caller of the contiguous KV path are
  unaffected — no behavior change, no new required arguments.
- `set_paged_kv`, `set_decode_position`, and `set_paged_kv_decode_slot` all
  default to `None` per layer; a layer only routes through the new paths
  once a caller explicitly attaches the corresponding tensor.
- Prefill (`seq_len > 1`) always uses the existing paths unconditionally;
  the position/decode-slot seams are rejected (return `Err`) if attached to
  a multi-token call, rather than silently producing wrong output.

## Testing

- `cargo fmt`, `cargo clippy --lib --tests -- -D warnings` (candle-core,
  candle-nn, candle-transformers).
- `cargo test` (candle-core, candle-nn, candle-transformers): CPU-only unit
  tests, including numerical equivalence between the new device-tensor
  paths and the existing host-derived paths for the same position/indices,
  and rejection of malformed/prefill input on each new seam.
- GPU tests (`--features cuda,flash-attn`, dispatched on a CUDA runner):
  - `paged_attention_matches_dense_causal_attention_on_gpu`: paged path
    matches the dense causal-attention reference.
  - `decode_position_stays_correct_across_cuda_graph_replay`: captures a
    decode-step rotary lookup in a real `CudaGraph`, replays it, updates the
    attached position tensor's contents in place, replays again, and
    asserts the output tracks the new position rather than the one live at
    capture time.
  - `paged_kv_decode_slot_stays_correct_across_cuda_graph_replay`: same
    property for the paged KV scatter destination — captures
    `PagedKvCache::write_new_kv`'s decode-step scatter, replays, updates the
    decode-slot tensor's contents in place, replays again, and asserts the
    scatter lands at the new slot.

Closes #3732


astorise:

Issue #3735, opened 2026-07-11
CPU conv2d backward produces incorrect gradients for non-contiguous kernels


EthanAlmloff:
## Summary

On CPU, `conv2d` backward can produce incorrect gradients when the convolution kernel is non-contiguous.

A transposed kernel and its contiguous copy represent the same logical values, so they should produce identical forward results and gradients. Instead, the gradients differ.

## Reproduction

```rs
use candle_core::Var;
use candle_core::{test_utils, Device, Var};

let device = &Device::Cpu;
let input_data: Vec<f32> = (1..=18).map(|value| value as f32).collect();
let kernel = Var::from_slice(
    &(1..=16).map(|value| value as f32).collect::<Vec<_>>(),
    (2, 2, 2, 2),
    device,
)?;
let non_contiguous_kernel = kernel.transpose(0, 1)?;

let input = Var::from_slice(&input_data, (1, 2, 3, 3), device)?;
let loss = input
    .conv2d(&non_contiguous_kernel, 0, 1, 1, 1)?
    .sqr()?
    .sum_all()?;
let grads = loss.backward()?;
let grad_input = grads.get(&input).unwrap();
let grad_kernel = grads.get(&kernel).unwrap();

let expected_input = Var::from_slice(&input_data, (1, 2, 3, 3), device)?;
let expected_kernel = Var::from_slice(
    &(1..=16).map(|value| value as f32).collect::<Vec<_>>(),
    (2, 2, 2, 2),
    device,
)?;
let expected_loss = expected_input
    .conv2d(&expected_kernel.transpose(0, 1)?.contiguous()?, 0, 1, 1, 1)?
    .sqr()?
    .sum_all()?;
let expected_grads = expected_loss.backward()?;
let expected_grad_input = expected_grads.get(&expected_input).unwrap();
let expected_grad_kernel = expected_grads.get(&expected_kernel).unwrap();

assert_eq!(
    test_utils::to_vec1_round(&grad_input.flatten_all()?, 4)?,
    test_utils::to_vec1_round(&expected_grad_input.flatten_all()?, 4)?,
);
assert_eq!(
    test_utils::to_vec1_round(&grad_kernel.flatten_all()?, 4)?,
    test_utils::to_vec1_round(&expected_grad_kernel.flatten_all()?, 4)?,
);
Ok(())
```

## Expected behavior

The gradients for input and kernel should match the gradients from the equivalent contiguous-kernel computation.

## Actual behavior

[...]

PR #3136, opened 2025-10-17
CPU Conv2d: separate module, tiled im2col, specialization

slckl:
# Changes
- new `cpu_backend/conv2d` module with all the conv2d cpu stuff in there
- new general conv2d impl, tentatively called `TiledIm2Col`
- conv2d entry point with specialization -> default general fallback kernel choice flow
- specialized 1x1 fast path as first case with decent perf
- makes tiledim2col + specialized 1x1 as the default

# What's missing?
- any kind of arm/neon benchmark numbers
- looking back at this, there seems to be an opening for parallelized im2col -> full im2col matrix -> gemm impl, but that will still suffer from slow strided copy at the end
- other small kernels not hitting the fast 1x1 path are probably still fastest on the full im2col - we could add a "specialization" here, but need more benchmarks... 

# Benchmarks 
Benchmarks obtained using `cargo bench conv2d -p candle-nn`, and switching the `DEFAULT` inside `conv2d.rs`.
I bolded (well, sonnet did) the winners in each bench.
Note that conv2d + bias benches are heavily dominated by the `+bias` term obscuring conv diff (as evidenced by the huge delta vs non-bias benches).

## i7-12700h

<table>
<thead>
<tr>
<th>Benchmark</th>
<th>Method</th>
<th>Time</th>
</tr>
</thead>
<tbody>
<tr>
<td rowspan="3">cpu_conv2d_F32_i128_k3x3_b/iter</td>
<td>Direct</td>
<td>[5.1890 ms 5.2280 ms 5.2659 ms]</td>
</tr>
<tr>
<td><strong>full im2col</strong></td>
<td><strong>[5.0741 ms 5.1176 ms 5.1603 ms]</strong></td>
</tr>
<tr>
<td>tiled im2col</td>
<td>[5.1679 ms 5.2891 ms 5.4236 ms]</td>
</tr>
<tr>
<td rowspan="4">cpu_conv2d_F32_i128_k1x1_nb/iter</td>
<td>Direct</td>
<td>[64.180 µs 64.335 µs 64.531 µs]</td>
</tr>
<tr>
<td>full im2col</td>
<td>[48.858 µs 49.143 µs 49.510 µs]</td>
</tr>
<tr>
<td>tiled im2col</td>
<td>[103.65 µs 104.83 µs 106.87 µs]</td>
</tr>
<tr>
<td><strong>Specialized</strong></td>
<td><strong>[31.947 µs 32.132 µs 32.384 µs]</strong></td>
</tr>
<tr>
[...]

PR #3736, opened 2026-07-12
[CPU] Fix wrong offset for non-contiguous conv2d im2col kernels

Sammy-Dabbas:
I ran into this while looking at #3735. In conv2d_im2col_gemm, a non-contiguous kernel gets copied into a fresh buffer (kernel_c) that starts at index 0, but the matmul layout was still built with the original kernel's start offset. If that kernel has a non zero offset, a channel slice of a larger pointwise weight, the matmul reads past kernel_c and the forward output and gradients come out wrong. Offset 0 lines the layout up with the buffer.

The test in conv_tests.rs runs a sliced 1x1 kernel and checks its forward result and both gradients against the same kernel made contiguous. It fails before the change and passes after.

The transposed-kernel repro in #3735 already passes on current main, probably from #3136 and #3405, so this is a different non-contiguous path, not that same case. This is a distinct case from the #3735 repro, so I did not link it to close the issue.


ivarflakstad (approved):
Thanks! 👌

PR #3742, opened 2026-07-16
fix(kernels): avoid redefining cuda fp16 cmp functions

### LINKED ISSUE #3737: regression in support for nvidia rtx 20xx GPUs with newer nvcc toolchains
Commit 1d7e9274591ef7eb101d965b71243a41994214d2 seems to have broken builds for me. Judging by the commit message it seems like the 750 -> 800 change was probably correct, but the removal of the cudacc version threshold seems to be a problem. I'm using nvcc version 12.9 (which is currently the latest available in Gentoo's repositories) which provides a conflicting definition for these functions. The intricacies of nvidia toolchain compatibility is a bit beyond me, but please feel free to ask if you need me to provide additional information about my system.

```
$ nvcc --version                                      
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2025 NVIDIA Corporation
Built on Tue_May_27_02:21:03_PDT_2025
Cuda compilation tools, release 12.9, V12.9.86
Build cuda_12.9.r12.9/compiler.36037853_0
```

```
$ nvidia-smi --query-gpu=name,compute_cap --format=csv
name, compute_cap
NVIDIA GeForce RTX 2070, 7.5
```
For what it's worth, I have confirmed that restoring the version ceiling at least lets me compile successfully:
```
#if (__CUDACC_VER_MAJOR__ < 12 || __CUDACC_VER_MINOR__ < 2) && __CUDA_ARCH__ < 800
__device__ __forceinline__ __half __hmax_nan(__half a, __half b) {
    return __hisnan(a) ? a : (__hisnan(b) ? b : __hmax(a, b));
}
__device__ __forceinline__ __half __hmin_nan(__half a, __half b) {
    return __hisnan(a) ? a : (__hisnan(b) ? b : __hmin(a, b));
}
#endif
```
If this seems like a reasonable fix, I can make a PR.

embrynod:
1d7e927 removed a macro check which prevented __hmax_nan and __hmin_nan from being defined for cudacc versions < 12.3. This introduced a regression for users with (relatively) older graphics cards (i.e. __CUDA_ARCH__ < 800) but new cudacc versions.

This change adds this check back, but keeps the 750 -> 800.

fixes #3737

### DIFF
--- a/candle-kernels/src/compatibility.cuh
+++ b/candle-kernels/src/compatibility.cuh
@@ -5,9 +5,13 @@
 // Table showing which features are supported on which compute capability
 // https://docs.nvidia.com/cuda/cuda-c-programming-guide/#features-and-technical-specifications
 
-// FIXME: the minimum compute capabilities are just guesses since the table is not specific enough
-
-#if __CUDA_ARCH__ < 800
+// NVIDIA's cuda_fp16.h started providing __hmax_nan/__hmin_nan for pre-Ampere
+// targets (__CUDA_ARCH__ < 800) in CUDA 12.2: in 12.1 and earlier these are
+// gated behind `__CUDA_ARCH__ >= 800`, in 12.2+ they are defined unconditionally.
+// So only define the fallback for toolkits strictly older than 12.2, otherwise
+// it collides with the native definition. Verified against the 12.1.105 and
+// 12.2.140 cuda_fp16.hpp headers.
+#if (__CUDACC_VER_MAJOR__ < 12 || (__CUDACC_VER_MAJOR__ == 12 && __CUDACC_VER_MINOR__ < 2)) && __CUDA_ARCH__ < 800
 __device__ __forceinline__ __half __hmax_nan(__half a, __half b) {
     return __hisnan(a) ? a : (__hisnan(b) ? b : __hmax(a, b));
 }


astorise:
Thanks for restoring this guard — the redefinition regression on pre-Ampere targets (#3737) is real and this correctly fixes it in spirit.

While integrating this into a fork, review turned up an issue with the version comparison itself:

#if (__CUDACC_VER_MAJOR__ < 12 || __CUDACC_VER_MINOR__ < 2) && __CUDA_ARCH__ < 800
This isn't a valid major.minor comparison — it ORs the minor-version check independently of the major version, so it misfires in both directions:

CUDA 13.0 (major=13, minor=0): 13 < 12 is false, but 0 < 2 is true → the OR is true, so the fallback gets defined even on CUDA 13.0, which already provides __hmax_nan/__hmin_nan natively — reintroducing the exact "redefinition" break this PR fixes, just on CUDA 13.x + a pre-800 arch instead of CUDA >= 12.3.
CUDA 12.2 (major=12, minor=2): 2 < 2 is false → the fallback is skipped even though (see below) 12.2 doesn't provide these natively for pre-800 targets — a missing-symbol break on that specific toolkit.
On the actual cutoff: I initially assumed the commit message's < 12.3 was correct and just fixed the comparison logic, but a reviewer (Codex) pointed out the real cutoff is 12.2, not 12.3. Rather than trust either claim, I checked NVIDIA's actual cuda_fp16.hpp (pulled the nvidia-cuda-runtime-cu12 wheels for 12.1.105 and 12.2.140 from PyPI and inspected the preprocessor guards around the __hmax_nan/__hmin_nan definitions directly):

CUDA 12.1.105: the definitions are enclosed by #if defined(_NVHPC_CUDA) || !defined(__CUDA_ARCH__) || (__CUDA_ARCH__ >= 800) — so a device build with __CUDA_ARCH__ < 800 does not get them. Fallback is required. (11.8 and 12.0 match this same guarded form.)
CUDA 12.2.140: no arch guard at all around the definitions (only the include guard and #if defined(__cplusplus)) — so they're defined unconditionally, for every arch including < 800. Fallback collides. (12.3 matches this same unguarded form.)
So the flip is exactly at 12.2, not 12.3. The corrected guard:

#if (__CUDACC_VER_MAJOR__ < 12 || (__CUDACC_VER_MAJOR__ == 12 && __CUDACC_VER_MINOR__ < 2)) && __CUDA_ARCH__ < 800
I verified all the resulting preprocessor branches with gcc -E across CUDA 11.8/12.0/12.1/12.2/12.3/13.0, both pre- and post-800 arch (no nvcc available to compile the real header end-to-end, but the branch selection logic matches the header evidence above in every case).

Pushed this on top of the PR in my fork if useful as a starting point: astorise/candle@1eac871 (also includes the major.minor comparison fix as a separate commit, 9c52a93). Happy to open it as a PR against this branch or main if that's easier to review than diff links.

embrynod:
Updated with the suggested changes.

PR #3747, opened 2026-07-17
Fix quantized Gemma 3 MLP activation (gelu_pytorch_tanh, not SiLU)

### LINKED ISSUE #3745: Gemma models use the wrong activation function
## Summary

`candle-transformers/src/models/quantized_gemma3.rs` hardcodes SiLU in the gated MLP:

```rust
let silu = candle_nn::ops::silu(&gate)?;
let gated = (silu * up)?;
self.feed_forward_down.forward(&gated)
```

Gemma 3's FFN activation is `gelu_pytorch_tanh`, not SiLU. Because GELU-tanh and SiLU are close in shape, nothing overflows and per-layer norms look healthy; the model just quietly produces lower-quality output, which is presumably why this has gone unnoticed.

Present in v0.11.0 and on current `main` (`candle-transformers/src/models/quantized_gemma3.rs`, the two lines in `impl Module for Mlp`).

Previously attempted in #3326; probably responsible for weirdness in #3492 and might be the root cause of #3299, as Gemma 4 uses the same activation function.

## Evidence that Gemma 3 is GELU-tanh

1. **HF transformers config default:** `src/transformers/models/gemma3/configuration_gemma3.py`:
   ```python
   hidden_activation: str = "gelu_pytorch_tanh"
   ```
2. **llama.cpp:** `src/models/gemma3.cpp` builds the FFN with `LLM_FFN_GELU` (not `LLM_FFN_SILU`).
3. **candle itself:**the non-quantized `models/gemma3.rs` reads `cfg.hidden_activation` from the checkpoint config and therefore runs GELU-tanh at inference. The quantized and non-quantized implementations of the same model currently disagree.

Sammy-Dabbas:
Fixes #3745

The quantized Gemma 3 MLP hardcoded SiLU, but Gemma 3 gates its feed-forward network with gelu_pytorch_tanh. The non-quantized gemma3.rs applies cfg.hidden_activation (gelu_pytorch_tanh for Gemma 3 checkpoints). The quantized model loads from GGUF metadata and has no config field, so it applies the activation directly, the same way quantized_glm4.rs hardcodes its Activation. This routes through the same code path as the non-quantized model (Activation::GeluPytorchTanh, the gelu tanh approximation).

SiLU and gelu_pytorch_tanh behave similarly, so the wrong one does not crash or produce obvious garbage; it just makes the output worse, which is why it went unnoticed. #3326 attempted a sliding-window plus GELU rework but was self-closed by its author, so this is the focused activation-only fix.

### DIFF
--- a/candle-transformers/src/models/quantized_gemma3.rs
+++ b/candle-transformers/src/models/quantized_gemma3.rs
@@ -57,8 +57,8 @@ impl Module for Mlp {
     fn forward(&self, xs: &Tensor) -> Result<Tensor> {
         let gate = self.feed_forward_gate.forward(xs)?;
         let up = self.feed_forward_up.forward(xs)?;
-        let silu = candle_nn::ops::silu(&gate)?;
-        let gated = (silu * up)?;
+        let gate = gate.apply(&candle_nn::Activation::GeluPytorchTanh)?;
+        let gated = (gate * up)?;
         self.feed_forward_down.forward(&gated)
     }
 }


jkantsios:
I was curious about the accuracy/performance impact this would have and produced the following perplexity score changes comparing SiLU and gelu_pytorch.

image
image

This definitely seems to confirm the efficacy of this change.

PR #3754, opened 2026-07-20
Fix clippy warnings for Rust 1.97

GregoryBolshakov:
Rust 1.97 promotes `useless_borrows_in_formatting`, `for_kv_map`, and `manual_filter` to warnings, which fail CI under `-D warnings`. Applied `cargo clippy --fix`; no behavior changes.

### DIFF
--- a/candle-transformers/src/generation/mod.rs
+++ b/candle-transformers/src/generation/mod.rs
@@ -29,7 +29,7 @@ impl LogitsProcessor {
     }
 
     pub fn new(seed: u64, temperature: Option<f64>, top_p: Option<f64>) -> Self {
-        let temperature = temperature.and_then(|v| if v < 1e-7 { None } else { Some(v) });
+        let temperature = temperature.filter(|&v| v >= 1e-7);
         let sampling = match temperature {
             None => Sampling::ArgMax,
             Some(temperature) => match top_p {
--- a/candle-transformers/src/generation/mod.rs
+++ b/candle-transformers/src/generation/mod.rs
@@ -29,7 +29,10 @@ impl LogitsProcessor {
     }
 
     pub fn new(seed: u64, temperature: Option<f64>, top_p: Option<f64>) -> Self {
-        let temperature = temperature.filter(|&v| v >= 1e-7);
+        // `manual_filter` suggests `.filter(|&v| v >= 1e-7)`, but `f64` is only `PartialOrd`:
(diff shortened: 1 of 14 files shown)


candle-transformers/src/generation/mod.rs
```diff
@@ -29,7 +29,7 @@ impl LogitsProcessor {
     }
 
     pub fn new(seed: u64, temperature: Option<f64>, top_p: Option<f64>) -> Self {
-        let temperature = temperature.and_then(|v| if v < 1e-7 { None } else { Some(v) });
+        let temperature = temperature.filter(|&v| v >= 1e-7);
```
tomsanbear: Could you update this change to preserve the original partial-order semantics? The current version handles NaN differently than the original code.

tomsanbear (changes requested):
Thank you for the fix, I just hit this myself as well. One change to preserve existing behaviour.

candle-transformers/src/generation/mod.rs
GregoryBolshakov: Good catch, you're right. `filter(|&v| v >= 1e-7)` drops NaN where the original kept it, since `f64` is only `PartialOrd`.

Reverted to the original expression with an `#[allow(clippy::manual_filter)]` and a comment, so behaviour is unchanged. (`!(v < 1e-7)` also preserves it but trips `neg_cmp_op_on_partial_ord`.) Happy to use `v.is_nan() || v >= 1e-7` instead if you'd rather not carry the allow.

The other `manual_filter` hunk in stable_diffusion/attention.rs is on `Option<usize>`, so it's unaffected.

tomsanbear (approved):

GregoryBolshakov:
@ivarflakstad could you approve the workflow run here? Approved by @tomsanbear, but CI hasn't run on the latest commit since fork PRs need a maintainer to kick it off. This unblocks clippy on the whole workspace for Rust 1.97.

GregoryBolshakov:
@tomsanbear looks like CI still needs someone with write access to approve the workflow run (fork PR). Could you kick it off and merge if green? ivarflakstad seems to be away from the repo at the moment.

ivarflakstad:
Hey!
Back from vacation :) 

Thanks for the PR. I kicked off CI  - we'll have this merged soon 👍

ivarflakstad (approved):
🙌

PR #3755, opened 2026-07-20
Fix Metal memory leak by draining autorelease pools (#2271)

### LINKED ISSUE #2271: Metal memory leak multiplying matrices
Running this code multiplying a 784x100 matrix times a 100x10 matrix seems to leak memory. The memory usage gradually increases to over a gigabyte when running with the metal feature enabled in release mode on commit `2b10aaa`:

```rust
use anyhow::Result;
use candle_core::{
    backend::BackendDevice,
    Device, MetalDevice, Tensor,
};

fn main() -> Result<()> {
    let device = Device::Metal(MetalDevice::new(0)?);
    let first = Tensor::randn(0f32, 1.0, (784, 100), &device)?;
    let second = Tensor::randn(0f32, 1.0, (100, 10), &device)?;
    loop {
        first.matmul(&second)?;
    }
}
```

On the CPU, memory usage stays steady at ~2 mb. This also seems to effect quantized matrix multiplication

fabianboesiger:
Fixes #2271. Running Metal ops in a loop (matmul, and any other op) leaked memory: the resident set grew without bound.

## Root cause

The leak is **not** in candle's own tracked structures. With the repro running, the buffer pool, the in-flight command buffers, and the `prev_ce_outputs` fence map all stay constant, yet RSS climbs linearly:

```
iter       0  RSS  27.57 MB  pool[sh=1 pr=4 0.56MB] inflight=0 prev_ce=1
iter   40000  RSS  46.32 MB  pool[sh=1 pr=4 0.56MB] inflight=0 prev_ce=1
iter   85000  RSS  66.39 MB  pool[sh=1 pr=4 0.56MB] inflight=0 prev_ce=1   ← still rising
```

The growth is **autoreleased Objective-C temporaries**. Command buffer / encoder creation, `commit`, and the waits all return autoreleased objects. In a tight Rust compute loop there is no run-loop turn to drain the thread's autorelease pool, so those references live for the whole program lifetime.

This matches @LaurentMazare's earlier diagnosis in the issue ("the issue is related to the autorelease pool not releasing memory fast enough").

## The fix

Wrap the Metal-object creation and synchronization paths in `Commands` — `command_encoder`, `blit_command_encoder`, `flush_and_wait`, `flush_and_wait_current`, `flush` — in `objc2::rc::autoreleasepool`. Objects candle keeps are held via `Retained` and survive the drain; only the pool's redundant references are freed.

## Verification

Added `candle-core/examples/matmul_leak.rs` (the issue's repro). With the fix, RSS is flat:

```
end RSS: 27.80 MB  (grew 13.73 MB over 200000 iterations)   # oscillates 27–31 MB, no upward trend
```

vs. unpatched, which passes 66 MB by 85k iterations and keeps climbing.

**Throughput** — no regression (fixed vs. `main`, same workload):

| workload | main | this PR |
|---|---|---|
| matmul + reduce + readback | 128.6 µs/iter | 125.3 µs/iter |
| encoder-bound (matmul only) | 3.98 µs/matmul | 3.81 µs/matmul |

Metal suites pass: `matmul_tests` (12), `metal_concurrent_tests` (2, exercises the command-buffer flush paths touched here), `quantized_tests` (67), `tensor_tests` (83). `cargo fmt --all --check`, CPU `cargo clippy --workspace --tests --examples --benches -D warnings`, and `cargo check --workspace --features metal` all pass.

Note: I used AI to identify and fix the issue.

### DIFF
--- a/candle-core/examples/matmul_leak.rs
+++ b/candle-core/examples/matmul_leak.rs
@@ -48,9 +48,10 @@ fn main() -> Result<()> {
     println!("start RSS: {:.2} MB", start_rss as f64 / 1e6);
 
     for i in 0..iterations {
-        let c = a.matmul(&b)?;
-        // Force the computation to actually run and read the result back to the CPU.
-        let _ = c.sum_all()?.to_scalar::<f32>()?;
+        // Exactly as in the issue reproducer: enqueue matmuls without ever reading a result back
+        // to the CPU and without wrapping the loop body in an autorelease pool. The backend must
+        // keep memory flat on its own; see huggingface/candle#2271.
+        let _ = a.matmul(&b)?;
 
         if i % 5_000 == 0 {
             let rss = rss_bytes();
--- a/candle-core/examples/matmul_leak.rs
+++ b/candle-core/examples/matmul_leak.rs
@@ -1,0 +1,72 @@
+// Regression check for https://github.com/huggingface/candle/issues/2271
+//
+// Repeatedly multiplying a 784x100 matrix by a 100x10 matrix used to leak memory
+// on the Metal backend: autoreleased command buffers / encoders accumulated in the
+// thread's autorelease pool, so the resident set grew without bound (>5 GB reported,
+// >100 GB for BERT-sized models). The CPU backend never leaked.
+//
+// With the fix the resident set stays flat on both backends.
+//
+// Run with:
+//   cargo run --release --example matmul_leak --features metal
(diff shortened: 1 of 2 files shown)


ivarflakstad (changes requested):
Nice!
I like the solution 👍 

As usual AI is way too verbose in comments - those should be trimmed down to a minimum.
Also after proving the matmul leak once we no longer need the example. Can also be removed :)

PR #3760, opened 2026-07-20
Fix Metal pipeline cache source identity

tomsanbear:
## Summary

The Metal pipeline cache is keyed by function name and constants, but pipeline loading also takes a `Source`. Different Metal libraries that expose the same function name and constants can therefore collide, returning a pipeline compiled from the wrong source.

## Fix

Include `Source` in the pipeline cache key and use `(source, name, constants)` consistently for lookup and insertion. Existing compilation and locking behavior remains unchanged.

## Test plan

Added a regression that primes the cache with `cos_f32` from the unary source, then requests the same name from the binary source, confirming the second source is consulted instead of returning the cached unary pipeline.

Ran:

- `cargo test --manifest-path candle-metal-kernels/Cargo.toml`
- `cargo test -p candle-core --features metal`
- `cargo test --workspace`
- `cargo check --workspace --features metal`

### DIFF
--- a/candle-metal-kernels/src/kernel.rs
+++ b/candle-metal-kernels/src/kernel.rs
@@ -59,7 +59,7 @@ impl From<String> for KernelName {
 }
 
 type Libraries = HashMap<Source, Library>;
-type Pipelines = HashMap<(KernelName, Option<ConstantValues>), ComputePipeline>;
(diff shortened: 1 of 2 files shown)


ivarflakstad (approved):
👍

PR #3761, opened 2026-07-20
Use shared locks for Metal cache hits

tomsanbear:
## Summary

Metal pipeline and library cache hits currently take an exclusive write lock, serializing concurrent readers even though cached entries are not modified.

Depends on #3760.

## Fix

Use a shared read lock for cache hits. On a miss, acquire the write lock and recheck the cache before compiling and inserting, preserving single-entry initialization when callers race.

Cold compilation and failure behavior remain unchanged.

## Test plan

Added a concurrent regression covering same-key cold loads, pipeline reuse, and warm pipeline/library cache hits.

Ran:

- `cargo test --manifest-path candle-metal-kernels/Cargo.toml`
- `cargo test -p candle-core --features metal`
- `cargo test --workspace`
- `cargo check --workspace --features metal`

### DIFF
--- a/candle-metal-kernels/src/kernel.rs
+++ b/candle-metal-kernels/src/kernel.rs
@@ -59,7 +59,7 @@ impl From<String> for KernelName {
 }
 
 type Libraries = HashMap<Source, Library>;
-type Pipelines = HashMap<(KernelName, Option<ConstantValues>), ComputePipeline>;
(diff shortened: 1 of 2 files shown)


ivarflakstad:
Needs a little rebase (or however you want to resolve the conflicts) ☺️

ivarflakstad (approved):
👌

Issue #3759, opened 2026-07-20
Add heterogeneous LoRA batching kernels for S-LoRA/Punica-style multi-adapter forward


astorise:
## Context

Tachyon Mesh issue #316 implements phase 1 of Multi-LoRA serving on the host side: requests for the same base model are now grouped into one scheduler step, then split into adapter-specific sub-batches before calling Candle.

This preserves correctness and avoids splitting scheduler steps by `adapter_id`, but it still executes each adapter group sequentially. To reach the target throughput for 5-10 concurrent agents using distinct LoRA adapters on the same resident base model, Candle needs a true heterogeneous LoRA forward path.

## Request

Please add S-LoRA/Punica-style heterogeneous LoRA batching support for Llama-family safetensors models.

Needed capability:

- One base model forward receives a batch containing multiple `adapter_id`s.
- Each sequence can select a different active LoRA adapter.
- LoRA A/B projections are applied through batched/segmented SGMV-style kernels rather than host-side per-adapter forwards.
- The API preserves the current no-adapter path unchanged.
- The implementation works with the existing Tachyon fork APIs around `LlamaLoadConfig`, `LoraConfig`, `set_active_adapter`, and paged-attention/CUDA graph seams where possible.

## Proposed API shape

A concrete shape can differ, but Tachyon needs something equivalent to:

```rust
model.forward_with_adapters(
    input_ids,
    index_pos,
    cache,
    adapter_assignments, // per sequence: None or adapter handle/id
)
```

or a cache/model method that binds a per-row adapter assignment tensor/table before `forward`.

## Acceptance criteria

- A single forward can process a batch with at least two different LoRA adapters and one no-adapter row.
- Outputs match the current sequential per-adapter execution within expected numeric tolerance.
- No-adapter rows are not affected by active adapter state.
- Adapter isolation is covered by a regression test.
- CUDA implementation includes a benchmark or test demonstrating improved aggregate throughput versus sequential sub-batches.
- The API remains additive so Tachyon can keep using the existing sequential fallback until this path is enabled.

Issue #3696, opened 2026-06-30
Add first-class LoRA adapter injection to candle-transformers model loaders


astorise:
## Motivation

`candle-nn` now has a useful `LoraLinear` primitive for PEFT-style adapters, but `candle-transformers` model implementations such as Llama cannot currently consume it without downstream projects reimplementing the model forward pass.

In `candle_transformers::models::llama::Llama`, the transformer blocks and projection fields are private, so a downstream runtime cannot replace `q_proj`, `k_proj`, `v_proj`, `o_proj`, `gate_proj`, `up_proj`, or `down_proj` with `LoraLinear` wrappers after loading the base model.

This forces consumers to duplicate Candle's Llama forward implementation just to apply LoRA adapters, which is fragile and makes it harder to stay aligned with upstream fixes.

## Use case

Tachyon-Mesh serves one shared foundation model for multiple tenants. Each request can include an `adapter_id`, resolved to a PEFT `.safetensors` LoRA adapter stored by the model broker.

Example flow:

1. Load one base Llama safetensors checkpoint once.
2. Tenant A sends a request with `adapter_id = "legal-v1"`.
3. Tenant B sends a request with `adapter_id = "support-v2"`.
4. The runtime hot-swaps the active LoRA adapter per request without reloading the base model.
5. Requests without `adapter_id` run against the unmodified base model.

Today, implementing this requires a local Llama forward copy because `candle-transformers::Llama` does not expose an adapter injection point.

## Proposed API direction

Please add a first-class way for `candle-transformers` loaders to build supported models with optional LoRA adapters.

One possible shape:

```rust
use candle_nn::lora::LoraConfig;
use candle_transformers::models::llama::{Llama, LlamaLoadConfig};

let base_vb = VarBuilder::from_mmaped_safetensors(&base_paths, DType::F32, &device)?;
let adapter_vb = VarBuilder::from_mmaped_safetensors(&adapter_paths, DType::F32, &device)?;

let load_config = LlamaLoadConfig::default().with_lora_adapter(
    "legal-v1",
    adapter_vb,
    LoraConfig {
        rank: 16,
        alpha: 32.0,
        target_modules: vec![
            "q_proj",
            "k_proj",
            "v_proj",
            "o_proj",
            "gate_proj",
            "up_proj",
            "down_proj",
        ],
    },
);

let mut model = Llama::load_with_config(base_vb, &config, load_config)?;
model.set_active_adapter(Some("legal-v1"))?;

[...]

Issue #3652, opened 2026-06-25
[feature] First-class LoRA adapters in candle-nn (load + merge + multi-adapter)


astorise:
### Motivation
LoRA is ubiquitous for serving fine-tuned variants, but candle has no in-tree
support, so every consumer re-implements `W·x + (B·A)·x·(α/r)` and PEFT loading.
A reusable `candle-nn` primitive would standardize this and enable multi-adapter
serving / hot-swap. (Out-of-tree prior art: `EricLBuehler/candle-lora`.)

### Gap in candle
- No `lora` module in `candle-nn`.

### Proposed implementation
- `candle-nn::LoraLinear` wrapping a base `Linear` with `A [r, in]`, `B [out, r]`
  and `alpha/r` scaling.
- Load PEFT `.safetensors` adapters via `VarBuilder`.
- `merge()` / `unmerge()` (fold weights for inference) and active-adapter
  selection for multi-LoRA.

### Use-case / example
```rust
// Apply a PEFT adapter on top of a base projection at load time.
let base = linear(in_dim, out_dim, vb.pp("q_proj"))?;
let lora = LoraLinear::from_peft(base, vb_adapter.pp("q_proj"), /*rank=*/16, /*alpha=*/32.0)?;
let y = lora.forward(&x)?; // base + LoRA delta
```

PR #3762, opened 2026-07-21
Add heterogeneous multi-LoRA batching (S-LoRA / Punica style) to candle-nn and the Llama model

astorise:
Part of #3759,.

This PR adds first-class LoRA support to candle-nn plus S-LoRA / [Punica](https://arxiv.org/abs/2310.18547)-style heterogeneous batching: a single base-model forward can process a batch in which every sequence selects its own LoRA adapter (or none), instead of grouping requests by adapter and running one forward per group. It also builds on the discussions in #3652 (LoRA layer support) and #3696 (adapter injection at model load time).

## What's included

**`candle-nn::lora::LoraLinear`** — a `Linear` wrapper with named low-rank adapters:
- PEFT-style checkpoint loading (`lora_A.weight` / `lora_B.weight` via `VarBuilder`), several adapters can coexist;
- runtime adapter switching (`set_active_adapter`), weight `merge` / `unmerge` for inference;
- **`forward_with_adapters(x, assignments)`** — the heterogeneous batched path, where `assignments` holds one `Option<&str>` per batch row.

**`candle-transformers` Llama integration**:
- `Llama::load_with_config` + `LlamaLoadConfig::with_lora_adapter(name, vb, LoraConfig { rank, alpha, target_modules })` injects adapters into the matching attention/MLP projections (standard PEFT `base_model.model.` layout by default, overridable prefix). `Llama::load` is unchanged and delegates with no adapters.
- `Llama::set_active_adapter(name)` hot-swaps the active adapter per request without reloading or duplicating base weights.
- **`Llama::forward_with_adapters(input_ids, index_pos, cache, adapter_assignments)`** — the API proposed in #3759: one entry per sequence, `Some(name)` or `None` for base-only rows. The assignments are threaded through the blocks down to each projection; the plain `forward` / `forward_input_embed` paths are untouched.

## How the batched path works

Instead of a host-side loop over adapter groups, the selected `A`/`B` matrices are gathered into per-row stacks and applied with two batched matmuls, SGMV-style:

- per-row `index_select` over stacked `A^T` / `B^T` (adapters of different ranks are zero-padded to the largest rank; the `alpha/r` scaling is folded into `B`);
- rows without an adapter hit an all-zero slot, so the base-only path is preserved exactly;
- `y = W·x + gather(Bᵀ)·gather(Aᵀ)·x` via `bmm`, works on CPU / CUDA / Metal with existing ops — no new kernels to maintain.
- The delta computation is factored behind a small internal seam (`batched_lora_delta(x, a_stack, b_stack, row_slots)`, with the stack layout documented there). A fused SGMV/BGMV kernel — e.g. from a dedicated kernel crate, following the `candle-flash-attn` pattern — can later replace this reference implementation to read the adapter stacks in place instead of materializing the per-row gather, without any change to the public API.

## Correctness

- Numerical equivalence with sequential per-row execution on mixed batches (multiple adapters of different ranks + base-only rows), adapter isolation, 2d/3d inputs, and error paths (assignment count mismatch, unknown adapter, merged adapter) — tested at both the `LoraLinear` and the `Llama` level.
- Model-level test also checks the batched path ignores (and does not leak) the currently active adapter.
- candle-nn tests are device-parameterized through `test_device!`; the full suite passes on CPU and on CUDA (validated on a self-hosted GPU runner: 12/12 lora tests green with `--features cuda`).

## Performance

Criterion benchmark (`candle-nn/benches`, batch=16, hidden=1024, rank=16, 4 adapters + base-only rows) comparing the batched path against grouped sequential execution:

| scenario | sequential | batched | speedup |
|---|---|---|---|
| decode-style (seq=1), CPU | 4.83 ms | 2.05 ms | ~2.4x |
| prefill (seq=128), CPU | 45.8 ms | 46.3 ms | ~1x (base matmul dominates) |

The win is precisely on decode-style multi-tenant batches, which is the S-LoRA / Punica serving scenario.

## Backward compatibility

Everything is additive: `Llama::load`, `Llama::forward`, `forward_input_embed` and the no-adapter code path are unchanged.

https://claude.ai/code/session_013jmpqNexwuhn9tQC3ayrfN

### DIFF
--- a/candle-nn/benches/bench_main.rs
+++ b/candle-nn/benches/bench_main.rs
@@ -4,5 +4,6 @@ use criterion::criterion_main;
 criterion_main!(
     benchmarks::norm::benches,
     benchmarks::softmax::benches,
-    benchmarks::conv::benches
+    benchmarks::conv::benches,
+    benchmarks::lora::benches
 );
--- a/candle-nn/benches/benchmarks/lora.rs
+++ b/candle-nn/benches/benchmarks/lora.rs
@@ -1,0 +1,122 @@
+use crate::benchmarks::{BenchDevice, BenchDeviceHandler};
+use candle::{DType, Device, Module, Tensor};
+use candle_nn::{linear_no_bias, LoraLinear, VarBuilder, VarMap};
+use criterion::{criterion_group, Criterion};
+use std::hint::black_box;
+use std::time::Instant;
+
+const HIDDEN: usize = 1024;
+const RANK: usize = 16;
+const N_ADAPTERS: usize = 4;
+const BATCH: usize = 16;
+
+fn make_lora(device: &Device) -> (LoraLinear, Vec<String>) {
+    let varmap = VarMap::new();
+    let vb = VarBuilder::from_varmap(&varmap, DType::F32, device);
+    let base = linear_no_bias(HIDDEN, HIDDEN, vb.pp("base")).unwrap();
+    let mut lora = LoraLinear::new(base);
+    let names: Vec<String> = (0..N_ADAPTERS).map(|i| format!("adapter_{i}")).collect();
+    for name in names.iter() {
+        lora.load_adapter(name, vb.pp(name), RANK, 16.0).unwrap();
+    }
+    (lora, names)
+}
+
+/// Round-robin over the adapters with every fourth row left on the base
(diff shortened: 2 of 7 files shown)


astorise:

PR #3763, opened 2026-07-21
Add candle-lora-kernels: fused BGMV kernels for heterogeneous multi-LoRA decode batching

### LINKED ISSUE #3759: Add heterogeneous LoRA batching kernels for S-LoRA/Punica-style multi-adapter forward

astorise:
Fixes #3759. Builds on #3762.

Follow-up to #3762, which added the heterogeneous multi-LoRA batched path
(`LoraLinear::forward_with_adapters` / `Llama::forward_with_adapters`) as a
pure-tensor `gather + batched-matmul` computation behind a small seam
(`batched_lora_delta`). This PR plugs a fused CUDA kernel into that seam for the
decode step, delivering the S-LoRA / [Punica](https://arxiv.org/abs/2310.18547)
BGMV fast path that #3759 asked for, with no change to the public API.

## What's included

**New crate `candle-lora-kernels`** (CUDA-only, excluded from the workspace like
`candle-flash-attn`):
- Two gather-GEMV passes computing the batched LoRA delta
  `delta[i] = x[i] · A_stack[slot(i)] · B_stack[slot(i)]`:
  - **shrink** (`in -> r`): `tmp[i] = x[i] · A_stack[slot(i)]`
  - **expand** (`r -> out`): `delta[i] = tmp[i] · B_stack[slot(i)]`
  
  Each pass reads the stacked adapter matrices in place through a per-row `u32`
  slot index, instead of materializing the per-row gather of `A`/`B` that the
  reference path builds before its matmuls. Slot 0 is the all-zero (no-adapter)
  slot, so base-only rows produce a zero delta.
- Layout mirrors `candle-flash-attn`: a `cudaforge` `build.rs` compiling
  `src/lora_sgmv.cu` into a static lib, an `extern "C"` ffi, and two `CustomOp3`
  wrappers (`LoraShrink` / `LoraExpand`) exposed via `bgmv_delta()`. f32/f16/bf16
  supported, accumulating in f32.

**candle-nn wiring**: a new `lora-cuda` feature (implies `cuda`, pulls in
`candle-lora-kernels`) dispatches `batched_lora_delta` through
`candle_lora_kernels::bgmv_delta` for a decode-step batch (`seq == 1`) on a CUDA
device with an f32/f16/bf16 input. Everything else — `seq > 1`, non-CUDA
devices, unsupported dtypes, and every build without the feature — keeps using
the pure-tensor reference path from #3762 unchanged.

## Why a fused kernel

The reference `gather + bmm` path already batches all adapters into one forward,
but it materializes per-row copies of the `A`/`B` matrices (`index_select` into
`[batch, in, r]` / `[batch, r, out]`) before the matmuls — pure memory traffic.
The BGMV kernel reads the stacks in place via the slot indices, removing that
traffic, which is exactly where the decode-step (GEMV-bound) win comes from.

## Correctness

`bgmv_delta` is numerically equivalent to `batched_lora_delta`, validated on a
CUDA GPU runner: `lora_batched_forward_matches_sequential_gpu` (kernel output ==
sequential per-row execution, mixed batch of two adapters of different ranks + a
base-only row) passes with `--features cuda,lora-cuda`, alongside the full lora
suite and the Llama integration tests. The crate compiles via `nvcc`
(`cudaforge`) in CI.

## Performance

GPU criterion bench (`candle-nn/benches`, batch=16, hidden=1024, rank=16,
4 adapters + base-only rows), same GPU, reference vs kernel:

| scenario | sequential (host loop) | batched reference (gather+bmm) | batched BGMV kernel |
|---|---|---|---|
| decode (seq=1) | 327 µs | 224 µs | **140 µs** |
| prefill (seq=128) | 1477 µs | 890 µs | 888 µs |

- Decode: the kernel is **−37% vs the gather+bmm reference** (criterion
  `change`, CI −34% to −41%, p<0.05), i.e. ~1.6x, and ~2.3x vs the sequential
  host loop — the S-LoRA / Punica multi-tenant decode scenario.
- Prefill (seq=128): unchanged, since the kernel only engages at `seq == 1` and
  falls back to the reference otherwise. This also confirms the seam dispatch.

## Backward compatibility

Fully additive and opt-in: without `lora-cuda` nothing changes, and the
reference path stays the default on every backend. The seam
(`batched_lora_delta`) and the public API are untouched.

https://claude.ai/code/session_013jmpqNexwuhn9tQC3ayrfN

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ exclude = [
     "candle-flash-attn",
     "candle-flash-attn-v3",
     "candle-kernels",
+    "candle-lora-kernels",
     "candle-metal-kernels",
     "candle-onnx",
 ]
@@ -40,6 +41,7 @@ candle-datasets = { path = "./candle-datasets", version = "0.11.0" }
 candle-flash-attn = { path = "./candle-flash-attn", version = "0.11.0" }
 candle-flash-attn-v3 = { path = "./candle-flash-attn-v3", version = "0.11.0" }
 candle-kernels = { path = "./candle-kernels", version = "0.11.0" }
+candle-lora-kernels = { path = "./candle-lora-kernels", version = "0.11.0" }
 candle-metal-kernels = { path = "./candle-metal-kernels", version = "0.11.0" }
 candle-nn = { path = "./candle-nn", version = "0.11.0" }
 candle-onnx = { path = "./candle-onnx", version = "0.11.0" }
--- a/candle-lora-kernels/Cargo.toml
+++ b/candle-lora-kernels/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "candle-lora-kernels"
+version = "0.11.0"
+edition = "2021"
+
+description = "Heterogeneous multi-LoRA batching CUDA kernels (S-LoRA / Punica style BGMV) for the candle ML framework."
+repository = "https://github.com/huggingface/candle"
+keywords = ["blas", "tensor", "machine-learning"]
+categories = ["science"]
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+
+[dependencies]
+candle = { path = "../candle-core", features = ["cuda"], package = "candle-core", version = "0.11.0" }
+half = { version = "2.3.1", features = ["num-traits"] }
+
+[build-dependencies]
+cudaforge = "0.1.2"
--- a/candle-lora-kernels/README.md
+++ b/candle-lora-kernels/README.md
@@ -1,0 +1,16 @@
+# candle-lora-kernels
+
+Heterogeneous multi-LoRA batching CUDA kernels (S-LoRA / Punica style BGMV) for
+the candle ML framework.
+
+This crate is the fused-kernel fast path for
+`candle_nn::lora::LoraLinear::forward_with_adapters`. A decode-step batch in
+which every row selects its own LoRA adapter is applied as two gather-GEMV
+passes — "shrink" (`in -> r`) then "expand" (`r -> out`) — that read the stacked
+adapter matrices in place through a per-row slot index, instead of materializing
+the per-row gather that the pure-tensor reference path builds before its batched
+matmuls.
+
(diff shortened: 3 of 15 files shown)


astorise:

PR #3765, opened 2026-07-22
Fix two residual f16/bf16 saturation gaps in PR #3717 (non-trailing-axis reductions, mean overflow)

astorise:
## Summary

Builds on #3717 (thanks @vanshverma-code for the original fix and for confirming these two gaps in review) — this branch is #3717's exact commits (`9a84be5`, `3c55f5f`, `63d40ad`) plus one additional commit fixing two residual precision issues found while integrating #3717 into a fork.

#3717 fixed f16/bf16 saturation in `softmax_last_dim`, `sum`/`mean`, and `avg_pool2d`, but only for the contiguous trailing-axis fast path. Two related cases still produce wrong results:

1. **Non-trailing-axis / non-contiguous reductions still saturate.** `ReduceSum::fold_impl`'s wide `vec_reduce_sum` accumulator only runs when reducing over the trailing axis; anything else (e.g. `Tensor::ones((4096, 2), DType::F16).sum(0)`, or a strided view) falls back to a plain in-dtype `+=` loop that still saturates the same way the pre-#3717 code did.

2. **`mean` can overflow to `inf` even when the true mean is representable.** `sum_impl` narrows its wide accumulator back to `T` before returning, and `mean`/`mean_keepdim` apply the `1/N` scale *after* that narrowing. So `Tensor::ones((65520,), DType::F16).mean_all()` overflows to `+inf` during the sum (65520 > f16's ~65504 max), even though the true mean (1.0) is trivially representable.

## What this adds

- A `VecOps::Accum` associated type (`f32` for f16/bf16, `Self` for everything else) with `to_accum`/`accum_add`/`from_accum`, used by `ReduceSum::fold_impl`'s two non-fast-path branches (contiguous-non-trailing, and non-contiguous) instead of a plain in-dtype accumulator.
- `mean`/`mean_keepdim`/`mean_all` now upcast f16/bf16 to f32, sum and scale there, and downcast once at the end — instead of dividing an already-narrowed (possibly infinite) sum.
- Regression tests for both: a non-last-dim column sum (`f16_sum_non_last_dim_no_saturation`, `bf16_sum_non_last_dim_no_saturation`) and a mean whose sum overflows f16 but whose true value doesn't (`f16_mean_all_overflowing_sum`).

## Test plan
- [x] All 5 tests in `f16_sum_check.rs` pass (2 pre-existing from #3717, 3 new)
- [x] Full `candle-core`/`candle-nn` test suites green on top of current `main`
- [x] `cargo clippy --all-targets -- -D warnings` / `cargo fmt --check` clean

### DIFF
--- a/candle-core/src/cpu/kernels.rs
+++ b/candle-core/src/cpu/kernels.rs
@@ -134,6 +134,19 @@ impl VecOps for half::f16 {
         super::vec_dot_f16(lhs, rhs, &mut res_f32, len);
         *res = half::f16::from_f32(res_f32);
     }
+
+    // Accumulate in f32: a sequential sum in f16 saturates on long axes (e.g. an
+    // f16 sum of 4096 ones stalls at 2048 once the running ULP exceeds 1),
+    // giving wrong `sum`/`mean` results that also disagree with the Metal
+    // backend. This mirrors the f32 accumulation already used by `vec_dot`.
+    #[inline(always)]
+    unsafe fn vec_reduce_sum(xs: *const Self, res: *mut Self, len: usize) {
+        let mut sum = 0f32;
+        for i in 0..len {
+            sum += (*xs.add(i)).to_f32();
+        }
+        *res = half::f16::from_f32(sum);
+    }
 }
 
 impl VecOps for f64 {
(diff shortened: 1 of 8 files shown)


astorise:

PR #3767, opened 2026-07-22
Add batch-native multi-prompt decode loop for heterogeneous LoRA serving

### LINKED ISSUE #3766: Add batch-native multi-prompt decode loop for heterogeneous LoRA serving
## Context

This request builds on #3759 and the open PR chain:

- #3762 adds heterogeneous multi-LoRA batching APIs such as `LoraLinear::forward_with_adapters` and `Llama::forward_with_adapters`.
- #3763 builds on #3762 and adds the CUDA BGMV fast path for decode-step heterogeneous LoRA batching.

Those PRs provide the model/projection side needed for S-LoRA / Punica-style execution, but serving integrations still need a full batch-native autoregressive decode loop that can drive multiple active prompt rows through prefill/decode without falling back to host-side per-adapter sub-batches.

## Request

Please expose or document a batch-native multi-prompt generation/decode path for Llama-family models where each active batch row can select a different LoRA adapter, or no adapter.

The important target is the full generation loop, not only one isolated model forward.

## Required behavior

- Batch prefill accepts multiple prompt rows for the same base model.
- Batch decode accepts one token per active sequence per step.
- Each row can select `Some(adapter_id)` or `None`.
- Logits are returned as `[batch, vocab]`, one row per active sequence.
- KV cache updates remain isolated per sequence.
- Rows with `None` remain strict base-model execution.
- Output matches sequential per-row/per-adapter generation within expected numeric tolerance.
- The path composes with the APIs from #3762 and the CUDA fast path from #3763.

## Acceptance criteria

- A test runs at least three active rows in one decode step: adapter A, adapter B, and no adapter.
- Batched decode output matches sequential reference output.
- The no-adapter row is equivalent to the plain base path.
- The decode loop does not split the batch by adapter on the host side.
- Existing `forward`, `forward_with_adapters`, and sequential generation behavior remain backward-compatible.

astorise:
## Summary

- Adds `Llama::generate_with_adapters`, a batch-native autoregressive decode loop where every row in the batch can select its own LoRA adapter (`Some(adapter_id)`) or run base-model-only (`None`). Prefill and every decode step go through a single `forward_with_adapters` call for the whole batch — the batch is never split or re-grouped by adapter on the host side.
- Requires a `Cache` with `use_kv_cache` set, since decode steps rely on each row's own past keys/values, isolated by the batch dimension, rather than recomputing them.
- Adds `lora_batched_forward_multistep_decode_gpu_matches_cpu` to `candle-nn/tests/lora.rs`, gated behind the `lora-cuda` feature: it drives several `seq == 1` decode-step batches (two adapters + a base-only row) back to back through `forward_with_adapters` on both CPU and CUDA, checking the `candle_lora_kernels` BGMV kernel path stays correct across repeated kernel launches, not just on a single isolated call.

This builds on #3762 (heterogeneous multi-LoRA batching APIs) and #3763 (CUDA BGMV kernels for the decode-step fast path), closing the gap #3759 identified: those PRs cover a single batched forward, but serving needs a full generation loop that drives prefill/decode across steps without host-side per-adapter sub-batches.

Closes #3766.

## Required behavior (from the issue)

- [x] Batch prefill accepts multiple prompt rows for the same base model.
- [x] Batch decode accepts one token per active sequence per step.
- [x] Each row selects `Some(adapter_id)` or `None`.
- [x] Logits are `[batch, vocab]`, one row per active sequence.
- [x] KV cache updates stay isolated per sequence.
- [x] `None` rows remain strict base-model execution.
- [x] Output matches sequential per-row/per-adapter generation within numeric tolerance.
- [x] Composes with the APIs from #3762 and the CUDA fast path from #3763 (decode steps are `seq == 1`, so they automatically route through the BGMV kernel when `lora-cuda` is enabled on a CUDA device).

## Acceptance criteria (from the issue)

- [x] A test runs at least three active rows in one decode step: adapter A, adapter B, and no adapter (`generate_with_adapters_matches_sequential_reference`, CPU).
- [x] Batched decode output matches sequential reference output.
- [x] The no-adapter row is equivalent to the plain base path (checked against `Llama::forward` directly, not only `forward_with_adapters(&[None])`).
- [x] The decode loop does not split the batch by adapter on the host side (`generate_with_adapters` issues exactly one `forward_with_adapters` call per step for the whole batch).
- [x] Existing `forward`, `forward_with_adapters`, and sequential generation behavior remain backward-compatible (no changes to existing APIs; all prior tests still pass).

## Testing

- `cargo test -p candle-transformers --lib models::llama` — 9 tests, including 4 new ones covering the multi-row decode loop (sequential-reference parity, `use_kv_cache` requirement, mismatched-count rejection).
- `cargo test -p candle-nn --test lora` — 8 tests (CPU-only build).
- `cargo test -p candle-nn --features cuda,lora-cuda --test lora` — 15 tests, run on a real GPU via CI (`arc-gpu-candle` runner), including the new multi-step BGMV kernel parity test:
  ```
  test lora_batched_forward_multistep_decode_gpu_matches_cpu ... ok
  test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
  ```
- `cargo fmt --check` and `cargo clippy` clean on both crates.

## Test plan

- [x] CI passes (CPU + GPU cuda jobs)
- [x] Review of `Llama::generate_with_adapters`'s KV-cache/index_pos bookkeeping across decode steps

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ exclude = [
     "candle-flash-attn",
     "candle-flash-attn-v3",
     "candle-kernels",
+    "candle-lora-kernels",
     "candle-metal-kernels",
     "candle-onnx",
 ]
@@ -40,6 +41,7 @@ candle-datasets = { path = "./candle-datasets", version = "0.11.0" }
 candle-flash-attn = { path = "./candle-flash-attn", version = "0.11.0" }
 candle-flash-attn-v3 = { path = "./candle-flash-attn-v3", version = "0.11.0" }
 candle-kernels = { path = "./candle-kernels", version = "0.11.0" }
+candle-lora-kernels = { path = "./candle-lora-kernels", version = "0.11.0" }
 candle-metal-kernels = { path = "./candle-metal-kernels", version = "0.11.0" }
 candle-nn = { path = "./candle-nn", version = "0.11.0" }
 candle-onnx = { path = "./candle-onnx", version = "0.11.0" }
--- a/candle-lora-kernels/Cargo.toml
+++ b/candle-lora-kernels/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "candle-lora-kernels"
+version = "0.11.0"
+edition = "2021"
+
+description = "Heterogeneous multi-LoRA batching CUDA kernels (S-LoRA / Punica style BGMV) for the candle ML framework."
+repository = "https://github.com/huggingface/candle"
+keywords = ["blas", "tensor", "machine-learning"]
+categories = ["science"]
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+
+[dependencies]
+candle = { path = "../candle-core", features = ["cuda"], package = "candle-core", version = "0.11.0" }
+half = { version = "2.3.1", features = ["num-traits"] }
+
+[build-dependencies]
+cudaforge = "0.1.2"
--- a/candle-lora-kernels/README.md
+++ b/candle-lora-kernels/README.md
@@ -1,0 +1,16 @@
+# candle-lora-kernels
+
+Heterogeneous multi-LoRA batching CUDA kernels (S-LoRA / Punica style BGMV) for
+the candle ML framework.
+
+This crate is the fused-kernel fast path for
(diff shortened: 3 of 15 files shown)


astorise:

PR #3768, opened 2026-07-22
feat(transformers): add PagedKvCache with per-layer block allocators

Vrinda12-tech:
Implements a block-structured KV cache for CPU inference, replacing `Tensor::cat` fragmentation with static pre-allocated pools and O(1) in-place writes.

## Architecture

- Per-layer `BlockAllocator` (free-list + reference counting)
- `BlockTable` maps logical sequence positions to physical pool blocks
- Flattened pool per layer: `[num_blocks * block_size, num_kv_heads, head_dim]`
- In-place writes via `Tensor::slice_set` (zero allocation per token)
- Cached GPU indices for `gather_kv`, rebuilt only on block boundary crossing
- Fast-path: single-block sequences use `narrow()` instead of `index_select`

## Benchmarks (CPU, Llama-7B dims, release build)

| Scenario | ConcatKvCache | PagedKvCache | Result |
|----------|--------------|--------------|--------|
| Write 100 tokens, 1 layer, 512 seq | 2.36s | 17.8ms | **133× faster** |
| Write 10 tokens, 32 layers, 512 seq | 4.21s | 177ms | **23× faster** |
| Write+gather 1 token, 32 layers, 512 seq | 1.29s | 210ms | **6× faster** |
| Write+gather 1 token, 32 layers, 32 seq | baseline | +72% | Fast-path overhead |
| COW fork ×100, 4096 ctx, 32 layers | ~8ms each | 433µs each | **18× faster** |

## Limitations (Phase 1)

`gather_kv` uses `index_select` to materialize contiguous tensors for attention. At 4096 tokens across 32 layers, this is slow on CPU (~69s). This makes Phase 1 unsuitable for long-context autoregressive decoding (&gt;2K tokens) until Phase 2 adds a backend `gather_blocks` op.

**Recommended for:** beam search, speculative decoding, prefill, short-context models.

## Files changed

- `candle-transformers/src/kv_cache/block.rs` (new)
- `candle-transformers/src/kv_cache/paged.rs` (new)
- `candle-transformers/src/kv_cache/mod.rs` (new)
- `candle-transformers/tests/paged_kv_cache.rs` (tests + benchmarks)
- `candle-transformers/src/lib.rs` (module export)

## Follow-up work

- Llama model integration
- Backend `gather_blocks` op (Phase 2)
- CUDA/Metal kernel fusion (Phase 3)


ivarflakstad (commented):
Thanks for the contribution! :)

This is a bit sloppy, but conceptually not a bad direction. It does have a couple technical issues though.

If the KV Cache were to use this it would be locked in as `f32`, which isn't really acceptable. You'll have to thread `<T: FloatDType>` through `BlockManager` and `VecSliceGuard` (if generics are unwieldy for some reason you can alternatively store type info via the `DType` enum).

The allocation calculation (`BlockManager::new`) is missing the number of layers it is supposed to represent, meaning you can either only represent a single layer or you have to allocate N different block managers, which is definitely not what you want.

I expect the block manager approach to be considerably faster than the old concat approach, but the benchmark is not an honest representation of actual kv cache interaction (resizing when initial allocation is not sufficient etc). Also there is already an in-place `KvCache` variant which is worth comparing against as well.

Other than that there is just some general coding/documentation style that doesn't align with the existing code in candle, but I'm not going to nitpick on that now ;)

Vrinda12-tech:
> Thanks for the contribution! :)
> 
> This is a bit sloppy, but conceptually not a bad direction. It does have a couple technical issues though.
> 
> If the KV Cache were to use this it would be locked in as `f32`, which isn't really acceptable. You'll have to thread `<T: FloatDType>` through `BlockManager` and `VecSliceGuard` (if generics are unwieldy for some reason you can alternatively store type info via the `DType` enum).
> 
> The allocation calculation (`BlockManager::new`) is missing the number of layers it is supposed to represent, meaning you can either only represent a single layer or you have to allocate N different block managers, which is definitely not what you want.
> 
> I expect the block manager approach to be considerably faster than the old concat approach, but the benchmark is not an honest representation of actual kv cache interaction (resizing when initial allocation is not sufficient etc). Also there is already an in-place `KvCache` variant which is worth comparing against as well.
> 
> Other than that there is just some general coding/documentation style that doesn't align with the existing code in candle, but I'm not going to nitpick on that now ;)

@ivarflakstad  Thanks for the detailed review . I have completely rewritten the implementation to address every point:

1. **Hardcoded f32:** Removed. The pool now stores raw bytes (`Vec&lt;u8&gt;`) and uses `DType` to determine element size and stride. The public API takes `DType` as a parameter.

2. **Missing num_layers:** Fixed. Each layer now has its own `BlockAllocator` and `Vec&lt;BlockTable&gt;`. One `PagedKvCache` serves all layers of a model. Block IDs are per-layer, so layer 0 block 5 and layer 1 block 5 are independent.

3. **Dishonest benchmarks:** Replaced. The old benchmark compared `append_slot` (a `Vec::push`) against `Tensor::cat`. The new benchmarks compare the full write+gather cycle. I included both the fast path (single block, `narrow()`) and the slow path (multi-block, `index_select`) so the numbers reflect real usage.

4. **Coding style:** Removed `unsafe impl Send/Sync`, `VecSliceGuard`, emoji, and `// FIXED:` comments. Errors use `candle::bail!` or proper `Result` propagation. Documentation follows existing Candle conventions.

5. **Scope:** This PR now contains only the `candle-transformers` module + tests. No changes to `candle-nn`. Llama integration is intentionally left for a follow-up to keep the diff reviewable.

The old `candle-nn/src/kv_cache.rs` changes have been reverted to upstream state.

I believe this is ready for review. Happy to make further changes.

PR #3770, opened 2026-07-23
Add post-construction LoRA adapter registration to Llama

### LINKED ISSUE #3769: Allow loading LoRA adapters into an existing Llama without reloading base weights
## Context

This follows #3759 and #3766.

Tachyon Mesh can now use Candle's heterogeneous LoRA forward/decode APIs from the fork tag `tachyon-v0.11.0-12`, but one integration gap remains for production serving: `LlamaLoadConfig` injects LoRA adapters only while constructing a new `Llama` with `Llama::load_with_config`.

For a long-running server, the base Llama model is already resident in GPU memory. When a batch arrives with tenant-specific LoRA adapters that were not known at base-model load time, the only available public shape is to construct a fresh `Llama` from the base safetensors with those adapters attached. On CUDA this creates a temporary second copy of all base weights plus the adapter weights, which can OOM on models sized near available VRAM. Even when it fits, each batch pays full model-load latency.

Tachyon currently has to disable the CUDA batch-native LoRA path and fall back rather than risk this second GPU allocation.

## Request

Please expose an additive API that allows registering, replacing, and optionally removing LoRA adapters on an already-loaded Llama-family model, without reloading or duplicating the base weights.

A concrete API can differ, but something equivalent to this would work:

```rust
impl Llama {
    pub fn load_lora_adapter_prefixed(
        &mut self,
        name: &str,
        vb: VarBuilder,
        config: LoraConfig,
        prefix: &str,
    ) -> Result<()>;

    pub fn remove_lora_adapter(&mut self, name: &str) -> Result<()>;
}
```

## Required Behavior

- The base model weights remain resident and are not reloaded from safetensors.
- New adapter tensors can be loaded onto the model device and attached to all targeted projections.
- `forward_with_adapters` can use adapters registered after `Llama` construction.
- `set_active_adapter` can select adapters registered after `Llama` construction.
- Replacing an adapter with the same name is either rejected or well-defined.
- Removing an adapter clears it from targeted projections and fails if the adapter is currently merged into base weights.
- Rows with `None` remain strict base-model execution.
- Existing `LlamaLoadConfig` construction-time loading remains backward-compatible.

## Acceptance Criteria

- A test loads a base Llama model, then registers two LoRA adapters after construction.
- A batched forward/decode step runs rows for adapter A, adapter B, and no adapter without constructing a second Llama.
- Outputs match the equivalent construction-time adapter loading path within expected numeric tolerance.
- A CUDA test or benchmark demonstrates that dynamic adapter registration does not allocate a second copy of the base weights.
- The API composes with the heterogeneous batching work in #3759 and the batch-native decode loop requested in #3766.

astorise:
## Summary

- Adds `Llama::load_lora_adapter` / `load_lora_adapter_prefixed` to register a new named LoRA adapter into an **already-loaded** model, without reloading or duplicating the base weights: projections targeted by the adapter's config that are still plain are promoted to LoRA-augmented ones in place, reusing the already-resident weight tensor (a cheap `Arc` clone, not a re-read from storage).
- Adds `Llama::remove_lora_adapter` to drop a registered adapter from every projection that carries it, again without touching the base weights.
- Loading is additive and inactive by default: registering an adapter never changes `forward()` output until `Llama::set_active_adapter` selects it, so existing callers are unaffected.
- Adds a small `with_tracing::Linear::inner()` accessor needed to hand a plain projection's existing weight off to `LoraLinear` during promotion.

## Motivation

Depends on / builds on top of:
- #3759 — heterogeneous multi-LoRA batching (`forward_with_adapters`, SGMV/BGMV-style batched kernels)
- #3766 — batch-native multi-prompt decode loop for heterogeneous LoRA serving (`generate_with_adapters`)

Both of the above let a single forward/decode pass serve rows on different adapters, but every adapter still had to be known at `Llama::load_with_config` time via `LlamaLoadConfig`. In a production serving scenario, the base model is already resident in GPU memory and a new tenant-specific adapter needs to be attached on the fly; the only option today is a full model reload, which risks a temporary duplicate of the base weights (and can OOM). This PR closes that gap with an additive API for registering/removing adapters on a live model.

Fixes #3769

## Changes

- `candle-transformers/src/models/llama.rs`
  - `Proj::load_lora_adapter` / `Proj::remove_adapter`
  - `CausalSelfAttention::load_lora_adapter` / `remove_adapter`
  - `Mlp::load_lora_adapter` / `remove_adapter`
  - `Block::load_lora_adapter` / `remove_adapter`
  - `Llama::load_lora_adapter`, `Llama::load_lora_adapter_prefixed`, `Llama::remove_lora_adapter`
- `candle-transformers/src/models/with_tracing.rs`
  - `Linear::inner()` accessor

## Testing

- `load_lora_adapter_matches_construction_time_loading`: registers two adapters ("a", "b") after construction on a model loaded with no adapters, then runs a batched `forward_with_adapters` call (rows on "a", "b", and none) and asserts the output exactly matches a reference model that baked both adapters in at construction time.
- `load_lora_adapter_does_not_affect_base_forward_until_active`: confirms loading an adapter leaves plain `forward()` output unchanged, activating it changes the output, and `remove_lora_adapter` reverts to the original output and makes the adapter name invalid again.
- `cargo test -p candle-transformers --lib models::llama::` — all 21 tests pass.
- `cargo fmt -p candle-transformers -- --check` — clean.
- `cargo clippy -p candle-transformers --lib --tests -- -D warnings` — clean.

## Test plan

- [x] Existing LoRA/Llama unit tests still pass
- [x] New unit tests cover post-hoc load, no-op-until-active, and remove
- [x] `cargo fmt` / `cargo clippy` clean


astorise:

PR #3776, opened 2026-07-25
fix(candle-core):Enable simd128 for target wasm32-wasip2

bharattech:
Problem Description:

Wasm-wasip2 SIMD128 does not support native f16 vector instructions or the CurrentCpuF16 type. 
Currently, vec_add_f16 in candle-core/src/cpu/mod.rs includes target_feature = "simd128" in its #[cfg(...)] guard. When compiling for wasm32-wasip2 with -C target-feature=+simd128, compilation fails with E0433: cannot find type CurrentCpuF16 in this scope.

Solution:

    Add not(target_arch = "wasm32") to the cfg attribute / remove target_feature = "simd128" on 
    vec_add_f16 , vec_add_bf16, vec_scalar_add_f16, vec_scalar_add_bf16 
    allowing WASM to use the scalar fallback for f16 while retaining SIMD acceleration for f32.
    
    Changes Made:

    1. candle-core/src/cpu/mod.rs: Excluded target_arch = "wasm32" from the #[cfg(...)] guards on f16 and bf16 SIMD functions.

    2. candle-core/src/cpu/kernels.rs: Guarded half::f16 and half::bf16 VecOps trait calls to the underlying SIMD helper functions with #[cfg(not(target_arch = "wasm32"))].

Signed-off-by: Bharat Shinde <bharattech@protonmail.com>

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,9 @@
 [build]
 rustflags = ["-C", "target-cpu=native"]
 
+[target.wasm32-wasip2]
+rustflags = ["-C", "target-feature=+simd128"]
+
 [target.wasm32-unknown-unknown]
 rustflags = ["-C", "target-feature=+simd128", "--cfg", 'getrandom_backend="wasm_js"']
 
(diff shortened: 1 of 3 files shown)


bharattech:
This also fix issue "bug_001 - wasm32+simd128 build break (candle-core/src/cpu/mod.rs) - HIGH confidence"
Mentioned https://github.com/toddwbucy/candle/issues/12

PR #3785, opened 2026-07-26
Add zero-copy (mmap-backed) Metal buffer support for quantized weights

alytaphoenix:
Adds a zero-copy path for loading quantized weights into Metal buffers, for the llama.cpp-style mmap-backed loading case: view an mmap'd GGUF file as a Metal buffer directly, instead of allocating a new buffer and copying the file's bytes into it.

## What

- `Device::new_buffer_with_bytes_no_copy` (`candle-metal-kernels`): a thin wrapper around `newBufferWithBytesNoCopy:length:options:deallocator:`, taking a raw pointer + length (e.g. an mmap'd region's base and size) and returning a `Buffer` that aliases that memory rather than copying it. No deallocator is registered — the caller owns the mapping and is responsible for keeping it alive at least as long as the buffer.
- `QMetalStorage::from_buffer` (`candle-core`): wraps an existing (possibly externally-created, possibly shared) buffer as a tensor's quantized storage, at a given `offset`/`length` within that buffer — the shape needed when one no-copy buffer backs an entire mmap'd file and many tensors are views into different offsets of it.
- `src0_offset`/`rhs_offset` parameters added to `call_quantized_matmul_mm_t`/`call_quantized_matmul_mv_t` so a matmul can read directly from such an offset view without a separate per-tensor buffer.

## Why

Without this, loading a GGUF's weights onto Metal means allocating a fresh buffer per tensor (or per file) and copying bytes in — for large quantized MoE models that's real load-time and peak-memory cost that llama.cpp's own mmap-based Metal loader avoids entirely by aliasing the OS page cache. This gives `candle-core` callers the same option: mmap the file, wrap the mapping in one `Buffer` via `new_buffer_with_bytes_no_copy`, then hand out `QMetalStorage::from_buffer` views into it per-tensor at zero additional copies.

This is the buffer-creation half of the mmap story; #3771 (residency-set batch registration for externally-created buffers) is the complementary keep-it-GPU-resident half — a caller using both would create the no-copy buffer here, then register it via `Device::register_buffers` there.

## Testing

`qmetalstorage_from_buffer_view_matches_owned_buffer` (`candle-core/tests/quantized_tests.rs`) validates that a tensor backed by a `from_buffer` view — at both a zero and a nonzero offset into a buffer shared with other tensors — produces identical dequantize and matmul output to the normal owned-buffer path. `cargo build -p candle-core --features metal` passes clean.

### DIFF
--- a/candle-core/src/quantized/metal.rs
+++ b/candle-core/src/quantized/metal.rs
@@ -8,6 +8,16 @@ pub struct QMetalStorage {
     dtype: GgmlDType,
     device: MetalDevice,
     buffer: Arc<Buffer>,
+    /// Byte offset of this tensor's data within `buffer`. Nonzero when
+    /// `buffer` is shared across several tensors (e.g. one no-copy buffer
+    /// wrapping an entire mmap'd file) via `from_buffer`; zero for every
+    /// storage that owns its buffer outright.
+    offset: usize,
+    /// This tensor's own byte length -- may be smaller than `buffer.length()`
+    /// when `buffer` is shared. Every read/blit against `buffer` must use
+    /// this, not `buffer.length()`, or it silently reads past this tensor's
+    /// data into a neighboring one.
+    length: usize,
 }
 
 impl QMetalStorage {
@@ -19,10 +29,32 @@ impl QMetalStorage {
             .with_label("qstorage_zeros")
             .build()?;
         Ok(Self {
+            length: buffer.length(),
             buffer,
             device: device.clone(),
             dtype,
+            offset: 0,
         })
+    }
+
+    /// Wraps an existing, externally-created buffer -- e.g. a no-copy Metal
+    /// buffer over an mmap'd region -- as this tensor's storage, with no
+    /// allocation and no copy. `offset`/`length` locate this tensor's bytes
+    /// within `buffer`, which may be shared with other tensors.
+    pub fn from_buffer(
+        buffer: Arc<Buffer>,
+        offset: usize,
+        length: usize,
+        device: MetalDevice,
+        dtype: GgmlDType,
+    ) -> Self {
+        Self {
+            buffer,
+            offset,
+            length,
+            device,
+            dtype,
+        }
     }
 
     pub fn dtype(&self) -> GgmlDType {
@@ -37,19 +69,31 @@ impl QMetalStorage {
         &self.buffer
     }
 
+    /// Byte offset of this tensor's data within `buffer()`. Zero unless this
+    /// storage was built via `from_buffer` over a shared buffer.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
+
+    /// This tensor's own byte length -- use this, not `buffer().length()`,
+    /// when `buffer()` may be shared with other tensors.
+    pub fn length(&self) -> usize {
+        self.length
+    }
+
     pub fn dequantize(&self, elem_count: usize) -> Result<MetalStorage> {
         use crate::quantized::k_quants::GgmlType;
 
         let buffer = self
             .device
             .new_buffer_builder()
-            .with_size(self.buffer.length())
+            .with_size(self.length)
             .with_label("qstorage_dequantize_blit")
             .build()?;
         {
             let mut blit = self.device.blit_command_encoder()?;
             blit.set_label("blit_to_cpu");
-            blit.copy_from_buffer(&self.buffer, 0, &buffer, 0, self.buffer.length());
+            blit.copy_from_buffer(&self.buffer, self.offset, &buffer, 0, self.length);
         }
         self.device.flush_and_wait_current()?;
         let mut out = vec![0.0; elem_count];
@@ -144,7 +188,13 @@ impl QMetalStorage {
             .with_data(&qcpu_storage.data()?)
             .with_label("qstorage_quantized")
             .build()?;
+        // A fresh full-ownership allocation replaces whatever view (offset,
+        // length) this storage previously had -- reset both, or a
+        // quantize-after-from_buffer would keep stale view metadata pointing
+        // into a buffer this storage no longer shares.
+        self.length = buffer.length();
         self.buffer = buffer;
+        self.offset = 0;
         Ok(())
     }
 
@@ -166,7 +216,13 @@ impl QMetalStorage {
             .with_data(&qcpu_storage.data()?)
             .with_label("qstorage_quantize_imatrix")
             .build()?;
+        // A fresh full-ownership allocation replaces whatever view (offset,
+        // length) this storage previously had -- reset both, or a
+        // quantize-after-from_buffer would keep stale view metadata pointing
+        // into a buffer this storage no longer shares.
+        self.length = buffer.length();
         self.buffer = buffer;
+        self.offset = 0;
         Ok(())
     }
 
@@ -192,7 +248,13 @@ impl QMetalStorage {
             .with_data(&qcpu_storage.data()?)
             .with_label("qstorage_quantize_imatrix_onto")
             .build()?;
+        // A fresh full-ownership allocation replaces whatever view (offset,
+        // length) this storage previously had -- reset both, or a
+        // quantize-after-from_buffer would keep stale view metadata pointing
+        // into a buffer this storage no longer shares.
+        self.length = buffer.length();
         self.buffer = buffer;
+        self.offset = 0;
         Ok(())
     }
 
@@ -213,12 +275,18 @@ impl QMetalStorage {
             .with_data(&qcpu_storage.data()?)
             .with_label("qstorage_quantize_onto")
             .build()?;
+        // A fresh full-ownership allocation replaces whatever view (offset,
+        // length) this storage previously had -- reset both, or a
+        // quantize-after-from_buffer would keep stale view metadata pointing
+        // into a buffer this storage no longer shares.
+        self.length = buffer.length();
         self.buffer = buffer;
+        self.offset = 0;
         Ok(())
     }
 
     pub fn storage_size_in_bytes(&self) -> usize {
-        self.buffer.length()
+        self.length
     }
 
     pub fn embedding(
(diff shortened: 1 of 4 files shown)


candle-core/src/quantized/metal.rs
```diff
@@ -8,6 +8,16 @@ pub struct QMetalStorage {
     dtype: GgmlDType,
     device: MetalDevice,
     buffer: Arc<Buffer>,
+    /// Byte offset of this tensor's data within `buffer`. Nonzero when
+    /// `buffer` is shared across several tensors (e.g. one no-copy buffer
+    /// wrapping an entire mmap'd file) via `from_buffer`; zero for every
+    /// storage that owns its buffer outright.
```
ivarflakstad:
```suggestion
    /// Byte offset of this tensor's data within `buffer`.
    /// Allows multiple quantized tensors to be backed by the same underlying metal buffer.
```

candle-core/src/quantized/metal.rs
```diff
@@ -8,6 +8,16 @@ pub struct QMetalStorage {
...
+    offset: usize,
+    /// This tensor's own byte length -- may be smaller than `buffer.length()`
+    /// when `buffer` is shared. Every read/blit against `buffer` must use
+    /// this, not `buffer.length()`, or it silently reads past this tensor's
+    /// data into a neighboring one.
```
ivarflakstad:
```suggestion
    /// This tensor's own byte length. May be smaller than `buffer.length()`
    /// when `buffer` is shared. All uses of `buffer` must use this, and not 
    /// `buffer.length()`, or risk silently reading past this tensor's data
    /// into a neighboring one.
```

candle-core/src/quantized/metal.rs
```diff
@@ -19,12 +29,34 @@ impl QMetalStorage {
...
             device: device.clone(),
             dtype,
+            offset: 0,
         })
     }
 
+    /// Wraps an existing, externally-created buffer -- e.g. a no-copy Metal
+    /// buffer over an mmap'd region -- as this tensor's storage, with no
+    /// allocation and no copy. `offset`/`length` locate this tensor's bytes
+    /// within `buffer`, which may be shared with other tensors.
```
ivarflakstad:
```suggestion
    /// Wraps an existing buffer as this tensor's storage.
    /// The buffer may be used as the shared underlying buffer of several tensors,
    /// so `offset` and `length` define the specific bytes that make up this tensor.
```

candle-core/src/quantized/metal.rs
```diff
@@ -144,7 +188,13 @@ impl QMetalStorage {
             .with_data(&qcpu_storage.data()?)
             .with_label("qstorage_quantized")
             .build()?;
+        // A fresh full-ownership allocation replaces whatever view (offset,
+        // length) this storage previously had -- reset both, or a
+        // quantize-after-from_buffer would keep stale view metadata pointing
+        // into a buffer this storage no longer shares.
+        self.length = buffer.length();
         self.buffer = buffer;
+        self.offset = 0;
```
ivarflakstad:
```suggestion
        // New backing allocation
        self.buffer = buffer;
        self.offset = 0;
        self.length = buffer.length();
```

Perhaps we should introduce a fn to update buffer, offset, and length together. Makes it harder to miss these edge cases.

candle-core/src/quantized/metal.rs
```diff
@@ -166,7 +216,13 @@ impl QMetalStorage {
             .with_data(&qcpu_storage.data()?)
             .with_label("qstorage_quantize_imatrix")
             .build()?;
+        // A fresh full-ownership allocation replaces whatever view (offset,
+        // length) this storage previously had -- reset both, or a
+        // quantize-after-from_buffer would keep stale view metadata pointing
+        // into a buffer this storage no longer shares.
+        self.length = buffer.length();
         self.buffer = buffer;
+        self.offset = 0;
```
ivarflakstad:
```suggestion
        // New backing allocation
        self.buffer = buffer;
        self.offset = 0;
        self.length = buffer.length();
```

candle-core/src/quantized/metal.rs
```diff
@@ -192,7 +248,13 @@ impl QMetalStorage {
             .with_data(&qcpu_storage.data()?)
             .with_label("qstorage_quantize_imatrix_onto")
             .build()?;
+        // A fresh full-ownership allocation replaces whatever view (offset,
+        // length) this storage previously had -- reset both, or a
+        // quantize-after-from_buffer would keep stale view metadata pointing
+        // into a buffer this storage no longer shares.
+        self.length = buffer.length();
         self.buffer = buffer;
+        self.offset = 0;
```
ivarflakstad:
```suggestion
        // New backing allocation
        self.buffer = buffer;
        self.offset = 0;
        self.length = buffer.length();
```

candle-core/src/quantized/metal.rs
```diff
@@ -213,12 +275,18 @@ impl QMetalStorage {
             .with_data(&qcpu_storage.data()?)
             .with_label("qstorage_quantize_onto")
             .build()?;
+        // A fresh full-ownership allocation replaces whatever view (offset,
+        // length) this storage previously had -- reset both, or a
+        // quantize-after-from_buffer would keep stale view metadata pointing
+        // into a buffer this storage no longer shares.
+        self.length = buffer.length();
         self.buffer = buffer;
+        self.offset = 0;
```
ivarflakstad:
```suggestion
        // New backing allocation
        self.buffer = buffer;
        self.offset = 0;
        self.length = buffer.length();
```

candle-metal-kernels/src/metal/device.rs
```diff
@@ -101,6 +101,42 @@ impl Device {
...
+    /// construction, so wrapping a whole mapping from its start needs no
+    /// adjustment).
+    ///
+    /// No deallocator is registered: nothing needs to be freed on this
+    /// side when the buffer is released, since the backing memory is
+    /// owned and will be unmapped by the caller, not by this buffer.
+    ///
+    /// # Safety
+    /// `pointer` must be valid for reads for `length` bytes for at least as
+    /// long as the returned `Buffer` is alive.
```
ivarflakstad:
```suggestion
```

candle-core/tests/quantized_tests.rs
```diff
@@ -1513,3 +1513,86 @@ test_device!(
...
     from_data_dequant_matches_canonical_when_caller_passes_cow_owned_metal
 );
+
+/// `QMetalStorage::from_buffer` is how a zero-copy (e.g. mmap-backed) loader
+/// wires a tensor's storage to a *view* -- possibly at a nonzero offset --
+/// into a buffer shared with other tensors, instead of a buffer this storage
+/// owns outright. This must produce identical dequantize and matmul output
+/// to the normal owned-buffer path, at both a zero and a nonzero offset --
+/// an offset-arithmetic bug here would silently corrupt weights rather than
+/// crash.
```
ivarflakstad:
```suggestion
```

The test name describes everything you need to know :)

candle-core/tests/quantized_tests.rs
```diff
@@ -1513,3 +1513,86 @@ test_device!(
...
+
+    let src_data: Vec<f32> = (0..n).map(|i| ((i as f32) * 0.013).sin()).collect();
+    let src = Tensor::from_vec(src_data, (n_out, k), &device)?;
+    let canonical = QTensor::quantize(&src, dtype)?;
+    let canonical_dequant = canonical.dequantize(&device)?.to_vec2::<f32>()?;
+    let raw_bytes: Vec<u8> = canonical.data()?.to_vec();
+
+    // QTensor isn't Clone and QMatMul::from_qtensor takes it by value, so
+    // compute the canonical matmul reference once, up front -- it doesn't
+    // depend on front_padding, only `view` needs rebuilding per iteration.
```
ivarflakstad:
```suggestion
```

candle-core/tests/quantized_tests.rs
```diff
@@ -1513,3 +1513,86 @@ test_device!(
...
+            .zip(view_dequant.iter().flatten())
+            .map(|(a, b): (&f32, &f32)| (a - b).abs())
+            .fold(0f32, f32::max);
+        assert_eq!(
+            max_diff, 0.0,
+            "from_buffer view (offset={front_padding}) dequant must be bit-identical to the owned-buffer path"
+        );
+
+        // Exercise the offset-aware matmul path (call_quantized_matmul_mv_t),
+        // not just dequantize -- these are two independent call sites.
```
ivarflakstad:
```suggestion
        // Also test matmul as it is a separate code path
```

candle-core/src/quantized/metal.rs
```diff
@@ -37,19 +69,31 @@ impl QMetalStorage {
...
     }
 
+    /// Byte offset of this tensor's data within `buffer()`. Zero unless this
+    /// storage was built via `from_buffer` over a shared buffer.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
+
+    /// This tensor's own byte length -- use this, not `buffer().length()`,
+    /// when `buffer()` may be shared with other tensors.
```
ivarflakstad:
```suggestion
    /// Byte length of the tensor. Use this, not `buffer().length()`,
    /// as the underlying buffer may be shared with other tensors.
```

ivarflakstad (commented):
Most of this looks good to me. Just got to update some comments, and possibly add a helper fn for updating buffer, offset, length.

alytaphoenix:
Applied all the suggested comment simplifications, plus introduced `replace_with_owned_buffer(&mut self, buffer)` to set buffer/offset/length together at all four `quantize*()` call sites, per your suggestion. Rebased onto the branch's latest merge from main. Tests still green (68/68 in `quantized_tests`).

AlpineVibrations:
nice one.

alytaphoenix:
Ping on this one. All twelve suggestions were applied in `b00cbbe2` — the doc wording is yours verbatim, and I took the `replace_with_owned_buffer` suggestion from the thread on the buffer-swap sites, so all four call sites now go through it rather than setting the three fields by hand. Happy to rebase if it's drifted.

PR #3786, opened 2026-07-27
Fix duplicated words in documentation

cupkk:
Fixes two duplicated words in user-facing documentation:

- `embeddings embeddings` in the Stella embedding example
- `single single tensor` in the `ModuleT` trait documentation

No APIs or runtime behavior change. Validated with exact text assertions and `git diff --check`.


ivarflakstad (approved):
Thanks 👌

Issue #3781, opened 2026-07-26
`quantized`: expose whether a `GgmlDType` supports matmul on a given device


astorise:
### Problem

There is no way to ask, ahead of loading a checkpoint, whether a given
`GgmlDType` can actually be multiplied on a given `Device`. The answer exists
only inside kernel dispatch, as the fallthrough arm of several `match`
statements:

- `candle-core/src/quantized/cuda.rs` — `dequantize_mul_mat_vec`,
  `mul_mat_vec_via_q8_1` and `mul_mat_via_q8_1` each end in
  `_ => crate::bail!("unsupported dtype for quantized matmul {dtype:?}")`.
- `candle-core/src/quantized/metal.rs` — covers every `GgmlDType`.

The sets genuinely differ, and not in an obvious direction:

| | CUDA | Metal |
|---|---|---|
| `Q4_0` `Q4_1` `Q5_0` `Q5_1` `Q8_0` `Q2K` `Q3K` `Q4K` `Q5K` `Q6K` | ✅ | ✅ |
| `Q8_1`, `Q8K` | dequantize only, **no matmul** | ✅ |

`Q8_1`/`Q8K` are the trap: they have CUDA *dequantize* kernels, so a checkpoint
carrying them loads and occupies VRAM, then fails at the first decode step with
a generic dtype error.

### Proposed API

```rust
impl GgmlDType {
    /// Whether a quantized matmul kernel exists for this block type on `device`.
    pub fn supports_matmul(self, device: &Device) -> bool;
}
```

Implemented next to the dispatch it describes, so the two cannot drift.

`F32`/`F16`/`BF16` report `true` on any device: `QMatMul::from_arc` dequantizes
them into a dense `Tensor` rather than dispatching to a quantized kernel, so
they never reach the matmul match at all — another piece of knowledge currently
only obtainable by reading the source.

There is deliberately no `supports_dequantize` counterpart: it would return
`true` for every dtype on every backend by construction, since the CUDA path
falls back to host-side dequantization for anything without a fast kernel. The
question worth asking there is a *performance* one ("is there a device-native
dequantize kernel"), which is not what this API answers and should not share
its name.

### Concrete caller

`astorise/tachyon-mesh` (`core-host/src/ai_inference/candle_llm_runtime.rs`)
validates a GGUF checkpoint's block types *before* uploading weights, so an
unrunnable model fails at load with an actionable message instead of at the
first decode step with VRAM already claimed:

```
[...]

PR #3787, opened 2026-07-27
Add dtype support validation for quantized matmul operations

astorise:
Fixes astorise/candle#3781.

## Summary
This PR adds validation methods to check whether a given GGML dtype is supported for quantized matmul operations and dequantization on different backends. It refactors existing dtype-checking logic in the MoE kernel to use a shared helper function, improving code reusability and maintainability.

## Key Changes

- **Added `GgmlDType::supports_matmul()` method** in `candle-core/src/quantized/mod.rs`:
  - Returns `true` for all dense types (F32, F16, BF16) and most quantized types
  - Returns `false` for Q8_1 and Q8K on CUDA/Metal backends (CPU-only types used internally)
  - Allows callers to validate checkpoint tensors before dispatching operations

- **Added `GgmlDType::supports_dequantize()` method** in `candle-core/src/quantized/mod.rs`:
  - Currently always returns `true` as all backends support dequantization
  - Separated from `supports_matmul()` to allow independent evolution of backend capabilities

- **Refactored MoE kernel dtype validation** in `candle-nn/src/moe.rs`:
  - Extracted `moe_gemm_gguf_dtype_code()` helper function to map supported dtypes to FFI codes
  - Added public `moe_gemm_gguf_supports_dtype()` function for pre-validation of expert weights
  - Simplified dtype matching logic in `moe_gemm_gguf()` using the new helper

- **Added comprehensive test** in `candle-core/tests/quantized_tests.rs`:
  - Validates that all GGML dtypes support both matmul and dequantize on CPU backend
  - Ensures CPU's generic `vec_dot` implementation covers all dtype variants

## Implementation Details

The dtype code mapping (Q8_0→0, Q4K→1, Q2K→2, Q3K→3, Q5K→4, Q6K→5) is now centralized in `moe_gemm_gguf_dtype_code()`, eliminating duplication and making it easier to maintain the FFI interface. The public `moe_gemm_gguf_supports_dtype()` function enables early validation of checkpoint compatibility before expensive forward pass operations.

### DIFF
--- a/candle-core/tests/quantized_tests.rs
+++ b/candle-core/tests/quantized_tests.rs
@@ -1559,7 +1559,6 @@ fn cpu_supports_qmatmul_matches_backend() -> Result<()> {
             ran,
             "supports_qmatmul disagrees with the CPU backend for {dtype:?}"
         );
-        assert!(device.supports_dequantize(dtype));
     }
     Ok(())
 }
--- a/candle-core/tests/quantized_tests.rs
+++ b/candle-core/tests/quantized_tests.rs
@@ -1513,3 +1513,38 @@ test_device!(
     from_data_dequant_matches_canonical_when_caller_passes_cow_owned_cuda,
     from_data_dequant_matches_canonical_when_caller_passes_cow_owned_metal
 );
+
+#[test]
+fn ggml_dtype_supports_matmul_and_dequantize_on_cpu() -> Result<()> {
+    let cpu = Device::Cpu;
+    let all_dtypes = [
+        GgmlDType::F32,
+        GgmlDType::F16,
+        GgmlDType::BF16,
+        GgmlDType::Q4_0,
+        GgmlDType::Q4_1,
+        GgmlDType::Q5_0,
+        GgmlDType::Q5_1,
+        GgmlDType::Q8_0,
+        GgmlDType::Q8_1,
+        GgmlDType::Q2K,
+        GgmlDType::Q3K,
+        GgmlDType::Q4K,
+        GgmlDType::Q5K,
+        GgmlDType::Q6K,
+        GgmlDType::Q8K,
+    ];
+    for dtype in all_dtypes {
+        // The CPU backend implements `vec_dot` generically for every GGML dtype, so it always
+        // supports both matmul and dequantize.
+        assert!(
+            dtype.supports_matmul(&cpu),
+            "expected {dtype:?} to support matmul on the CPU backend"
+        );
+        assert!(
+            dtype.supports_dequantize(&cpu),
+            "expected {dtype:?} to support dequantize on the CPU backend"
+        );
+    }
+    Ok(())
+}
--- a/candle-core/tests/quantized_tests.rs
+++ b/candle-core/tests/quantized_tests.rs
@@ -1514,37 +1514,52 @@ test_device!(
     from_data_dequant_matches_canonical_when_caller_passes_cow_owned_metal
 );
 
+/// The exhaustive `match` in `Device::supports_qmatmul` forces a decision whenever a variant is
+/// added, and this list makes the test fail to compile until it is updated too.
+const ALL_GGML_DTYPES: [GgmlDType; 15] = [
+    GgmlDType::F32,
+    GgmlDType::F16,
+    GgmlDType::BF16,
+    GgmlDType::Q4_0,
+    GgmlDType::Q4_1,
+    GgmlDType::Q5_0,
+    GgmlDType::Q5_1,
+    GgmlDType::Q8_0,
+    GgmlDType::Q8_1,
+    GgmlDType::Q2K,
+    GgmlDType::Q3K,
+    GgmlDType::Q4K,
+    GgmlDType::Q5K,
+    GgmlDType::Q6K,
+    GgmlDType::Q8K,
+];
+
+/// Check the predicate against what the backend actually does rather than restating it.
 #[test]
-fn ggml_dtype_supports_matmul_and_dequantize_on_cpu() -> Result<()> {
-    let cpu = Device::Cpu;
-    let all_dtypes = [
-        GgmlDType::F32,
-        GgmlDType::F16,
-        GgmlDType::BF16,
-        GgmlDType::Q4_0,
-        GgmlDType::Q4_1,
-        GgmlDType::Q5_0,
-        GgmlDType::Q5_1,
-        GgmlDType::Q8_0,
-        GgmlDType::Q8_1,
-        GgmlDType::Q2K,
-        GgmlDType::Q3K,
-        GgmlDType::Q4K,
-        GgmlDType::Q5K,
-        GgmlDType::Q6K,
-        GgmlDType::Q8K,
-    ];
-    for dtype in all_dtypes {
-        // The CPU backend implements `vec_dot` generically for every GGML dtype, so it always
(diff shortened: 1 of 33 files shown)


ivarflakstad (commented):
Looks pretty good!

In my mind it does make more sense if the device checks if it supports qmatmul of a given dtype though, ala `device.supports_qmatmul(dtype)` and `device.supports_dequantize(dtype)`.

But I'll still give some feedback as I think it will be useful.

Could you take a look at the comments of other functions that `GgmlDType` has?
The LLM style comments are way too verbose. Even just the first sentence is too much. It could simply be `/// Wether the provided device support quantized matmul with this dtype`
Same applies to `supports_dequantize`.

I'm not convinced `ggml_dtype_supports_matmul_and_dequantize_on_cpu` adds a lot of value. It would be more valuable if we could iterate over all the enum variants using a macro / reflection. Then the test could be written as a guard that ensures all ggml dtypes are always valid on cpu. But that requires work that is out of scope for this PR, so not asking you to do that here.

astorise:
> Looks pretty good!
> 
> In my mind it does make more sense if the device checks if it supports qmatmul of a given dtype though, ala `device.supports_qmatmul(dtype)` and `device.supports_dequantize(dtype)`.
> 
> But I'll still give some feedback as I think it will be useful.
> 
> Could you take a look at the comments of other functions that `GgmlDType` has? The LLM style comments are way too verbose. Even just the first sentence is too much. It could simply be `/// Wether the provided device support quantized matmul with this dtype` Same applies to `supports_dequantize`.
> 
> I'm not convinced `ggml_dtype_supports_matmul_and_dequantize_on_cpu` adds a lot of value. It would be more valuable if we could iterate over all the enum variants using a macro / reflection. Then the test could be written as a guard that ensures all ggml dtypes are always valid on cpu. But that requires work that is out of scope for this PR, so not asking you to do that here.

Thanks — all three points addressed.

**`device.supports_qmatmul(dtype)`**: agreed, and it lines up with the existing
`Device::supports_bf16`, so the check now lives on `Device` and the `GgmlDType`
methods are gone rather than kept in parallel. Note this departs from the API
sketched in #3781, which put it on the dtype — following your shape here.

**Docs**: trimmed to three lines.

**Test**: you were right that it didn't earn its place — it hardcoded a list and
asserted the values the implementation returns by construction, so it couldn't
catch anything. It now runs a real qmatmul per dtype (`m = 2`, so the general
path rather than mat-vec) and asserts the predicate agrees with whether the
backend actually accepted it; it fails if the two ever drift apart. For the
new-variant coverage you had in mind with macro iteration, the exhaustive
`match` already forces a decision at compile time, so I'd rather not pull in a
derive/macro dependency just to enumerate.

Two further changes:

**Dropped `supports_dequantize`** — #3781 explicitly rules out a counterpart
since every dtype dequantizes on every backend, so it carried no information.

**Corrected the Metal rationale.** My original comment claimed Metal had no
kernels for `Q8_1`/`Q8K`, which is wrong: `call_quantized_matmul_mv_t` handles
both, only `call_quantized_matmul_mm_t` rejects them, and `fwd` dispatches
between the two on `m == 1`. So on Metal these work for mat-vec and fail from
`m > 1` onwards. I deliberately kept the predicate at `false` there: the
signature has no `m` to condition on, and returning `true` would let decode pass
and prefill fail — precisely the "loads fine, dies during inference" mode #3781
is meant to eliminate. A false negative only costs a dequantize + dense fallback.

On CUDA the original reasoning holds — all three kernel selectors
(`dequantize_mul_mat_vec`, `mul_mat_vec_q`, `mul_mat_q`) reject both dtypes.

For context on the intended consumer: `astorise/tachyon-mesh` currently
hardcodes these dtype checks mirroring the CUDA dispatch, which is the
duplicated invariant this is meant to retire.

candle-core/tests/quantized_tests.rs
```diff
@@ -1513,3 +1513,52 @@ test_device!(
...
+    GgmlDType::Q8_1,
+    GgmlDType::Q2K,
+    GgmlDType::Q3K,
+    GgmlDType::Q4K,
+    GgmlDType::Q5K,
+    GgmlDType::Q6K,
+    GgmlDType::Q8K,
+];
+
+/// Check the predicate against what the backend actually does rather than restating it.
```
ivarflakstad:
```suggestion
```

candle-core/tests/quantized_tests.rs
```diff
@@ -1513,3 +1513,52 @@ test_device!(
     from_data_dequant_matches_canonical_when_caller_passes_cow_owned_cuda,
     from_data_dequant_matches_canonical_when_caller_passes_cow_owned_metal
 );
+
+/// The exhaustive `match` in `Device::supports_qmatmul` forces a decision whenever a variant is
+/// added, and this list makes the test fail to compile until it is updated too.
```
ivarflakstad: This comment doesn't really add anything of value, and this part

> and this list makes the test fail to compile until it is updated too.

is simply not true. The test will compile fine.

```suggestion
```

candle-nn/src/moe.rs
```diff
@@ -2,9 +2,31 @@
...
+    }
+}
+
+/// Whether `dtype` is one of the GGML quant types accepted as expert weights by
+/// [`moe_gemm_gguf`] (`Q2K`, `Q3K`, `Q4K`, `Q5K`, `Q6K` or `Q8_0`). `moe_gemm_gguf` is CUDA-only
+/// and fails at runtime for any other dtype; checking this ahead of time lets callers validate a
+/// checkpoint's expert tensors before dispatching a MoE forward pass.
+pub fn moe_gemm_gguf_supports_dtype(dtype: GgmlDType) -> bool {
+    moe_gemm_gguf_dtype_code(dtype).is_some()
+}
```
ivarflakstad: Let's hold off on adding this until it is actually used somewhere.

```suggestion
```

astorise:
Updated in 635791f — all three inline comments addressed:

- Removed `moe_gemm_gguf_supports_dtype`.
- Removed the comment above `ALL_GGML_DTYPES`. You're right that the claim was
  simply false: a fixed-size array of explicit variants keeps compiling when a
  variant is added, so only the `match` in `supports_qmatmul` actually forces a
  decision. The test list can still go stale silently — I didn't think that was
  worth a macro or a derive dependency to close.
- Removed the redundant comment above the test.

I also reverted `candle-nn/src/moe.rs` entirely. That extraction only existed to
share the dtype -> FFI code mapping with the helper you asked me to hold off on;
with it gone there's a single caller left, in a file unrelated to what this PR is
for. The diff is now just `candle-core/src/device.rs` and
`candle-core/tests/quantized_tests.rs`, +61/-0.

One note on the rule itself, since the issue body has been corrected in the
meantime: on Metal `Q8_1`/`Q8K` are mat-vec only — `call_quantized_matmul_mm_t`
and `call_quantized_get_rows` both return `UnsupportedDTypeForOp`. A checkpoint
stored in either dtype decodes at batch 1, then fails at prefill and on any
embedding lookup, so `Q8_1 | Q8K => self.is_cpu()` is the correct rule rather
than a conservative one. `supports_dequantize` has been dropped from both the
proposal and the PR.

The PR description predates these review rounds and is out of date — this comment
is the current state.

---
_Generated by [Claude Code](https://claude.ai/code)_

astorise:

PR #3789, opened 2026-07-27
Add `IncrementalDecoder` to `candle-transformers::generation`

### LINKED ISSUE #3782: `candle-transformers::generation`: provide a correct incremental detokenizer
### Problem

Every autoregressive decode loop needs the text generated so far on each step —
for stop-sequence matching and for streaming deltas. The two available options
are both unsatisfying:

1. Re-decode the whole token sequence each step. Correct, but O(n²) in the
   generated length. Invisible at 256 tokens, millions of redundant per-token
   decodes at several thousand.
2. `candle-examples`'s `TokenOutputStream`. O(1) amortised, but its
   `is_alphanumeric` heuristic changes emission timing and it can hold text back
   indefinitely; it is an example, not a supported API.

The naive fix does not work, which is what makes this worth having in the
library rather than in each consumer. A BPE/SentencePiece decode is not the
concatenation of its tokens' decodes: a multi-byte character can span several
tokens (the prefix decodes to U+FFFD, which a later token retroactively
replaces), and decoding a sub-sequence can gain or lose a leading space.

### What a correct implementation needs

We built one in Tachyon-Mesh and it took three rounds of review to get right.
The non-obvious parts:

- **Window + delta, not append.** Keep a bounded trailing window of tokens,
  re-decode only that, and append the difference against the *previous* window
  decode — so a sub-sequence artifact appears in both decodes and cancels.
- **Revision, not just extension.** When a new token rewrites text inside the
  window, replace that suffix wholesale.
- **Verified re-anchoring.** Only move the window when the shorter decode is
  provably a suffix of the current one. A tokenizer that never offers a clean
  split must degrade to whole-sequence decoding, not corrupt output.
- **Bounded decoder context.** The suffix check proves the shortened decode
  matches *now*; it does not prove a future token cannot rewrite text before the
  anchor. A `Replace` decoder with a regex pattern can do exactly that. Windowing
  must be gated on the decoder having a bounded context, and for a literal
  `Replace` the retained tail must be at least as long as the pattern.

That last point in particular is only checkable from inside — `tokenizers` keeps
the decoder fields private, so we had to inspect the *serialized* decoder and
treat anything unrecognized as unbounded.

### Proposed API

```rust
// candle-transformers::generation, next to LogitsProcessor
pub struct IncrementalDecoder { /* … */ }

impl IncrementalDecoder {
    pub fn new(tokenizer: &Tokenizer) -> Self;   // resolves the bounded-context property
    pub fn push(&mut self, token: u32) -> Result<&str>;  // newly available text
    pub fn text(&self) -> &str;
}
```

The invariant worth asserting in tests: the incremental text equals a
whole-sequence decode at **every** step.

### Note

This arguably belongs in `tokenizers` rather than candle, since nothing about
it is candle-specific and the decoder introspection would be far cleaner with
access to the private fields. Filing here because that is where the decode loops
live; happy to move it upstream if you prefer.

astorise:
Implements the incremental detokenizer requested in #3782.

## Problem

Every autoregressive decode loop needs the text generated so far on each step, for
stop-sequence matching and for streaming deltas. Today there are two options and both are
unsatisfying:

1. Re-decode the whole token sequence each step — correct, but O(n²) in the generated
   length.
2. `candle-examples`'s `TokenOutputStream` — O(1) amortised, but its `is_alphanumeric`
   heuristic changes emission timing and can hold text back indefinitely; it is an
   example, not a supported API.

The naive fix does not work, which is what makes this worth having in the library. A
BPE/SentencePiece decode is not the concatenation of its tokens' decodes: a multi-byte
character can span several tokens (the prefix decodes to U+FFFD, which a later token
retroactively replaces), and decoding a sub-sequence can gain or lose a leading space.

## API

```rust
// candle-transformers::generation, next to LogitsProcessor
pub struct IncrementalDecoder { /* … */ }

impl IncrementalDecoder {
    pub fn new(tokenizer: &Tokenizer) -> Self;        // clones
    pub fn from_tokenizer(tokenizer: Tokenizer) -> Self;
    pub fn with_skip_special_tokens(self, skip: bool) -> Self;

    pub fn push(&mut self, token: u32) -> Result<&str>;  // newly settled text
    pub fn text(&self) -> &str;                          // whole decoded text
    pub fn finish(&mut self) -> &str;                    // release what is held back
    pub fn pending(&self) -> &str;                       // held back, without consuming

    pub fn is_windowed(&self) -> bool;
    pub fn window_len(&self) -> usize;
    pub fn retracted(&self) -> usize;
    pub fn tokens(&self) -> &[u32];
    pub fn tokenizer(&self) -> &Tokenizer;
    pub fn into_tokenizer(self) -> Tokenizer;
    pub fn clear(&mut self);
}
```

Two invariants:

- **`text()` equals a whole-sequence decode of every token pushed so far, at every step** —
  the one the issue asks to assert in tests.
- **Emission is append-only**: concatenating every `push()` result, then `finish()`,
  reproduces `text()` exactly. `push` never returns text it has to take back.

Typical use:

```rust
let mut decoder = IncrementalDecoder::new(&tokenizer);
for _ in 0..max_tokens {
    let token = logits_processor.sample(&logits)?;
    print!("{}", decoder.push(token)?);
    if decoder.text().ends_with(stop_sequence) {
        break;
    }
}
print!("{}", decoder.finish());
```

## How it works

- **Window + delta, not append.** A bounded trailing window of tokens (re-anchored to 8,
  re-anchoring attempted at 64) is re-decoded on each push, and the delta is taken against
  the *previous* decode of the same window — so a sub-sequence artifact appears in both
  decodes and cancels.
- **Revision, not just extension.** Text a future token could still rewrite is withheld
  rather than emitted and later corrected.
- **Verified re-anchoring.** The window only moves forward when the shorter decode is
  provably a suffix of the current window decode, is long enough that a rewriting decoder
  cannot reach past it, and does not start inside an open `ByteFallback` marker run. When
  no split qualifies the window keeps growing, degrading towards whole-sequence decoding
  rather than dropping or duplicating text.
- **Bounded decoder context.** The suffix check proves the shortened decode matches *now*;
  it does not prove a future token cannot rewrite text before the anchor. Since
  `tokenizers` keeps the decoder fields private, the decoder is inspected through its
  serialized form, and anything unrecognized is treated as unbounded.

The decoder scan is **order-aware**, which is the part that is easy to get wrong. It tracks
three properties of the decoders seen so far, and reads each subsequent decoder against
them:

| Scan state | Meaning |
| --- | --- |
| `fused` | The token list has been collapsed into one string, so later string rewrites apply to the whole text instead of one token. |
| `suppresses` | The list can lose entries, so a push does not necessarily hand later decoders one more element. |
| `alters_markers` | A token's string can be rewritten, so whether a `<0xXX>` byte marker reaches `ByteFallback` can no longer be told from the raw id. |

| Serialized decoder | Verdict |
| --- | --- |
| *(none)* — tokens joined with a space | bounded |
| `Fuse` | bounded; sets `fused` |
| `ByteLevel` | bounded and sets `fused`; **unbounded** if already `fused` |
| `ByteFallback` | bounded, enables run tracking, sets `suppresses`; **unbounded** if `fused` or `alters_markers` |
| `Metaspace` | bounded; sets `alters_markers` unless `prepend_scheme` is `never` and its replacement is marker-neutral |
| `WordPiece` | bounded, sets `alters_markers`; if `fused` and `cleanup`, tail += longest cleanup pattern |
| `CTC` | bounded, sets `suppresses` and `alters_markers`; if `fused`, tail += `pad_token` + cleanup pattern + `word_delimiter_token` |
| `BPEDecoder` | sets `alters_markers`; if `fused`, tail += `suffix`; else one-step lag, or **unbounded** if `suppresses` |
| `Strip` | bounded, sets `alters_markers`; if `fused`, tail += `stop` × `content` |
| `Replace`, literal pattern | bounded; sets `alters_markers` unless marker-neutral; if `fused`, tail += `pattern` |
| `Replace`, regex pattern | sets `alters_markers`; **unbounded** if `fused` |
| `Sequence` | recurse in order, threading the scan state |
| anything else | **unbounded** |

"Marker-neutral" is decided by the alphabet a `<0xXX>` marker is made of — `<`, `>`, `x`
and hex digits. A rewrite whose *output* holds none of those characters cannot forge a
marker, since its result always carries a character no marker has; and one whose *input
pattern* holds none of them cannot match inside a marker, so it cannot destroy one either.
Both halves have to hold. The SentencePiece chain replaces `▁` with a space and satisfies
both, so it keeps the fast path.

`push` withholds text that is not yet settled: a trailing run of U+FFFD from an incomplete
multi-byte character, an open `ByteFallback` marker run (the run is converted as a whole,
so one invalid byte re-renders all of it), the reach of a rewriting decoder, the last token
of an unfused `BPEDecoder`, and — for a decoder that is not windowed — everything, since
nothing about it can be shown to be settled.

`retracted()` is a diagnostic that is always zero. A non-zero value means some decoder's
rewrite reach was under-estimated in the scan, i.e. a bug here rather than something a
caller has to handle.

## Dependency

`candle-transformers` gains a direct dependency on `tokenizers`, but **no crate is added to
the dependency graph**: `candle-core` already depends on `tokenizers` unconditionally
(`candle-core/Cargo.toml:38`, used by `candle-core/src/quantized/tokenizer.rs` for GGUF
tokenizer support), and `candle-transformers` depends on `candle-core` through `candle-nn`.
Every `candle-transformers` user compiles it today.

A feature gate was considered and dropped for that reason: it would have bought nothing in
compile time or graph size, while making the API harder to discover, hiding it from
docs.rs by default, and leaving the tests reachable only through feature unification. No
`tokenizers` feature is requested here, so the workspace's `default-features = false`
applies — the decoder introspection only uses `serde_json` on the serialized form and never
touches the regex engine.

Happy to put it behind a feature instead if you would rather `candle-transformers` not name
`tokenizers` directly, or if that `candle-core` dependency is something you plan to remove.

## `TokenOutputStream`

`candle-examples`'s `TokenOutputStream` becomes a thin adapter over `IncrementalDecoder`,
with its public API unchanged, so no example needed touching. Behaviour change: the
`is_alphanumeric` heuristic is gone, so text streams as soon as it is provably settled
instead of waiting for the next alphanumeric character, and nothing is held back
indefinitely. `decode_all()` returns the maintained text rather than re-decoding.

## Testing

`candle-transformers/tests/incremental_decoder_tests.rs` — 21 tests, each building a
tokenizer from an explicit vocabulary and decoder so nothing is downloaded. They cover the
base behaviour (multi-byte characters spanning tokens, the SentencePiece chain and its
leading-space strip, `WordPiece`, no decoder, `BPEDecoder`'s deferred last token, window
boundedness over a 2000-token run) and one regression test per composition that was found
to break the invariants.

Separately, not in the test suite since it needs network, the invariants were checked
against four real `tokenizer.json` files — GPT-2, Qwen2.5-0.5B, bert-base-uncased and
Mistral-7B-Instruct-v0.3 — over 3000 random tokens each. In every case `text()` matched a
whole-sequence decode at every step, the emitted text was always a prefix of it, there were
no retractions, the window stayed at or below 64 tokens, and nothing was left held back at
the end.

`cargo fmt --check`, `cargo clippy --tests -- -D warnings` and the test suite pass.

## Notes for reviewers

The decoder scan started out considerably more permissive, and review hardened it through
thirteen findings. Two of them broke `text()` itself rather than just emission, and are
worth knowing about if you review this area:

- Re-anchoring could start the window inside an open `ByteFallback` marker run. The shorter
  decode is a genuine suffix at that moment, but the next marker re-renders the part left
  outside the window, so 65 markers decoded as 56 characters plus 9 replacements.
- `ByteLevel` and `ByteFallback` placed after a `Fuse` match a shape against the whole
  accumulated string, so appending a single token can rewrite the entire prefix.

That rate is itself the strongest argument for the point the issue raises: this logic would
be far more robust inside `tokenizers`, where the decoders' behaviour is readable directly
rather than inferred from their serialized form. Every one of those defects came from a
subtlety only visible in a `decode_chain` implementation. Filing here because that is where
the decode loops live; happy to move it upstream if you prefer.

## Note on prior art

`tokenizers` 0.22 ships `Tokenizer::decode_stream` / `DecodeStream`, which solves the same
problem with an unverified re-anchor: it drains to the previous prefix index and returns a
`DecodeStreamError::InvalidPrefix` when the assumption breaks, with no gate on the
decoder's context. This implementation differs in verifying the re-anchor before taking it,
degrading to whole-sequence decoding instead of erroring, and exposing the full text so
stop-sequence matching does not need a second buffer.

Closes #3782

### DIFF
--- a/candle-examples/src/token_output_stream.rs
+++ b/candle-examples/src/token_output_stream.rs
@@ -1,86 +1,69 @@
 use candle::Result;
+use candle_transformers::generation::IncrementalDecoder;
 
 /// This is a wrapper around a tokenizer to ensure that tokens can be returned to the user in a
 /// streaming way rather than having to wait for the full decoding.
+///
+/// It is a thin adapter over [`IncrementalDecoder`], which does the actual work; new code is
+/// better off using that directly.
 pub struct TokenOutputStream {
-    tokenizer: tokenizers::Tokenizer,
-    tokens: Vec<u32>,
-    prev_index: usize,
-    current_index: usize,
+    decoder: IncrementalDecoder,
 }
 
 impl TokenOutputStream {
     pub fn new(tokenizer: tokenizers::Tokenizer) -> Self {
         Self {
-            tokenizer,
-            tokens: Vec::new(),
-            prev_index: 0,
-            current_index: 0,
+            decoder: IncrementalDecoder::from_tokenizer(tokenizer),
         }
     }
 
     pub fn into_inner(self) -> tokenizers::Tokenizer {
-        self.tokenizer
+        self.decoder.into_tokenizer()
     }
 
-    fn decode(&self, tokens: &[u32]) -> Result<String> {
-        match self.tokenizer.decode(tokens, true) {
-            Ok(str) => Ok(str),
-            Err(err) => candle::bail!("cannot decode: {err}"),
-        }
-    }
-
-    // https://github.com/huggingface/text-generation-inference/blob/5ba53d44a18983a4de32d122f4cb46f4a17d9ef6/server/text_generation_server/models/model.py#L68
+    /// Returns the text that this token made available, if any. `None` means the token did not
+    /// settle any new text yet, e.g. it carries the first byte of a multi-byte character.
+    ///
+    /// The output is append-only: [`IncrementalDecoder::push`] only ever hands out text it has
+    /// shown cannot be rewritten, so what this returns is always safe to print or append. For a
+    /// tokenizer whose decoder can rewrite text it already produced, that means nothing is
+    /// streamed at all and the whole text arrives through [`Self::decode_rest`].
     pub fn next_token(&mut self, token: u32) -> Result<Option<String>> {
-        let prev_text = if self.tokens.is_empty() {
-            String::new()
-        } else {
-            let tokens = &self.tokens[self.prev_index..self.current_index];
-            self.decode(tokens)?
-        };
-        self.tokens.push(token);
-        let text = self.decode(&self.tokens[self.prev_index..])?;
-        if text.len() > prev_text.len() && text.chars().last().unwrap().is_alphanumeric() {
-            let text = text.split_at(prev_text.len());
-            self.prev_index = self.current_index;
-            self.current_index = self.tokens.len();
-            Ok(Some(text.1.to_string()))
-        } else {
+        let text = self.decoder.push(token)?;
+        if text.is_empty() {
             Ok(None)
+        } else {
+            Ok(Some(text.to_string()))
         }
     }
 
+    /// The text held back by [`Self::next_token`], to be printed once generation is over.
     pub fn decode_rest(&self) -> Result<Option<String>> {
-        let prev_text = if self.tokens.is_empty() {
-            String::new()
-        } else {
-            let tokens = &self.tokens[self.prev_index..self.current_index];
-            self.decode(tokens)?
-        };
-        let text = self.decode(&self.tokens[self.prev_index..])?;
-        if text.len() > prev_text.len() {
-            let text = text.split_at(prev_text.len());
-            Ok(Some(text.1.to_string()))
-        } else {
+        let rest = self.decoder.pending();
+        if rest.is_empty() {
             Ok(None)
+        } else {
+            Ok(Some(rest.to_string()))
         }
     }
 
     pub fn decode_all(&self) -> Result<String> {
-        self.decode(&self.tokens)
+        Ok(self.decoder.text().to_string())
     }
 
     pub fn get_token(&self, token_s: &str) -> Option<u32> {
-        self.tokenizer.get_vocab(true).get(token_s).copied()
+        self.decoder
+            .tokenizer()
+            .get_vocab(true)
+            .get(token_s)
+            .copied()
     }
 
     pub fn tokenizer(&self) -> &tokenizers::Tokenizer {
-        &self.tokenizer
+        self.decoder.tokenizer()
     }
 
     pub fn clear(&mut self) {
-        self.tokens.clear();
-        self.prev_index = 0;
-        self.current_index = 0;
+        self.decoder.clear()
     }
 }
--- a/candle-transformers/Cargo.toml
+++ b/candle-transformers/Cargo.toml
@@ -23,6 +23,7 @@ rayon = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 serde_plain = { workspace = true }
+tokenizers = { workspace = true }
 tracing = { workspace = true }
 
 [features]
--- a/candle-transformers/src/generation/incremental_decoder.rs
+++ b/candle-transformers/src/generation/incremental_decoder.rs
@@ -1,0 +1,651 @@
+//! Incremental detokenization for autoregressive decode loops.
+//!
+//! An autoregressive loop needs the text produced so far on every step, both to match stop
+//! sequences and to stream deltas back to the caller. Re-decoding the whole token sequence on
+//! every step is correct but quadratic in the generated length. Appending the decode of each
+//! token is linear but wrong: a BPE/SentencePiece decode is *not* the concatenation of its
+//! tokens' decodes. A multi-byte character can span several tokens (the prefix decodes to
+//! U+FFFD, and a later token retroactively replaces it), and decoding a sub-sequence can gain
+//! or lose a leading space.
+//!
+//! [`IncrementalDecoder`] keeps a bounded trailing window of tokens, re-decodes only that
+//! window, and appends the difference against the *previous* window decode, so any artifact
+//! caused by decoding a sub-sequence appears in both decodes and cancels out. The window is
+//! only moved forward when the shorter decode is provably a suffix of the current one, and
+//! only when the tokenizer's decoder has been shown to have a bounded context; otherwise the
+//! decoder degrades to whole-sequence decoding rather than corrupting the output.
+//!
+//! The invariant is that [`IncrementalDecoder::text`] equals a whole-sequence decode of every
+//! token pushed so far, at every step.
+//!
+//! ```no_run
+//! use candle_transformers::generation::IncrementalDecoder;
+//!
+//! # fn main() -> candle::Result<()> {
+//! # let tokenizer: tokenizers::Tokenizer = unimplemented!();
+//! # let next_token = |_: &[u32]| -> candle::Result<u32> { unimplemented!() };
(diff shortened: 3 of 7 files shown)


astorise:
Follow-up on the addendum in #3782: the decoder no longer forces a tokenizer clone.

`IncrementalDecoder::new(&tokenizer)` clones, and that clone deep-copies the vocabulary. A
continuous-batching loop needs one decoder per row, so it paid N vocabulary clones per batch,
plus N re-resolutions of the decoder-context property — the one thing that made the batched
path measurably worse than a hand-rolled decoder.

Since the decoder only ever reads the tokenizer, I went with the generic form rather than
adding `Arc` to the public API:

```rust
pub struct IncrementalDecoder<T = Tokenizer> { /* ... */ }

impl<T: Borrow<Tokenizer>> IncrementalDecoder<T> {
    pub fn from_tokenizer(tokenizer: T) -> Self;
}
```

So a caller supplies the tokenizer in whatever form it already holds it, and pays only for
that:

```rust
let owning    = IncrementalDecoder::from_tokenizer(tokenizer.clone());
let borrowing = IncrementalDecoder::from_tokenizer(&tokenizer);
let sharing   = IncrementalDecoder::from_tokenizer(Arc::clone(&tokenizer));
```

Notes:

- The default type parameter keeps `IncrementalDecoder` naming the owning form, so it stays
  usable as a plain struct field and nothing existing has to change. `new(&Tokenizer)` is
  unchanged for callers that do want the clone.
- `into_tokenizer` now returns `T`, handing back the `Arc` rather than a copy.
- The decoder is `Clone`, which covers the second half of the cost: cloning carries over the
  tokenizer handle *and* the already-resolved `DecoderContext`, so a batch resolves both once
  and clones a decoder per row instead of rebuilding one. Cloning mid-stream forks the stream,
  buffered state included, which is also what speculative decoding wants for a rollback point.

Three tests added (27 total): the owning, borrowing and shared forms produce identical deltas
step by step; a batch of four cloned decoders leaves `Arc::strong_count` at 6 and returns a
tokenizer that is `ptr_eq` to the shared one, i.e. the vocabulary is never copied; and a clone
taken with a multi-byte character still open diverges correctly from its parent.

astorise:
Follow-up for the stop-criteria and finish-reason half of the generation loop is available in https://github.com/astorise/candle/pull/41. It is based on this PR's current head and resolves https://github.com/huggingface/candle/issues/3810. It adds StopCriteria / FinishReason, including split stop-sequence, UTF-8 boundary, earliest-match, and EOS-at-token-budget-boundary tests.

astorise:
Closing — these two PRs are divergent branches of the same change. Closing both; this work is moving to a separate crate.

PR #3790, opened 2026-07-27
Use fancy-regex instead of onig in candle-core (drops the C dependency)

fbbbkaya:
`candle-core` pulls `tokenizers` with the `onig` feature, which builds the Oniguruma **C** library through `onig_sys`. That forces a C toolchain on every downstream user, even though `candle-core` needs a tokenizer in exactly one place: `quantized/tokenizer.rs`.

Cargo features are additive, so downstream crates cannot turn `onig` off. The only escapes today are `[patch.crates-io]` or vendoring the whole candle stack — both heavy for what is an unused code path in most builds.

### Why `fancy-regex` is the right swap

- `tokenizers` already treats `onig` and `fancy-regex` as interchangeable backends — `src/utils/mod.rs` picks one or the other behind the same `SysRegex` type.
- The wasm target in this repo already builds without `onig`, so there is precedent.
- Plain `regex` would **not** work: the GPT-2 splitting pattern uses a `\s+(?!\S)` lookahead, which `regex` does not support by design. `fancy-regex` does.

### Verification

```
cargo build -p candle-core                    # ok
cargo test  -p candle-core --lib              # 17 passed
cargo tree  -p candle-core -e normal,build    # no onig, onig_sys or cc
```

Tokenization is unchanged. Token IDs were compared against Python HuggingFace `tokenizers` 0.23.1 (which uses onig) over a fixed corpus covering the GPT-2 splitting pattern, Unicode letter/number classes, the `\s+(?!\S)` lookahead, Turkish dotted/dotless `İ i I ı`, digits, punctuation and whitespace runs — **identical IDs** for both `whisper-large-v3-turbo` and Qwen3 tokenizers.

As a side effect this also fixes the long-standing Windows build pain with `onig_sys` (huggingface/tokenizers#1581).


AlpineVibrations:
good one.

ivarflakstad:
I just merged #3952
Added you as co-author in the commit message because your solution was identical, I just didn't see this was open.

PR #3792, opened 2026-07-27
onnx: add Softplus, NonZero, GatherND, ConvTranspose, RandomNormalLike and Pad constant mode

fbbbkaya:
These are the operators a VITS/Piper text-to-speech model needs. With them `candle-onnx` runs Piper checkpoints end to end; without them the graph fails at load with `unsupported op_type ... for op ...`.

| Op | Notes |
|---|---|
| `Softplus` | `ln(1 + e^x)`, used by the duration predictor. Written as `max(x,0) + ln(1 + e^-|x|)` so it does not overflow for large `x`. |
| `NonZero` | Returns coordinates of non-zero elements, shape `[rank, count]` as the spec requires (one **column** per element). Used for masking in the flow layers. |
| `GatherND` | `batch_dims=0` only, which is what the model uses; other values bail out. Negative indices are handled; out-of-range indices produce an error rather than a silently wrong slice. |
| `ConvTranspose` | 1D only — that is the vocoder's upsampling path. 2D bails out explicitly instead of producing a wrong result. Optional bias is broadcast over the channel axis. |
| `RandomNormalLike` | The duration predictor samples noise to decide speech rhythm. `RandomNormal` already existed; only the `Like` variant was missing. |
| `Pad` `"constant"` | The mode was rejected even though it is the ONNX **default**, and the optional third input carrying the constant value was refused outright (`expected 2` inputs). Negative padding is rejected explicitly. |

### Verification

```
cargo build --manifest-path candle-onnx/Cargo.toml   # ok
cargo fmt  --manifest-path candle-onnx/Cargo.toml -- --check   # clean
```

The resulting graph was run against `tr_TR-dfki-medium` (Piper, VITS architecture) and produced audio matching the reference implementation. The same runtime also runs YOLOv8 detection graphs (14 op types, 300 nodes, none unsupported).

Happy to split this into separate PRs per operator if that is easier to review.

### DIFF
--- a/candle-onnx/src/eval.rs
+++ b/candle-onnx/src/eval.rs
@@ -1174,13 +1174,21 @@ fn simple_eval_(
                 let mode = get_attr_opt(node, "mode")?.unwrap_or("constant");
                 let data = get(&node.input[0])?;
                 let pads = get(&node.input[1])?;
-                if node.input.len() > 2 {
+                // Opset 11+ allows an optional third input holding the value
+                // used by the "constant" mode, and an optional fourth input
+                // selecting the axes. Only the former is handled here.
+                if node.input.len() > 3 {
                     bail!(
-                        "unsupported number of inputs {} for Pad node {:?}, expected 2",
+                        "unsupported number of inputs {} for Pad node {:?}, expected 2 or 3",
                         node.input.len(),
                         node.name
                     );
                 }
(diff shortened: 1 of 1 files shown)


cryptomilk:
> Happy to split this into separate PRs per operator if that is easier to review.

I guess the answer to this is always yes, please do ;-)

PR #3794, opened 2026-07-28
Add a GGUF architecture dispatcher to `candle-transformers`

### LINKED ISSUE #3783: `candle-transformers`: dispatch GGUF checkpoints by `general.architecture`
### Problem

`candle-transformers` ships a growing set of `quantized_*` modules —
`quantized_llama`, `quantized_qwen2`, `quantized_qwen3`, `quantized_qwen3_moe`,
`quantized_gemma3`, `quantized_phi3`, `quantized_glm4`, … — but no way to load
"whatever GGUF this is". Every consumer writes the same dispatcher.

Two things make that dispatcher more annoying than it should be:

**Signatures diverge.** They are close enough to look uniform and different
enough to need a wrapper enum:

```rust
quantized_llama::ModelWeights::from_gguf(ct, reader, device)
quantized_qwen2::ModelWeights::from_gguf(ct, reader, device)
quantized_qwen3::ModelWeights::from_gguf(ct, reader, device)
quantized_gemma3::ModelWeights::from_gguf(ct, reader, device)
quantized_qwen3_moe::GGUFQWenMoE::from_gguf(ct, reader, device, dtype)   // + dtype
quantized_phi3::ModelWeights::from_gguf(use_flash_attn, ct, reader, device) // leading flag
```

The MoE type is also named differently from every sibling (`GGUFQWenMoE` vs
`ModelWeights`), which is easy to trip over.

**Metadata keys are architecture-prefixed.** `llama.context_length` vs
`qwen3.context_length` vs `phi3.context_length`. A consumer that hardcodes the
`llama.` prefix silently falls back to a default context window for every other
family — no error, just a model that rejects prompts it should accept.

### Proposed API

```rust
pub trait QuantizedLm {
    fn forward(&mut self, input: &Tensor, index_pos: usize) -> Result<Tensor>;
}

/// Select and construct the backend registered for `ct`'s `general.architecture`.
    ct: gguf_file::Content,
    reader: &mut R,
    device: &Device,
) -> Result<Box<dyn QuantizedLm>>;
```

Every family already exposes `forward(&mut self, input, index_pos) -> Result<Tensor>`,
so the trait costs nothing to implement. The per-family extras (`dtype`,
`use_flash_attn`) can move into an options struct with sane defaults.

An accompanying `context_length(&Content) -> Option<usize>` that reads the right
namespace would remove the second footgun.

### Related

`quantized_qwen3_moe` cannot currently be used outside CUDA:
`candle_nn::moe_gemm_gguf` returns "only implemented for the cuda backend" on
CPU and Metal, and the prefill kernel accepts only F16/BF16 working dtypes. We
had to unwire that family after advertising it, because nothing surfaces those
constraints before the first expert layer runs. A dispatcher would be the
natural place to reject an unusable (family, device, dtype) combination at load.

astorise:
Closes #3783.

`candle-transformers` ships one `quantized_*` module per model family but no way to
load "whatever GGUF this is", so every consumer ends up writing the same dispatcher.
This adds `models::quantized_lm`, which does it once.

## API

```rust
pub trait QuantizedLm {
    fn forward(&mut self, input: &Tensor, index_pos: usize) -> Result<Tensor>;
}

    ct: gguf_file::Content,
    reader: &mut R,
    device: &Device,
) -> Result<Box<dyn QuantizedLm>>;

pub fn from_gguf_with<R: Read + Seek>(
    ct: gguf_file::Content,
    reader: &mut R,
    device: &Device,
    options: &Options,
) -> Result<Box<dyn QuantizedLm>>;

pub fn context_length(ct: &gguf_file::Content) -> Option<usize>;
pub fn arch_metadata<'a>(ct: &'a gguf_file::Content, suffix: &str) -> Option<&'a gguf_file::Value>;
```

Usage:

```rust
let content = gguf_file::Content::read(&mut file)?;
let ctx = quantized_lm::context_length(&content);
let mut model = quantized_lm::from_gguf(content, &mut file, &device)?;
let logits = model.forward(&input, 0)?;
```

## Divergent signatures

The trait costs nothing to implement: every dispatched family already exposes
`forward(&mut self, input, index_pos) -> Result<Tensor>` inherently, so the impls are
one-line delegations. The constructor extras move into an `Options` struct rather than
into the trait:

```rust
pub struct Options {
    pub use_flash_attn: bool,   // read by quantized_phi3 only
    pub dtype: Option<DType>,   // read by quantized_glm4 and quantized_qwen3_moe
}
```

`dtype: None` resolves to BF16 on CUDA and Metal, F32 elsewhere — the same choice the
`quantized-glm4` example already makes by hand. `use_flash_attn` stays inert for the
families that don't read it, so a generic loader can set it once and load anything.

## Dispatch table

| `general.architecture` | backend |
| --- | --- |
| `llama` | `quantized_llama::ModelWeights` |
| `gemma`, `gemma2`, `gemma3`, `gemma-embedding` | `quantized_gemma3::ModelWeights` |
| `glm4` | `quantized_glm4::ModelWeights` |
| `lfm2` | `quantized_lfm2::ModelWeights` |
| `phi2` | `quantized_phi::ModelWeights` |
| `phi3` | `quantized_phi3::ModelWeights` |
| `qwen2` | `quantized_qwen2::ModelWeights` |
| `qwen3` | `quantized_qwen3::ModelWeights` |
| `qwen3moe` | `quantized_qwen3_moe::GGUFQWenMoE` |

The four gemma names share one entry point because `quantized_gemma3::from_gguf` already
probes those metadata prefixes itself. `Architecture` and `SUPPORTED_ARCHITECTURES` are
public so callers can probe support without opening a file, and an unknown architecture
fails with the supported list rather than a metadata-key error.

## Architecture-prefixed metadata

`arch_metadata` reads through the file's own namespace instead of a hardcoded prefix, and
`context_length` is built on it. A Qwen3 file with no `qwen3.context_length` now yields
`None` — a caller falls back to its own default instead of silently picking up a
`llama.context_length` that happens to be present. There is a test pinning exactly that.

## Pre-flight validation

Both checks run before a single tensor is read, addressing the "Related" section of the
issue:

- **`qwen3moe` off CUDA.** `candle_nn::moe_gemm_gguf` is CUDA-only and its prefill kernel
  casts the input to F16 or BF16, so the family is rejected on a non-CUDA device or a
  non-F16/BF16 working dtype. Previously this only surfaced once the first expert layer
  ran.
- **Unhonorable flash-attention.** `quantized_phi3`'s `flash_attn` shim is
  `unimplemented!()` without the `flash-attn` feature, so `use_flash_attn: true` on such a
  build used to panic at the first attention layer. It is now a load-time error, and
  requires a CUDA device.

Both are exposed as `Architecture::check_device_support` and
`Architecture::check_flash_attn_support` so a caller can pre-flight a combination without
attempting a load.

## Tests

13 unit tests, all CPU-only and requiring no weights — they build `gguf_file::Content`
in memory. They cover name resolution and the gemma aliases, the missing- and
unknown-architecture errors, `context_length` namespacing and integer-width upcasting,
and that both pre-flight checks fire before any read. One further test is gated on an
available CUDA device and no-ops otherwise; it covers the `qwen3moe` accept path, its
dtype rejection, and the BF16 default, none of which a CPU-only run can reach.

## Notes for reviewers

- Additive only: no existing module is modified, and no existing signature changes. The
  per-family constructors stay public and usable directly.
- `QuantizedLm` requires `Send`, so a loaded model can be held as
  `Mutex<Box<dyn QuantizedLm>>` and handed to a worker pool. All nine backends already
  satisfied it — type erasure was the only thing dropping it, and `Box<dyn QuantizedLm>`
  cannot be widened to `Box<dyn QuantizedLm + Send>` after the fact. `Sync` is not
  required, since `forward` takes `&mut self`.
- `Architecture::name()` is deliberately not a round-trip of `from_name` for the aliased
  gemma family — a `gemma2` file resolves to `Architecture::Gemma3`, whose name is
  `"gemma3"`. Documented on the method.
- No end-to-end load of a real GGUF is covered, as there are no checkpoint fixtures in
  the repo for these families. Happy to add an example wiring if useful.

### DIFF
--- a/candle-transformers/src/models/mod.rs
+++ b/candle-transformers/src/models/mod.rs
@@ -10,6 +10,8 @@
 //!  - Computer vision models: [`dinov2`], [`convmixer`], [`efficientnet`], ...
 //!  
 //! Some of the models also have quantized variants, e.g.  [`quantized_blip`], [`quantized_llama`] and  [`quantized_qwen2`].
+//! [`quantized_lm`] loads any of the quantized language models from a GGUF file by dispatching on
+//! its `general.architecture`.
 //!
 //! The implementations aim to be readable while maintaining good performance. For more information
 //! on each model see the model's module docs in the links below.
@@ -92,6 +94,7 @@ pub mod quantized_glm4;
 pub mod quantized_lfm2;
 pub mod quantized_llama;
 pub mod quantized_llama2_c;
+pub mod quantized_lm;
 pub mod quantized_metavoice;
 pub mod quantized_mistral;
 pub mod quantized_mixformer;
--- a/candle-transformers/src/models/quantized_lm.rs
+++ b/candle-transformers/src/models/quantized_lm.rs
@@ -160,12 +160,49 @@ impl Architecture {
         }
         Ok(())
     }
+
+    /// Whether this backend reads [`Options::use_flash_attn`].
+    ///
+    /// The flag is ignored by every other family, so a generic loader can leave it set.
+    pub fn supports_flash_attn(&self) -> bool {
+        *self == Self::Phi3
+    }
+
+    /// Reject a flash-attention request this build or device cannot honor.
+    ///
+    /// Only checked for the backends that read the flag; see
+    /// [`Architecture::supports_flash_attn`]. `quantized_phi3` calls a `flash_attn` shim that is
+    /// `unimplemented!()` without the `flash-attn` feature, so an unhonorable request would
+    /// otherwise panic at the first attention layer rather than fail at load.
+    pub fn check_flash_attn_support(&self, device: &Device, use_flash_attn: bool) -> Result<()> {
+        if !use_flash_attn || !self.supports_flash_attn() {
+            return Ok(());
+        }
+        if !cfg!(feature = "flash-attn") {
+            candle::bail!(
+                "the {} backend cannot use flash-attention: build candle-transformers with \
+                 --features flash-attn",
+                self.name()
+            )
+        }
+        if !device.is_cuda() {
+            candle::bail!(
+                "the {} backend can only use flash-attention on a cuda device",
+                self.name()
+            )
+        }
+        Ok(())
+    }
 }
 
 /// Per-family construction extras, with defaults that suit every backend.
 #[derive(Debug, Clone, Copy, Default)]
 pub struct Options {
-    /// Use flash-attention where the backend supports it. Only `phi3` reads this today.
+    /// Use flash-attention where the backend supports it. Only `phi3` reads this today; the
+    /// other families ignore it, so a generic loader can leave it set.
+    ///
+    /// Setting it for a backend that does read it requires the `flash-attn` feature and a CUDA
+    /// device — see [`Architecture::check_flash_attn_support`].
     pub use_flash_attn: bool,
     /// Working dtype for the backends that take one (`glm4`, `qwen3moe`).
     ///
@@ -253,6 +290,7 @@ pub fn from_gguf_with<R: Read + Seek>(
     let arch = Architecture::from_content(&ct)?;
     let dtype = options.dtype_for(device);
     arch.check_device_support(device, dtype)?;
+    arch.check_flash_attn_support(device, options.use_flash_attn)?;
     let model: Box<dyn QuantizedLm> = match arch {
         Architecture::Llama => Box::new(quantized_llama::ModelWeights::from_gguf(
             ct, reader, device,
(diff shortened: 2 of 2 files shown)


astorise:
Good catch, and thank you for the precise diagnosis — applied in `dd0977d`.

```rust
pub trait QuantizedLm: Send {
    fn forward(&mut self, input: &Tensor, index_pos: usize) -> Result<Tensor>;
}
```

Your reasoning holds exactly: **no implementation needed changing**. All nine backends
compile unmodified under the new supertrait, which is the proof that they were already
`Send` and that only type erasure was losing it. The bound costs nothing today and buys
back the use case the module exists for — if a consumer still has to keep the per-family
enum in order to hold the model, the dispatcher has not actually deleted it.

Agreed on `Sync` as well, and not just because `&mut self` serializes access: requiring it
would also rule out a future backend using interior mutability for, say, a scratch buffer,
with no caller benefit.

I added a type-level test alongside it:

```rust
fn assert_send<T: Send>() {}
assert_send::<Box<dyn QuantizedLm>>();
assert_send::<std::sync::Mutex<Box<dyn QuantizedLm>>>();
```

The interesting failure mode is not today's code but a backend added later that is not
`Send` — without this, the breakage surfaces in a consumer's worker pool rather than here.
The rationale is also on the trait's doc comment, since a bound with no visible use in the
crate itself is the kind of thing a later change removes as dead weight.

The caveat about this in the PR description is now stale and no longer applies.

astorise:

PR #3797, opened 2026-07-29
Metal MoE: indexed/routed matmul dispatch (mm_id, mv_id, chunked mm_id)

alytaphoenix:
Adds Metal kernel support for indexed/routed matmul (ggml's `mul_mat_id`), used by Mixture-of-Experts models to dispatch each token to its selected experts' weights in one batched call instead of per-expert loops. Includes three follow-on prefill throughput optimizations on top of the initial wiring, since the first working version measured 97.9x slower than llama.cpp on MoE prefill and each phase closes part of that gap on real hardware, not just synthetically.

## What

- `call_quantized_matmul_mm_id` (`candle-metal-kernels`): matrix-*matrix* dispatch for `kernel_mul_mm_id_*` — one weight buffer holding all experts stacked on their leading dim, an `ids` routing table mapping each (token, selected-expert) pair to a row of that buffer, and a dispatch/threadgroup-memory sizing scheme mirrored from ggml's own Metal backend (`ggml_metal_encode_node`'s `GGML_OP_MUL_MAT_ID` case).
- `call_quantized_matmul_mv_id` + `mv_id_eligible`: matrix-*vector* dispatch for the decode case (`batch == 1`), which ggml routes to a separate, faster kernel below a device- and dtype-dependent threadgroup-memory ceiling. `mv_id_eligible` is the single source of truth for which dtypes qualify and the minimum contraction-dim each needs, so eligibility can't drift out of sync with the per-dtype tuning table.
- `call_quantized_matmul_mm_id_chunked` + `mm_id_max_nei1`: `mm_id`'s rowids scratch is sized off the whole-batch token count, so a large enough prefill batch overflows the device's threadgroup-memory budget regardless of tuning. The chunked wrapper splits such a batch into sequential dispatches of at most `mm_id_max_nei1` tokens each, on the same command encoder.
- **Row discovery parallelized across the threadgroup's 128 threads** (was a single serial thread scanning the whole routing table) — 10.7x prefill on its own. Ordering matters here: an earlier atomic-based version permuted slot assignment and made greedy decoding nondeterministic (three identical prompts, three different completions against a bit-stable baseline); the shipped version uses contiguous per-thread ranges that reconstruct the original loop's exact order.
- **Doomed token tiles skipped before the matmul** — a token tile only ever holds `nei1*nei0/ne02` of an expert's rows (24 of 768 at top-8-of-256), so `kernel_mul_mm_id_impl`'s own `r1 * BLOCK_SIZE_N >= ne1` guard, already computing the exact per-tile count, is hoisted to bail before wasted matmul work.
- **Per-expert row counts precomputed once per dispatch** (`kernel_mm_id_zero_counts` + `kernel_mm_id_expert_counts`, a small atomic histogram over the routing table) so `kernel_mul_mm_id` can skip the row-discovery scan entirely for a token tile that holds none of its expert's rows, instead of running the O(nei0\*nei1) scan to find out — that scan was the single largest remaining prefill cost by measured ablation (3.61s of a 9.42s prefill on the profiled case). Counts are computed **per chunk**, not once for the whole batch: the guard's `tgpig.x` is chunk-local, so a chunk-local count is what makes the skip actually fire on chunks after the first. Deliberately counts only, not ordered row lists — producing those in scan order would need a stable counting sort, and ordering is exactly what caused the earlier nondeterminism bug; counts are order-independent and safe with a plain atomic histogram.
- `QTensor::indexed_moe_forward`'s Metal arm (`candle-core`), wiring all of the above together and routing `FusedMoeGGUF::forward`'s Metal path through it.

## Why

Without this, Metal has no way to dispatch a quantized indexed matmul at all — `FusedMoeGGUF`'s Metal path had no MoE-specific kernel to call. This adds the three dispatch shapes ggml itself distinguishes (matrix-matrix for prefill, matrix-vector for decode, chunked matrix-matrix for oversized prefill batches), plus the throughput work needed to make matrix-matrix prefill actually fast rather than only correct — a naive first version measured 97.9x slower than llama.cpp, entirely localized to this one kernel (a dense model on the same hardware was only 2.2x off), which is what motivated digging further rather than shipping the initial wiring alone.

## Measured (M4 Pro, `Qwen3.6-35B-A3B` UD-Q4_K_M, 837-token prompt)

| stage | prefill | vs. llama.cpp (699.89 tok/s) |
|---|---|---|
| naive `mm_id` wiring | 7.38 tok/s | 97.9x slower |
| + parallel row discovery | 78.88 tok/s | 9.2x slower |
| + skip doomed token tiles | 96.45 tok/s | 7.5x slower |
| + counts-only precompute | **151.9 tok/s** | **4.6x slower** |

A `dispatchThreadgroupsWithIndirectBuffer` lever was profiled and ruled out along the way: launch overhead for the 49,152 threadgroups this shape dispatches measured at 60ms of a 9.42s prefill (0.6%), not the dominant cost it looked like on paper.

## Testing

- `candle-metal-kernels/src/tests.rs`: pipeline-load de-risk spikes for every `kernel_mul_mm_id_*`/`kernel_mul_mv_id_*` kernel name; bit-exact-equality differential tests between the chunked and unchunked `mm_id` wrapper (single-chunk, multi-chunk-via-forced-low-threshold, and a case against the real device's own threshold); a `mv_id`-vs-`mm_id` differential test (single-token and broadcast shapes) using `mm_id`'s own validated output as the correctness oracle.
- `candle-core/tests/quantized_tests.rs`: `indexed_moe_forward` exercised through the real public API with real quantized weights — the decode-routes-to-`mv_id` path for each of `mv_id_eligible`'s four dtypes, and the chunked-prefill path with a batch large enough to exceed the device's real threadgroup-memory ceiling.
- Two live differential tests (not in this repo — in the downstream project this was built for) replay real prompts against a real model on real Metal hardware and assert byte-identical, deterministic output against a `llama-server /completion`-verified reference: one at the greedy-decode boundary, one specifically forcing multi-chunk prefill (837 tokens against a 768-token chunk threshold, so it exercises a ragged final chunk). Both pass on every phase above, including the counts-only precompute — this kernel fails *fluently* rather than loudly on a routing bug, so a differential against a real model was treated as non-negotiable for every change, not just the initial wiring.
- `cargo test -p candle-core --features metal --test quantized_tests` (74 passed) and `cargo test` in `candle-metal-kernels` (76 passed) both green; `cargo clippy`/`cargo fmt --check` clean across `candle-core` and `candle-metal-kernels`.

Independent of #3785 (mmap-backed Metal buffers) — every `QMetalStorage` here always owns its buffer outright, so this doesn't need that PR's `offset`/`length` fields and can merge in either order.

### DIFF
--- a/candle-core/src/quantized/metal.rs
+++ b/candle-core/src/quantized/metal.rs
@@ -37,6 +37,20 @@ impl QMetalStorage {
         &self.buffer
     }
 
+    /// Converts an element-stride array to bytes for this storage's
+    /// (possibly sub-byte-per-element, block-quantized) dtype. Shared by
+    /// `fwd` and `indexed_moe_forward` so the two matmul paths can't
+    /// silently diverge on how quantized weight strides are computed.
+    fn quantized_byte_strides(&self, stride: &[usize]) -> Vec<usize> {
+        stride
+            .iter()
+            .map(|x| {
+                (*x as f32 * (self.dtype.type_size() as f32 / self.dtype.block_size() as f32))
+                    as usize
+            })
+            .collect()
+    }
+
     pub fn dequantize(&self, elem_count: usize) -> Result<MetalStorage> {
         use crate::quantized::k_quants::GgmlType;
 
@@ -395,14 +409,7 @@ impl QMetalStorage {
         let src0_l = crate::Layout::contiguous(
             [vec![1; 4 - self_shape.rank()], self_shape.dims().to_vec()].concat(),
         );
-        let src0_stride = src0_l
-            .stride()
-            .iter()
-            .map(|x| {
-                (*x as f32 * (self.dtype.type_size() as f32 / self.dtype.block_size() as f32))
-                    as usize
-            })
-            .collect::<Vec<_>>();
+        let src0_stride = self.quantized_byte_strides(src0_l.stride());
 
         if src_shape.rank() > 4 {
             crate::bail!("weight rank ({}) must be <= 4", src_shape.rank())
@@ -463,7 +470,16 @@ impl QMetalStorage {
         if !ids_l.is_contiguous() {
             crate::bail!("indexed_moe_forward ids is not contiguous {ids_l:?}")
         }
-        assert_eq!(input.dtype(), DType::F32);
+        if input.dtype() != DType::F32 {
+            crate::bail!("indexed_moe_forward input must be F32, got {:?}", input.dtype())
+        }
+        // The kernel unconditionally reads `ids` as raw int32_t regardless of
+        // the Rust-side dtype tag; U32 is bit-compatible for the small
+        // non-negative expert indices this carries, but anything else would
+        // silently misread the buffer at the byte-stride computed below.
+        if ids.dtype() != DType::U32 {
+            crate::bail!("indexed_moe_forward ids must be U32, got {:?}", ids.dtype())
+        }
 
         let (_num_experts, n, k) = self_shape.dims3()?;
         let in_shape = input_l.shape();
@@ -498,14 +514,7 @@ impl QMetalStorage {
         let encoder = device.command_encoder()?;
 
         let src0_l = crate::Layout::contiguous(self_shape.dims());
-        let src0_stride = src0_l
-            .stride()
-            .iter()
-            .map(|x| {
-                (*x as f32 * (self.dtype.type_size() as f32 / self.dtype.block_size() as f32))
-                    as usize
-            })
-            .collect::<Vec<_>>();
+        let src0_stride = self.quantized_byte_strides(src0_l.stride());
 
         let input_stride = input_l
             .stride()
@@ -526,7 +535,7 @@ impl QMetalStorage {
             self_shape.dims(),
             &src0_stride,
             &self.buffer,
-            self.offset,
+            0, // this storage always owns its buffer outright, so its offset is always 0
             in_shape.dims(),
             &input_stride,
             input.buffer(),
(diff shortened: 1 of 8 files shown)


alytaphoenix:
Updated (rebased onto current `main`, no conflicts against the base wiring) to include three follow-on prefill throughput phases: parallel row discovery, skipping doomed token tiles, and a per-chunk expert-count precompute that lets the kernel skip row discovery entirely on tiles with no work. Measured on real hardware against a real model: prefill went from 7.38 tok/s (97.9x slower than llama.cpp) to 151.9 tok/s (4.6x slower) — see the updated PR description for the full table and methodology. All correctness claims are backed by live differential tests against a llama.cpp-verified reference, not just synthetic kernel tests.

alytaphoenix:
Pushed an update: extends `indexed_moe_forward`'s `mv_id` fast path from `batch == 1` to `batch <= 8`. The mat-vec kernel's dispatch depth (`nei0 * nei1`) is proportional to real work regardless of batch size, while `mm_id`'s (`ne02`) is effectively fixed — so small multi-token batches (e.g. a speculative-decode verify step) were paying `mm_id`'s per-expert-count dispatch tax against very little real work. Verified batch-general correctness via a bit-exact self-differential against N independent batch-1 calls, plus a tolerance-based cross-check against `mm_id`, both at production expert/top-k shape. A timing table shows `mv_id` beating `mm_id` by 10-20% through batch 16 before crossing over past 32; 8 keeps a safety margin below that crossover.

PR #3811, opened 2026-07-31
Add IncrementalDecoder and stop criteria to candle-transformers::generation

### LINKED ISSUE #3782: `candle-transformers::generation`: provide a correct incremental detokenizer

### LINKED ISSUE #3810: `candle-transformers::generation`: provide stop criteria and a finish reason
## Problem

`generation` covers producing a token — `Sampling`, `LogitsProcessor`,
`apply_frequency_presence_penalty`, and `IncrementalDecoder` (#3789). It has
nothing for the other half of a decode loop: deciding when to stop, and saying
why it stopped.

So every serving loop writes that part itself, and it is harder than it looks.
A stop sequence is a property of the *text*, not of the token ids, so it has to
be scanned over incrementally decoded output; it can straddle a token boundary,
so a suffix must be withheld until it is known not to be the start of one; what
is emitted must land on a char boundary; and with several stop sequences the
earliest match has to win, not the first one checked.

The finish reason is the part that is quietly wrong most often. Inferring it
from the token count alone — `completion_tokens >= max_new_tokens` means
`length` — is wrong exactly at the boundary, because a generation whose EOS or
stop sequence lands on the budget's last token has both spent its budget and
ended normally. Reporting `length` there tells an OpenAI-compatible client that
a complete answer was truncated, which is the same misreport as the opposite
error, in the other direction.

This is not hypothetical. In Tachyon-Mesh four decode loops — ordinary,
speculative, continuous-batch, and a batch-native LoRA loop — each grew their
own copy of this logic, and a review pass found the same class of defect in
every one of them: the boundary case above in all four, and in two of them a
withheld separator that was either dropped from the stream or emitted twice,
because "bytes accounted for" and "bytes actually sent" had been conflated into
one counter. Each loop got it wrong differently, which is the signature of
logic that should exist once.

## What a correct implementation needs

1. **Scan text, not ids.** The needle is a string; tokenization is irrelevant to
   whether it matched.
2. **Withhold a suffix** at least as long as the longest stop sequence, so a
   match split across two tokens is still found.
3. **Emit only to a char boundary**, since the held-back window is measured in
   bytes and multi-byte characters straddle it.
4. **Earliest match wins** across all configured sequences — checking in
   configuration order truncates at the wrong place.
5. **Distinguish "the model ended it" from "the budget ended it"**, and let the
   former win when both are true on the same token.
6. **Report nothing rather than guess.** A loop that stopped for a reason it
   cannot name (a wall-clock deadline, a departed consumer) should leave the
   reason absent instead of picking the closest label.

## Proposed API

```rust
pub struct StopCriteria { /* … */ }

impl StopCriteria {
    pub fn new(stop: impl IntoIterator<Item = String>, eos: Vec<u32>) -> Self;

    /// Bytes to withhold: the longest stop sequence, minus one.
    pub fn hold(&self) -> usize;

    /// Byte offset of the earliest match in `text`, if any.
    pub fn matched(&self, text: &str) -> Option<usize>;

    /// Largest char-boundary offset safe to emit from `text` right now.
    pub fn safe_emit_end(&self, text: &str) -> usize;
}

#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum FinishReason {
    /// EOS, or a configured stop sequence.
    Stop,
    /// The token budget ran out first.
    Length,
}
```

Used with `IncrementalDecoder`, a loop becomes: push the token, ask
`matched()`, emit up to `safe_emit_end()`, and report `Stop` whenever the model
ended the sequence itself — with `Length` reserved for the case where it did
not.

## Note

This pairs with #3789: `IncrementalDecoder` gives the text, and this decides
what to do with it. The two were built together and it may be worth reviewing
them that way.

I am happy to submit the PR, including the boundary cases above as tests —
particularly EOS landing on the budget's last token, and a stop sequence split
across two tokens.

astorise:
This PR provides the missing generation-loop building blocks in
`candle-transformers::generation`:

- `IncrementalDecoder`: incrementally detokenizes generated tokens while preserving
  whole-sequence decoding correctness and append-only streaming.
- `StopCriteria`: matches text stop sequences, recognises EOS token ids, and computes
  a UTF-8-safe streaming boundary.
- `FinishReason`: distinguishes a model-controlled stop (`EOS` or a configured stop
  sequence) from a token-budget exhaustion.

Stop sequences are matched against decoded text rather than token ids. The API retains
the longest possible stop prefix, emits only at character boundaries, and selects the
earliest matching sequence irrespective of configuration order.

`finish_reason` gives `Stop` priority over `Length` when an EOS token or stop sequence
lands exactly on the final allowed token. Loops that stop for an unrelated reason can
continue to report no finish reason.

The incremental decoder is generic over its tokenizer ownership, so callers can borrow
or share a tokenizer across batch rows without cloning its vocabulary.

Tests cover:

- multi-byte characters and decoder rewrites across token boundaries;
- append-only incremental streaming and bounded decoder windows;
- split stop sequences;
- UTF-8-safe withheld suffixes;
- earliest-match selection across several stop sequences;
- EOS and stop sequences on the token-budget boundary.

Closes #3782
Closes #3810

### DIFF
--- a/candle-examples/src/token_output_stream.rs
+++ b/candle-examples/src/token_output_stream.rs
@@ -1,86 +1,69 @@
 use candle::Result;
+use candle_transformers::generation::IncrementalDecoder;
 
 /// This is a wrapper around a tokenizer to ensure that tokens can be returned to the user in a
 /// streaming way rather than having to wait for the full decoding.
+///
+/// It is a thin adapter over [`IncrementalDecoder`], which does the actual work; new code is
+/// better off using that directly.
 pub struct TokenOutputStream {
-    tokenizer: tokenizers::Tokenizer,
-    tokens: Vec<u32>,
-    prev_index: usize,
-    current_index: usize,
+    decoder: IncrementalDecoder,
 }
 
 impl TokenOutputStream {
     pub fn new(tokenizer: tokenizers::Tokenizer) -> Self {
         Self {
-            tokenizer,
-            tokens: Vec::new(),
-            prev_index: 0,
-            current_index: 0,
+            decoder: IncrementalDecoder::from_tokenizer(tokenizer),
         }
     }
 
     pub fn into_inner(self) -> tokenizers::Tokenizer {
-        self.tokenizer
+        self.decoder.into_tokenizer()
     }
 
-    fn decode(&self, tokens: &[u32]) -> Result<String> {
-        match self.tokenizer.decode(tokens, true) {
-            Ok(str) => Ok(str),
-            Err(err) => candle::bail!("cannot decode: {err}"),
-        }
-    }
-
-    // https://github.com/huggingface/text-generation-inference/blob/5ba53d44a18983a4de32d122f4cb46f4a17d9ef6/server/text_generation_server/models/model.py#L68
+    /// Returns the text that this token made available, if any. `None` means the token did not
+    /// settle any new text yet, e.g. it carries the first byte of a multi-byte character.
+    ///
+    /// The output is append-only: [`IncrementalDecoder::push`] only ever hands out text it has
+    /// shown cannot be rewritten, so what this returns is always safe to print or append. For a
+    /// tokenizer whose decoder can rewrite text it already produced, that means nothing is
+    /// streamed at all and the whole text arrives through [`Self::decode_rest`].
     pub fn next_token(&mut self, token: u32) -> Result<Option<String>> {
-        let prev_text = if self.tokens.is_empty() {
-            String::new()
-        } else {
-            let tokens = &self.tokens[self.prev_index..self.current_index];
-            self.decode(tokens)?
-        };
-        self.tokens.push(token);
-        let text = self.decode(&self.tokens[self.prev_index..])?;
-        if text.len() > prev_text.len() && text.chars().last().unwrap().is_alphanumeric() {
-            let text = text.split_at(prev_text.len());
-            self.prev_index = self.current_index;
-            self.current_index = self.tokens.len();
-            Ok(Some(text.1.to_string()))
-        } else {
+        let text = self.decoder.push(token)?;
+        if text.is_empty() {
             Ok(None)
+        } else {
+            Ok(Some(text.to_string()))
         }
     }
 
+    /// The text held back by [`Self::next_token`], to be printed once generation is over.
     pub fn decode_rest(&self) -> Result<Option<String>> {
-        let prev_text = if self.tokens.is_empty() {
-            String::new()
-        } else {
-            let tokens = &self.tokens[self.prev_index..self.current_index];
-            self.decode(tokens)?
-        };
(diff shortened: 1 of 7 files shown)


astorise:
Closing — these two PRs are divergent branches of the same change. Closing both; this work is moving to a separate crate.

PR #3813, opened 2026-07-31
Fix concrete shape element count validation

### LINKED ISSUE #3812: Tensor::from_slice / from_vec silently accept a concrete shape whose element count ≠ data length
## Summary

`Tensor::from_vec` and `Tensor::from_slice` both document that *"the number of elements in this vector must be the same as the number of elements defined by the shape"*, but that invariant is **not enforced when the shape is fully concrete**. A slice/vec shorter (or longer) than the shape's element count is accepted without error, producing a tensor whose logical `elem_count()` does not match its backing storage.

## Repro

```rust
use candle_core::{Tensor, Device};

    // 4 values, but the shape claims 2 * 2 * 2 = 8 elements.
    let t = Tensor::from_slice(&[1f32, 2., 3., 4.], (2, 2, 2), &Device::Cpu)?;
    // Constructs successfully despite the mismatch:
    assert_eq!(t.elem_count(), 8); // tensor built from only 4 values
    Ok(())
}
```

`from_vec` behaves identically.

## Root cause

Both constructors funnel the data length into `ShapeWithOneHole::into_shape(data.len())` (`from_vec` / `from_slice` → `from_vec_impl` in `candle-core/src/tensor.rs`). For a fully-concrete shape the blanket impl throws the length away:

```rust
// candle-core/src/shape.rs
impl<S: Into<Shape>> ShapeWithOneHole for S {
    fn into_shape(self, _el_count: usize) -> Result<Shape> {
        Ok(self.into()) // el_count ignored — no validation
    }
}
```

The *one-hole* impls, by contrast, do validate via `hole_size` (which rejects `el_count % prod_d != 0`). So `((), 2, 2)` is checked but the far more common `(2, 2, 2)` is not.

## Impact

- **Under-sized data** → the tensor claims more elements than its storage holds. Downstream ops trust `shape.elem_count()` and read past the buffer: a panic on CPU, out-of-range/garbage reads on GPU kernels (no bounds check).
- **Over-sized data** → silently truncated.
- This is the **root cause of the recurring "batched causal mask" bug** (issue #3582, now closed, plus the near-duplicate PRs #3780 / #3484 / #3743 / #3646 / #3624 / #3610 / #3586 / #3611 / #3609): every model built `from_slice(&mask, (b, 1, tgt, tgt + offset))` from a `tgt * (tgt + offset)` buffer, and the mismatch was accepted silently instead of erroring at construction — surfacing as garbage/NaN for `batch > 1`, far from the cause.

## Suggested fix

Enforce the documented contract for concrete shapes: when the shape carries no hole, verify `shape.elem_count() == el_count` and return an error otherwise (mirroring the `hole_size` check the one-hole impls already use). This turns the whole class of "wrong output for batch > 1"-style bugs into a clear error at the construction site.

Note this is a behavior change: code relying on the current silent truncation/over-claim would start erroring — arguably surfacing latent bugs, but worth a deliberate decision.

## Related (secondary)

The batch-mask bug fanned out into ~9 near-duplicate PRs because causal/sliding-window mask construction is copy-pasted across 8+ model files (`qwen3`, `quantized_qwen3`, `qwen3_moe`, `quantized_qwen3_moe`, `glm4`, `smollm3`, `z_image`, …), so the same fix must be replicated per model. Enforcing the length check above catches every copy at once; a shared mask helper (separate discussion) would prevent the divergence in the first place.

ccyyy1023:
## Summary

- validate fully concrete shapes against the input element count in `ShapeWithOneHole::into_shape`
- return the existing structured `ShapeMismatch` error before creating inconsistent tensors
- add regression coverage for `from_slice` and `from_vec`, including shorter and longer inputs, valid inferred shapes, and zero-sized concrete shapes

## Testing

- `cargo clippy -p candle-core --no-default-features --tests -- -D warnings`
- `cargo test -p candle-core --no-default-features --tests` (172 passed)
- `cargo test -p candle-core --no-default-features --doc` (45 passed, 1 ignored)

Closes #3812

### DIFF
--- a/candle-core/src/shape.rs
+++ b/candle-core/src/shape.rs
@@ -472,8 +472,16 @@ pub trait ShapeWithOneHole {
 }
 
 impl<S: Into<Shape>> ShapeWithOneHole for S {
-    fn into_shape(self, _el_count: usize) -> Result<Shape> {
-        Ok(self.into())
+    fn into_shape(self, el_count: usize) -> Result<Shape> {
+        let shape = self.into();
+        if shape.elem_count() != el_count {
+            return Err(Error::ShapeMismatch {
+                buffer_size: el_count,
+                shape,
+            }
+            .bt());
+        }
+        Ok(shape)
     }
 }
 
--- a/candle-core/tests/tensor_tests.rs
+++ b/candle-core/tests/tensor_tests.rs
@@ -2003,6 +2003,28 @@ fn test_flip_3d_channels() -> Result<()> {
 }
 
 #[test]
+fn tensor_from_data_validates_concrete_shape() -> Result<()> {
+    let values = [1f32, 2., 3., 4.];
+
+    assert!(Tensor::from_slice(&values, (2, 2, 2), &Device::Cpu).is_err());
+    assert!(Tensor::from_slice(&values, (2, 1), &Device::Cpu).is_err());
+    assert!(Tensor::from_vec(values.to_vec(), (2, 2, 2), &Device::Cpu).is_err());
+    assert!(Tensor::from_vec(values.to_vec(), (2, 1), &Device::Cpu).is_err());
+
+    let tensor = Tensor::from_slice(&values, (2, 2), &Device::Cpu)?;
+    assert_eq!(tensor.dims(), &[2, 2]);
+
+    let tensor = Tensor::from_vec(values.to_vec(), ((), 2), &Device::Cpu)?;
+    assert_eq!(tensor.dims(), &[2, 2]);
+
+    let empty: &[f32] = &[];
+    let tensor = Tensor::from_slice(empty, (0, 3), &Device::Cpu)?;
+    assert_eq!(tensor.dims(), &[0, 3]);
+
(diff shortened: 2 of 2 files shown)


astorise:
Thanks for the fix — this looks good to me.

The change enforces the documented `from_vec` / `from_slice` contract at the shared `ShapeWithOneHole` boundary, preventing inconsistent tensor shapes before storage is created. Using the existing `Error::ShapeMismatch` variant is also consistent with `Tensor::from_array`.

The regression test coverage is thorough: it exercises both constructors, undersized and oversized inputs, inferred shapes, and zero-sized concrete shapes.

Reviewed-by: astorise

ccyyy1023:
Thanks for the thoughtful review. I hadn’t realized you were also preparing #3814 when I started working on this, and I’m sorry about the duplicated effort. Your detailed issue report and root-cause analysis were very helpful, and I really appreciate you taking the time to review #3813.

ivarflakstad (approved):
👍

PR #3818, opened 2026-08-01
Expose Llama blocks for per-layer composition

### LINKED ISSUE #3817: `candle-transformers::models::llama`: expose `Block` for per-block composition
## Problem

`Llama::load_with_mlp_factory` (#3809) closed the case where a caller wants to
change a block's *MLP* and inherit everything else. It does not help a caller
that needs less than a whole model.

Pipeline parallelism is the case. A stage is a contiguous range of transformer
layers placed on one device: stage 0 owns the token embedding, the last stage
owns the final norm and the LM head, and every stage in between owns **neither**
— just its blocks, its slice of the KV cache, and the activations handed to it
by the previous stage. There is no `Llama` to construct: `load_with_mlp_factory`
builds embeddings, all `num_hidden_layers` blocks, the norm and the head, which
is precisely what a stage must not materialise. A four-stage deployment that
called it would load the full model four times.

The same gap appears one step away, in mixture-of-experts. A checkpoint whose
layers are individually dense or MoE needs attention composed with a per-layer
choice of feed-forward. The MLP factory's `usize` argument expresses the choice
— but only inside a whole-model load, so a caller that wants the block itself
still has to build one.

Both therefore reimplement the block: `RmsNorm` → attention → `RmsNorm` → MLP,
plus the causal mask, the rotary tables and the KV cache that attention needs.
In Tachyon-Mesh that is ~269 lines of copied attention/cache/mask math, kept
honest only by a test asserting bit-for-bit equality against the dense
reference. The MLP seam did not reduce it, because what these callers duplicate
is not the MLP — it is everything *around* it.

What makes this worth a second look is how close the code already is.
`Block::load_with_mlp` is what `load_with_mlp_factory` calls per layer, and
`Cache` — rotary tables, masks, per-layer KV — is already public and already
indexed by `block_idx`. The composition these callers need is the one the
crate performs internally; only the visibility differs.

## What a usable seam needs

1. **Construct one block**, at a known layer index, from that layer's
   `VarBuilder` — with the same MLP substitution `load_with_mlp_factory`
   already offers, so the two seams compose rather than compete.
2. **Run it against the public `Cache`.** `Block::forward` already takes
   `block_idx` and indexes the cache with it, which is exactly what a stage
   holding layers 12..18 needs; no new cache shape is required.
3. **Require neither embedding nor head.** A middle stage has no `wte` and no
   `lm_head`, and must not be made to load them.
4. **Stay additive.** `Llama::load`, `load_with_mlp_factory` and `forward` must
   behave exactly as they do now — the shape #3721 and #3809 established.

## Proposed API

Visibility, not new machinery:

```rust
pub struct Block { /* … */ }

impl Block {
    /// Load layer `layer_idx`, with the dense MLP.
    pub fn load(vb: VarBuilder, cfg: &Config, layer_idx: usize) -> Result<Self>;

    /// Load layer `layer_idx`, substituting the feed-forward computation —
    /// the per-layer half of `Llama::load_with_mlp_factory`.
    pub fn load_with_block_mlp(
        vb: VarBuilder,
        cfg: &Config,
        layer_idx: usize,
        mlp: Box<dyn BlockMlp>,
    ) -> Result<Self>;

    pub fn forward(
        &self,
        x: &Tensor,
        index_pos: usize,
        block_idx: usize,
        cache: &mut Cache,
        adapters: Option<&[Option<&str>]>,
    ) -> Result<Tensor>;
}
```

`load_with_block_mlp` takes the trait object rather than the private
`BlockMlpImpl`, so the enum stays an implementation detail; `load_with_mlp_factory`
can keep using the enum internally for the dense case.

## Note

This is the second half of #3809. That one let a caller replace a block's MLP
inside a whole model; this one lets a caller assemble blocks without one. With
both, a tensor-parallel model, a pipeline stage and a mixed dense/MoE stack are
all compositions of upstream parts, and the copied attention goes away.

I am happy to submit the PR, with the equality test against the dense reference
extended to a block assembled this way. Related: #3809, #3721.

astorise:
## Summary

- Expose `llama::Block` so callers can load and run individual transformer layers.
- Add `Block::load_with_block_mlp` for per-layer feed-forward substitution.
- Retain `Llama::load_with_mlp_factory` for whole-model composition.
- Add equivalence tests for independently composed blocks and custom dense MLP loading.

## Motivation

Pipeline-parallel and mixed dense/MoE callers need contiguous layer ranges without loading embeddings, the final norm, or the LM head. This public block seam reuses Candle’s attention, RoPE, masks, and KV cache rather than requiring downstream reimplementations.

Closes #3817

## Testing

- `cargo test -p candle-transformers --lib models::llama::tests`

### DIFF
--- a/candle-transformers/src/models/llama.rs
+++ b/candle-transformers/src/models/llama.rs
@@ -423,17 +423,48 @@ impl Mlp {
     }
 }
 
-#[derive(Debug, Clone)]
-struct Block {
+/// A transformer's feed-forward computation within a Llama block.
+///
+/// Implement this trait to replace a block's dense MLP while retaining Candle's
+/// attention, RoPE, causal masking, KV cache, norms, and residual wiring.
+pub trait BlockMlp: Send + Sync {
+    fn forward(&self, xs: &Tensor) -> Result<Tensor>;
+}
+
+impl BlockMlp for Mlp {
+    fn forward(&self, xs: &Tensor) -> Result<Tensor> {
+        Self::forward(self, xs)
+    }
+}
+
+/// A single Llama transformer layer.
+///
+/// `Block` can be loaded and run independently, which allows callers to build
+/// a contiguous range of layers without materialising Llama's embeddings, final
+/// norm, or language-model head.
+#[derive(Clone)]
+pub struct Block {
     rms_1: RmsNorm,
     attn: CausalSelfAttention,
     rms_2: RmsNorm,
-    mlp: Mlp,
+    mlp: std::sync::Arc<dyn BlockMlp>,
     span: tracing::Span,
 }
 
+impl std::fmt::Debug for Block {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Block")
+            .field("rms_1", &self.rms_1)
+            .field("attn", &self.attn)
+            .field("rms_2", &self.rms_2)
+            .field("mlp", &"dyn BlockMlp")
+            .finish()
+    }
+}
+
 impl Block {
-    fn forward(
+    /// Runs this block using `block_idx`'s entry in `cache`.
+    pub fn forward(
         &self,
         x: &Tensor,
         index_pos: usize,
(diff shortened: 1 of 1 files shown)


astorise:

PR #3822, opened 2026-08-01
Store the first gradient directly instead of adding it into fresh zeros

PCfVW:
## What this changes

`Tensor::backward` accumulates gradients through `GradStore::or_insert`, which initializes every node's slot with `zeros_like` and lets the call site add into it:

```rust
let sum_grad = grads.or_insert(arg)?;
*sum_grad = sum_grad.add(&arg_grad)?;
```

Every gradient edge therefore pays a full size zero fill (one write) plus a full size add (two reads and a write, into another fresh allocation). This includes the case where the node has a single consumer and the sum has exactly one term, which in a transformer is nearly every node.

This PR adds two private helpers to `GradStore` and converts the call sites to them:

```rust
fn accumulate(&mut self, tensor: &Tensor, grad: &Tensor) -> Result<()> {
    match self.0.entry(tensor.id()) {
        Entry::Occupied(mut entry) => {
            let sum_grad = entry.get_mut();
            *sum_grad = sum_grad.add(grad)?;
        }
        Entry::Vacant(entry) => {
            entry.insert(grad.contiguous()?);
        }
    }
    Ok(())
}
```

The first gradient reaching a node is stored directly; nodes with real fan-in accumulate exactly as before from the second gradient onward. An `accumulate_neg` twin covers the six gradient rules that subtract (`Neg`, `Cos`, `Tanh`, `Recip`, and the right hand sides of `Sub` and `Div`), so their fan-in path keeps the single `sub` it has today. This is the accumulation strategy PyTorch uses: `InputBuffer::add` stores the first gradient into the empty slot and only adds from the second onward.

66 of the 71 `or_insert` call sites convert. Four keep the accumulator by design: `Gather` and `IndexSelect` scatter/index-add into a dense base, and the two `UpsampleNearest` sites overwrite their slot, a pre-existing behaviour this PR deliberately does not touch. Net diff: 115 insertions, 133 deletions, one file. The public API is unchanged and `or_insert` stays for the sites that need it.

## Why

Profiled with Nsight Systems on a 10.7M parameter transformer training step (6 layers, d=384, batch 128, f32, RTX 5060 Ti), the initialization traffic is about 900 `badd_f32` launches per step and roughly a third of total GPU time. The kernels themselves are fine, they run at about 80% of the card's spec bandwidth; the cost is purely the number of full tensor round trips.

After the change, same machine, same protocol, wall clock differenced over 20 vs 120 step runs:

| | before | after |
|---|---|---|
| `backward()` (per phase sync profile) | 413.7 ms | 251.8 ms |
| whole step, wall clock | 0.521 s | 0.391 s |
| training throughput | 31.4k tok/s | 41.9k tok/s |
| `badd_f32` launches per profiled iteration, nsys (one training step plus one forward only control pass) | 976, median 196 us | 401, median 8.4 us |
| gemm time (nsys) | unchanged | unchanged |

The launch count and the median collapse are the mechanism check: the surviving adds are small bias gradient reductions and genuine fan-in, and the full tensor swarm is gone. Every candle training loop pays this tax today; nothing in the change or the measurement is specific to the workload used here.

## Two details that keep the change behaviour preserving

The insert is `grad.contiguous()?`. The old zeros plus add always produced a contiguous gradient, while a directly stored first gradient may be a strided view (a transpose backward, for instance). For already contiguous gradients, the common case, `contiguous` returns self and costs nothing; for a view it costs one copy where the old path cost a zeros write plus an add.

A few gradient expressions read the accumulator's metadata (`sum_grad.dims()` in the broadcast rules, `sum_grad.device()` in `ToDevice`). These are rewritten to read the target tensor's, which is what the accumulator's shape and device are defined to be.

## Evidence that it works

`cargo test -p candle-core` passes, 171 tests including the `grad_tests` suite, and `cargo test -p candle-nn` passes unchanged against the patched core.

The strongest check comes from the training setup this was measured on: a whole loop parity oracle that replays 20 recorded optimizer steps of the real model in PyTorch, bit exact batches and identical init on both sides. With this change the worst per step loss disagreement over the 20 steps is 1.78e-07, which is below the 2.0e-07 disagreement of unpatched candle with itself across CPU and GPU on the same rig. A downstream 64 test training suite also passes unchanged, including a staged equals continuous resume property held at 1e-5.

## One caveat worth documenting

Signed zeros. `0.0 + (-0.0)` is `+0.0`, so the old path could turn a `-0.0` gradient into `+0.0` where the new path preserves it. Values compare equal and this matches PyTorch, which also stores the first gradient, but a test asserting bit patterns on exact zeros could in principle notice.

## Alternatives considered

Keeping `or_insert` everywhere and pooling the allocations would remove the allocation cost but not the two full tensor round trips per edge, which are the measured majority of the cost. Making the inserted gradient non contiguous (skipping the `contiguous` call) would be slightly faster still for view gradients, but it would hand downstream consumers a layout the old code never produced, and `matmul` rejects general strides; the conservative form seemed the right first step.

## Notes

The four kept call sites and the two metadata rewrites are the only places where the conversion is not mechanical, and they are commented in the diff. Happy to split the `accumulate_neg` sites into a follow up if a smaller diff is preferred.

### DIFF
--- a/candle-core/src/backprop.rs
+++ b/candle-core/src/backprop.rs
@@ -183,32 +183,24 @@ impl Tensor {
             if let Some(op) = node.op() {
                 match op {
                     Op::Binary(lhs, rhs, BinaryOp::Add) => {
-                        let lhs_sum_grad = grads.or_insert(lhs)?;
-                        *lhs_sum_grad = lhs_sum_grad.add(&grad)?;
-                        let rhs_sum_grad = grads.or_insert(rhs)?;
-                        *rhs_sum_grad = rhs_sum_grad.add(&grad)?;
+                        grads.accumulate(lhs, &grad)?;
+                        grads.accumulate(rhs, &grad)?;
                     }
                     Op::Binary(lhs, rhs, BinaryOp::Sub) => {
-                        let lhs_sum_grad = grads.or_insert(lhs)?;
-                        *lhs_sum_grad = lhs_sum_grad.add(&grad)?;
-                        let rhs_sum_grad = grads.or_insert(rhs)?;
-                        *rhs_sum_grad = rhs_sum_grad.sub(&grad)?;
+                        grads.accumulate(lhs, &grad)?;
+                        grads.accumulate_neg(rhs, &grad)?;
                     }
                     Op::Binary(lhs, rhs, BinaryOp::Mul) => {
                         let lhs_grad = grad.mul(rhs)?;
-                        let lhs_sum_grad = grads.or_insert(lhs)?;
-                        *lhs_sum_grad = lhs_sum_grad.add(&lhs_grad)?;
+                        grads.accumulate(lhs, &lhs_grad)?;
                         let rhs_grad = grad.mul(lhs)?;
-                        let rhs_sum_grad = grads.or_insert(rhs)?;
-                        *rhs_sum_grad = rhs_sum_grad.add(&rhs_grad)?;
+                        grads.accumulate(rhs, &rhs_grad)?;
                     }
                     Op::Binary(lhs, rhs, BinaryOp::Div) => {
                         let lhs_grad = grad.div(rhs)?;
-                        let lhs_sum_grad = grads.or_insert(lhs)?;
-                        *lhs_sum_grad = lhs_sum_grad.add(&lhs_grad)?;
+                        grads.accumulate(lhs, &lhs_grad)?;
                         let rhs_grad = grad.mul(lhs)?.div(&rhs.sqr()?)?;
-                        let rhs_sum_grad = grads.or_insert(rhs)?;
-                        *rhs_sum_grad = rhs_sum_grad.sub(&rhs_grad)?;
+                        grads.accumulate_neg(rhs, &rhs_grad)?;
                     }
                     Op::Binary(lhs, rhs, BinaryOp::Minimum)
                     | Op::Binary(lhs, rhs, BinaryOp::Maximum) => {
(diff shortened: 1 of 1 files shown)


PCfVW:

Also a heads-up about this PR's CI, because it was about to become misleading.
The two runs queued when this was filed,
[30705605070](https://github.com/huggingface/candle/actions/runs/30705605070)
and [30705605092](https://github.com/huggingface/candle/actions/runs/30705605092),
were still `action_required`, waiting for a maintainer to approve a fork pull
request's workflow. GitHub expires an unapproved run after thirty days and stamps
it `conclusion: failure` with zero jobs executed. These were queued at 15:17 UTC
on 1 August, so that deadline fell today at 15:17 UTC. The sibling PR
[#3819](https://github.com/huggingface/candle/pull/3819) crossed the same line
earlier today: its two runs were stamped `failure` having never started, which
reported nothing about the patch either way, and clearing that took the same
rebase this PR has just had. This push re-queues the workflows here ahead of the
deadline rather than after it, and resets the clock. They still need one
maintainer click to actually execute.

`36074897` before and after, and `candle-core/src/backprop.rs` was not touched by
the 20 commits in between, so this is still the same +115/-133 in one file.

`rust-ci.yml`:

- `cargo test -p candle-core`, 17 targets, 225 passed, 0 failed, 1 ignored, being
  180 in the test binaries plus 45 doctests. `grad_tests` is 9 of 9
- `cargo test -p candle-nn`, 14 targets, 64 passed, 0 failed, 2 ignored

I ran those same two commands on unpatched `638a819a` for a baseline: candle-core
17 targets and 225 passed, candle-nn 14 targets and 64 passed. Identical to the
patched tree on both counts, so no test changed status in either direction, which
is what a behaviour preserving change should look like.

The 171 quoted in the description was measured on 1 August against a different
`main`. The throughput figures predate the rebase as well, but nothing in the 20
intervening commits touches `backprop.rs`, and the patch is byte for byte the one
that produced them.

Issue #3752, opened 2026-07-19
[Invalid — misfiled downstream issue] LayerNorm fast path silently severs autograd graph (apply_op3_no_bwd)


ynishi:
# [Bug] LayerNorm fast path silently severs autograd graph (apply_op3_no_bwd)

## Summary

candle-nn 0.11's `LayerNorm::forward` fast path dispatches through `crate::ops::layer_norm`, which is a `CustomOp3` registered via `apply_op3_no_bwd`. Tensors produced by `apply_op3_no_bwd` carry `BackpropOp::none()`, severing the autograd graph at every LayerNorm invocation.

Any trainable parameter upstream of a LayerNorm (in practice: every transformer-block Var) receives no gradient during training, while forward + loss + optimizer.step run to completion with no error signal. Forward produces numerically correct outputs, `.backward()` does not error, `optimizer.step()` does not error, and checkpoints save without complaint. The failure mode is only visible via a before/after weight-diff check.

This is a silent zero-learning bug in a widely-used primitive.

## Reproduction

Downstream reproduction on GPT-2 medium (355M params) using `candle-nn::LayerNorm` for `ln_1` / `ln_2` per block and `ln_f` on the final projection:

- Environment: candle-core 0.11 / candle-nn 0.11
- Devices reproducing: RunPod A40 CUDA 12.4, and CPU (Apple M4 + Linux/gcc)
- 200-step LoRA training on a periodic synthetic corpus (cycle length 25)
- Track `optimizer.step()` invocation count = 199, `.backward()` succeeds with no error, checkpoint save succeeds
- Compare all 192 LoRA A/B parameter tensors (rank 16 × 6 target modules × 24 layers) before/after training with `abs_diff_mean` + cosine similarity

Observed:

- 192 tensors: `abs_diff_mean = 0.000e+00` and `cos_sim = 1.000000` — every trainable parameter bit-identical before/after 199 optimizer steps
- Loss trajectory bit-exact within each dataset cycle position (step 0 = step 25 = step 175 = `135.11166381835938` to 15 significant digits)
- Base-frozen check passes (expected, since no updates land anywhere)
- `loss.sorted_nodes()` reports 20 nodes, of which exactly 1 is a Var: the tied LM head via `wte`. The 15 transformer-block Vars (LoRA A/B + LN affine + Linear) are absent from the sorted-node set because the LayerNorm hop drops them from the graph.

## Root cause

[...]

Issue #3569, opened 2026-05-28
candle-nn: ops::softmax_last_dim is apply_op1_no_bwd; severs autograd (same pattern as #2168 / PR #3526)


nxrobins:
## Severity: HIGH — silently severs autograd through attention weights

\`candle_nn::ops::softmax_last_dim\` at \`candle-nn/src/ops.rs:430\` uses \`apply_op1_no_bwd\`. The typical use site is attention weights inside a model's forward pass, and the typical loss is computed downstream of those — so a \`loss.backward()\` call returns with no error, but \`grads.get(var)\` returns \`None\` for any \`Var\` upstream of the softmax.

The workaround \`softmax(t, D::Minus1)\` is pure-Candle differentiable and produces the same output. Same naming-footgun pattern as #2168 / PR #3526 (RmsNorm) and #3568 (rope): users reach for the obvious name and silently lose gradients.

**Verified present at candle main HEAD \`1d7e9274591ef7eb101d965b71243a41994214d2\`** (committed 2026-05-26).

## Reproducer

https://github.com/nxrobins/candle-bug-4-softmax-no-bwd

\`\`\`bash
git clone https://github.com/nxrobins/candle-bug-4-softmax-no-bwd && cd candle-bug-4-softmax-no-bwd && cargo run --release
# exit 1: BUG PRESENT (softmax_last_dim severs autograd; softmax(t, dim) propagates)
# exit 0: bug appears fixed
\`\`\`

\`cargo run\` and \`cargo test\` (2 tests) pass. Loss shape is \`(out * out).sum_all()\` rather than \`out.sum_all()\` because the latter is mathematically a constant for any softmax (sums to 1 per row) so its gradient is legitimately zero — the squared sum is non-uniform w.r.t. the input and gives a meaningful gradient signal for the positive control.

## Suggested fix

Two options:

**Option A (doc-only floor)** — \`#[doc]\` warning on \`softmax_last_dim\` pointing at \`softmax(t, dim)\`.

**Option B (supremum)** — implement \`bwd\` for the \`SoftmaxLastDim\` Op. Math:

\`\`\`
grad_in = softmax * (grad_out - sum(grad_out * softmax, dim=-1))
\`\`\`

This would mirror the shape of PR #3526 (\`bwd\` impl on the existing \`CustomOp\`). Watch the cuDNN-backed path if compiled with \`--features cudnn\`.

## Related

[...]

PR #3823, opened 2026-08-01
Give the fused softmax_last_dim and layer_norm an analytic backward

PCfVW:
## What this changes

`candle_nn::ops::softmax_last_dim` and the fused `layer_norm` are built with `apply_op*_no_bwd`: their output records no backprop op, so a backward pass silently stops there. Every parameter upstream of such a call trains as if frozen, with no error, no warning, and a loss that still decreases because whatever remains connected keeps learning. The module level constructors route into these ops when their gates hold (`LayerNorm` when the input is contiguous, the mean is removed and a bias is present), so a user who never writes `ops::` is affected too.

This change implements `CustomOp::bwd` for both ops and switches them to `apply_op1` / `apply_op3`:

```rust
fn bwd(&self, _arg: &Tensor, res: &Tensor, grad_res: &Tensor) -> Result<Option<Tensor>> {
    Ok(Some(Self::bwd_impl(res, grad_res)?))
}
```

Untracked inputs are byte identical to before: the same fused kernels run and nothing else happens. The only difference is that the op is now recorded when an input requires a gradient, and the analytic backward runs when `backward()` reaches it.

## Relationship to the open PRs and issues

This is a known problem with a long history: #2168 is the root report (RmsNorm, 2024), #2977 named the module level gates, #3011 and #3752 report it through LayerNorm, #3569 through softmax. Four fix PRs are open: #3526 (RmsNorm), #3612 (rope), #3613 (layer_norm) and #3724 (softmax_last_dim). We derived the backward formulas independently from PyTorch's reference decompositions and arrived at the same mathematics as #3613, which we take as evidence the formulas are right rather than as a priority claim. What this PR adds over the open ones:

- both ops in one review, with the same dtype policy (half precision computes in f32 end to end, mirroring both the fused forward kernels and the composed fall throughs);
- element wise gradient parity tests against the autograd of the composed forms, for all gradients the ops produce (dx at 1e-6 for softmax; dx, dgamma and dbeta at 1e-5 for layer_norm), on deterministic sign varied inputs with a non trivial cotangent, plus a canary test asserting the gradients exist at all, which is the regression this class of bug needs;
- a whole loop, real training check against PyTorch (below).

If the maintainers prefer to land #3613 and #3724, we are happy to close this and port the tests to those PRs instead.

## Where the formulas come from

PyTorch's reference decompositions, which are the specification its fused kernels are tested against:

- softmax (`_softmax_backward_data`): `dx = y * (dy - sum_h(dy * y))`, needing only the saved output;
- layer norm (`native_layer_norm_backward`): with `g = dy * gamma` and statistics over the hidden axis, `dx = rstd / N * (N * g - sum_h(g) - x_hat * sum_h(g * x_hat))`, `dgamma = sum_outer(dy * x_hat)`, `dbeta = sum_outer(dy)`.

## Evidence that it works

`cargo test -p candle-nn` passes on this branch: 61 tests, including the three new ones in `tests/fused_grad_tests.rs`.

The strongest check comes from a real training setup: a 10.7M parameter transformer (6 layers, d=384, batch 128, f32, RTX 5060 Ti) whose training loop replays 20 recorded optimizer steps in PyTorch with bit exact batches and identical init on both sides. With these ops carrying the training forward and their analytic backward, the step 0 loss is bit identical to PyTorch (0.00e+00 difference), because both sides now run fused kernels with the same reduction structure, and the worst per step disagreement over the 20 steps is 1.78e-07, below the 2.0e-07 disagreement of stock candle with itself across CPU and GPU on the same rig.

Throughput on that workload, measured on top of #3822 (the two changes are independent; the numbers stack): 0.391 s to 0.333 s per step, 41.9k to 49.2k tokens per second. In a per phase synced profile of the same runs, the forward went from 122.2 ms to 95.6 ms, which is the composed norm and softmax chains collapsing into single kernels; the figure reproduced at 95.8 ms when re-measured for this write up.

## One deviation worth documenting

ATen's layer norm backward consumes the mean and rstd saved by the forward. A `CustomOp` has a single output, so this implementation recomputes them from the input inside `bwd`: three extra reductions over hidden sized data plus a few small per row ops. Saving the statistics needs a multi output native op and is left as a follow up; the recomputation is still far cheaper than the composed chain it replaces.

## Costs and risks

- Code that unintentionally relied on the fused ops as a gradient barrier will now see gradients flow through them. That is the fix working, but such users would observe changed training behaviour.
- A tracked fused call now keeps its inputs alive for the backward, as any recorded op does. Inference is unaffected.
- `rms_norm` (#3526) and `ops::sdpa` still carry `no_bwd` after this change; the same treatment applies and can follow.

## Notes

The measurement instrument, for reproduction: `nsys profile -t cuda` plus `nsys stats --report cuda_gpu_kern_sum` on any training loop shows the composed chains (`fast_sum`, `bmul`, `bdiv`, `bsub` on norm sized tensors) before, and single `layernorm_f32` / `softmax_f32` launches after. Nothing in the change or the measurement is specific to the workload used here.

### DIFF
--- a/candle-nn/src/ops.rs
+++ b/candle-nn/src/ops.rs
@@ -289,6 +289,10 @@ impl candle::CustomOp1 for SoftmaxLastDim {
         "softmax-last-dim"
     }
 
+    fn bwd(&self, _arg: &Tensor, res: &Tensor, grad_res: &Tensor) -> Result<Option<Tensor>> {
+        Ok(Some(Self::bwd_impl(res, grad_res)?))
+    }
+
     fn cpu_fwd(&self, storage: &CpuStorage, layout: &Layout) -> Result<(CpuStorage, Shape)> {
         fn softmax<T: candle::WithDType + num_traits::Float>(
             src: &[T],
@@ -435,7 +439,26 @@ impl candle::CustomOp1 for SoftmaxLastDim {
 }
 
 pub fn softmax_last_dim(xs: &Tensor) -> Result<Tensor> {
-    xs.apply_op1_no_bwd(&SoftmaxLastDim)
+    xs.apply_op1(SoftmaxLastDim)
+}
+
+impl SoftmaxLastDim {
+    /// The softmax jacobian-vector product, `dx = y * (dy - sum_h(dy * y))` — the formula of
+    /// PyTorch's `_softmax_backward_data` reference decomposition. Only the saved output `y` is
+    /// needed; nothing is recomputed. Half-precision inputs run the two reductions in `f32`,
+    /// mirroring the composed path.
+    fn bwd_impl(res: &Tensor, grad_res: &Tensor) -> Result<Tensor> {
+        let x_dtype = res.dtype();
+        let internal = match x_dtype {
+            candle::DType::F16 | candle::DType::BF16 => candle::DType::F32,
+            d => d,
+        };
+        let y = res.to_dtype(internal)?;
+        let dy = grad_res.to_dtype(internal)?;
+        let t = (&dy * &y)?;
+        let s = t.sum_keepdim(D::Minus1)?;
+        (t - y.broadcast_mul(&s)?)?.to_dtype(x_dtype)
+    }
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


PCfVW:

Also a heads-up about this PR's CI, because it was about to become misleading.
The two runs queued when this was filed,
[30714971505](https://github.com/huggingface/candle/actions/runs/30714971505)
and [30714971556](https://github.com/huggingface/candle/actions/runs/30714971556),
were still `action_required`, waiting for a maintainer to approve a fork pull
request's workflow. GitHub expires an unapproved run after thirty days and stamps
it `conclusion: failure` with zero jobs executed. These were queued at 19:32 UTC
on 1 August, so that deadline fell today at 19:32 UTC. The sibling PR
[#3819](https://github.com/huggingface/candle/pull/3819) crossed the same line
earlier today: its two runs were stamped `failure` having never started, which
reported nothing about the patch either way, and clearing that took the same
rebase this PR has just had. This push re-queues the workflows here ahead of the
deadline rather than after it, and resets the clock. They still need one
maintainer click to actually execute.

`c15ce21d` before and after, and neither `candle-nn/src/ops.rs` nor
`candle-nn/tests/fused_grad_tests.rs` was touched by the 20 commits in between,
so this is still the same +230/-2.

`rust-ci.yml`:

- `cargo test -p candle-nn`, 15 targets, 67 passed, 0 failed, 2 ignored, with the
  new suite green:

```
running 3 tests
test fused_backward_exists_at_all ... ok
test fused_softmax_backward_matches_autograd ... ok
test fused_layer_norm_backward_matches_autograd ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

- `cargo test -p candle-core`, 17 targets, 225 passed, 0 failed, 1 ignored

I ran `cargo test -p candle-nn` on unpatched `638a819a` for a baseline: 14
targets and 64 passed. The difference is exactly the one added target and its
three tests, so nothing that already existed changed status.

The 61 quoted in the description was measured on 1 August against a different
`main`.

PR #3825, opened 2026-08-02
Add candle-nvfp4-kernels for NVIDIA ModelOpt FP4 weights

astorise:
## Summary

Adds a new `candle-nvfp4-kernels` crate for executing NVIDIA ModelOpt W4A16/NVFP4 quantized linear layers on CUDA.

The crate provides:

- `linear_f32_host_batched` for `[tokens, cols] @ [rows, cols]^T`
- `linear_f32_host` as a single-token convenience wrapper
- Validation of activation, packed-weight, and E4M3 scale buffer sizes before kernel launch
- An explicit runtime capability gate: native NVFP4 execution requires CUDA compute capability 10.0 (Blackwell) or newer
- CUDA-side decoding of packed E2M1 nibbles, E4M3FN per-block scales (block size 16), and the tensor-wide F32 scale

The CUDA feature is opt-in so normal CPU-only workspace builds remain unaffected.

## Why

ModelOpt checkpoints increasingly use W4A16/NVFP4 weights. Decoding them on the host materializes dense weights repeatedly and defeats the format's memory and throughput benefits. This crate keeps unpacking and the linear operation on the GPU, while offering a batched entry point for prefill workloads.

## Validation

- `cargo test -p candle-nvfp4-kernels`
- `cargo clippy -p candle-nvfp4-kernels -- -D warnings`

Native CUDA compilation was not run locally because the Windows CUDA toolchain could not locate `cl.exe` in `PATH`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
     "candle-core",
     "candle-datasets",
     "candle-examples",
+    "candle-nvfp4-kernels",
     "candle-nn",
     "candle-pyo3",
     "candle-transformers",
--- a/candle-nvfp4-kernels/Cargo.toml
+++ b/candle-nvfp4-kernels/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "candle-nvfp4-kernels"
+version = "0.11.0"
+edition = "2021"
+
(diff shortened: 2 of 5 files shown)


astorise:

PR #3827, opened 2026-08-02
feat(generation): add JSON Schema decoding constraints

### LINKED ISSUE #3826: `candle-transformers::generation`: constrained decoding from a JSON schema
## Problem

`generation` now covers producing a token (`LogitsProcessor`, `Sampling`),
turning tokens back into text (`IncrementalDecoder`, #3789) and deciding when to
stop (`StopCriteria`, #3811). What it cannot do is constrain *what* the model is
allowed to produce.

That is the missing half of every "JSON mode" or "structured output" feature.
The technique is settled — compile the schema to a state machine, and before
each sample, mask to `-inf` every token whose text the machine would reject from
its current state — but it is nowhere in this crate, so each serving stack
writes it again. Getting it wrong is quiet: a model that emits one invalid token
produces a response that fails to parse, and the usual reaction is to retry the
whole generation rather than to notice the constraint never applied.

The pieces it needs are already here. Masking is a `Tensor` op over the logits
row; the token-to-text step is `IncrementalDecoder`'s; `StopCriteria` already
establishes that per-step decisions of this kind belong in `generation` rather
than in each caller's loop. Only the machine is missing.

## What a usable implementation needs

1. **Compile a schema once, use it per request.** Schema compilation is
   expensive relative to a decode step and is shared by every request naming the
   same schema, so the compiled form must be cheap to clone or share and the
   per-request state must be separate from it.
2. **Mask by decoded text, not by token id.** Whether a token is legal depends
   on the string it produces, which is a property of the tokenizer, not of the
   model — so the mask has to be computed against a decode callback rather than
   a fixed id set.
3. **Let EOS through only when the value is complete.** A machine that permits
   EOS mid-object produces truncated JSON that parses as a schema violation
   rather than as a truncation; one that never permits it hangs to the token
   limit.
4. **Advance on what was actually emitted.** The state moves with the token the
   sampler chose, which is not necessarily the highest-probability legal one.
5. **Fail compilation loudly.** An unsupported schema construct must be an error
   at compile time, not a constraint silently dropped at decode time — the
   caller asked for a guarantee and would otherwise be told it holds.

## Proposed API

```rust
pub struct CompiledFsm { /* … */ }

pub enum SchemaCompileError { /* invalid JSON, unsupported construct, … */ }

/// Compile a JSON Schema into a decoding constraint.
pub fn compile_schema(schema: &str) -> Result<CompiledFsm, SchemaCompileError>;

pub struct FsmLogitProcessor { /* … */ }

impl FsmLogitProcessor {
    pub fn new(fsm: Arc<CompiledFsm>) -> Self;

    /// Token ids the machine will accept next. `decode_one` maps an id to its
    /// text, so the caller's tokenizer decides — this crate does not assume one.
    pub fn allowed_token_ids(
        &self,
        vocab_size: usize,
        eos_tokens: &[u32],
        decode_one: impl Fn(u32) -> String,
    ) -> Vec<u32>;

    /// Advance on the token that was actually sampled.
    pub fn commit(&mut self, token_text: &str);
}
```

A caller then masks the logits row to the allowed set before
`LogitsProcessor::sample`, and `commit`s the result — which composes with
`IncrementalDecoder` and `StopCriteria` without either knowing about it.

## Note

Tachyon-Mesh has run this in production against real checkpoints; the
implementation is ~700 lines including the schema compiler and its tests, and I
am happy to submit it.

One design question worth settling in the issue rather than the PR: the
`allowed_token_ids` shape above builds a vector per step, which is simple and
measurably fine at the vocabulary sizes we serve, but a caller that wants a
bitmask or an in-place `Tensor` mask would rather have that. If a masking form
is preferred I would rather write it that way from the start than add it after.

astorise:
## Summary

Adds JSON Schema-aware constrained decoding to `candle-transformers::generation`.

The constraint is compiled once into a shareable `CompiledFsm`, while
`FsmLogitProcessor` holds independent per-request state. It evaluates
candidate token ids through caller-provided decoded token text, keeping the
implementation tokenizer-agnostic.

New public API: `compile_schema`, `CompiledFsm`, `SchemaCompileError`, and
`FsmLogitProcessor`.

EOS is allowed only after the generated text is a complete JSON value matching
the schema. Unsupported schema constructs fail at compile time.

Supported schema constructs currently include:

- `type`: `object`, `array`, `string`, `number`, `integer`, `boolean`, `null`
- object `properties`, `required`, and boolean `additionalProperties`
- array `items`
- `enum` and `const`

This branch also includes `IncrementalDecoder` and `StopCriteria`, the
preceding generation primitives referenced by #3826 and needed to bring this
branch from `sync/upstream-main` to the intended generation API.

## Tests

- `cargo test -p candle-transformers --test json_schema_tests`
- `cargo test -p candle-transformers --test incremental_decoder_tests --test stop_criteria_tests`

Closes #3826

### DIFF
--- a/candle-transformers/Cargo.toml
+++ b/candle-transformers/Cargo.toml
@@ -23,6 +23,8 @@ rayon = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 serde_plain = { workspace = true }
+tokenizers = { workspace = true }
+thiserror = { workspace = true }
 tracing = { workspace = true }
 
 [features]
--- a/candle-transformers/src/generation/incremental_decoder.rs
+++ b/candle-transformers/src/generation/incremental_decoder.rs
@@ -1,0 +1,726 @@
+//! Incremental detokenization for autoregressive decode loops.
+//!
+//! An autoregressive loop needs the text produced so far on every step, both to match stop
+//! sequences and to stream deltas back to the caller. Re-decoding the whole token sequence on
+//! every step is correct but quadratic in the generated length. Appending the decode of each
+//! token is linear but wrong: a BPE/SentencePiece decode is *not* the concatenation of its
+//! tokens' decodes. A multi-byte character can span several tokens (the prefix decodes to
+//! U+FFFD, and a later token retroactively replaces it), and decoding a sub-sequence can gain
+//! or lose a leading space.
+//!
+//! [`IncrementalDecoder`] keeps a bounded trailing window of tokens, re-decodes only that
+//! window, and appends the difference against the *previous* window decode, so any artifact
+//! caused by decoding a sub-sequence appears in both decodes and cancels out. The window is
+//! only moved forward when the shorter decode is provably a suffix of the current one, and
+//! only when the tokenizer's decoder has been shown to have a bounded context; otherwise the
+//! decoder degrades to whole-sequence decoding rather than corrupting the output.
+//!
+//! The invariant is that [`IncrementalDecoder::text`] equals a whole-sequence decode of every
+//! token pushed so far, at every step.
+//!
+//! ```no_run
+//! use candle_transformers::generation::IncrementalDecoder;
+//!
+//! # fn main() -> candle::Result<()> {
+//! # let tokenizer: tokenizers::Tokenizer = unimplemented!();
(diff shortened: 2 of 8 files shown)


astorise:

PR #3829, opened 2026-08-02
candle-nn: add tensor-parallel linear layers

### LINKED ISSUE #3828: `candle-nn`: column- and row-parallel linear layers for multi-GPU inference
## Problem

There is no tensor parallelism in this crate. A model too large for one device
can be loaded layer-by-layer, or quantized until it fits, but it cannot be
*sharded* — and #1813 ("Quantized models on multi-GPU") has been open since
March 2024 without an answer.

The missing piece is small and old. Megatron-style tensor parallelism needs
exactly two layers: a column-parallel linear that splits its output features
across devices and concatenates, and a row-parallel linear that splits its input
features and all-reduces. Everything else — which layers to shard, how devices
are discovered, how ranks talk to each other — is the caller's, and should stay
the caller's.

The MLP seam merged for #3809 makes this the natural next step: a caller can now
substitute a block's feed-forward computation, but has to write the sharded
linear layers itself to have anything worth substituting. Tachyon-Mesh wrote
them, and they are not mesh-specific: they are two `Tensor` operations and a
device list.

## What the layers need to get right

1. **Refuse an uneven split.** `out_features` (column) or `in_features` (row)
   not divisible by the device count must be an error at shard time. Silently
   dropping the remainder produces a model that runs and is wrong, which is the
   worst available outcome.
2. **Contiguity before the matmul.** `narrow`, `to_device` and `t()` all yield
   non-contiguous views; the CPU matmul tolerates them and the CUDA one does
   not, so the shards must be made contiguous inside the layer rather than by
   every caller who happens to hit the GPU path.
3. **A single all-reduce per row-parallel layer**, not per shard — that
   reduction is the whole communication cost of the scheme, and doing it
   per-shard is what turns tensor parallelism into a slowdown.
4. **Degenerate cleanly to one device.** A single-element device list must
   produce exactly the dense result, so the same code path serves both and a
   caller does not branch.
5. **Leave the communicator to the caller.** Whether the all-reduce goes through
   NCCL, through host memory, or through something else is a deployment
   decision; the layer should accept a reduction strategy rather than assume
   one.

## Proposed API

```rust
pub struct ColumnParallelLinear { /* … */ }

impl ColumnParallelLinear {
    /// Split `weight` (`[out_features, in_features]`) into `devices.len()`
    /// equal column shards.
    pub fn shard(weight: &Tensor, devices: &[Device]) -> Result<Self>;

    /// Run every shard and gather onto `gather_device`.
    pub fn forward(&self, x: &Tensor, gather_device: &Device) -> Result<Tensor>;
}

pub struct RowParallelLinear { /* … */ }

impl RowParallelLinear {
    /// Split `weight` (`[out_features, in_features]`) into `devices.len()`
    /// equal row shards.
    pub fn shard(weight: &Tensor, devices: &[Device]) -> Result<Self>;

    /// Run every shard on its own slice of `x` and reduce onto
    /// `reduce_device`.
    pub fn forward(&self, x_shards: &[Tensor], reduce_device: &Device) -> Result<Tensor>;
}

/// Split a `[rows, in_features]` activation the way `RowParallelLinear`
/// expects it.
pub fn split_for_row_parallel(x: &Tensor, devices: &[Device]) -> Result<Vec<Tensor>>;
```

With these plus #3809's `BlockMlp`, a tensor-parallel Llama is a factory closure
and nothing else — which is how we run it today.

## Note

Deliberately *not* proposed here: cluster topology discovery, NCCL bootstrap,
NUMA binding, pipeline-stage transports. Those are deployment concerns and they
belong in whatever is orchestrating the devices, not in `candle-nn`. Keeping the
boundary there is what makes this two types rather than a subsystem.

I am happy to submit the PR, with the equality test we use — sharded output must
match the dense reference bit for bit on a real checkpoint, which is also the
only way to catch an uneven-split bug that a shape check misses.

Related: #3809, #1813.

astorise:
## Summary

Adds Megatron-style tensor-parallel linear building blocks to `candle-nn`:

- `ColumnParallelLinear`, which shards output features across devices and gathers the concatenated output.
- `RowParallelLinear`, which shards input features across devices and reduces partial outputs.
- `split_for_row_parallel`, to partition activations consistently with row-parallel weights.
- `TensorReducer`, allowing callers to provide their own collective implementation (e.g. NCCL).
- `SumReducer`, a portable default reducer for single-device and host-mediated use cases.

The implementation rejects empty and uneven device splits, makes matmul inputs contiguous internally, and invokes the row-parallel reducer once per layer.

## Tests

- `cargo test -p candle-nn tensor_parallel`

Closes #3828

### DIFF
--- a/candle-nn/src/lib.rs
+++ b/candle-nn/src/lib.rs
@@ -36,6 +36,7 @@ pub mod rnn;
 pub mod rotary_emb;
 pub mod sampling;
 pub mod sequential;
+pub mod tensor_parallel;
 pub mod var_builder;
 pub mod var_map;
 /// Re-export of [`attention::varlen`] for backward compatibility.
@@ -62,6 +63,9 @@ pub use ops::Dropout;
 pub use optim::{AdamW, Optimizer, ParamsAdamW, SGD};
 pub use rnn::{gru, lstm, GRUConfig, LSTMConfig, GRU, LSTM, RNN};
 pub use sequential::{seq, Sequential};
+pub use tensor_parallel::{
+    split_for_row_parallel, ColumnParallelLinear, RowParallelLinear, SumReducer, TensorReducer,
+};
 pub use var_builder::VarBuilder;
 pub use var_map::VarMap;
 
--- a/candle-nn/src/tensor_parallel.rs
+++ b/candle-nn/src/tensor_parallel.rs
@@ -1,0 +1,227 @@
+//! Building blocks for Megatron-style tensor parallel linear layers.
+//!
+//! This module deliberately does not discover devices or create a communicator.
+//! Callers provide both the devices used to shard a weight and, when needed, a
+//! [`TensorReducer`] that performs their distributed reduction.
+
+use candle::{bail, Device, Result, Tensor, D};
+
+/// Reduces the partial outputs of a [`RowParallelLinear`] once per layer.
+///
+/// Implement this trait to connect a row-parallel layer to a collective such as
+/// NCCL. The tensors passed to `all_reduce` live on their respective shard
+/// devices and the returned tensor must live on `reduce_device`.
+pub trait TensorReducer: Send + Sync {
+    fn all_reduce(&self, shards: &[Tensor], reduce_device: &Device) -> Result<Tensor>;
+}
+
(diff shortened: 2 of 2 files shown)


astorise:

PR #3830, opened 2026-08-02
Speed up non-contiguous where_cond on CPU by migrating to NdIter

Fabioomega:
## What this changes

On the non-contiguous path, the CPU `where_cond` zipped three `strided_index()` iterators,
recomputing full multi-dimensional index arithmetic per element, per tensor. In the same vein
as `binary_map`, switching to an `NdIter` that walks all three tensors at once is
significantly faster.

This adds `where_impl`, which is generic over how each operand's offset is computed, so
specializing a new stride combination is just another match arm. This PR specializes the 0/1
combinations and leaves everything else on the generic path. The combination that matters in
practice is a causal/attention mask against a broadcast scalar fill.

The fully contiguous fast path is unchanged, so the pre-existing all-contiguous benchmarks
are not listed below.

## Results

Benchmarked on CPU via `cargo bench --bench bench_main -- where_cond`, criterion medians.
All four scenarios are new benchmarks added in this PR.

### f32

| benchmark | before | after | change |
|---|---|---|---|
| `where_cond_masked_fill` | 4.49 ms | 755 µs | −83.2% |
| `where_cond_masked_fill_bcast` | 4.76 ms | 757 µs | −84.1% |
| `where_cond_transposed` | 5.60 ms | 2.64 ms | −52.9% |
| `where_cond_narrow_inner` | 2.07 ms | 1.10 ms | −46.8% |

### bf16

| benchmark | before | after | change |
|---|---|---|---|
| `where_cond_masked_fill` | 4.24 ms | 387 µs | −90.9% |
| `where_cond_masked_fill_bcast` | 4.56 ms | 412 µs | −91.0% |
| `where_cond_transposed` | 5.22 ms | 2.16 ms | −58.6% |
| `where_cond_narrow_inner` | 1.99 ms | 592 µs | −70.3% |

### f16

| benchmark | before | after | change |
|---|---|---|---|
| `where_cond_masked_fill` | 4.28 ms | 396 µs | −90.7% |
| `where_cond_masked_fill_bcast` | 4.39 ms | 393 µs | −91.0% |
| `where_cond_transposed` | 5.71 ms | 1.94 ms | −66.0% |
| `where_cond_narrow_inner` | 1.57 ms | 603 µs | −61.7% |

## New benchmarks

There were no benchmarks for the non-contiguous branches. The added ones are:

- `where_cond_masked_fill`: causal `u8` mask, broadcast scalar `on_true`, contiguous
  `on_false`. Strides `(1, 0, 1)`. This is the attention-mask shape.
- `where_cond_masked_fill_bcast`: same, but the mask itself is broadcast across the batch
  and head dims.
- `where_cond_transposed`: `on_false` transposed, so the inner stride is neither 0 nor 1
  and the generic path is exercised.
- `where_cond_narrow_inner`: narrowed innermost dim.

## Tests

Extended the existing `ternary_op` test (so it runs on CPU, CUDA and Metal) to cover the
broadcast (`0`) and contiguous (`1`) inner-stride combinations, plus the transposed and
narrowed layouts that land on the generic path. Each case asserts the strided result matches
the same op run on `.contiguous()` inputs.

## Checks

- `cargo test -p candle-core`
- `cargo clippy -p candle-core --all-targets`

CUDA and Metal backends are untouched; the new tests do exercise them via `test_device!`.

### DIFF
--- a/candle-core/benches/benchmarks/where_cond.rs
+++ b/candle-core/benches/benchmarks/where_cond.rs
@@ -18,6 +18,18 @@ const fn create_cond_arr<const N: usize>() -> [u8; N] {
     arr
 }
 
+const fn create_causal_arr<const S: usize, const N: usize>() -> [u8; N] {
+    let mut arr = [0u8; N];
+    let mut i = 0;
+    while i < N {
+        let row = (i / S) % S;
+        let col = i % S;
+        arr[i] = if col > row { 1 } else { 0 };
+        i += 1;
+    }
+    arr
+}
+
 const B: usize = 1;
 const M: usize = 1024;
 const K: usize = 1024;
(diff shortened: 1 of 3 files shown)


ivarflakstad (approved):
Nice! This is exactly why I added `NdIter`, just haven't gotten to all ops yet.

With the amount of generated code these days it's rare that PRs are good to go first try. Good stuff 👍

PR #3833, opened 2026-08-03
feat(nvfp4): add device-resident Tensor linear layer

### LINKED ISSUE #3831: candle-nvfp4-kernels: device-resident weights and a Tensor-level matmul
## Problem

The crate's entry points are host-in, host-out:

```rust
pub fn linear_f32_host(input: &[f32], packed_weight: &[u8], weight_scale_e4m3: &[u8],
                       tensor_scale: f32, rows: usize, cols: usize) -> Result<Vec<f32>>;
pub fn linear_f32_host_batched(/* … */, tokens: usize) -> Result<Vec<f32>>;
```

The batched form fixed the launch-per-token problem and is a real improvement.
What it cannot fix is the shape of the call: every operator takes a `&[f32]`
from the host, uploads it, uploads the packed weight and its scales, computes,
and downloads the result. A model has hundreds of operators per token, so the
activation makes a full round trip through PCIe on each of them, and the
weights — which never change — are re-uploaded every time.

The consequence is architectural rather than incremental. A caller cannot keep
its hidden state on the device, because the only way to apply a weight brings
the state back to the host. So the entire runtime around these kernels is
forced to be host-scalar: `Vec<f32>` hidden states, scalar norms, scalar
attention, scalar recurrences. Not because any of that is hard on a device, but
because the operator boundary insists on it.

That is what makes this different from a performance ticket. Nothing else in
the crate's neighbourhood can be used from a runtime built this way — not
`gated_delta_rule_chunked`, not the `Qwen3_5*` layers, not any `Tensor` op —
because all of them speak `Tensor` and this speaks `Vec<f32>`.

The other quantized formats in this repo do not have this boundary:
`candle-fp8-kernels`, `candle-awq-kernels` and `candle-gptq-kernels` all
produce something a `Tensor` graph can hold.

## What it needs

1. **The weight resident on the device, uploaded once.** The packed nibbles,
   the E4M3 block scales and the F32 tensor scale are immutable for the model's
   lifetime; they belong in device memory from load, not in the argument list of
   every call.
2. **A `Tensor` in, a `Tensor` out**, with the activation staying where it is.
   `[tokens, cols] x [rows, cols]^T -> [tokens, rows]`, no host staging on
   either side.
3. **Construction from a checkpoint's tensors**, so a loader can build the
   operator from the safetensors it already mapped rather than by first
   flattening them to host slices.
4. **The existing host entry points kept**, as the CPU fallback they already
   are. A device that cannot execute FP4 must still be able to run the model
   slowly rather than not at all.

## Proposed API

Mirroring how the other quantized formats present themselves:

```rust
pub struct Nvfp4Linear { /* device-resident packed weight, scales, tensor scale */ }

impl Nvfp4Linear {
    /// Upload a ModelOpt operator to `device` once.
    pub fn new(
        packed_weight: &[u8],      // [rows, cols / 2], E2M1 nibbles
        weight_scale_e4m3: &[u8],  // [rows, cols / BLOCK]
        tensor_scale: f32,
        rows: usize,
        cols: usize,
        device: &Device,
    ) -> Result<Self>;

    /// `xs`: [.., cols] -> [.., rows]. No host round trip.
    pub fn forward(&self, xs: &Tensor) -> Result<Tensor>;
}

impl Module for Nvfp4Linear { /* … */ }
```

`Module` is the part that matters beyond convenience: it is what lets an
NVFP4 operator be substituted anywhere a layer accepts one, which is the
subject of the sibling issue about a projection seam in `qwen3_5`.

## Note

Tachyon-Mesh runs Qwen 3.5 NVFP4 checkpoints on the current host entry points
and would move to this immediately; the host-scalar runtime we have exists
because of the boundary described above, not by preference.

One question worth settling here rather than in the PR: whether `Nvfp4Linear`
should own a bias and an optional output dtype, or stay a bare matmul with the
caller adding both. I would lean to bare — ModelOpt checkpoints keep the bias
unquantized, so it is an ordinary `Tensor` add the caller already has — but the
other kernel crates do not all agree, and consistency with them is worth more
than my preference.

Related: #3824.

astorise:
## Summary

Adds a Tensor-native NVFP4 linear layer to `candle-nvfp4-kernels`.

- Adds `Nvfp4Linear`, which uploads packed ModelOpt NVFP4 weights and E4M3 scales to CUDA once at construction.
- Implements `Module`, so it can be used anywhere a Candle layer accepts a module.
- Adds a CUDA custom op that accepts `[tokens, cols]` f32 activations and returns `[tokens, rows]` f32 outputs without host staging.
- Preserves arbitrary leading input dimensions: `[..., cols] -> [..., rows]`.
- Launches the kernel on Candle’s active CUDA stream.
- Keeps `linear_f32_host` and `linear_f32_host_batched` unchanged as the host fallback.

`Nvfp4Linear` intentionally stays a bare matmul: checkpoint biases remain ordinary unquantized Candle tensors and can be added by callers.

## Validation

- `cargo test -p candle-nvfp4-kernels`
- `cargo check -p candle-nvfp4-kernels --features cuda`

Fixes #3831.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
     "candle-core",
     "candle-datasets",
     "candle-examples",
+    "candle-nvfp4-kernels",
     "candle-nn",
     "candle-pyo3",
     "candle-transformers",
--- a/candle-nvfp4-kernels/Cargo.toml
+++ b/candle-nvfp4-kernels/Cargo.toml
@@ -11,7 +11,10 @@ license = "MIT OR Apache-2.0"
 
 [features]
 default = []
-cuda = []
+cuda = ["candle-core/cuda"]
+
+[dependencies]
+candle-core = { path = "../candle-core", version = "0.11.0", optional = false }
 
 [build-dependencies]
 anyhow = "1"
--- a/candle-nvfp4-kernels/Cargo.toml
+++ b/candle-nvfp4-kernels/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "candle-nvfp4-kernels"
+version = "0.11.0"
+edition = "2021"
+
+description = "CUDA kernels for NVIDIA ModelOpt NVFP4 quantized linear layers in Candle."
+repository = "https://github.com/huggingface/candle"
+keywords = ["blas", "tensor", "machine-learning", "cuda", "quantization"]
+categories = ["science"]
+license = "MIT OR Apache-2.0"
+
+[features]
+default = []
+cuda = []
+
+[build-dependencies]
+anyhow = "1"
+cudaforge = "0.1.2"
--- a/candle-nvfp4-kernels/build.rs
+++ b/candle-nvfp4-kernels/build.rs
@@ -1,0 +1,27 @@
+use std::path::PathBuf;
+
+fn main() -> anyhow::Result<()> {
+    println!("cargo::rerun-if-changed=build.rs");
+    if std::env::var("CARGO_FEATURE_CUDA").is_err() {
+        return Ok(());
+    }
+
+    println!("cargo::rerun-if-changed=kernels/nvfp4_gemm.cu");
+    let out_dir = PathBuf::from(std::env::var("OUT_DIR").expect("OUT_DIR not set"));
+    let mut builder = cudaforge::KernelBuilder::new()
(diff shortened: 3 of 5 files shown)


astorise:

Issue #3832, opened 2026-08-03
qwen3_5: a projection seam, so a non-GGUF quantized checkpoint can reuse the layers


astorise:
## Problem

`qwen3_5.rs` builds its projections concretely:

```rust
pub struct Qwen3_5GatedDeltaNet {
    // …
    in_proj_qkv: Linear,
    in_proj_z: Linear,
    in_proj_b: Linear,
    in_proj_a: Linear,
    out_proj: Linear,
    // …
}

impl Qwen3_5GatedDeltaNet {
    pub fn new(cfg: &Config, vb: VarBuilder) -> Result<Self> {
        let in_proj_qkv = linear_no_bias(hidden_size, conv_dim, vb.pp("in_proj_qkv"))?;
        // …
    }
}
```

`Qwen3_5Attention` and `Qwen3_5MLP` are the same. There are exactly two ways to
get a Qwen 3.5 through this crate: dense weights via `qwen3_5.rs`, or GGUF via
`quantized_qwen3_5.rs`. A checkpoint in a third format has no way in.

That third format is not hypothetical. NVIDIA's ModelOpt publishes Qwen 3.5
checkpoints as NVFP4 safetensors — packed E2M1 nibbles with per-block E4M3
scales — and the tensor names, the config, the layer topology and every
non-projection operation are byte-for-byte the ones `qwen3_5.rs` already
implements. The *only* thing that differs is how a `[rows, cols]` weight is
multiplied.

So a caller with such a checkpoint reimplements the whole file — the conv1d
short convolution, the GQA repeat, the L2 normalisation of q and k, the decay
and beta derivation, the gated RMSNorm, the whole `Qwen3_5DecoderLayer`
structure — to change one line of arithmetic. We did, and it is roughly a
thousand lines of duplicated model semantics whose only job is to be identical
to this file, including the risk that it silently stops being.

This is also what keeps `gated_delta_rule_chunked` out of reach for those
callers. That module is the good part — an O(T/chunk) prefill scan with a
sequential reference test — and a reimplementation cannot call it, because
everything around it that feeds it (conv, projections, head layout) is inside
the layer they could not reuse.

## Why the same seam as #3809

`llama.rs` had this exact problem and solved it: `BlockMlp` plus
`load_with_mlp_factory` let a caller substitute the feed-forward computation
while keeping every other line of the reference implementation. #3828 did the
same for the reduction in a tensor-parallel linear. The pattern is established
[...]

PR #3834, opened 2026-08-03
Qwen 3.5 projection factory

astorise:
Adds a projection seam to the dense Qwen 3.5 implementation so callers can
reuse Candle's model semantics with non-dense projection weights, such as
NVFP4 safetensors.

### Changes

- add a public `Projection` trait with a blanket implementation for Candle
  `Module`s;
- add `Model::new_with_projections`, accepting a factory for every attention,
  linear-attention, and MLP projection;
- retain the existing dense `Model::new` path unchanged;
- cover `q_proj`, `k_proj`, `v_proj`, `o_proj`, GatedDeltaNet projections, and
  MLP projections;
- add a CPU equivalence test: a factory returning dense `linear_no_bias`
  projections produces exactly the same output as `Model::new`.

This lets external projection implementations own their weights and device
placement while retaining Qwen 3.5's RoPE, norms, causal masking, KV cache,
GatedDeltaNet scan, and residual wiring.

### Validation

- `cargo test -p candle-transformers qwen3_5::tests::projection_factory_matches_dense_model --lib`
- `cargo clippy -p candle-transformers --lib -- -D warnings`

Related: #3832, #3831


astorise:

PR #3838, opened 2026-08-04
Add Qwen 3.5 sparse MoE and partial RoPE support

### LINKED ISSUE #3837: qwen3_5: the sparse-MoE layers and partial rotary the released checkpoints need
## Problem

Both Qwen 3.5 implementations in this crate — `qwen3_5.rs` and
`quantized_qwen3_5.rs` — are the **dense** variant. Neither mentions experts:

```
$ grep -ic expert candle-transformers/src/models/qwen3_5.rs
0
$ grep -ic expert candle-transformers/src/models/quantized_qwen3_5.rs
0
```

`Qwen3_5MLP` is a plain SwiGLU over `intermediate_size`, and `TextConfig` has
no `num_experts`, `num_experts_per_tok`, `moe_intermediate_size`,
`shared_expert_intermediate_size`, or `partial_rotary_factor`.

The published Qwen 3.5 checkpoints are mixture-of-experts with a shared expert.
So the model in this file is not the model the weights describe, and the two
config gaps fail in different ways:

- **The MoE gap fails loudly.** A sparse layer stores
  `mlp.experts.{i}.{gate,up,down}_proj`, `mlp.gate` and `mlp.shared_expert.*`;
  `Qwen3_5MLP::new` asks for `mlp.gate_proj.weight` and gets a missing-tensor
  error. Unpleasant, but honest.

- **`partial_rotary_factor` fails silently.** `TextConfig` derives plain
  `Deserialize` with no `deny_unknown_fields`, so the field is dropped on the
  floor, and `Qwen3_5TextRotaryEmbedding::new` builds its table over
  `dim = cfg.head_dim()` — the whole head. A checkpoint that rotates only the
  first `head_dim * partial_rotary_factor` dimensions loads without a word and
  produces wrong numbers. This is the same shape of problem as the
  `attention_bias` hole fixed in #3832: the config says something the model
  never reads.

## Why this is worth closing now

#3832 gave the layers a projection seam and #3831 gave `candle-nvfp4-kernels` a
device-resident `Module`. Between them, a caller with a ModelOpt NVFP4
checkpoint can supply its own projections and run everything else — attention,
linear attention, `gated_delta_rule_chunked`, norms, cache, residuals — out of
this crate.

Everything except the layer the checkpoint actually has. So the seam currently
opens onto a model those checkpoints cannot be loaded into, and a caller is
back to reimplementing the file it was meant to reuse. Closing this gap is what
turns those two changes from a good API into a working path.

## What the sparse layer needs

The semantics below are the ones we run in production against real Qwen 3.5
NVFP4 checkpoints, with a scalar reference implementation and its tests.

1. **Router top-k over renormalised probabilities.** Softmax the router logits,
   take the top `num_experts_per_tok`, then renormalise the selected weights so
   they sum to one. Ties broken by ascending expert index, so a run is
   reproducible.
2. **A shared expert on every token**, over `shared_expert_intermediate_size`,
   scaled by `sigmoid` of the `mlp.shared_expert_gate` logit and added to the
   routed sum. It is not one of the `num_experts`.
3. **Routed experts over `moe_intermediate_size`**, which differs from the
   dense `intermediate_size` and is what sizes the expert weights.
4. **Grouping tokens by expert, not looping tokens.** Each expert's weights are
   read once for all the tokens routed to it in a chunk. Done per token, an
   expert's operator is re-read for every activation, and on a quantized
   checkpoint that means re-decoding its weights — which for us was the single
   largest cost in a prefill.

And for the rotary: `rotary_dim = (head_dim as f64 * partial_rotary_factor) as
usize`, rotating the leading `rotary_dim` values of each head and passing the
rest through. A missing field should default to `1.0` so dense checkpoints are
unaffected.

## Shape

Following `layer_types`, which already selects the token mixer per layer, the
natural form is a second arm on the same kind of switch:

```rust
enum FeedForward {
    Dense(Qwen3_5MLP),
    Sparse(Qwen3_5SparseMoeBlock),
}
```

with the projection seam from #3832 reaching the experts by name
(`experts.{i}.gate_proj`, `shared_expert.up_proj`, `gate`,
`shared_expert_gate`) exactly as it reaches the dense ones — so a quantized
checkpoint gets its experts through the same factory, with no new API.

Alternatively, a `BlockMlp`-style trait as `llama.rs` has (#3809) would let a
caller supply the whole mixture. I would rather have the mixture in the crate:
the routing rule, the renormalisation and the shared-expert gate are model
semantics, and every caller reimplementing them is how they drift.

## Note

Happy to submit the PR. The tests I would bring are the two that caught real
bugs for us: a batched mixture must equal running the same tokens one at a
time, bit for bit, so chunk size stays a performance knob; and each expert must
be invoked once per chunk rather than once per token, which is the property the
grouping exists for and the one a refactor silently loses.

Related: #3809, #3831, #3832.

astorise:
Fixes #3837.

This PR adds the pieces required to load and run released Qwen 3.5 sparse-MoE checkpoints.

## Changes

- add Qwen 3.5 support and the injectable projection seam required for non-dense projections;
- add sparse MoE feed-forward blocks:
  - router softmax followed by top-k selection;
  - selected weights are renormalized;
  - deterministic ascending expert-index tie breaking;
  - routed tokens are grouped by expert, so each selected expert runs once per chunk;
  - add the shared expert, gated with `sigmoid(shared_expert_gate)`;
- add the MoE config fields:
  `num_experts`, `num_experts_per_tok`, `moe_intermediate_size`,
  and `shared_expert_intermediate_size`;
- honor `partial_rotary_factor` in the safetensors model, preserving the
  unrotated tail of each head;
- honor `qwen3.rope.dimension_count` in the GGUF quantized model, with the
  full head dimension retained as the backward-compatible default.

## Tests

- `cargo test -p candle-transformers qwen3_5`

The targeted tests cover partial RoPE tail preservation, deterministic router
tie-breaking, and grouped expert execution.


oetiker:
Hi @astorise — thanks for this, it's been genuinely useful to build on. We've been running your branch against a real Qwen3.6-35B-A3B Q4_K_M GGUF (~22 GB) on the Metal backend, and after the three fixes below we get token-for-token agreement with llama.cpp on the same checkpoint and prompt. Before them we did not. Offering the findings here in case they're useful; all three are small and cherry-pick cleanly onto this PR's head (`6d94f28d`).

One thing up front so the credential isn't overstated: the 35B run also needed quantized-MoE support that neither this PR nor candle currently has, so that end-to-end match is *this PR plus those three fixes plus additional local work*. The three fixes below are the parts that belong to your code.

---

### 1. `ssm_a` is already `-exp(A_log)` — the loader applies `-exp()` a second time

The GGUF converter pre-applies the transform: `conversion/qwen.py:381` does `data_torch = -torch.exp(data_torch)`, and llama.cpp then consumes the tensor verbatim (`qwen35.cpp:377`, `ggml_mul(alpha_softplus, ssm_a)`).

Checked on the checkpoint rather than inferred: **all 18 `ssm_a` tensors are strictly negative**, which a raw `A_log` would not be. Since `x ↦ −e^x` is monotonically decreasing, the second application both compresses every decay coefficient into (−1, 0) **and inverts their order** — the longest-memory head goes `-0.0014 → -0.9986`, 713× off.

It hides easily: the decay multiplies only the carried state, which is zeros for the first token, so a 1-token prompt matches llama.cpp to 3–4 decimals. The error compounds from the second token on.

→ [`fix/qwen35-ssm-a-double-exp`](https://github.com/oetiker/candle/tree/fix/qwen35-ssm-a-double-exp) (`cf2361ff`)

### 2. RoPE angle table is built in the model dtype (f16 by default)

`RotaryEmbedding::new` casts positions and `inv_freq` to `dtype`, which is f16 whenever the GGUF has no `general.dtype` key. The angle `position * inv_freq` grows without bound while f16's absolute resolution degrades with magnitude. Against an f64 reference at `rope_theta = 1e7`, position 5304 is off by **0.812 rad — a cos error of 0.727 on a quantity bounded in [−1, 1]**.

Real, but we should say plainly that it is **insufficient on its own** to explain the wrong output; fix 1 is the dominant term. `apply()` casting sin/cos to the activation dtype at use is fine and we left it alone — they're bounded by 1, so that costs ~1e−3.

→ [`fix/qwen35-rope-f32`](https://github.com/oetiker/candle/tree/fix/qwen35-rope-f32) (`8468a9a6`)

### 3. q/k broadcast is interleaved where ggml tiles

`ggml_repeat_4d` is cyclic (`dst[i] = src[i % src_ne]`), so v-head `h` reads k-head `h % num_k_heads` (`qwen35.cpp:441-445`). `repeat_interleave` gives `h / repeat_n` instead, permuting the pairing against the layout the converter wrote.

This **only bites when `num_v_heads != num_k_heads`**. It is provably a no-op at `repeat_n == 1`, so the 0.8B is unaffected; on the 35B, `group_count` 16 vs `inner_size/state_size = 4096/128 = 32` gives `repeat_n == 2`.

→ [`fix/qwen35-qk-broadcast-tiled`](https://github.com/oetiker/candle/tree/fix/qwen35-qk-broadcast-tiled) (`dc308dcc`)

---

### Limits of what we actually verified

- **Metal only.** The CPU arm on a 0.8B reaches 22/128. We chased that down and it is a knife-edge argmax tie, not a bug — a CPU gate needs a tie allowance to be meaningful. We have not built one.
- **One prompt, one probe.** The MoE router is exercised on exactly one routing pattern.
- The 41 `candle-transformers` tests pass on each branch, but **none of them touch the quantized GGUF loader**. That is non-regression, not correctness.

### Two notes on existing code, not on this PR

Separating these clearly since they're about files that predate your changes:

- `qwen3_5.rs:484` appears to build its angle table the same way, so the f16 argument in fix 2 would seem to apply there too. We have **not** measured it on the dense path and did not touch it.
- `qwen3_5.rs:209`'s `-exp(A_log)` is **correct** and should not be "fixed" — safetensors stores the untransformed `A_log`. The double application is specific to the GGUF path, and the asymmetry is the whole point.

---

Happy to open these as three PRs against your fork if that's easier to take than cherry-picking, or to rework any of them. Also glad to be told I've misread something — the 35B is the only checkpoint we have where fix 3 is live, so that one rests on a source comparison rather than an isolated measurement.

astorise:
Thanks for the deep dive and for testing against a real Q4_K_M checkpoint — very useful, and precise enough to act on directly.

All three fixes have been applied on top of this PR (cherry-picked from your branches, verified they still build and pass the existing qwen3_5 test suite):

ssm_a double -exp() — fixed in quantized_qwen3_5.rs, using the GGUF-converter-transformed tensor verbatim instead of re-applying -exp().
RoPE angle-table precision — RotaryEmbedding/Qwen3_5TextRotaryEmbedding now build the sin/cos table in f32 unconditionally, casting to the activation dtype only at use in apply(). Applied to both the quantized and the dense/safetensors path for consistency, since the dense path builds the table in the model dtype the same way (the ssm_a fix does not carry over there — the dense loader reads the genuine A_log from safetensors, not the GGUF-converter's pre-transformed ssm_a, so -exp(A_log) is correct as-is).
Q/K broadcast tiling — replaced repeat_interleave with a repeat_tiled helper matching ggml's cyclic dst[i] = src[i % src_ne] semantics, again in both the quantized and dense paths. No-op when num_v_heads == num_k_heads (e.g. the 0.8B), only diverges when repeat_n > 1 as you noted for the 35B.
On the credential caveat: agreed that quantized-MoE support is a separate, unaddressed gap — the quantized loader still explicitly rejects qwen35moe checkpoints. Opened [#3865](https://github.com/astorise/candle/issues/3865) to track adding it as a follow-up rather than scope-creeping this PR.

AlpineVibrations:
this sounds great. thanks for the work y'all.

astorise:
@oetiker your three fixes are in, but I squashed them into my own commits and you ended up with no attribution — sorry about that. I'll add Co-authored-by: oetiker <429279+oetiker@users.noreply.github.com> to the commits carrying them unless you'd rather I use a different address.

oetiker:
thanks!

PR #3840, opened 2026-08-04
fix(core): reject overflowing shapes/byte sizes in GGUF/GGML loaders

### LINKED ISSUE #3816: Shape::elem_count() silently wraps in release, producing a Tensor with corrupted metadata (dims disagree with backing storage)
## Summary
Loading a GGUF file with attacker/model-controlled dimensions that overflow `usize` (e.g. `[2^32, 2^32]`) causes `Shape::elem_count()` to silently wrap to a small value while `Shape::dims()` continues to report the original (huge) dimensions. The result is a `Tensor`/`QTensor` whose reported shape and actual backing storage disagree — a silently corrupted tensor, not a crash and not (as far as I can tell) an out-of-bounds access.

Not filed as a security report: every code path I could find that actually touches the mismatched data is bounds-checked and panics before any memory access happens (see "Why it never becomes unsafe" below), so I believe this is contained. Filing as a correctness/robustness issue because the corrupted `Tensor` state itself is visible and surprising.

## Root cause
`src/shape.rs:131`
```rust
pub fn elem_count(&self) -> usize {
    self.0.iter().product()      // unchecked usize product
}
```
This wraps in `usize` with no overflow check. **The wrap only exists in release builds** — a debug build panics at the multiply (`attempt to multiply with overflow`), so this is a release-only defect that debug testing will never catch.

## Reproduction
A 384-byte GGUF file declaring a tensor with `dims = [4294967296, 4294967296]`:
```
declared dims        = [4294967296, 4294967296]
Shape::elem_count()  = 0                          <- wrapped (2^32 * 2^32 mod 2^64 = 0)
QTensor accepted: storage_size_in_bytes = 0
dequantize() -> Ok
Tensor reports dims=[4294967296, 4294967296] but elem_count=0
```
The file is accepted, candle allocates 0 bytes, and hands back a `Tensor` claiming ~4.3 billion × 4.3 billion elements over an empty backing store. `dequantize()` returns `Ok` rather than an error.

The desync propagates through ordinary ops that all return `Ok` on the corrupted tensor: `flatten_all`, `t()`, `contiguous()`, `copy()`, `narrow()`, `reshape()`, `affine()`, `matmul()`. Sharpest case — `narrow()` followed by `contiguous()`:
```
base:      dims=[2^32, 2^32]  elem_count=0
narrowed:  dims=[1, 2^32]     elem_count=4294967296   <- no longer zero
contiguous() -> dims=[1, 4294967296]  elem_count=4294967296
  real storage len = 0 bytes
  peak RSS ~1.9 MB   <- no 16 GiB allocation ever happened
```
`narrow()` is the interesting step because it defeats the `elem_count() == 0` guard elsewhere: the narrowed view's element count no longer wraps to zero, even though the storage behind it is still empty.

## Why it never becomes unsafe (three independent reasons, so I'm not filing this as a vulnerability)
1. `elem_count()` is used for *both* buffer sizing and empty-tensor short-circuits (e.g. `tensor.rs:1516`: `if c_shape.elem_count() == 0 || k == 0 { return Tensor::zeros(...) }`), so the same wrap that corrupts the shape also routes most ops onto a no-op path. `matmul` never reaches gemm on the base (unwrapped-view) case.
2. Every CPU backend path that actually reads/writes the mismatched data panics on a Rust slice bounds check before any access — e.g. `narrowed.affine(...)` panics at `cpu_backend/utils.rs:331` with "range end index 4294967296 ... length 0", not a memory fault.
3. Confirmed with Miri (`-C overflow-checks=off` to force release semantics, `-Zmiri-disable-alignment-check` to isolate from an unrelated alignment issue I filed separately, #3815): no out-of-bounds access, no aliasing/provenance UB, across the read -> dequantize -> narrow -> contiguous -> flatten -> transpose -> matmul chain.

## Suggested fix
`checked_product` (or fold with `checked_mul`, returning a `Result`/erroring shape) in `Shape::elem_count()`, or reject a `Shape` at construction time whose dimension product overflows `usize`. Either closes the silent-corruption case; I don't think a panic-on-overflow is even necessary given containment above, but a `Result`-returning path would let GGUF loaders reject the file cleanly instead of accepting a tensor with impossible metadata.

Happy to share the crafted GGUF + a Miri test harness if useful.

CacinieP:
Closes #3816.

## Problem

`Shape::elem_count()` computes the dimension product with an unchecked `usize` multiply, which silently wraps in release builds. GGUF/GGML loaders feed attacker/model-controlled dimensions straight into this, so a tensor declaring `[2^32, 2^32]` is accepted with `elem_count == 0`, yielding a `Tensor` whose reported shape disagrees with its (empty) backing storage. As the reporter notes the corrupted state stays contained (every data path is bounds-checked), but the desync is surprising rather than a clean rejection.

The defect is **release-only**: a debug build panics at the multiply, so it is invisible to debug test runs.

## Fix

- **`Shape::elem_count_checked()`** — a new checked variant using `checked_mul`. A zero dimension short-circuits to `Ok(0)` before any multiply, so the result is order-independent (`[MAX, 2, 0]` and `[0, MAX, 2]` both return `Ok(0)`). `elem_count()` itself is left unchanged (105 call sites) but now documents the release-only overflow and points at the checked variant.
- **Loader hardening** — `gguf_file::TensorInfo::read` and the two `ggml_file` loaders switch to `elem_count_checked()`, and the trailing byte-size multiply becomes `checked_mul` too: a valid element count can still overflow once multiplied by `type_size`. The `read_one_tensor` byte-size computation was reordered to divide by `block_size` *before* the `checked_mul(type_size)`, so quantized dtypes (large `block_size`, small `type_size`) are not falsely rejected.
- **32-bit safety** — GGUF v2/v3 `u64` dimensions now go through `usize::try_from` instead of `as usize`, so a 32-bit / `wasm32` target rejects a `2^32` dimension rather than truncating it to `0`. (v1 `u32` stays `as usize`, which is always safe since `usize` is ≥ 32 bits.)
- **Distinct error variants** — `Error::ShapeElementCountOverflow` for the dimension product and `Error::TensorSizeOverflow` for the byte-size product, so the two failure modes carry accurate messages.

## Tests

- `elem_count_checked_matches_unchecked` — parity with the unchecked path, plus scalar/zero edge cases.
- `elem_count_checked_rejects_overflow` — `[usize::MAX, usize::MAX]` errors.
- `elem_count_checked_zero_dim_is_order_independent` — three orderings of `[MAX, 2, 0]`.
- `gguf_rejects_overflowing_shape` — end-to-end; asserts the specific `ShapeElementCountOverflow` variant (not just `is_err()`).
- `gguf_rejects_overflowing_byte_size` — F32 with `[usize::MAX / 4 + 1]` so the unchecked `* 4` would wrap to `0` in release; asserts `TensorSizeOverflow`. The dimension is chosen so the wrap reaches `0` (the file-length check would otherwise still catch the empty reader and mask a missing guard).

## Verification

- `cargo test -p candle-core` (debug) — all green, no regressions
- `cargo clippy -p candle-core --tests --examples --benches -- -D warnings`
- `cargo test -p candle-core --release --lib` — the 5 overflow tests pass with overflow checks disabled, confirming the guards hold in release builds

## Notes / limitations

- The 32-bit `usize::try_from(u64)` path is statically correct, but I did **not** run it on a `wasm32`/32-bit target locally (no such toolchain on this machine), and I could not find an explicit 32-bit job in candle's existing CI. Flagging this so CI can confirm or a maintainer can decide whether a dedicated 32-bit test is wanted.
- `elem_count()` is intentionally left unchecked to avoid a sweeping 105-call-site API change; only the untrusted-input (loader) paths are hardened.

### DIFF
--- a/candle-core/src/error.rs
+++ b/candle-core/src/error.rs
@@ -95,6 +95,14 @@ pub enum Error {
         n_parts: usize,
     },
 
+    #[error(
+        "element count overflows usize for shape {shape:?}; the dimension product does not fit"
+    )]
+    ShapeElementCountOverflow { shape: Shape },
+
+    #[error("tensor byte size overflows usize for shape {shape:?} with dtype {dtype:?}")]
+    TensorSizeOverflow { shape: Shape, dtype: String },
+
     #[error("{op} can only be performed on a single dimension")]
     OnlySingleDimension { op: &'static str, dims: Vec<usize> },
 
--- a/candle-core/src/quantized/ggml_file.rs
+++ b/candle-core/src/quantized/ggml_file.rs
@@ -141,14 +141,20 @@ pub fn qtensor_from_ggml(
     dims: Vec<usize>,
     device: &Device,
 ) -> Result<super::QTensor> {
-    let tensor_elems = dims.iter().product::<usize>();
+    let shape = crate::Shape::from(dims.as_slice());
+    let tensor_elems = shape.elem_count_checked()?;
     let block_size = ggml_dtype.block_size();
     if tensor_elems % block_size != 0 {
         crate::bail!(
             "the number of elements {tensor_elems} is not divisible by the block size {block_size}"
         )
     }
-    let size_in_bytes = tensor_elems / block_size * ggml_dtype.type_size();
+    let size_in_bytes = (tensor_elems / block_size)
+        .checked_mul(ggml_dtype.type_size())
+        .ok_or_else(|| crate::Error::TensorSizeOverflow {
+            shape: shape.clone(),
+            dtype: format!("{:?}", ggml_dtype),
+        })?;
 
     match ggml_dtype {
         GgmlDType::F32 => from_raw_data::<f32>(raw_data, size_in_bytes, dims, device),
@@ -211,8 +217,21 @@ fn read_one_tensor<R: std::io::Seek + std::io::Read>(
         reader.seek(std::io::SeekFrom::Current(((32 - pos % 32) % 32) as i64))?;
     }
     let dims = dims.iter().map(|&u| u as usize).collect::<Vec<_>>();
-    let tensor_elems = dims.iter().product::<usize>();
-    let size_in_bytes = tensor_elems * ggml_dtype.type_size() / ggml_dtype.block_size();
+    let shape = crate::Shape::from(dims.as_slice());
+    let tensor_elems = shape.elem_count_checked()?;
+    let type_size = ggml_dtype.type_size();
+    let block_size = ggml_dtype.block_size();
+    if tensor_elems % block_size != 0 {
+        crate::bail!(
+            "the number of elements {tensor_elems} is not divisible by the block size {block_size}"
+        )
+    }
+    let size_in_bytes = (tensor_elems / block_size)
+        .checked_mul(type_size)
+        .ok_or_else(|| crate::Error::TensorSizeOverflow {
+            shape: shape.clone(),
+            dtype: format!("{:?}", ggml_dtype),
+        })?;
     // TODO: Mmap version to avoid copying the data around?
     let mut raw_data = vec![0u8; size_in_bytes];
     reader.read_exact(&mut raw_data)?;
--- a/candle-core/src/quantized/gguf_file.rs
+++ b/candle-core/src/quantized/gguf_file.rs
@@ -91,14 +91,19 @@ impl TensorInfo {
         tensor_data_offset: u64,
         device: &Device,
     ) -> Result<QTensor> {
-        let tensor_elems = self.shape.elem_count();
+        let tensor_elems = self.shape.elem_count_checked()?;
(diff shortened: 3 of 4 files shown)


CacinieP:
Rebased onto latest main (the original CI run failed with no jobs executed, likely a workflow-file issue). Could a maintainer approve the workflow run? Thanks!

CacinieP:
Rebased onto the latest `main` (`fee1592d`) — no conflicts, and the new unit tests (`elem_count_checked_*`, `gguf_rejects_overflowing_*`) pass locally on top of it.

Note: the earlier `CI / cuda` run ([30927906486](https://github.com/huggingface/candle/actions/runs/30927906486)) never actually executed — it timed out after 30 days waiting for workflow approval since this PR comes from a fork. Could a maintainer approve the workflow run (or review directly)? Happy to make any changes.

PR #3845, opened 2026-08-05
Avoid unstable AArch64 FP16 vector type

jamesbrink:
## Summary

- store FP16 lane bits in stable `uint16x8_t` while inline assembly supplies floating-point semantics
- avoid `stdarch_neon_f16`, which fails on stable Rust when `target-feature=+fp16` is enabled
- preserve the existing optimized FMLA.8h implementation and register layout

## Test plan

- `RUSTFLAGS="-C target-feature=+fp16" cargo check -p candle-core` on aarch64-darwin with stable Rust

This also unblocks Nix desktop builds that target Apple Silicon FP16.

### DIFF
--- a/candle-core/src/cpu/neon.rs
+++ b/candle-core/src/cpu/neon.rs
@@ -157,7 +157,7 @@ mod fp16 {
         pub struct CurrentCpuF16 {}
 
         impl CurrentCpuF16 {
-            fn reduce_one(x: float16x8_t) -> f32 {
+            fn reduce_one(x: uint16x8_t) -> f32 {
                 let result: f32;
                 unsafe {
                     asm!(
@@ -185,8 +185,10 @@ mod fp16 {
         }
 
         impl CpuF16 for CurrentCpuF16 {
-            type Unit = float16x8_t;
-            type Array = [float16x8_t; Self::ARR];
+            // Keep the f16 lane bits in the stable integer vector type. The
+            // inline assembly assigns the floating-point interpretation.
+            type Unit = uint16x8_t;
+            type Array = [uint16x8_t; Self::ARR];
 
             const STEP: usize = 32;
             const EPR: usize = 8;


ivarflakstad:
`float16x8_t` is not unstable ([ref](https://doc.rust-lang.org/core/arch/aarch64/struct.float16x8_t.html))
I'm on a mac and have no issues running this code with stable rust.
Is this nix specific?

jamesbrink:
Thanks for flagging this. I verified it again on Apple Silicon, including an independent peer review.

You are correct that `float16x8_t` works on **current stable Rust**. This is not inherently Nix-specific; it is a Rust-version compatibility issue:

- Rust 1.93 rejects `float16x8_t` with `E0658: stdarch_neon_f16`.
- Rust 1.94.1 and 1.95 accept it.
- On this Apple Silicon host, `fp16` is a default target feature, so no custom `RUSTFLAGS` are required to reproduce the Rust 1.93 failure.

Reproduction from a clean checkout:

```bash
rustup toolchain install 1.93.0

git switch --detach 8b7aad6bdfb52bafa24cb5f6d3298a3e3c0f4702
rustc +1.93.0 --print cfg | grep fp16
cargo +1.93.0 check -p candle-core
```

The final command fails at `candle-core/src/cpu/neon.rs:160,188,189` with three `E0658` errors.

The same check passes at the PR head:

```bash
git switch --detach 8a9bc6bafa2d4330b84bea67ccefa49761bc3156
cargo +1.93.0 check -p candle-core
```

For an AArch64 target where FP16 is not enabled by default, the equivalent explicit reproduction is:

```bash
RUSTFLAGS="-C target-feature=+fp16" cargo +1.93.0 check -p candle-core
```

As a control, the unpatched base succeeds with Rust 1.94.1:

```bash
RUSTFLAGS="-C target-feature=+fp16" cargo +1.94.1 check -p candle-core
```

I also verified the patch itself: changing the Rust carrier from `float16x8_t` to `uint16x8_t` preserves the same 128 bits and 16-bit lanes. The inline assembly operands remain `vreg`, and the `.8h`/`.4h` instruction suffixes (`fadd`, `fmla`, `fcvtl`, `fcvtl2`, and `st1`) provide the FP16 interpretation. PR-head library tests passed 17/17 under Rust 1.93.

Candle currently has no declared `rust-version`, no pinned `rust-toolchain`, and CI tracks latest stable. So this patch is compatibility for Rust <=1.93, not a fix required by current stable. If supporting compilers older than latest stable is not intended, I am happy to close the PR.

ivarflakstad (approved):
Compelling argument. If there are no cons to the change, and it let's us support older rust versions, then this is strictly an improvement 👍

PR #3849, opened 2026-08-06
Add a ModelOpt weight source connecting candle-nvfp4-kernels to models

### LINKED ISSUE #3846: ModelOpt/NVFP4: nothing connects candle-nvfp4-kernels to the models that can use them
## Summary

candle now has both halves of NVFP4 inference, and nothing joins them.

- `candle-nvfp4-kernels` can **multiply** a packed operator: `Nvfp4Linear::new(packed_weight, weight_scale_e4m3, tensor_scale, rows, cols, device)`.
- `candle_transformers::models::qwen3_5` can **accept** a custom operator: the `Projection` trait plus `new_with_projections`.
- Nothing in candle can **find** a packed operator in a checkpoint.

`Nvfp4Linear::new` takes `&[u8]` for both the packed weight and the block scales. Every argument has to be produced by the caller: locate the shard from `model.safetensors.index.json`, read the safetensors header, group `<path>.weight` / `.weight_scale` / `.weight_scale_2` / `.input_scale`, check that the packed tensor is `[rows, cols/2]` and the scales `[rows, cols/BLOCK_SIZE]`, read the scalar tensor scale, and keep the packed bytes away from anything that would interpret them as `f32`.

That is a checkpoint-format concern, not an application concern, and every consumer will re-derive it identically.

## Proposal

A ModelOpt weight source in candle. Given a Hugging Face-style directory (sharded `.safetensors`, `model.safetensors.index.json`, `config.json` / `hf_quant_config.json`), it would:

1. detect NVFP4 from `W4A16_NVFP4` metadata, or from ModelOpt tensor names such as `.weight_scale_2`;
2. answer, for a linear path, what backs it — packed NVFP4, FP8, or plain dense;
3. build a `Box<dyn Projection>` suitable for `new_with_projections`;
4. degrade to a dense `Linear` where the kernel is unreachable, under an explicit memory bound.

Roughly:

```rust
let weights = ModelOptWeights::open(path)?;          // parses index + quant metadata
let vb = VarBuilder::from_mmaped_safetensors(&weights.shards(), dtype, &device)?;
let model = qwen3_5::ModelForCausalLM::new_with_projections(
    &cfg,
    vb,
    |path, vb, in_dim, out_dim, bias| weights.projection(path, vb, in_dim, out_dim, bias, &device),
)?;
```

Everything that is not a projection — embeddings, norms, the short convolution, `dt_bias`, `A_log` — keeps going through the ordinary `VarBuilder`, unquantized, exactly as it sits in the checkpoint.

## Why this should not stay per-model

`new_with_projections` exists on `qwen3_5` and nowhere else. Its factory signature — `Fn(&str, VarBuilder, usize, usize, bool) -> Result<Box<dyn Projection>>` — is architecture-independent, but the seam is not: the next architecture to get a ModelOpt release needs the same closure written again, with the same header parsing, the same tensor grouping and the same shape checks.

A weight source that answers *"what backs the linear at path P"* has no model in it. That the split is real is visible in practice: in our downstream adapter, the only model-specific code is a config translation, and every non-projection tensor is untouched.

## Prior art, and what we can contribute

We run exactly this downstream in [Tachyon-Mesh](https://github.com/astorise/Tachyon-Mesh): about 2.6k lines of ModelOpt directory parsing with 28 tests, plus a 417-line adapter that hands `qwen3_5` its projections. We are happy to upstream both, shaped however maintainers prefer.

Points worth stating up front, because each one cost us time:

- Packed FP4 bytes must never be read as `f32`. This is the failure mode that produces plausible-looking garbage rather than an error.
- Block size 16, scales in FP8 E4M3, one scalar `weight_scale_2` per operator.
- Nibble order is a real contract and deserves its own test — `nvfp4_encoding_matches_modelopt_layout` is the right shape, it just has no checkpoint above it.
- Checkpoints are **mixed**. A single graph carries FP8 projections, W4A16 NVFP4 projections and dense BF16 tensors. The source has to answer "dense" as legitimately as "packed"; a loader with no dense path fails on real files.
- Unpacking to dense is 8x the packed footprint. For a mixture-of-experts checkpoint that fits nowhere, so it needs an explicit memory limit rather than being a silent fallback.

## Testing

`candle-nvfp4-kernels` has three unit tests, all at the shape/encoding level, and there is no NVFP4 checkpoint anywhere in the test suite. A synthetic fixture would be cheap to add and would cover the loader end to end: every 4-bit code is a valid E2M1 value and E4M3 scales only need to be finite, so a tiny packed checkpoint can be generated without a quantizer and without trained weights. We are building one downstream and can contribute the generator.

## Open questions for maintainers

1. **Where does it live?** `candle-transformers`, next to the models, or a separate `candle-modelopt` crate — the parsing has no model dependency.
2. **Relationship to the GGUF path.** `quantized_var_builder` is the closest existing shape. Is a sibling `VarBuilder` backend preferable to a standalone type?
3. **Should `Projection` move out of `qwen3_5`** into a shared location, so other architectures can adopt the same seam without duplicating the trait?

## Context

Follows #3824 (kernel moved into candle), #3831 (compute-capability floor removed), #3832 and #3837.

astorise:
## Summary

Closes #3846.

`candle-nvfp4-kernels` already provides an `Nvfp4Linear` operator for NVIDIA
ModelOpt NVFP4 (W4A16) checkpoints, but nothing in Candle could locate a
packed NVFP4 operator inside a checkpoint: every consumer would have had to
parse tensor names, verify shapes, and extract packed bytes itself.

This adds the missing weight source, `candle_transformers::quantized_nvfp4`:

- **`detect_backing(vb)`** inspects a `VarBuilder`'s tensor names to tell
  dense, block-wise FP8, and packed NVFP4 layers apart, so a single mixed
  checkpoint (dense + FP8 + NVFP4 layers side by side) can be loaded
  layer-by-layer without special-casing.
- **`dequantize_nvfp4()`** decodes packed E2M1 nibbles and E4M3 block scales
  explicitly — never via a numeric dtype cast — so packed FP4 bytes are never
  misread as `f32`. It's bounded by an explicit `max_dense_bytes` budget, so
  an ~8x unpack fails loudly with a clear error instead of silently blowing
  past available memory.
- **`auto_linear()`** ties detection and construction together: on a CUDA
  device with the new `nvfp4-cuda` feature enabled, it keeps NVFP4 weights
  packed and device-resident via `candle-nvfp4-kernels::Nvfp4Linear`;
  otherwise it degrades gracefully to a dense `Linear` within the configured
  memory bound.

New `nvfp4-cuda` feature on `candle-transformers`, mirroring the existing
`fp8-cuda`/`gptq-cuda` pattern.

## Test plan

- [x] `cargo test -p candle-transformers --lib quantized_nvfp4` — 4 unit
      tests (dequantization correctness against a manual reference
      computation, E4M3 decode, backing detection, memory-budget rejection).
- [x] `cargo build`/`cargo fmt`/`cargo clippy -- -D warnings` clean across
      the touched crates.
- [x] `cargo test -p candle-nvfp4-kernels --features cuda` and
      `cargo test -p candle-transformers --features nvfp4-cuda quantized_nvfp4::`
      on a CUDA runner (device-resident `Nvfp4LinearCuda` path). This also
      caught and fixed a pre-existing test
      (`cuda_capability_signals_are_false_without_cuda_feature`) that only
      held without the `cuda` feature but wasn't gated on it, so it failed
      whenever run with `--features cuda` on real hardware.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/ci_cuda.yaml
+++ b/.github/workflows/ci_cuda.yaml
@@ -2,11 +2,6 @@ name: CI / cuda
 
 on:
   workflow_dispatch:
-    inputs:
-      test_ref:
-        description: 'Branch/tag/SHA to checkout and test (defaults to the dispatched ref)'
-        required: false
-        type: string
   pull_request:
 
 jobs:
@@ -14,11 +9,11 @@ jobs:
     concurrency:
       group: ${{ github.workflow }}-${{ github.job }}-${{ github.head_ref || github.run_id }}
       cancel-in-progress: true
-    runs-on: arc-gpu-candle
+    runs-on:
+      group: aws-g5-4xlarge-cache
     container:
       image: nvidia/cuda:13.0.2-cudnn-devel-ubuntu24.04
-      options: --gpus all
-    if: ${{ github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name }}
+    if: ${{ github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name }}
     permissions:
       contents: write
       packages: write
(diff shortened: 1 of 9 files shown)


astorise:

PR #3851, opened 2026-08-06
fix(metal): cap how far above its size a request may reuse a pooled buffer

oetiker:
## The problem

`find_available_buffer` returns the smallest **available** pooled buffer of at least the requested size — with no upper bound on how much larger that buffer may be:

```rust
if buffer_size >= &size && buffer_size < &best_buffer_size {
```

Since `buf_size` rounds every allocation up to a power of two, the pool is a set of size classes. A **long-lived** tensor that allocates at a moment when only much larger buffers happen to be free captures one for the rest of its life, and every later allocation of that larger size must then allocate a fresh buffer.

In LLM prefill this collides two tensors with very different lifetimes:

- **attention scores** — large, short-lived, freed and reallocated every layer;
- **the KV cache** — small, but alive for the whole forward, and *growing*, so it changes size class as the sequence grows.

## Measured instance

Qwen2.5-7B Q4_K_M on Metal, ctx 5831, prefill chunked at 512 (an M1 Max, 64 GB).

At chunk `pos=4096` the attention scores are `28 × 512 × 4608 × 4` = 264,241,152 B → the **256 MiB** class, and earlier chunks had already created and freed buffers there. In the same step the KV cache crosses a power-of-two boundary: the model is GQA (4 KV heads × 128 × f32 = 2048 B per token per layer), so each of the 56 K/V tensors grows from exactly 8 MiB at seq 4096 to 9,437,184 B at seq 4608. No 16/32/64/128 MiB buffer was free, so **44 KV-cache tensors were each served a 256 MiB buffer** (28.4× over-service) and held it permanently. The attention scores then had to allocate **46 fresh 256 MiB buffers**.

An allocation trace of that one chunk:

```
46 x new     requested=264241152  bucket=268435456
44 x reuse   requested=  9437184  served=268435456  waste=28.4x
44 x reuse   requested= 10485760  served=268435456  waste=25.6x
44 x reuse   requested= 11534336  served=268435456  waste=23.3x
```

The pool went from 8.51 GB to 20.01 GB in that single chunk — **11.3 GB of buffers holding 0.45 GB of data.**

Because the collision depends on *which size classes happen to be free* when the KV cache grows, the effect is not monotonic in the prefill chunk size — chunk 512 cost more than twice chunk 1024:

| prefill chunk | 128 | 256 | **512** | 1024 | 2048 |
|---|--:|--:|--:|--:|--:|
| activations, main | 2.17 | 5.04 | **15.12** | 6.81 | 12.86 |
| activations, this PR | 2.19 | 3.34 | **4.38** | 6.91 | 13.12 |

(GB; peak `currentAllocatedSize` during prefill minus the weights baseline.) So the conventional chunk size of 512 — the value llama.cpp and mlx use — was the worst available choice, and on a 32B model it was the difference between running and an out-of-memory abort.

## The fix

Cap reuse at 2 power-of-two size classes above the request. After it the curve is monotonic and no chunk size regresses by more than 2 %.

The cap is deliberately loose. Restricting reuse to the exact size class also removes the pathology, but costs **+36 % at chunk 128** (2.95 GB vs 2.19 GB) by forbidding harmless reuse; allowing only 1 class above costs +29 %. 2 classes is the knee of that curve.

Only power-of-two steps are representable here: every pooled buffer is allocated at `buf_size`, so the map is keyed by powers of two and a cap of `3 * buf_size` would admit exactly the same buffers as `2 * buf_size`. The knob is therefore a shift, and the swept values 1, 2, 4 are 0, 1, 2 classes above.

## Verification

- `cargo test --release --features metal -p candle-core` passes. (`conv2d_grad_noncontiguous_kernel_metal` fails, but it fails identically on unmodified `main` — unrelated to this change.)
- Generated token ids are byte-identical with and without the cap on Qwen2.5-0.5B and 7B, from one binary with a runtime override, so the change is numerically inert.
- The added test `test_allocator_does_not_serve_a_small_request_from_a_huge_free_buffer` exercises `allocate_buffer` rather than the helper, and was sabotage-checked: with the cap removed it is the only failure, reporting `9MB request was served a 268435456-byte buffer`.

Measurements were taken on `main` at ab9c3ab1 and on this branch, from separate builds, with the binary checksum recorded for each arm.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Run macos with metal
         if: matrix.os == 'macOS-latest' 
-        run: cargo check --workspace --features metal
+        run: cargo check --workspace --features metal --all-targets
 
       - name: Run normal cpu
         if: matrix.os == 'ubuntu-latest' || matrix.os == 'windows-latest'
--- a/candle-core/src/metal_backend/device.rs
+++ b/candle-core/src/metal_backend/device.rs
@@ -353,9 +353,9 @@ fn buf_size(size: usize) -> usize {
     size.next_power_of_two()
 }
 
-/// How many size classes above its own a request may be served from when reusing a
-/// pooled buffer. See [`reuse_cap`].
-const REUSE_SLACK: usize = 4;
+/// How many power-of-two size classes above its own a request may be served from when
+/// reusing a pooled buffer. See [`reuse_cap`].
+const REUSE_SLACK: u32 = 2;
 
 /// The largest pooled buffer a request of `size` bytes may be reused from.
 ///
@@ -368,8 +368,16 @@ const REUSE_SLACK: usize = 4;
 /// tensors of 9-12 MB each were served from the 256 MB class that the attention scores
 /// use, forcing 46 new 256 MB allocations. Peak activations 15.12 GB, against 4.38 GB
 /// with this cap in place, and the memory-vs-chunk-size curve was non-monotonic.
+///
+/// The cap is deliberately loose. Restricting reuse to the exact size class also removes
+/// the pathology, but costs +36 % at chunk 128 (2.95 GB against 2.19 GB) by forbidding
+/// harmless reuse; allowing only one class above costs +29 %. Two classes is the knee.
+///
+/// Every pooled buffer is allocated at [`buf_size`], so the map is keyed by powers of two
+/// and only power-of-two steps are representable: a cap of `3 * buf_size` would admit
+/// exactly the same buffers as `2 * buf_size`. Hence the shift.
 fn reuse_cap(size: usize) -> usize {
-    REUSE_SLACK.saturating_mul(buf_size(size))
+    buf_size(size).saturating_mul(1usize << REUSE_SLACK)
 }
 
 /// Applies the [`BufferBuilder`] label, clearing any stale label on a reused pooled buffer.
@@ -513,6 +521,9 @@ mod tests {
         // not force exact-size matching, which measured worse (2.95 GB vs 2.19 GB at
         // chunk 128) because it also forbids harmless reuse.
         assert!(reuse_cap(kv) >= buf_size(kv));
+        // 9 MB rounds to the 16 MB class, and REUSE_SLACK = 2 classes above it is 64 MB.
+        assert_eq!(buf_size(kv), 16 << 20);
+        assert_eq!(reuse_cap(kv), buf_size(kv) << REUSE_SLACK);
         assert_eq!(reuse_cap(kv), 64 << 20);
     }
 
--- a/candle-core/src/metal_backend/device.rs
+++ b/candle-core/src/metal_backend/device.rs
@@ -353,6 +353,25 @@ fn buf_size(size: usize) -> usize {
     size.next_power_of_two()
 }
 
+/// How many size classes above its own a request may be served from when reusing a
+/// pooled buffer. See [`reuse_cap`].
+const REUSE_SLACK: usize = 4;
+
+/// The largest pooled buffer a request of `size` bytes may be reused from.
+///
+/// Without a cap, [`find_available_buffer`] hands out the smallest *available* buffer of
+/// any size, however much larger than the request. A long-lived tensor that allocates at a
+/// moment when only much larger buffers happen to be free captures one for the rest of its
+/// life, and every later allocation of that larger size must then allocate a fresh buffer.
+///
+/// Measured on a Qwen2.5-7B Q4_K_M prefill (Metal, ctx 5831, chunk 512): 44 KV-cache
+/// tensors of 9-12 MB each were served from the 256 MB class that the attention scores
+/// use, forcing 46 new 256 MB allocations. Peak activations 15.12 GB, against 4.38 GB
+/// with this cap in place, and the memory-vs-chunk-size curve was non-monotonic.
+fn reuse_cap(size: usize) -> usize {
+    REUSE_SLACK.saturating_mul(buf_size(size))
+}
+
 /// Applies the [`BufferBuilder`] label, clearing any stale label on a reused pooled buffer.
 #[cfg(feature = "metal-debug-labels")]
 #[inline]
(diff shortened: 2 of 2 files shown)


4ktLuffy:
Nice write-up — the measured before/after and the fact that you tried exact-size matching and reported it *worse* (2.95 GB vs 2.19 GB at chunk 128) is what makes the choice of a cap convincing rather than arbitrary. `test_allocator_does_not_serve_a_small_request_from_a_huge_free_buffer` being called out as the gate that fails when the cap isn't consulted is also the right instinct; the pure `reuse_cap` tests would pass either way.

One thing on the units, since it will matter to whoever tunes this next.

`REUSE_SLACK` is documented as *"How many size classes above its own a request may be served from"*, but it is used as a plain multiplier: `reuse_cap = REUSE_SLACK * buf_size(size)`. Because `buf_size` rounds to a power of two, a multiplier of 4 is **2 size classes**, not 4:

```
request    9,437,184 B   (9 MiB)
buf_size  16,777,216 B   (16 MiB)  <- its own class
  +1                      32 MiB
  +2                      64 MiB   <- reuse_cap, = 4 * buf_size
```

which is consistent with the assertion in the test (`reuse_cap(kv) == 64 << 20`). So the behaviour is fine, but the name and doc comment describe a different quantity than the code computes. Someone who later wants "allow 3 classes" will reasonably write `REUSE_SLACK = 3` and get 1.58 classes.

Either wording works:

- keep the multiplier and say so — `REUSE_SLACK` is *"the largest multiple of its own size class a request may be served from"*; or
- make it a class count and shift instead — `buf_size(size) << REUSE_SLACK`, with `REUSE_SLACK = 2` to preserve today's behaviour.

Minor, but the second form would also make the empirical tuning easier to reason about later, since the knob and the unit in the measurements would then match.

One question rather than a request: `test_allocator_does_not_serve_a_small_request_from_a_huge_free_buffer` returns early when there's no Metal device, so on a non-Metal runner the gate passes without executing. Is metal CI running for this PR, or is that assertion effectively local-only for now?

oetiker:
> One question rather than a request: `test_allocator_does_not_serve_a_small_request_from_a_huge_free_buffer` returns early when there's no Metal device, so on a non-Metal runner the gate passes without executing. Is metal CI running for this PR, or is that assertion effectively local-only for now?

this was on my m1max ... no idea what the ci runs

4ktLuffy:
Had a look, since it is cheap to check.

The test lands in `candle-core/src/metal_backend/device.rs` inside `#[cfg(test)] mod tests`, so three things have to line up for CI to run it: the `metal` feature on, test targets actually built, and a Metal device present.

On main, in `.github/workflows/rust-ci.yml`:

* Check job, macOS (line 29): `cargo check --workspace --features metal`. The feature is on, but plain `cargo check` does not build test targets, so the test is not even type-checked. I confirmed that separately rather than assuming it: a deliberate type error inside a `#[cfg(test)] mod tests` gives 0 errors under `cargo check` and 2 under `cargo check --all-targets`.
* Test job, macOS (line 80): `cargo test --workspace`, with no `--features metal`. `lib.rs:70` gates `pub mod metal_backend` on that feature and `default = []`, so the whole module is compiled out. `cargo test -p candle-core --list` lists no metal tests at all.

So right now it neither compiles nor runs on CI, and your m1max is the only place it has ever executed.

The cheapest thing that keeps it from quietly rotting is `--all-targets` on the check step that already has the feature:

```yaml
- name: Run macos with metal
  if: matrix.os == 'macOS-latest'
  run: cargo check --workspace --features metal --all-targets
```

That gets it type-checked on every PR at no real risk. Actually executing the assertion would need the test job to enable `metal`, and I have not verified whether the GitHub macOS runners expose a usable Metal device, so I cannot promise that would do anything more than hit the early return.

oetiker:
Both fixed in acb9b910.

**Units.** You were right, and checking it turned out to constrain the design rather than just the wording. Every pooled buffer is allocated at `buf_size` (`device.rs:233`, `:316`), so the map is keyed by powers of two, and the cap is compared against those keys. That means intermediate multipliers are not merely awkward, they are dead values: `REUSE_SLACK = 3` admits exactly the same buffers as `2`, and `5..7` the same as `4`. So it is not that someone asking for 3 classes gets 1.58 — they get 1 class, with nothing to indicate the edit did nothing.

Taking your second option for that reason. `REUSE_SLACK: u32 = 2`, `buf_size(size).saturating_mul(1 << REUSE_SLACK)`. The cap is unchanged (64 MB for the 9 MB request), and now every integer value of the knob is a distinct setting.

`saturating_mul` stays rather than a bare `<<`: `test_reuse_cap_does_not_overflow` requires `reuse_cap(usize::MAX / 2) == usize::MAX`, and a plain shift there wraps to 0, which would silently disable all reuse — the opposite of the intended failure mode.

The measurements were swept over the multiplier, at 1, 2 and 4 only, so they map onto classes exactly, with no arm lost. The PR body is relabelled accordingly:

| swept `REUSE_SLACK` | classes above | result |
|---|--:|---|
| 1 | 0 (exact class) | 2.95 GB, +36 % at chunk 128 |
| 2 | 1 | +29 % |
| 4 | **2** ← shipped | 2.19 GB |

Same numbers, corrected labels. The tuning rationale now lives on `reuse_cap` alongside the cap itself.

**CI.** Confirmed your reading, and it is worth stating plainly for the record: the test has never run on CI and never even compiled there. `cargo check --workspace --features metal --all-targets` passes locally (exit 0, one pre-existing unused-import warning in another target), so I have taken the `--all-targets` change.

That gets it type-checked. Whether it *executes* still depends on the test job enabling `metal` and the runner exposing a device, neither of which I can verify — so for now the assertion remains local-only, on an M1 Max, and I would rather that be written down here than assumed.

**Scope, since it came up locally:** this pathology is Metal-only. `find_available_buffer`, `BufferMap` and the size-class rounding exist nowhere else in `candle-core` — CUDA allocates per tensor straight through `cudarc` (`cuda_backend/device.rs:78-89`) with no candle-side reuse pool, and CPU is a plain `Vec`. There is no sibling backend to port either the cap or the test to.

4ktLuffy:
@oetiker could you please check your email. Thank you!!

oetiker:
Rebased onto current `main` so this stops drifting.

- old head: `acb9b91037eddf2eca7edb48b96f37819ffc83dc`
- new head: `01d46285249e719a100014b03fc71fd711b2321a`
- new base: `6f74e7c390c717f8fd34f23ce02aceb058173370` ("Use shared locks for Metal cache hits", #3761)

The old head is preserved as the tag `backup/metal-buffer-reuse-cap-prerebase` on my fork, so nothing in this discussion has become unreachable.

In the interest of not overstating it: this was housekeeping, not a fix. The previous base was exactly one commit behind `main`, that commit touched only `candle-metal-kernels/src/kernel.rs`, and this PR touches `candle-core/src/metal_backend/device.rs` and the CI workflow — no overlap, and GitHub already reported it mergeable. I rebased so the PR is reviewed against what will actually be merged into, not because anything conflicted.

Both commits are patch-identical to their originals (verified with `git patch-id --stable`), so the substance of the change — capping reuse at `REUSE_SLACK = 2` size classes above the request — is untouched. `cargo check -p candle-core --features metal` is clean on the rebased branch; the one remaining warning (`unused import` in `metal_backend/mod.rs:7`) is pre-existing on `main` and in a file this PR does not touch.

@4ktLuffy no action needed from you on this — the review comments still apply unchanged.

PR #3856, opened 2026-08-06
Fix the Metal f16/f32 quantized matvec dispatch, add bf16

tomsanbear:
## Summary

The dense f16/f32 arm of `call_quantized_matmul_mv_t` does not work.

`kernel_mul_mv_impl` takes `r0 = tgpig.x` as the dst row, so it covers one row per threadgroup, and it addresses the weights as `offset0 = r0*nb01 + ...`. The host passed `nb00..nb02 = nb10..nb12 = 0` and `align = 8`:

- with zeroed strides every output row reads weight row 0;
- with `align = 8` the grid is `ceil(n/8)` wide, so seven eighths of `dst` is never written;
- the grid height ignores the kernel's `N_MV_T_T = 4` src1 tiling, so `m > 1` launches four times the threadgroups it needs. The kernel breaks on `r1 >= ne11`, so that one is wasted work rather than a wrong result.

`BF16` named `kernel_mul_mv_bf16_f32`, which the shader never instantiated, even though `kernel_mul_mm_bf16_f32` and `kernel_get_rows_bf16` both exist. bf16 weights worked for mat-mat and embeddings but fell over at mat-vec.

`Q8_1` and `Q8K` named kernels that do not exist either, but unlike bf16 these are not weight formats. In ggml they are the activation side of a dot product — `VecDotType` in `k_quants.rs` pairs Q4_1/Q5_1 with Q8_1 and the K-quants with Q8_K — and nothing in the shader references `block_q8_1` or `block_q8_K`, so there is no machinery to hook a kernel onto. `call_quantized_matmul_mm_t` and `call_quantized_get_rows` already reject both.

None of this is reachable through `QMatMul`, which dequantizes F32/F16/BF16 in `from_arc`. It is reachable by anything calling `candle-metal-kernels` directly.

## Fix

Real byte strides from the dtype's element size, one threadgroup per dst row, and a grid height divided by the src1 tiling.

Instantiate the existing `kernel_mul_mv` template for `bfloat` — one line, and the shader already requires `bfloat` unconditionally for `kernel_mul_mm_bf16_f32`.

Return `UnsupportedDTypeForOp` for Q8_1/Q8K from the name match, matching `call_quantized_matmul_mm_t` and `call_quantized_get_rows`.

## Test plan

Two tests in `candle-metal-kernels`, which is where the kernel can be dispatched directly.

`quantized_matmul_mv_float_weights` fills dst with NaN and checks every row for f32, f16 and bf16 against a reference that rounds the weights the same way the kernel does, so the tolerance stays tight across all three. On `main` the f32 case gives:

```
left:  [-0.5938, -0.5938, -0.5938, -0.5938, NaN, NaN, ... NaN]
right: [-0.5938,  0.5938,  0.3438, -1.3438, 2.7188, ...]
```

The first four rows are row 0 repeated, and the remaining 28 were never written.

`quantized_matmul_mv_rejects_activation_only_dtypes` covers Q8_1 and Q8K.

CI will not exercise either: `cargo test --workspace` on macOS carries no metal feature, and `candle-metal-kernels` is not a workspace member, so its test module never runs there. The shader is also compiled at runtime, so the new bf16 instantiation is only validated by running on a device.

Ran on an M3:

- `cargo nextest run --manifest-path candle-metal-kernels/Cargo.toml` (62/62)
- `cargo nextest run -p candle-core --features metal -E 'test(quantized)'` (34/34)

This does not touch the quantized dtypes. The mat-vec tail out-of-bounds behaviour for the K-quants is a separate fix.

### DIFF
--- a/candle-metal-kernels/src/kernels/quantized.rs
+++ b/candle-metal-kernels/src/kernels/quantized.rs
@@ -43,18 +43,10 @@ pub fn call_quantized_matmul_mv_t(
     let ne02 = b as i64;
     let ne03 = 1i64;
 
-    let nb00 = 0i64;
-    let nb01 = 0i64;
-    let nb02 = 0i64;
-
     let ne10 = k as i64;
     let ne11 = m as i64;
     let ne12 = b as i64;
     let ne13 = 1i64;
-
-    let nb10 = 0i64;
-    let nb11 = 0i64;
-    let nb12 = 0i64;
 
     let ne0 = n as i64;
     let ne1 = m as i64;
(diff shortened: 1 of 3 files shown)


4ktLuffy:
The diagnosis reads cleanly, and two things in the test plan are worth calling out because they are the parts that make it checkable:

Filling `dst` with `NaN` before dispatch is the right construction here. The failure mode is *rows never written*, and a zero-filled buffer would have hidden that behind a plausible-looking zero. The pasted `main` output showing row 0 repeated four times followed by 28 `NaN`s makes both halves of the bug visible at once — the zeroed strides and the `align = 8` grid.

And thank you for stating outright that CI will not exercise either test. I checked that independently and it holds: `candle-metal-kernels` is in the workspace `exclude` list, not `members`, so its test module is not reached by `cargo test --workspace` regardless of features.

That is the one thing I would raise, not as a change request on this PR but as a question about where it leaves the fix. Once this merges, the corrected dispatch is protected only by tests that no automated run will ever execute, so a future regression here comes back silently — which is roughly how the original got in. #3853 has the same shape from the other side (`test-cuda` skipped on fork PRs), so this does not look specific to Metal.

Is there any appetite for a metal-gated job that runs `candle-metal-kernels`' own tests, or for pulling it into the workspace so `--features metal` reaches it? Happy to look at that separately rather than load it onto this PR.

One small question on the Q8_1/Q8K arm: returning `UnsupportedDTypeForOp` matches `call_quantized_matmul_mm_t` and `call_quantized_get_rows`, which is clearly the right target. Was the previous behaviour a kernel-lookup failure at pipeline creation, i.e. is this strictly turning a late confusing error into an early clear one, or could anything have been relying on the old error type?

PR #3858, opened 2026-08-06
Add NVFP4 support with a packed CPU fallback (fixes #3857)

### LINKED ISSUE #3857: NVFP4 has no execution path off CUDA: no packed CPU operator, and no Metal backend like its sibling crates
## Summary

NVFP4 is the only quantized format in candle with no way to execute packed weights off CUDA.

- `candle-nvfp4-kernels` has exactly one feature: `cuda`. Its sibling crates each carry a `metal` one — `candle-fp8-kernels`, `candle-gptq-kernels` and `candle-awq-kernels` all declare `metal = ["candle/metal", "dep:candle-metal-kernels", "dep:objc2-metal"]` with an implementation in `src/metal.rs`, commented "Fused dequant+GEMM Metal kernel … for Apple Silicon". NVFP4 is the odd one out.
- `linear_f32_host` and `linear_f32_host_batched` look like a host path, but `host` describes the *buffers*, not the compute. Both open with `if !is_available() { return Err("NVFP4 CUDA kernels require an active CUDA device") }` and are `unreachable!()` without the `cuda` feature.
- The only non-CUDA path is `quantized_nvfp4::nvfp4_linear`, which dequantizes the whole operator to dense `f32` at load — roughly 8x the packed footprint, bounded by `Nvfp4Config::max_dense_bytes` (4 GiB by default).

So on a CPU host, or on Apple Silicon, an NVFP4 checkpoint either expands 8x or does not run. For a mixture-of-experts checkpoint, that means it does not run.

## What this asks for

**1. A CPU operator that consumes the packed weight.**

The same block walk `dequantize_nvfp4` already performs — unpack the nibble, apply the block scale and the tensor scale — but fused into the accumulation instead of materializing a dense tensor first. Plus a `Module` holding the packed bytes: the CPU twin of `Nvfp4Linear`.

The scalar reference is already in the tree. `e2m1_to_f32` and `e4m3fn_to_f32` sit in `candle-nvfp4-kernels/src/lib.rs` under `#[cfg(test)]`. Much of this is promoting them and writing the accumulation loop.

**2. An explicit preference order in `auto_linear`.**

Device-resident packed (CUDA today, Metal later) → packed CPU → dense dequantization only when explicitly asked for. Today the third is the only fallback that exists, which makes it the default on every non-CUDA host — silently, and at 8x.

**3. Room for `nvfp4-metal`, even if it lands later.**

Not asking for the Metal kernel in the same change. But the three sibling crates already establish the shape it should take, so the CPU operator wants to be introduced as *one backend among several* rather than as *the fallback*. Apple Silicon is where an NVFP4 checkpoint is most attractive — unified memory makes the 4-bit footprint the entire point — and it is currently the platform with the least NVFP4 support in candle.

## Why this matters beyond performance

**Memory is the blocker, not speed.** A packed CPU operator keeps a checkpoint at its on-disk footprint instead of 8x. That is the difference between "runs slowly" and "does not run at all".

**It also unlocks testing.** Today any test that executes an NVFP4 model needs a CUDA runner. `candle-nvfp4-kernels` has three unit tests, all at the shape/encoding level, and nothing anywhere executes a packed model. With a CPU operator, a tiny synthetic checkpoint exercises the whole path — weight source, projection factory, model forward — in ordinary CI. Such a fixture is cheap to build: every 4-bit code is a valid E2M1 value and E4M3 scales only need to be finite, so it needs no quantizer and no trained weights.

## Notes on the CPU implementation

- **Scalar first is fine.** The comparison is not against the CUDA kernel; it is against an 8x dense expansion that does not fit in memory. SIMD and rayon can come later, inside candle, where every consumer benefits rather than each downstream reinventing it.
- **Keep the accumulation order stable.** Parity tests naturally compare a packed run against a dense-dequantized run, and a bit-exact comparison is far more useful than one with a tolerance.
- **Match `Nvfp4Linear::forward`'s contract** — `[..., cols] @ dequant([rows, cols]).T`, preserving the leading dimensions — so the two are interchangeable behind `Projection` without the caller knowing which one it holds.

## Context

Follows #3846 (ModelOpt weight source).

Downstream in [Tachyon-Mesh](https://github.com/astorise/Tachyon-Mesh) we carry a scalar Qwen 3.5 runtime — a second implementation of an architecture candle already owns — for exactly this reason: our loader refuses a non-CUDA host rather than expand a MoE checkpoint 8x, so the reimplementation is the only thing that runs there. A packed CPU operator removes that duplicate outright, and the same argument will apply to anyone loading an NVFP4 checkpoint on a laptop.

astorise:
## Summary

Closes #3857.

NVFP4 (ModelOpt / compressed-tensors W4A16) currently has no execution
path off CUDA hardware in candle: the CUDA-only `candle-nvfp4-kernels`
crate provides a fused, device-resident operator, but any non-CUDA
host (CPU, Apple Silicon) has no way to run an NVFP4 checkpoint except
dequantizing every packed weight to a dense `f32` tensor at load time
— an 8x footprint expansion that OOMs or fails outright on exactly the
hosts this issue is about. As the issue states: memory is the
blocker, not speed.

This PR adds:

- `candle-nvfp4-kernels`: a CUDA crate exposing a fused NVFP4 linear
  operator (`Nvfp4Linear`) that keeps the packed weight
  device-resident and never materializes a dense tensor.
- `candle_transformers::quantized_nvfp4`: a ModelOpt/compressed-tensors
  aware weight source —
  - `detect_backing()` distinguishes dense, block-wise FP8, and packed
    NVFP4 layers from a `VarBuilder`'s tensor names alone, so mixed
    checkpoints load layer by layer.
  - `dequantize_nvfp4()` decodes packed E2M1 nibbles and E4M3 block
    scales explicitly (never via a numeric dtype cast), bounded by an
    explicit `max_dense_bytes` budget so an 8x unpack fails loudly
    instead of exceeding available memory.
  - **`Nvfp4LinearPacked`** (new in this PR): a CPU operator that keeps
    the packed weight and block scales resident in their
    checkpoint-native, ~1x-footprint form and decodes each output row
    on the fly during `forward`, instead of pre-expanding the whole
    matrix to dense `f32`. This is what actually resolves #3857 for
    non-CUDA hosts.
  - `auto_linear()` ties detection and construction together with an
    explicit execution preference order:
    1. `nvfp4-cuda` fused kernel — CUDA device, packed and
       device-resident.
    2. `Nvfp4LinearPacked` — the new default off CUDA: packed, ~1x
       footprint, decoded on the fly.
    3. Dense dequantization (`nvfp4_linear`) — opt-in via
       `Nvfp4Config::force_dense_fallback`, for callers who'd rather
       trade memory for forward-pass throughput and know the
       checkpoint fits within `max_dense_bytes`.

A Metal backend is still future work (tracked by the sibling-crate gap
noted in #3857) — `Nvfp4LinearPacked` is device-agnostic CPU code, so
on a Metal device it currently round-trips through host memory rather
than running on-device. That's a reasonable interim step: it turns an
outright failure/OOM into a working (if not fully accelerated) path,
and a future Metal kernel can slot into the same preference order
without changing `auto_linear`'s public API.

## Why a new operator instead of just relaxing the dense fallback

Raising or removing `max_dense_bytes` doesn't fix the underlying
problem — a 70B-parameter NVFP4 checkpoint still needs ~8x its packed
size in RAM once dequantized, whether or not you allow the allocation.
`Nvfp4LinearPacked` avoids the expansion entirely by never storing a
dense copy of the weight, at the cost of decoding each row on every
forward pass instead of once at load time. That tradeoff matches what
the issue asks for explicitly: "Memory is the blocker, not speed."

## Test plan

- [x] `cargo test -p candle-transformers quantized_nvfp4` — unit tests
      cover `detect_backing`, both ModelOpt and compressed-tensors
      naming conventions, the `max_dense_bytes` budget guard, and a
      new test asserting `Nvfp4LinearPacked` matches
      `dequantize_nvfp4`'s dense output on a batched, multi-dimensional
      input.
- [x] `cargo build -p candle-transformers -p candle-nvfp4-kernels`
- [x] `cargo fmt --check` / `cargo clippy --tests -- -D warnings` on
      the touched crates.
- [ ] CUDA fused path (`--features nvfp4-cuda`) exercised on a GPU
      runner — not available in this environment; the CUDA kernel
      itself is unchanged by this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
     "candle-core",
     "candle-datasets",
     "candle-examples",
+    "candle-nvfp4-kernels",
     "candle-nn",
     "candle-pyo3",
     "candle-transformers",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ candle-flash-attn-v3 = { path = "./candle-flash-attn-v3", version = "0.11.0" }
 candle-kernels = { path = "./candle-kernels", version = "0.11.0" }
 candle-metal-kernels = { path = "./candle-metal-kernels", version = "0.11.0" }
 candle-nn = { path = "./candle-nn", version = "0.11.0" }
+candle-nvfp4-kernels = { path = "./candle-nvfp4-kernels", version = "0.11.0", default-features = false }
 candle-onnx = { path = "./candle-onnx", version = "0.11.0" }
 candle-transformers = { path = "./candle-transformers", version = "0.11.0" }
 candle-ug = { path = "./candle-ug", version = "0.11.0" }
--- a/candle-nvfp4-kernels/Cargo.toml
+++ b/candle-nvfp4-kernels/Cargo.toml
@@ -11,7 +11,10 @@ license = "MIT OR Apache-2.0"
 
 [features]
 default = []
-cuda = []
+cuda = ["candle-core/cuda"]
+
+[dependencies]
+candle-core = { path = "../candle-core", version = "0.11.0", optional = false }
 
 [build-dependencies]
 anyhow = "1"
--- a/candle-nvfp4-kernels/Cargo.toml
+++ b/candle-nvfp4-kernels/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "candle-nvfp4-kernels"
+version = "0.11.0"
+edition = "2021"
+
+description = "CUDA kernels for NVIDIA ModelOpt NVFP4 quantized linear layers in Candle."
+repository = "https://github.com/huggingface/candle"
+keywords = ["blas", "tensor", "machine-learning", "cuda", "quantization"]
+categories = ["science"]
+license = "MIT OR Apache-2.0"
+
+[features]
+default = []
+cuda = []
+
+[build-dependencies]
+anyhow = "1"
+cudaforge = "0.1.2"
--- a/candle-nvfp4-kernels/build.rs
+++ b/candle-nvfp4-kernels/build.rs
@@ -1,0 +1,27 @@
+use std::path::PathBuf;
+
+fn main() -> anyhow::Result<()> {
+    println!("cargo::rerun-if-changed=build.rs");
+    if std::env::var("CARGO_FEATURE_CUDA").is_err() {
+        return Ok(());
+    }
+
+    println!("cargo::rerun-if-changed=kernels/nvfp4_gemm.cu");
+    let out_dir = PathBuf::from(std::env::var("OUT_DIR").expect("OUT_DIR not set"));
+    let mut builder = cudaforge::KernelBuilder::new()
+        .source_files(vec!["kernels/nvfp4_gemm.cu"])
+        .out_dir(&out_dir)
+        .arg("-std=c++17")
+        .arg("-O3");
+    if !std::env::var("TARGET").is_ok_and(|target| target.contains("msvc")) {
+        builder = builder.arg("-Xcompiler").arg("-fPIC");
+    }
+    builder.build_lib(out_dir.join("libnvfp4kernels.a"))?;
+    println!("cargo::rustc-link-search={}", out_dir.display());
+    println!("cargo::rustc-link-lib=nvfp4kernels");
+    println!("cargo::rustc-link-lib=dylib=cudart");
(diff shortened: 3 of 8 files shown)


astorise:

PR #3861, opened 2026-08-08
Decode NVFP4 block scales once per block, not once per column

### LINKED ISSUE #3860: Nvfp4LinearPacked decodes each block scale once per column, not once per block
## Summary

NVFP4 is the only quantized format in candle with no way to execute packed weights off CUDA.

- `candle-nvfp4-kernels` has exactly one feature: `cuda`. Its sibling crates each carry a `metal` one — `candle-fp8-kernels`, `candle-gptq-kernels` and `candle-awq-kernels` all declare `metal = ["candle/metal", "dep:candle-metal-kernels", "dep:objc2-metal"]` with an implementation in `src/metal.rs`, commented "Fused dequant+GEMM Metal kernel … for Apple Silicon". NVFP4 is the odd one out.
- `linear_f32_host` and `linear_f32_host_batched` look like a host path, but `host` describes the *buffers*, not the compute. Both open with `if !is_available() { return Err("NVFP4 CUDA kernels require an active CUDA device") }` and are `unreachable!()` without the `cuda` feature.
- The only non-CUDA path is `quantized_nvfp4::nvfp4_linear`, which dequantizes the whole operator to dense `f32` at load — roughly 8x the packed footprint, bounded by `Nvfp4Config::max_dense_bytes` (4 GiB by default).

So on a CPU host, or on Apple Silicon, an NVFP4 checkpoint either expands 8x or does not run. For a mixture-of-experts checkpoint, that means it does not run.

## What this asks for

**1. A CPU operator that consumes the packed weight.**

The same block walk `dequantize_nvfp4` already performs — unpack the nibble, apply the block scale and the tensor scale — but fused into the accumulation instead of materializing a dense tensor first. Plus a `Module` holding the packed bytes: the CPU twin of `Nvfp4Linear`.

The scalar reference is already in the tree. `e2m1_to_f32` and `e4m3fn_to_f32` sit in `candle-nvfp4-kernels/src/lib.rs` under `#[cfg(test)]`. Much of this is promoting them and writing the accumulation loop.

**2. An explicit preference order in `auto_linear`.**

Device-resident packed (CUDA today, Metal later) → packed CPU → dense dequantization only when explicitly asked for. Today the third is the only fallback that exists, which makes it the default on every non-CUDA host — silently, and at 8x.

**3. Room for `nvfp4-metal`, even if it lands later.**

Not asking for the Metal kernel in the same change. But the three sibling crates already establish the shape it should take, so the CPU operator wants to be introduced as *one backend among several* rather than as *the fallback*. Apple Silicon is where an NVFP4 checkpoint is most attractive — unified memory makes the 4-bit footprint the entire point — and it is currently the platform with the least NVFP4 support in candle.

## Why this matters beyond performance

**Memory is the blocker, not speed.** A packed CPU operator keeps a checkpoint at its on-disk footprint instead of 8x. That is the difference between "runs slowly" and "does not run at all".

**It also unlocks testing.** Today any test that executes an NVFP4 model needs a CUDA runner. `candle-nvfp4-kernels` has three unit tests, all at the shape/encoding level, and nothing anywhere executes a packed model. With a CPU operator, a tiny synthetic checkpoint exercises the whole path — weight source, projection factory, model forward — in ordinary CI. Such a fixture is cheap to build: every 4-bit code is a valid E2M1 value and E4M3 scales only need to be finite, so it needs no quantizer and no trained weights.

## Notes on the CPU implementation

- **Scalar first is fine.** The comparison is not against the CUDA kernel; it is against an 8x dense expansion that does not fit in memory. SIMD and rayon can come later, inside candle, where every consumer benefits rather than each downstream reinventing it.
- **Keep the accumulation order stable.** Parity tests naturally compare a packed run against a dense-dequantized run, and a bit-exact comparison is far more useful than one with a tolerance.
- **Match `Nvfp4Linear::forward`'s contract** — `[..., cols] @ dequant([rows, cols]).T`, preserving the leading dimensions — so the two are interchangeable behind `Projection` without the caller knowing which one it holds.

## Context

Follows #3846 (ModelOpt weight source).

Downstream in [Tachyon-Mesh](https://github.com/astorise/Tachyon-Mesh) we carry a scalar Qwen 3.5 runtime — a second implementation of an architecture candle already owns — for exactly this reason: our loader refuses a non-CUDA host rather than expand a MoE checkpoint 8x, so the reimplementation is the only thing that runs there. A packed CPU operator removes that duplicate outright, and the same argument will apply to anyone loading an NVFP4 checkpoint on a laptop.

astorise:
## Summary

`Nvfp4LinearPacked::forward` and `dequantize_nvfp4` (`candle-transformers/src/quantized_nvfp4.rs`)
both decode a block's `F8E4M3` `weight_scale` byte to `f32` **inside** the
per-column loop, instead of once per `block_size`-wide block:

```rust
for (col, val) in row_buf.iter_mut().enumerate() {
    let byte = packed_row[col / 2];
    let nibble = if col % 2 == 0 { byte & 0xf } else { byte >> 4 };
    let block_scale = scale_row[col / self.block_size].to_f32(); // redone every column
    *val = unpack_e2m1(nibble) * block_scale * self.weight_scale_2;
}
```

With `block_size = 16`, that's 16x more `F8E4M3::to_f32` conversions than
the format requires — the same cost as unpacking a nibble, redone on every
column of the block instead of once per block.

Fixes #3860.

## Change

Both loops now walk block-by-block: decode `block_scale` once per block,
then iterate only the columns inside that block reusing it. Behavior is
unchanged — this is a decode-cost fix, not a numeric one.

## Testing

- `cargo test -p candle-transformers --lib quantized_nvfp4` — all 8 tests
  pass, including `nvfp4_linear_packed_matches_dense_dequantization`
  (bit-exact packed-vs-dense parity) and
  `dequantize_nvfp4_matches_manual_computation`.
- `cargo clippy -p candle-transformers --lib` — no new warnings.
- `cargo fmt -p candle-transformers -- --check` — clean.

This only touches the CPU decode path (`Nvfp4LinearPacked` and the dense
`dequantize_nvfp4` fallback); the `nvfp4-cuda` fused kernel path
(`cuda::Nvfp4LinearCuda`) is untouched, so no GPU testing is needed for
this change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
     "candle-core",
     "candle-datasets",
     "candle-examples",
+    "candle-nvfp4-kernels",
     "candle-nn",
     "candle-pyo3",
     "candle-transformers",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ candle-flash-attn-v3 = { path = "./candle-flash-attn-v3", version = "0.11.0" }
 candle-kernels = { path = "./candle-kernels", version = "0.11.0" }
 candle-metal-kernels = { path = "./candle-metal-kernels", version = "0.11.0" }
 candle-nn = { path = "./candle-nn", version = "0.11.0" }
+candle-nvfp4-kernels = { path = "./candle-nvfp4-kernels", version = "0.11.0", default-features = false }
 candle-onnx = { path = "./candle-onnx", version = "0.11.0" }
 candle-transformers = { path = "./candle-transformers", version = "0.11.0" }
 candle-ug = { path = "./candle-ug", version = "0.11.0" }
--- a/candle-nvfp4-kernels/Cargo.toml
+++ b/candle-nvfp4-kernels/Cargo.toml
@@ -11,7 +11,10 @@ license = "MIT OR Apache-2.0"
 
 [features]
 default = []
-cuda = []
+cuda = ["candle-core/cuda"]
+
+[dependencies]
+candle-core = { path = "../candle-core", version = "0.11.0", optional = false }
 
 [build-dependencies]
 anyhow = "1"
--- a/candle-nvfp4-kernels/Cargo.toml
+++ b/candle-nvfp4-kernels/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "candle-nvfp4-kernels"
+version = "0.11.0"
+edition = "2021"
+
+description = "CUDA kernels for NVIDIA ModelOpt NVFP4 quantized linear layers in Candle."
+repository = "https://github.com/huggingface/candle"
+keywords = ["blas", "tensor", "machine-learning", "cuda", "quantization"]
+categories = ["science"]
+license = "MIT OR Apache-2.0"
+
+[features]
+default = []
+cuda = []
+
+[build-dependencies]
+anyhow = "1"
+cudaforge = "0.1.2"
--- a/candle-nvfp4-kernels/build.rs
+++ b/candle-nvfp4-kernels/build.rs
@@ -1,0 +1,27 @@
+use std::path::PathBuf;
+
+fn main() -> anyhow::Result<()> {
+    println!("cargo::rerun-if-changed=build.rs");
+    if std::env::var("CARGO_FEATURE_CUDA").is_err() {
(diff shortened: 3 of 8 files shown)


astorise:

PR #3862, opened 2026-08-08
fix(metal): surface command buffer errors that occur during execution

### LINKED ISSUE #3850: Metal: command buffer execution errors are silently ignored, producing wrong results instead of an error
## Summary

`Commands::ensure_completed` inspects a command buffer's status, waits, and then returns `Ok(())` without re-checking. Since the status it matched on is the one *before* the wait, a command buffer that fails **during execution** is never reported. Candle then goes on to read output buffers the GPU never wrote, so a hard error becomes silently wrong numbers.

I hit this as "quantized inference is broken on Metal": the same GGUF weights and the same prompt decoded correct German on CPU and emitted garbled Chinese from the first token on Metal, with **exit code 0 and no error of any kind**. It took a while to find that the GPU was aborting the work and Candle was not saying so.

## Environment

- `candle-core` / `candle-nn` / `candle-transformers` `0.11.0`, and also reproduced on `main` at `ab9c3ab`
- M1 Max, 64 GB, macOS 26.6 (25G72)
- Qwen2.5-32B-Instruct GGUF (Q4_K_M and Q4_0), `quantized_qwen2::ModelWeights`, 5831-token prompt

## The root cause

`candle-metal-kernels/src/metal/commands.rs`:

```rust
fn ensure_completed(cb: &CommandBuffer) -> Result<(), MetalKernelError> {
    match cb.status() {
        MTLCommandBufferStatus::NotEnqueued | MTLCommandBufferStatus::Enqueued => {
            cb.commit();
            cb.wait_until_completed();          // <-- may fail here
        }
        MTLCommandBufferStatus::Committed | MTLCommandBufferStatus::Scheduled => {
            cb.wait_until_completed();          // <-- or here
        }
        MTLCommandBufferStatus::Completed => {}
        MTLCommandBufferStatus::Error => { /* ...reported... */ }
        _ => unreachable!(),
    }
    Ok(())                                       // <-- status never re-checked
}
```

The `Error` arm only fires for a buffer that had *already* failed before `ensure_completed` looked at it. In the CPU-readback path (`flush_and_wait_current`) the buffer is freshly committed, so its pre-wait status is `Committed` — never `Error` — and any execution failure is discarded.

`flush_and_wait` has the same gap for the buffer it waits on: it error-checks `to_wait[..len-1]` explicitly, but the last one goes through `ensure_completed` and so is only checked before its wait.

## Evidence

The GPU was aborting the work the whole time. From `log show`, during exactly the runs that produced garbage — and during none of the runs that produced correct output:

```
qwen-spike (Metal) Execution of the command buffer was aborted due to an error during
execution. Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
kernel (IOGPUFamily) IOGPUCommandDescriptor::prepare: This command buffer is nopped
```

87 such aborts, all from the candle process, all `OutOfMemory`, perfectly correlated with the failing runs (32B at 5831 tokens) and absent from the passing ones (0.5B / 7B / 14B, and 32B with a short prompt).

For context on why it ran out — measured with a background thread polling `currentAllocatedSize()` on the same `MTLDevice`, taking the high-water mark across one prefill forward (`recommendedMaxWorkingSetSize` here is 51.8 GB, shared with the whole system):

| model | ctx | weights | peak | activations |
|---|--:|--:|--:|--:|
| 0.5B | 5831 | 1.02 GB | 8.95 GB | 7.93 GB |
| 7B | 5831 | 6.40 GB | 27.04 GB | 20.64 GB |
| 14B | 4096 | 11.28 GB | 36.66 GB | 25.38 GB |
| 32B | 5831 | 21.40 GB | — | OOM |

Activations dominate the weights well before the model is large. That part looks like the same behaviour as #3464 (transient outputs staying resident while command buffers are outstanding), measured on a bigger machine — but this issue is specifically that the resulting failure is **not reported**.

## Fix

Re-check the status after waiting: https://github.com/oetiker/candle/tree/fix/metal-cb-error-after-wait

Same binary, same weights, same prompt, 32B Q4_K_M on Metal:

| | result |
|---|---|
| before | garbled output from token 1, **exit 0** |
| after | `kIOGPUCommandBufferCallbackErrorOutOfMemory`, **exit 1** |

The healthy path is unaffected (0.5B at the same 5831-token context still decodes correctly, exit 0), and the patched run fails immediately at the first wait rather than grinding through the remaining layers.

Happy to open a PR if the approach looks right.

One caveat: I am not proposing a regression test. There is no public API to force a command buffer into `Error`, and the only trigger I know of is a genuine GPU OOM, which is not something CI can reasonably induce.

oetiker:
Closes #3850.

### The bug

`Commands::ensure_completed` inspects a command buffer's status, and then, for the
not-yet-finished states, waits:

```rust
fn ensure_completed(cb: &CommandBuffer) -> Result<(), MetalKernelError> {
    match cb.status() {
        NotEnqueued | Enqueued  => { cb.commit(); cb.wait_until_completed(); }
        Committed | Scheduled   => { cb.wait_until_completed(); }
        Completed               => {}
        Error                   => { /* ...reported... */ }
        _ => unreachable!(),
    }
    Ok(())        // status never re-checked
}
```

The status that is matched is the status *before* the wait. A buffer that was
`Committed` or `Scheduled` at the moment of the check can fail during
execution — the GPU aborts it, and every buffer queued behind it is nopped —
ending in `Error`. That transition happens inside `wait_until_completed()`, after
the match arm has already been chosen, so nothing looks at it and the function
returns `Ok`.

The `Error` arm therefore only ever fires for a buffer that had *already* failed
before `ensure_completed` looked at it. In the CPU-readback path
(`flush_and_wait_current`) the buffer is freshly committed, so its pre-wait status
is `Committed` — never `Error` — and any execution failure is discarded. The
caller then reads output buffers the GPU never wrote.

This also covers the gap #3850 notes in `flush_and_wait`: it error-checks
`to_wait[..len-1]` explicitly, but the last buffer goes through
`ensure_completed` and so was only ever checked before its wait.

### Why it matters

It turns a hard, diagnosable failure into wrong numbers. A GPU out-of-memory —
routine when a model does not fit — comes back as garbled output and **exit status
0**, with nothing anywhere saying the GPU aborted.

It is also a force multiplier on every other Metal bug: any kernel fault that
aborts the command buffer is reported as bad output rather than as a fault, so an
investigation starts by suspecting the numerics instead of the abort.

### The fix

Re-check the status after waiting, and factor the error construction so both
sites share it:

```rust
        MTLCommandBufferStatus::Error => return Err(Self::cb_error(cb)),
        _ => unreachable!(),
    }

    // The status matched above is the status *before* waiting. A buffer that was still
    // NotEnqueued/Enqueued/Committed/Scheduled can fail during execution -- the GPU aborts
    // it, and every buffer queued behind it is nopped -- ending in `Error`. Without this
    // second check the failure is silently swallowed and the caller goes on to read output
    // buffers the GPU never wrote, turning a hard error into wrong numbers.
    if cb.status() == MTLCommandBufferStatus::Error {
        return Err(Self::cb_error(cb));
    }

    Ok(())
}
```

No API change and no new error variant — `MetalKernelError::CommandBufferError`
already exists and already carries the message from `cb.error()`. The only
behavioural difference is that a failure which previously returned `Ok` now
returns that error.

### Testing, and what I did *not* test

- Compiles clean: `cargo check -p candle-metal-kernels` on macOS (M1 Max, macOS
  26.6), against `main` at `6f74e7c3`.
- **No automated test.** Provoking this needs a command buffer that fails *during*
  execution, which in practice means a real GPU OOM or a deliberately faulting
  kernel — neither belongs in the test suite. The change is a status re-read on a
  path that previously had none, so regression risk is confined to callers that
  were relying on errors being swallowed.
- **I have not re-verified the runtime symptom against this patch.** The symptom
  in #3850 — `kIOGPUCommandBufferCallbackErrorOutOfMemory` on a 32B quantized GGUF
  emitting garbled output with exit 0 on Metal, while the same weights and prompt
  decoded correctly on CPU — is as reported there. What I confirmed for this PR is
  the cherry-pick onto current `main` and the compile.

Happy to rework any of it.

### DIFF
--- a/candle-metal-kernels/src/metal/commands.rs
+++ b/candle-metal-kernels/src/metal/commands.rs
@@ -328,11 +328,6 @@ impl Commands {
             _ => unreachable!(),
         }
 
-        // The status matched above is the status *before* waiting. A buffer that was still
-        // NotEnqueued/Enqueued/Committed/Scheduled can fail during execution -- the GPU aborts
-        // it, and every buffer queued behind it is nopped -- ending in `Error`. Without this
-        // second check the failure is silently swallowed and the caller goes on to read output
-        // buffers the GPU never wrote, turning a hard error into wrong numbers.
         if cb.status() == MTLCommandBufferStatus::Error {
             return Err(Self::cb_error(cb));
         }
--- a/candle-metal-kernels/src/metal/commands.rs
+++ b/candle-metal-kernels/src/metal/commands.rs
@@ -324,17 +324,28 @@ impl Commands {
                 cb.wait_until_completed();
             }
             MTLCommandBufferStatus::Completed => {}
-            MTLCommandBufferStatus::Error => {
-                let msg = cb
-                    .error()
-                    .map(|e| e.to_string())
-                    .unwrap_or_else(|| "unknown error".to_string());
-                return Err(MetalKernelError::CommandBufferError(msg));
-            }
+            MTLCommandBufferStatus::Error => return Err(Self::cb_error(cb)),
             _ => unreachable!(),
         }
 
+        // The status matched above is the status *before* waiting. A buffer that was still
+        // NotEnqueued/Enqueued/Committed/Scheduled can fail during execution -- the GPU aborts
+        // it, and every buffer queued behind it is nopped -- ending in `Error`. Without this
+        // second check the failure is silently swallowed and the caller goes on to read output
+        // buffers the GPU never wrote, turning a hard error into wrong numbers.
+        if cb.status() == MTLCommandBufferStatus::Error {
+            return Err(Self::cb_error(cb));
+        }
+
         Ok(())
+    }
+
+    fn cb_error(cb: &CommandBuffer) -> MetalKernelError {
+        let msg = cb
+            .error()
+            .map(|e| e.to_string())
+            .unwrap_or_else(|| "unknown error".to_string());
+        MetalKernelError::CommandBufferError(msg)
     }
 
     fn end_encoding(&self, encoder: ComputeCommandEncoder) {


candle-metal-kernels/src/metal/commands.rs
```diff
@@ -324,19 +324,30 @@ impl Commands {
...
-            }
+            MTLCommandBufferStatus::Error => return Err(Self::cb_error(cb)),
             _ => unreachable!(),
         }
 
+        // The status matched above is the status *before* waiting. A buffer that was still
+        // NotEnqueued/Enqueued/Committed/Scheduled can fail during execution -- the GPU aborts
+        // it, and every buffer queued behind it is nopped -- ending in `Error`. Without this
+        // second check the failure is silently swallowed and the caller goes on to read output
+        // buffers the GPU never wrote, turning a hard error into wrong numbers.
```
ivarflakstad:
```suggestion
```

ivarflakstad (commented):
Code changes look fine.

The PR description is a pain to read, so is the documentation. Way too verbose.

ivarflakstad (approved):
👍

PR #3867, opened 2026-08-08
Add a directory-level ModelOptCheckpoint API

### LINKED ISSUE #3866: ModelOpt checkpoints: the weight source stops at the projection; detection, shards and config are still on every consumer
## Summary

#3846 asked for a ModelOpt weight source. What landed covers the *projection*: given a `VarBuilder` already positioned at a linear layer, `quantized_nvfp4::auto_linear` reads the packed weight, the block scales and the tensor scale, and picks the operator. That is the hard half, and it works.

The other half is still on every consumer: deciding that a directory *is* a ModelOpt NVFP4 checkpoint, and reading the config it ships.

Concretely, before `auto_linear` can be called even once, a caller must:

- recognise NVFP4 from `W4A16_NVFP4` metadata in `config.json` or `hf_quant_config.json`, or from ModelOpt tensor names such as `.weight_scale_2`;
- read `model.safetensors.index.json`, resolve the shard set, and hand the right paths to `VarBuilder::from_mmaped_safetensors`;
- deserialize the checkpoint's own `config.json` into the model's `Config`.

None of that is application logic, and all of it is identical for every consumer.

## The sharpest symptom: two config types for one model

`qwen3_5::Config` and `TextConfig` both derive `serde::Deserialize`, so they look ready to read a checkpoint. They are not, for a ModelOpt export: the nesting and several field names differ from what the file carries. So a downstream ends up with a *second* config type that deserializes the real file, plus a translation function into candle's.

That is what we have. The translation is 35 lines of field-by-field copying whose only reason to exist is that the two shapes disagree:

```rust
pub(crate) fn upstream_config(config: &Qwen35MoeConfig) -> UpstreamConfig {
    UpstreamConfig { text_config: TextConfig {
        vocab_size: config.vocab_size,
        head_dim: Some(config.head_dim),
        hidden_act: candle_nn::Activation::Silu,   // not in the file
        attention_bias: false,                     // not in the file
        // … 20 more
    }}
}
```

Two of those fields are not in the checkpoint at all and have to be pinned by the caller from knowledge of the architecture — which is precisely the knowledge candle has and the caller should not need.

Any second consumer will write this same struct and this same function, and the two copies will disagree the first time a field is added upstream.

## What this asks for

A directory-level entry point beside the projection-level one. Roughly:

```rust
// Detects the format and reads what the directory states about itself.
let ckpt = ModelOptCheckpoint::open(path)?;          // None if not ModelOpt/NVFP4
let cfg: qwen3_5::Config = ckpt.model_config()?;     // the file's own config.json
let vb = ckpt.var_builder(dtype, &device)?;          // index + shards resolved

let model = qwen3_5::ModelForCausalLM::new_with_projections(
    &cfg, vb,
    |name, vb, in_dim, out_dim, bias| {
        Ok(Box::new(auto_linear(in_dim, out_dim, bias, vb.pp(name), cfg)?))
    },
)?;
```

The shape matters more than the names. Three things are being asked for, and they are separable:

1. **Detection** — is this directory a ModelOpt NVFP4 checkpoint? Today every consumer re-derives it from metadata keys and tensor-name suffixes.
2. **Shard resolution** — index to shard paths. Mechanical, and identical everywhere.
3. **Config deserialization** — the checkpoint's `config.json` into the model's `Config`, including the fields the file does not carry.

(3) is the one that removes a whole parallel type from downstreams. (1) and (2) are smaller but are pure duplication.

## Why this is candle's and not the application's

The line we have been drawing: candle owns the model, the kernels and the checkpoint format; the application owns policy. Detection, shard resolution and config parsing are all format facts. What legitimately stays with a consumer is which producer versions it has qualified, what it does when a checkpoint fails validation, and where it will run — none of which candle should decide.

Our own `modelopt_nvfp4.rs` is ~2.6k lines. Most of it is (1), (2) and the tensor grouping that `auto_linear` already made redundant; the part that is genuinely ours — a fail-closed compatibility profile that refuses a producer version we have not qualified — is a few hundred lines and belongs where it is.

## Notes from having written it downstream

- Detection has to accept both spellings of the packed weight (`weight` for ModelOpt, `weight_packed` for compressed-tensors) — the same distinction #3846 settled at the projection level applies to the directory-level probe.
- `hf_quant_config.json` is not always present; some exports carry the quantization block inside `config.json`. Both are real.
- A checkpoint is mixed: dense BF16, FP8 and W4A16 NVFP4 tensors in one graph. Detection must not require that every linear be NVFP4.
- Sharded and single-file checkpoints both occur; `model.safetensors.index.json` is absent in the second case.
- Fields like `hidden_act` and `attention_bias` are architecture facts rather than file contents. If `Config` deserialization owns them via `#[serde(default = …)]`, no consumer has to guess them.

## Context

Follows #3846 (projection-level weight source) and #3857 (packed CPU operator). Downstream in [Tachyon-Mesh](https://github.com/astorise/Tachyon-Mesh) we have just deleted a local reimplementation of Qwen 3.5 in favour of candle's; what remains model-specific on our side is almost entirely this loading gap and the policy around it.

astorise:
## Summary

- Adds `candle_transformers::modelopt_checkpoint::ModelOptCheckpoint`, a directory-level loader for NVIDIA ModelOpt NVFP4 checkpoint exports.
- Fixes #3866.

A ModelOpt NVFP4 export mixes dense, block-wise FP8, and packed NVFP4 tensors, and lays them out as either a single `model.safetensors` or several shards indexed by `model.safetensors.index.json`. Telling such a checkpoint apart from a plain dense one, resolving its shard paths, and parsing `config.json` into a model's own `Config` type were, before this, left to every consumer to reimplement.

`ModelOptCheckpoint::open(dir)` wraps those steps:

- `is_nvfp4()` checks `config.json`'s `quantization_config` first, falling back to scanning each shard's safetensors header (not its weight data) for a `weight_scale_2` tensor — the unambiguous per-tensor NVFP4 signal.
- `shard_paths()` resolves `model.safetensors.index.json`'s `weight_map` into deduplicated, sorted shard paths, or falls back to a single `model.safetensors`.
- `model_config::<T>()` deserializes `config.json` into any `T`; most model `Config` types already derive `serde::Deserialize` against the same HF-format `config.json` ModelOpt emits, so no per-model translation layer is needed.
- `var_builder()` builds a `VarBuilder` over the resolved shards.

```rust
let ckpt = ModelOptCheckpoint::open(path)?;
let cfg: SomeConfig = ckpt.model_config()?;
let vb = unsafe { ckpt.var_builder(dtype, &device)? };
```

## Test plan

- [x] `cargo test -p candle-transformers modelopt_checkpoint` — 7 unit tests covering shard resolution (single-file and indexed/sharded), detection (config-declared and tensor-name-declared), config deserialization ignoring unknown fields, and `VarBuilder` construction.
- [x] `cargo test -p candle-transformers --doc modelopt_checkpoint` — module doc example compiles.
- [x] `cargo clippy -p candle-transformers --lib --tests` — no warnings.
- [x] `cargo fmt -p candle-transformers -- --check` — clean.

### DIFF
--- a/candle-transformers/src/lib.rs
+++ b/candle-transformers/src/lib.rs
@@ -1,5 +1,6 @@
 pub mod fused_moe;
 pub mod generation;
+pub mod modelopt_checkpoint;
 pub mod models;
 pub mod object_detection;
 pub mod pipelines;
--- a/candle-transformers/src/modelopt_checkpoint.rs
+++ b/candle-transformers/src/modelopt_checkpoint.rs
@@ -1,0 +1,346 @@
+//! Directory-level loading for NVIDIA ModelOpt NVFP4 checkpoint exports.
+//!
+//! A ModelOpt NVFP4 export mixes dense, block-wise FP8, and packed NVFP4
+//! tensors, and can lay them out as either a single `model.safetensors` or
+//! several shards indexed by `model.safetensors.index.json`. Telling such a
+//! checkpoint apart from a plain dense one, resolving its shard paths, and
+//! parsing `config.json` into a model's own `Config` type were, before this
+//! module, reimplemented by every consumer (huggingface/candle#3866): a
+//! checkpoint's `config.json` carries fields a model's `Config` doesn't
+//! need, but most `Config` types already derive `serde::Deserialize`
+//! directly against the same HF-format `config.json` ModelOpt emits, so no
+//! per-model translation layer is actually required -- consumers just
+//! needed the detection and shard-resolution work done once.
+//! [`ModelOptCheckpoint`] wraps those steps behind one directory-level API:
+//!
+//! ```no_run
+//! # #[derive(serde::Deserialize)]
+//! # struct Config { hidden_size: usize }
+//! # fn run() -> candle::Result<()> {
+//! use candle_transformers::modelopt_checkpoint::ModelOptCheckpoint;
+//!
+//! let ckpt = ModelOptCheckpoint::open("path/to/checkpoint")?;
+//! let cfg: Config = ckpt.model_config()?;
+//! // Safety: the shard files must outlive the returned `VarBuilder`
+//! // unmodified -- see `VarBuilder::from_mmaped_safetensors`.
+//! let vb = unsafe { ckpt.var_builder(candle::DType::BF16, &candle::Device::Cpu)? };
+//! # let _ = (cfg, vb);
+//! # Ok(())
+//! # }
+//! ```
+
+use candle::{DType, Device, Result};
+use candle_nn::VarBuilder;
+use std::collections::HashSet;
+use std::path::{Path, PathBuf};
+
+/// A directory holding an NVIDIA ModelOpt export: `config.json` plus one or
+/// more `.safetensors` shards, optionally indexed by
+/// `model.safetensors.index.json`.
+#[derive(Debug, Clone)]
+pub struct ModelOptCheckpoint {
+    dir: PathBuf,
+    config: serde_json::Value,
+    shard_paths: Vec<PathBuf>,
+}
+
+impl ModelOptCheckpoint {
+    /// Opens `dir`, reading `config.json` and resolving its safetensors
+    /// shard(s) -- from `model.safetensors.index.json` if present, otherwise
+    /// a single `model.safetensors` -- without loading any tensor data.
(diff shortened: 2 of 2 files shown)


astorise:

PR #3869, opened 2026-08-09
qwen3_5: add dense Qwen 3.5 model support with load-time Config validation

### LINKED ISSUE #3868: qwen3_5::Config deserializes without checking its fields agree, so an incoherent one fails deep instead of at load
## Summary

`qwen3_5::Config` and `TextConfig` deserialize without checking that the fields agree with each other. An incoherent config is accepted, and the model is then built from it — so the failure surfaces deeper, at a tensor shape or a slice bound, in a message that names neither the field that was wrong nor the layer it broke.

Every consumer that wants an early, legible refusal writes the same checks. They are not policy: they are facts about what a buildable config is.

## The invariants

These are the ones a downstream ends up writing, in the order they bite:

| invariant | what goes wrong without it |
|---|---|
| `layer_types.len() == num_hidden_layers` | layers are built from a shorter or longer schedule than the model declares |
| `num_experts_per_tok >= 1` and `<= num_experts` | top-k selects more experts than exist, or none |
| `num_attention_heads % num_key_value_heads == 0` | GQA repeat factor is not integral |
| `linear_num_value_heads % linear_num_key_heads == 0` | see below — this one is now silent |
| `hidden_size`, `vocab_size`, `num_hidden_layers` non-zero | empty tensors, or an lm_head over no vocabulary |
| `head_dim` consistent with `hidden_size / num_attention_heads` when present | attention reshapes against the wrong split |

## Why the linear-attention head ratio matters more since 66bdacc

The Q/K broadcast in the gated delta net moved from `repeat_interleave` to `repeat_tiled`, matching `ggml_repeat_4d`'s cyclic `dst[i] = src[i % src_ne]`. That was the right fix for llama.cpp parity.

It also makes a wrong `linear_num_value_heads / linear_num_key_heads` ratio *quieter*. Interleaving a non-integral ratio tends to fail on a shape; tiling wraps, produces a full-size tensor, and returns numbers. A checkpoint whose config declares a ratio the tensors do not have will now generate plausible-looking output instead of erroring — which is the worst of the three possible behaviours.

Validating the ratio at config load is a one-line check that restores the loud failure the broadcast used to give for free.

## Two shapes this could take

**A `Config::validate(&self) -> Result<()>`**, called by `new`/`new_with_projections` before anything is built. Smallest change, no API break, and a consumer that wants to validate before constructing can call it directly.

**Or validation inside `Deserialize`**, via `#[serde(try_from = "…")]`. Stronger — an invalid `Config` becomes unrepresentable — but it changes what `serde_json::from_str::<Config>` does, so it is the more intrusive of the two.

Either is better than the status quo; the first is likely enough.

## The related half: naming what a projection failure broke

`new_with_projections` hands the factory a name and dimensions. When a factory fails — a missing tensor, a shape that disagrees — the error propagates as whatever the factory returned, and the layer and expert index that were being built are already out of scope by the time a human reads it.

Wrapping the factory's error with the path being resolved would turn

```
cannot find tensor model.layers.7.mlp.experts.113.up_proj.weight_scale_2
```

into something that also says which layer and expert of how many, and what kind of component was expected. Downstream we do this by hand and it is the difference between a five-minute and a fifty-minute diagnosis on a 40-layer, 256-expert checkpoint.

This is separable from the config validation above and lower priority.

## What is deliberately not being asked for

The producer-version pin. Downstream we refuse anything that is not `modelopt 0.44.0` because that is the version we have qualified — that is a statement about our test coverage, not about the checkpoint, and candle blocking a working export on our behalf would be wrong. Likewise the memory limits and the device policy: those belong to whoever is deploying.

The line being drawn is: candle refuses what it *cannot build*; the application refuses what it *has not qualified*.

## Context

Follows #3866 (directory-level `ModelOptCheckpoint`), which removed the detection and shard-resolution duplication. Config coherence is the piece still reimplemented downstream. In [Tachyon-Mesh](https://github.com/astorise/Tachyon-Mesh) it lives in a fail-closed compatibility profile; roughly a third of that profile is the checks above, and would go away if `Config` held them.

astorise:
## Summary

- Adds a dense (safetensors) implementation of **Qwen 3.5** (`candle-transformers/src/models/qwen3_5.rs`), including the hybrid full-attention / gated-delta-net linear-attention layers, sparse MoE feed-forward, and a `Projection` seam so callers can swap in custom projection backends (e.g. quantized linear layers) without duplicating the attention/cache/residual wiring.
- Adds `qwen3_5_linear_attn_scan.rs`, the chunked scan implementation backing the linear-attention token mixer.
- Adds `quantized_qwen3_5.rs`, a GGUF-loadable variant reusing the same architecture.
- Fixes `Config`/`TextConfig` for Qwen 3.5 to validate their invariants at load time instead of failing deep inside layer construction or a tensor op with a cryptic error.

Fixes #3868

## Problem

`qwen3_5::Config`/`TextConfig` deserialize without checking that their fields agree with each other. An incoherent config (e.g. `layer_types.len() != num_hidden_layers`, `num_experts_per_tok > num_experts`, `num_attention_heads` not a multiple of `num_key_value_heads`, mismatched linear-attention head ratios, or an invalid `partial_rotary_factor`/`head_dim` combination) previously surfaced as a panic or an opaque tensor-shape error somewhere inside layer/expert construction, with no indication of which config field was at fault.

## Changes

- `Config::validate()` / `TextConfig::validate()`: checks, with a message naming the offending field:
  - `layer_types.len() == num_hidden_layers`, and every entry is `"full_attention"` or `"linear_attention"`.
  - `num_attention_heads` is non-zero and a multiple of `num_key_value_heads`.
  - `head_dim`, when set, is non-zero; when unset, `hidden_size` must be a multiple of `num_attention_heads`.
  - When a `linear_attention` layer is present: `linear_num_key_heads`, `linear_num_value_heads`, `linear_key_head_dim`, `linear_value_head_dim`, `linear_conv_kernel_dim` are non-zero, and `linear_num_value_heads` is a multiple of `linear_num_key_heads`.
  - When `num_experts > 0`: `1 <= num_experts_per_tok <= num_experts` and `moe_intermediate_size != 0`; otherwise `intermediate_size != 0`.
  - `hidden_size`, `vocab_size`, `num_hidden_layers` are non-zero.
  - `partial_rotary_factor` combined with `head_dim` yields a valid, even, non-zero rotary dimension.
- `Model::new` and `Model::new_with_projections` call `cfg.validate()?` before any tensor op runs.
- Layer and expert construction errors are now wrapped with `.with_context(...)` naming the failing `layer_idx` / expert index, so a bad config on a large model (many layers/experts) is easy to localize.

## Testing

- `cargo test -p candle-transformers --lib qwen3_5::` — 13 passed (includes the new invariant tests: `validate_accepts_a_coherent_config`, `validate_rejects_layer_types_length_mismatch`, `validate_rejects_unknown_layer_type`, `validate_rejects_num_attention_heads_not_multiple_of_kv_heads`, `validate_rejects_num_experts_per_tok_above_num_experts`, `validate_rejects_zero_num_experts_per_tok_with_moe`, `validate_rejects_linear_heads_not_multiple`, `validate_rejects_zero_hidden_size`, and `model_new_reports_incoherent_config_at_load_time` asserting `Model::new` now fails fast with a message naming the bad field).
- `cargo fmt -p candle-transformers -- --check`
- `cargo clippy -p candle-transformers --lib` — no warnings on the changed file.


astorise:
Closing — I opened four overlapping PRs for Qwen 3.5 support instead of iterating on one branch. Closing all four; this work is moving to a separate crate.

PR #3873, opened 2026-08-10
Fix Metal rank-5+ strided reduce reading out of bounds

### LINKED ISSUE #3847: Metal: `Tensor::expand` backward intermittently produces NaN gradients under repeated execution
On the Metal backend, repeatedly running the backward of `Tensor::expand` (broadcasting a size-1 dimension) intermittently produces non-finite (NaN/inf) gradients. A single backward is correct; the corruption only appears when the identical backward is repeated, and it appears even with the input held fixed, so it is not value-dependent, it looks like a state/buffer issue in the Metal backend.

I found this as diverging training (NaN loss) in a grouped-query-attention transformer and bisected it down to a single `expand`.

# Minimal repro (one op)

A single `Var`, broadcast a size-1 dim with `expand`, mean-squared loss, backward — repeated. No linear/matmul/softmax/attention.

```rust
use candle_core::{DType, Device, Result, Tensor, Var};

fn nonfinite(t: &Tensor) -> Result<bool> {
    Ok(!t.to_dtype(DType::F32)?.sqr()?.sum_all()?.to_scalar::<f32>()?.is_finite())
}

fn main() -> Result<()> {
    let dev = Device::new_metal(0)?;          // Device::Cpu is finite every time
    let dt = DType::BF16;
    let mut bad = 0;
    for _ in 0..16 {
        let w = Var::from_tensor(&Tensor::randn(0f32, 1f32, (1usize, 1, 512, 64), &dev)?.to_dtype(dt)?)?;
        let mut nan = false;
        for _ in 0..80 {
            // broadcast the interior (unsqueezed) dim 1 -> 2, then reduce
            let e = w.as_tensor().unsqueeze(2)?.expand((1usize, 1, 2, 512, 64))?;
            let g = (e.sqr()?.mean_all()? * 0.5)?.backward()?;
            let gw = g.get(&w).unwrap();
            if nonfinite(gw)? { nan = true; break; }
            w.set(&(w.as_tensor() - (gw * 1e-3)?)?)?; // (removing this SGD step still reproduces)
        }
        bad += nan as usize;
    }
    println!("{bad}/16 trials produced non-finite gradients");
    Ok(())
}
```

`Cargo.toml`: `candle-core = "0.11.0"` (built with the `metal` feature).

## Result
Metal (bf16) `16/16`; CPU `0/16`.

# What's going on (bisected)

- Single backward is finite; repeated backward is not. One `expand` backward is always correct.
  Corruption appears only after the identical op runs many times (~tens of iterations).
- Not value/weight dependent. Holding the input fixed (no SGD update - the exact same backward
  every iteration) still reproduces (`16/16`). So it isn't an input that drives values to NaN.
- Only the interior broadcast dimension. `w(1,1,512,64).unsqueeze(2).expand((1,1,2,512,64))`
  reproduces; broadcasting a leading dim of the 4-D tensor,
  `w(1,1,512,64).expand((1,2,512,64))`, does not.
- CPU is always correct, in f32 and (Metal-only) bf16.

# dtype

- The one-op repro above reproduces in bf16 only (f32 is finite for the bare `expand`).
- In a realistic setting it also fails in f32: a single grouped-query attention block
  (`linear -> reshape -> expand (1 kv head -> 2) -> attention -> linear`) NaNs its backward on Metal
  in both bf16 and f32 (`16/16` each), CPU `0/16`. So `expand` is necessary (replacing it with
  plain MHA makes the bug vanish) but, in f32, not sufficient on its own — it needs the surrounding
  matmuls.

# Ruled out

Not grad-accumulation, the optimizer, RoPE/RMSNorm/QK-norm/FFN, embedding, cross-entropy, or a specific batch/seq shape — all removed and it still reproduces. Not bf16 overflow (the loss is convex and minimized; grads shrink). Removing the `expand` broadcast (plain MHA) makes it vanish.

# Environment

- candle version: 0.11.0
- Device: Apple M3 Pro (12-core), 18 GB unified memory
- OS: macOS 26.5.2
- rust 1.97.1

waamm:
Fixes #3847.

`get_strided_idx_fallback` in `reduce.metal` decomposes the linear index over the wrong dimensions. This affects all strided reduces (sum/min/max/argmin/argmax) on rank-5+ inputs.

Concrete example from the issue's repro: the backward of an interior-dim broadcast on a
`(1, 1, 2, 512, 64)` tensor is a rank-5 strided sum with kernel dims
`[1, 1, 512, 64, 2]` / strides `[65536, 65536, 64, 1, 32768]`. The fallback maps logical
index 65535 to storage offset 63 + 32704 + 65536 = 98303 (should be 65535) in a 65536-element buffer.

The `expand_grad` test is distilled from this repro; on an M4 Max it fails before the fix and passes after.

### DIFF
--- a/candle-core/tests/grad_tests.rs
+++ b/candle-core/tests/grad_tests.rs
@@ -38,6 +38,24 @@ fn sum_grad(device: &Device) -> Result<()> {
     Ok(())
 }
 
+fn expand_grad(device: &Device) -> Result<()> {
+    // Computing the gradient sums over a middle dim of a 5D tensor, which
+    // takes the strided-index fallback path in the Metal reduce kernels,
+    // see issue #3847.
+    let data: Vec<f32> = (0..24).map(|v| v as f32).collect();
+    let x = Var::from_vec(data, (1, 1, 4, 6), device)?;
+    let e = x.as_tensor().unsqueeze(2)?.expand((1, 1, 3, 4, 6))?;
+    let loss = (e.sqr()?.sum_all()? * 0.5)?;
+    let grads = loss.backward()?;
+    let grad_x = grads.get(&x).context("no grad for x")?;
+    // loss = 0.5 * sum over the 3 broadcast copies of x^2, so dloss/dx = 3.x
+    assert_eq!(
+        grad_x.flatten_all()?.to_vec1::<f32>()?,
+        (0..24).map(|v| 3. * v as f32).collect::<Vec<f32>>()
+    );
+    Ok(())
+}
+
 fn matmul_grad(device: &Device) -> Result<()> {
     let data: Vec<_> = (0..12).map(|i| i as f32).collect();
     let x = Var::from_slice(&data, (2, 2, 3), device)?;
@@ -587,6 +605,12 @@ test_device!(
     simple_grad_metal
 );
 test_device!(sum_grad, sum_grad_cpu, sum_grad_gpu, sum_grad_metal);
+test_device!(
+    expand_grad,
+    expand_grad_cpu,
+    expand_grad_gpu,
+    expand_grad_metal
+);
 test_device!(
     matmul_grad,
     matmul_grad_cpu,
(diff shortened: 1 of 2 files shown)


ivarflakstad (approved):
Nice catch!
If you need performant strided indexing of 5D tensors we could have uncommented `//case 5: return strided_indexer<5, IndexT>{dims, strides}(idx);`, but regardless the fallback needs to be correct.

PR #3875, opened 2026-08-11
Fix CPU argsort for offset tensor views

### LINKED ISSUE #3874: CPU argsort/sort_last_dim ignores the view's start offset, silently mis-sorting narrow'd tensors
## Summary

On the CPU backend, `sort_last_dim` / `arg_sort_last_dim` read their input from the **start of the underlying storage**, ignoring the view's start offset. For a view produced by `narrow` (or any op that yields an offset but stride-contiguous layout), the sort permutation is computed from the wrong storage window, while the `gather` that `sort_last_dim` performs afterwards respects the layout and reads the right window — so the result is the *correct* rows permuted by the *wrong* rows' orderings. The output looks plausible and no error is raised.

Metal and CUDA are not affected (they apply the layout offset). Observed on 0.9.2; the same code is present on current `main`.

## Reproduction

```rust

    // Row 0 is already ascending; row 1 is not.
    let t = Tensor::new(&[[1f32, 2., 3.], [30., 10., 20.]], &Device::Cpu)?;
    let row1 = t.narrow(0, 1, 1)?; // offset view; strides are contiguous

    let (sorted, idx) = row1.sort_last_dim(true)?;
    println!("row1                 : {:?}", row1.to_vec2::<f32>()?);
    println!("sort_last_dim values : {:?}", sorted.to_vec2::<f32>()?);
    println!("arg_sort indices     : {:?}", idx.to_vec2::<u32>()?);

    // .contiguous() does not help: the view's strides pass the check.
    let (sorted_c, _) = row1.contiguous()?.sort_last_dim(true)?;
    println!("after .contiguous()  : {:?}", sorted_c.to_vec2::<f32>()?);

    // Materializing through an elementwise op works around it.
    let (sorted_m, _) = row1.affine(1.0, 0.0)?.sort_last_dim(true)?;
    println!("after affine copy    : {:?}", sorted_m.to_vec2::<f32>()?);
    Ok(())
}
```

Output (candle-core 0.9.2, macOS, CPU):

```
row1                 : [[30.0, 10.0, 20.0]]
sort_last_dim values : [[30.0, 10.0, 20.0]]
arg_sort indices     : [[0, 1, 2]]
after .contiguous()  : [[30.0, 10.0, 20.0]]
after affine copy    : [[10.0, 20.0, 30.0]]
```

The argsort was computed on row 0 (`[1, 2, 3]`, identity permutation), then gathered from row 1.

## Root cause

In `candle-core/src/sort.rs`, `ArgSort::asort` iterates the storage slice from index 0:

```rust
sort_indexes
    .par_chunks_exact_mut(self.last_dim)
    .zip(vs.par_chunks_exact(self.last_dim))
```

`layout.start_offset()` is never consulted, and `cpu_fwd` passes the full storage slice. By contrast, the CUDA path slices by `layout.contiguous_offsets()` and the Metal path obtains the buffer through `buffer_o(...)`, both of which apply the offset.

Two things make it silent:

- `Layout::is_contiguous` checks strides only, so an offset-but-stride-contiguous view returns `true` and `.contiguous()` is a no-op — there is no way to normalize the view short of an elementwise copy.
- `sort_last_dim` follows the argsort with a `gather`, which handles the layout correctly — so the values returned are real values from the right rows, just mis-ordered, rather than obviously alien data.

## How we hit it

A model head's output narrowed along the sequence dimension (`out.narrow(1, train, t - train)`) fed into a per-row sort of predicted quantiles. Because the rows were *nearly* sorted already and neighboring rows had similar orderings, the corruption stayed below fp-level test tolerances until an out-of-distribution input exposed it.

## Suggested fix

Slice the CPU storage by `layout.contiguous_offsets()` before chunking, as the CUDA implementation does (bailing on non-contiguous layouts, which `cpu_fwd` already effectively assumes).

LegenDairy93:
## Summary

- make CPU `arg_sort_last_dim` sort only the tensor view's contiguous storage window
- add ascending and descending regression coverage for a narrowed tensor with a non-zero offset

This matches the existing CUDA path, which already applies `Layout::contiguous_offsets()` before sorting.

Fixes #3874

## Tests

- `cargo test -p candle-core --test tensor_tests`
- `cargo clippy -p candle-core --tests -- -D warnings`

### DIFF
--- a/candle-core/src/sort.rs
+++ b/candle-core/src/sort.rs
@@ -8,7 +8,11 @@ struct ArgSort {
 }
 
 impl ArgSort {
-    fn asort<T: crate::WithDType>(&self, vs: &[T], layout: &crate::Layout) -> Vec<u32> {
+    fn asort<T: crate::WithDType>(&self, vs: &[T], layout: &crate::Layout) -> Result<Vec<u32>> {
+        let vs = match layout.contiguous_offsets() {
+            None => crate::bail!("input has to be contiguous"),
+            Some((o1, o2)) => &vs[o1..o2],
+        };
         #[allow(clippy::uninit_vec)]
         // Safety: indexes are set later in the parallelized section.
         let mut sort_indexes = unsafe {
@@ -48,7 +52,7 @@ impl ArgSort {
                     })
                 });
         }
-        sort_indexes
+        Ok(sort_indexes)
     }
 }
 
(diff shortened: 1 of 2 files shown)


candle-core/src/sort.rs
```diff
@@ -8,7 +8,11 @@ struct ArgSort {
 }
 
 impl ArgSort {
-    fn asort<T: crate::WithDType>(&self, vs: &[T], layout: &crate::Layout) -> Vec<u32> {
+    fn asort<T: crate::WithDType>(&self, vs: &[T], layout: &crate::Layout) -> Result<Vec<u32>> {
+        let vs = match layout.contiguous_offsets() {
+            None => crate::bail!("input has to be contiguous"),
```
ivarflakstad, line 13: Technically it doesn't have to be contiguous, but we actually enforce this in `Tensor::arg_sort_last_dim` already so for now we don't have to support the strided case.

ivarflakstad (approved):
lgtm 👍

PR #3877, opened 2026-08-12
Fix repeat with zero factors

charpty:
`Tensor::repeat` treated a zero repeat factor as one.

```rust
let tensor = Tensor::new(&[[1u32, 2], [3, 4]], &Device::Cpu)?;
let repeated = tensor.repeat((0, 2))?;
assert_eq!(repeated.dims(), &[0, 4]);
```

Before this fix, the result shape was `[2, 4]`.

The fix returns an empty view for zero factors, updates ColPali to use `1` when preserving a dimension, and adds regression tests for `repeat` and empty-axis `meshgrid`.

### DIFF
--- a/candle-core/src/tensor.rs
+++ b/candle-core/src/tensor.rs
@@ -696,9 +696,11 @@ impl Tensor {
             self.clone()
         };
         for (idx, &repeat) in repeats.iter().enumerate() {
-            if repeat > 1 {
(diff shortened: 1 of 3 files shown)


ivarflakstad (approved):
Thanks! This gives `repeat` functional parity with pytorch 👍

Issue #3863, opened 2026-08-08
Metal: K-quant mul_mv kernels store past ne01; Q2K over-dispatches 2x on top of it


oetiker:
The Metal K-quant `mul_mv` kernels store past `ne01`. There appear to be two
independent causes: a missing bound check that affects the whole family, and a
dispatch-table mismatch specific to Q2K that makes it much worse. There is a
self-contained failing test on a branch cut from current `main`.

I have deliberately **not** written a fix — the two obvious fixes are not
equivalent and interact, so the choice looks like a maintainer's call.

**Environment:** Apple M1 Max, macOS 26.6, `main` at
`6f74e7c390c717f8fd34f23ce02aceb058173370`, debug profile, Metal backend.
Reproduction branch:
[`oetiker/candle@repro/metal-q2k-mv-store-overrun`](https://github.com/oetiker/candle/tree/repro/metal-q2k-mv-store-overrun)
(commits `e4f69b54` and `60d1c2b5`; adds two tests to
`candle-metal-kernels/src/tests.rs`, changes no kernel and no dispatch code).

### Defect 1 — missing store guard, family-wide

None of the K-quant `mul_mv` kernels bound-check their store:

```c
for (int row = 0; row < N_DST; ++row) {
    all_sum = simd_sum(sumf[row]);
    if (tiisg == 0) {
        dst[r1*ne0 + im*ne0*ne1 + first_row + row] = all_sum;   // quantized.metal:4667
    }
}
```

The legacy-quant kernels in the same file do carry the check, at `:2374` and
`:2551`:

```c
if (tiisg == 0 && first_row + row < ne01) {
```

Consequence: any `ne01` that is not a multiple of the kernel's row stride
overruns by up to `stride − 1` rows. This affects q3_K, q4_K, q5_K and q6_K.

### Defect 2 — dispatch-table mismatch, Q2K only

`kernel_mul_mv_q2_K_f32_impl` advances **8** rows per threadgroup —
`first_row = (r0 * N_SIMDGROUP + sgitg) * N_DST` with `N_SIMDGROUP == 2` and
`N_DST == 4`, both defined once at `quantized.metal:2306-2307` with no `#undef`
anywhere in the file. But `call_quantized_matmul_mv_t` gives Q2K `align = 4`
(`candle-metal-kernels/src/kernels/quantized.rs:76-83`) and dispatches
`ceil(ne01 / align)` threadgroups (`:115`, `divide` = `div_ceil` at `:359`).

[...]

PR #3880, opened 2026-08-13
Metal: bound-check K-quant mul_mv stores and fix the Q2K dispatch

Shreyas-jk:
## Problem

On Metal, the five reachable K-quant `mul_mv` kernels store past the end of `dst`. For Q2K
the overrun is as large as the output itself: at `ne01 = 12`, `kernel_mul_mv_q2_K_f32`
writes 12 `f32` values past the end.

`candle-metal-kernels` is `exclude`d from the workspace (`Cargo.toml:19`), so the tests
have to be run from the crate directory:

```
cd candle-metal-kernels
cargo test q2k_mul_mv_writes_past_ne01
```

Overrun in `f32` elements past `ne01` on the unpatched tree, measured on an Apple M4,
macOS 26.2, debug profile. Every cell matches `stride * ceil(ne01 / align) - ne01`:

| dtype | ne01=8 | 10 | 12 | 16 | 17 |
|-------|-------:|---:|---:|---:|---:|
| Q2K   | 8      | 14 | 12 | 16 | 23 |
| Q3K   | 0      |  2 |  0 |  0 |  3 |
| Q4K   | 0      |  2 |  0 |  0 |  3 |
| Q5K   | 0      |  2 |  0 |  0 |  3 |
| Q6K   | 0      |  0 |  0 |  0 |  1 |

The in-range rows are correct in every case; this is purely a write outside `[0, ne01)`.

Reported by @oetiker in #3863, with a self-contained reproduction branch. The two test
commits here are theirs, cherry-picked.

## Root cause

Two independent defects.

**1. The K-quant stores are not bound-checked.** The legacy quant kernels carry
`if (tiisg == 0 && first_row + row < ne01)` (`quantized.metal:2374` for
q4_0/q4_1/q5_0/q5_1, `:2551` for q8_0). The K-quant kernels store unconditionally:
`:4667` q2_K, `:4854` q3_K, `:4996` q4_K, `:5153` q5_K, `:5264` q6_K. Any `ne01` that is
not a multiple of the kernel's row stride overruns by up to `stride - 1` rows.

The comment at `:2308-2311` still says the template "does not guard against the number of
rows not being divisible by N_DST", which stopped being true when `:2374` was added. The
guard looks like it was retrofitted there and never propagated to the K-quants.

**2. Q2K's dispatch disagrees with its own kernel.** `kernel_mul_mv_q2_K_f32_impl` uses
`first_row = (r0 * N_SIMDGROUP + sgitg) * N_DST` (`:4596`) with `N_DST` 4 and
`N_SIMDGROUP` 2, and `nth0 * nth1 = 64` threads is two simdgroups, so one threadgroup
covers 8 rows. `call_quantized_matmul_mv_t` dispatched `ceil(ne01 / 4)` — twice the
threadgroups needed. Q2K is the only dtype where the table and the kernel disagree:

| kernel | row index | rows/simdgroup | simdgroups | stride | old `align` |
|--------|-----------|---------------:|-----------:|-------:|------------:|
| q2_K   | `(r0*2+sgitg)*4` | 4 | 2 | 8 | **4** |
| q3_K   | `(r0*2+sgitg)*2` | 2 | 2 | 4 | 4 |
| q4_K   | `r0*4`           | 4 | 1 | 4 | 4 |
| q5_K   | `(r0*2+sgitg)*2` | 2 | 2 | 4 | 4 |
| q6_K   | `2*r0+sgitg`     | 1 | 2 | 2 | 2 |

## On whether these are alternatives

@oetiker deliberately left this unpatched, on the grounds that the two fixes are not
equivalent and the choice looked like a maintainer's call. That framing deserves a direct
answer rather than a silent decision, so: I don't think it is a choice, and the reason is a
third thing that isn't in the issue.

Every one of these kernels derives its `src0` base pointer from the row index *before*
anything bounds-checks that index — q2_K at `:4596` → `:4597` → `:4604`, q6_K at `:5215` →
`:5222`. The surplus threadgroups therefore do not only store out of bounds, they read out
of bounds, and a store guard cannot bound a read.

The clearest evidence is the template that is already guarded. `mul_vec_q_n_f32_impl`
computes `offset0` from `first_row` at `:2339` and reads `x + ib + row*nb` at `:2366`,
with the guard only at the store on `:2374`. So "a store guard leaves the read unbounded"
is not a prediction about a hypothetical fix — it is a property of the fix this file
already carries.

With a correct dispatch that read is bounded by `stride - 1` rows, the same tail every
kernel here already tolerates. With Q2K's 2x over-dispatch it is not: it reads roughly
`ne01` rows past the weights. That is why the dispatch change is in this PR as a
correctness fix rather than as the ~2x launch saving it also happens to be.

Demonstrated rather than asserted: `q2k_mul_mv_output_does_not_depend_on_rows_past_ne01`
over-allocates `src0`, keeps rows `[0, ne01)` byte-identical between two runs, and poisons
only the rows past `ne01`. On the unpatched tree the output changes — `dst[12..24]` reads
`520320.0` under one poison byte and `0.0` under the other. The output depends on weight
rows the kernel should never have touched. (The buffer is over-allocated deliberately, so
the overread lands inside a real allocation and the test itself stays well-defined.)

One limit, stated plainly: that test proves the read happens, but it cannot prove the read
*survives* a store guard, because a guarded kernel discards the values and the read stops
being observable through `dst`. That half rests on the source (`:4604`, `:5222`, `:2339`).
I tried `MTL_SHADER_VALIDATION=1` and it did not flag anything, which I'm treating as
inconclusive rather than as evidence of absence. If you know a reliable way to observe a
discarded device-memory read on Metal, I'd like to add it as a test.

If you read it the other way and want these as two PRs, I'm happy to split it — say which
you'd rather review first and I'll cut it that way. My only concern with landing the guard
on its own is the ordering: after it, nothing observable detects the dispatch mismatch (the
table in Testing below shows my own overread test going green while the bug is still
there), so a follow-up PR would have no failing test to point at.

## Fix

- Add the `:2374` guard to the five K-quant stores. q6_K needs `row < ne01` rather than
  `first_row + row < ne01` because it stores `dst[... + row]` with `row = 2*r0 + sgitg`
  (`:5264`), so the guard is the same idea in a different shape. Cost is one compare in
  the `tiisg == 0` tail.
- Set Q2K's `align` to 8 so the dispatch matches the kernel's row stride. Only the Q2K arm
  changes; `align` is selected per dtype, so no other dtype's launch geometry moves.

Deliberately untouched:

- The ten `kernel_mul_mv_iq*` kernels share the unguarded store, and `iq4_nl:6238` guards
  in its loop condition while its otherwise-identical twin `iq4_xs:6333` does not. But
  `GgmlDType` has no iq variants and no Rust code references those kernels, so they are
  unreachable from candle. Left alone rather than widening the diff into dead code.
- The CUDA and CPU paths. This is a Metal kernel and a Metal dispatch table; neither is
  shared.
- Pre-existing clippy findings in this crate. `cargo clippy --all-targets` reports the
  same 13 locations on this branch as on `main`, none in the changed code.

One part is separable: extracting the geometry table into `mul_mv_dispatch_geometry` is
what makes defect 2 testable at all (see below), but if you'd rather have a one-line
`align` change, say so and I'll drop the extraction.

## Testing

**The threadgroup-count assertion is the important one.** Once the stores are guarded, an
`align` that is too small stops being observable: the surplus threadgroups compute rows
that are then discarded, and every output-correctness test goes green. I measured this
rather than assuming it — with the guards in and `align` back at 4, three of the four
tests pass, *including the overread test above*. Only
`mul_mv_dispatch_align_matches_kernel_row_stride` catches it:

| test | unpatched | store guard only | this PR |
|------|-----------|------------------|---------|
| `q2k_mul_mv_writes_past_ne01` (@oetiker) | fail | pass | pass |
| `kquant_mul_mv_store_overrun_survey` (@oetiker) | fail | pass | pass |
| `q2k_mul_mv_output_does_not_depend_on_rows_past_ne01` | fail | **pass** | pass |
| `mul_mv_dispatch_align_matches_kernel_row_stride` | fail | **fail** | pass |

So without that assertion, landing the store guard alone would make the dispatch mismatch
permanently undetectable — it would look fixed and stay broken. The assertion derives the
expected stride from the thread geometry the function itself returns, so changing
`nth0`/`nth1` is caught too, not just a hand-edited `align`.

Red to green, from the crate directory on an Apple M4, macOS 26.2, debug profile:

- `q2k_mul_mv_writes_past_ne01`: 10/10 fail unpatched, 10/10 pass patched.
- `q2k_mul_mv_output_does_not_depend_on_rows_past_ne01`: 10/10 fail unpatched, 10/10 pass
  patched. Deterministic, not flaky.
- `kquant_mul_mv_store_overrun_survey` now asserts instead of only printing; it reported
  the table above and passed silently, which is why it needed the assertion.
- Full crate suite: 65 passed, 0 failed.
- `cargo fmt -- --check` clean. The cherry-picked test commits needed reformatting; that
  is the only reason `tests.rs` shows formatting churn.

Worth flagging: none of this runs in CI. `candle-metal-kernels` is excluded from the
workspace, so `cargo test --workspace` never reaches it, and the Metal job in
`rust-ci.yml` only runs `cargo check`. These tests protect the code only for someone
running them locally in the crate directory.

### DIFF
--- a/candle-metal-kernels/src/kernels/quantized.rs
+++ b/candle-metal-kernels/src/kernels/quantized.rs
@@ -24,6 +24,47 @@ pub enum GgmlDType {
     BF16,
 }
 
+/// Threadgroup geometry for the `kernel_mul_mv_*_f32` family: `(nth0, nth1, align)`.
+///
+/// `align` must equal the number of `dst` rows a single threadgroup writes, because
+/// the caller dispatches `ceil(ne01 / align)` threadgroups. An `align` smaller than
+/// the kernel's real row stride over-dispatches, and the surplus threadgroups compute
+/// rows past `ne01` — reading `src0` out of bounds on the way, since every one of these
+/// kernels derives its `src0` base pointer from the row index before any bound check.
+///
+/// Row stride per kernel, from `metal_src/quantized.metal` (`N_DST` 4, `N_SIMDGROUP` 2):
+///
+/// | kernel | row index          | rows/simdgroup | simdgroups | stride |
+/// |--------|--------------------|----------------|------------|--------|
+/// | q2_K   | `(r0*2+sgitg)*4`   | 4              | 2          | 8      |
+/// | q3_K   | `(r0*2+sgitg)*2`   | 2              | 2          | 4      |
+/// | q4_K   | `r0*4`             | 4              | 1          | 4      |
+/// | q5_K   | `(r0*2+sgitg)*2`   | 2              | 2          | 4      |
+/// | q6_K   | `2*r0+sgitg`       | 1              | 2          | 2      |
+pub(crate) fn mul_mv_dispatch_geometry(dtype: GgmlDType) -> (usize, usize, usize) {
+    match dtype {
+        GgmlDType::Q4_0
+        | GgmlDType::Q4_1
+        | GgmlDType::Q5_0
+        | GgmlDType::Q5_1
+        | GgmlDType::Q8_0
+        | GgmlDType::Q8_1 => (8, 8, 8),
+        // Fixing a bug in Metal for GGML
+        // https://github.com/ggerganov/llama.cpp/blob/b8109bc0139f15a5b321909f47510b89dca47ffc/ggml-metal.m#L1576
+        //
+        // `align` was 4 here, but the kernel row stride is 8:
+        // `first_row = (r0 * N_SIMDGROUP + sgitg) * N_DST` with two simdgroups per
+        // threadgroup (nth0 * nth1 = 64 threads).
+        GgmlDType::Q2K => (2, 32, 8),
+        GgmlDType::Q4K => (4, 8, 4),
+        GgmlDType::Q3K | GgmlDType::Q5K => (2, 32, 4),
+        GgmlDType::Q6K => (2, 32, 2),
+        // Original implem uses rows
+        GgmlDType::F16 | GgmlDType::BF16 | GgmlDType::Q8K => (32, 1, 8),
+        GgmlDType::F32 => (32, 1, 8),
+    }
+}
+
 #[allow(clippy::too_many_arguments)]
 pub fn call_quantized_matmul_mv_t(
     device: &Device,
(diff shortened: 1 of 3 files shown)


Shreyas-jk:
@ivarflakstad This has no check runs on the head commit (afe62f7e), which I think means the workflows are waiting on approval. Could you kick them off when you get a chance?

Worth flagging in advance: candle-metal-kernels is excluded from the workspace, so CI won't exercise these tests even once it runs. They pass 10/10 from the crate directory on an M4 with cargo test --features metal, on the reporter's repro branch commits which are cherry-picked here so their authorship is preserved.

Shreyas-jk:
@ivarflakstad @EricLBuehler Quick process question: CI on this PR has never run. The head commit (`afe62f7e`) still has no check runs, so I think the workflows are waiting on first-time-contributor approval. Who can approve them? The PR can't really be evaluated until they run. Happy to rebase first if you'd rather run CI against current main.

PR #3883, opened 2026-08-14
Refs/heads/claude/candle issue 3882 nvfp4 fp8tensor

### LINKED ISSUE #3882: ModelOpt per-tensor FP8 falls through to a dense read, dropping `weight_scale`
## Where this is

`candle-transformers/src/quantized_nvfp4.rs` does not exist on `main` — it
arrives with the open NVFP4 series (#3846 → #3849, #3857 → #3858). This is
feedback on that series, raised as an issue so it can be tracked and referenced
rather than buried in a review thread, matching how #3846 and #3857 were
handled.

## Summary

`detect_backing` recognises two quantized layouts and treats everything else as
dense:

```rust
if vb.contains_tensor("weight_scale_2") {
    Backing::Nvfp4
} else if vb.contains_tensor("weight_scale_inv") {
    Backing::Fp8Block
} else {
    Backing::Dense
}
```

NVIDIA ModelOpt exports a third: **per-tensor FP8** — `weight` in `F8E4M3`
beside a single *scalar* `weight_scale` (plus an `input_scale` on
activation-quantized operators). It carries neither marker, so it takes the
`Dense` arm, and `auto_linear` hands it to `candle_nn::linear_b`, which reads
the `F8E4M3` tensor at the `VarBuilder`'s dtype and discards `weight_scale`.

Nothing errors. The layer loads, runs, and is wrong by a constant factor on
every element — typically `amax / 448`, the value the exporter divided by.

## Why it is reachable in this series

A ModelOpt Qwen 3.5 MoE checkpoint — the checkpoint family this series targets
— is mixed precision, and the split is structural rather than incidental:

| operator | layout |
|---|---|
| `mlp.experts.*`, `mlp.shared_expert.*`, `lm_head` | NVFP4, `weight_scale_2` |
| `self_attn.{q,k,v,o}_proj` | per-tensor FP8, scalar `weight_scale` |
| `linear_attn.{in_proj_qkv,in_proj_z,out_proj}` | per-tensor FP8, scalar `weight_scale` |

So on exactly the checkpoint the NVFP4 path was built for, the NVFP4 operators
work and every attention projection silently loses its scale.

The checkpoint states this itself: `hf_quant_config.json` carries
`quant_algo: "MIXED_PRECISION"` with a `quantized_layers` map naming `FP8` or
`W4A16_NVFP4` per operator. The information is present; the reader does not
look for it.

## The second-order trap

Detecting per-tensor FP8 on `weight_scale` alone is correct only with the
preference order made explicit, because two other layouts also carry that
tensor:

- **NVFP4** carries a per-block `weight_scale` alongside `weight_scale_2`.
  Testing `weight_scale` before `weight_scale_2` misclassifies NVFP4 as
  per-tensor FP8.
- **GPTQ/AWQ-style** layouts carry `weight_scale` *and* `weight_zero_point`.
  Once `weight_scale` means per-tensor FP8, such a layer stops falling through
  to dense and starts being read as FP8 — trading one silent misread for a
  worse one.

Which is the general point: `else → Dense` converts "I do not recognise this
format" into "I will compute something". A layer carrying quantization tensors
no backing accounts for should fail to load and name what it saw. A model that
refuses to load costs five minutes; a model that answers wrong is not noticed.

## Shape of the change

- `Backing::Fp8Tensor`, detected on `weight_scale` after `weight_scale_2`,
  `weight_scale_inv` and `weight_zero_point`.
- `fp8_tensor_linear`: read the `F8E4M3` weight, cast to `vb.dtype()`, apply
  the scale with `Tensor::affine` on-device. Unlike `fp8_block_linear` there is
  no per-block lookup, so `dequantize_fp8_block`'s host round-trip is not
  needed.
- `read_scalar_f32` is reused as-is: exporters have written the scale as both a
  rank-0 scalar and a rank-1 `[1]`, and it already tolerates both for
  `weight_scale_2`.
- `input_scale` stays deliberately unread — this dequantizes the weight only —
  and is documented as the attachment point for a future activation-quantized
  path, so its presence is never mistaken for a different layout.
- `weight_zero_point` fails to load, naming the tensor.

One API note for reviewers: refusing means `detect_backing` returns
`Result<Backing>`. It is `pub`, but within this series it has a single caller
(`auto_linear`) plus tests.

## Acceptance

- A layer with `weight: F8E4M3` and a scalar `weight_scale` builds an operator
  that **applies** that scale. The test that matters compares against an
  expected value: with a weight of all 1.0, an input of all 1.0 and scale `s`,
  the output must be `in_dim * s` and not `in_dim`. A test asserting only that
  the layer loads would have passed on the original bug.
- `weight_scale_2` still wins over `weight_scale`, so NVFP4 keeps classifying
  as NVFP4.
- `weight_zero_point` fails to load and names itself, rather than classifying
  as per-tensor FP8 or dense.
- Block-wise FP8 and NVFP4 classify exactly as before.
- A mixed-precision ModelOpt checkpoint loads with its NVFP4 and FP8 operators
  each on their own path, and neither on the dense one.

## Status

Implemented and tested against the above; the change is ready to fold into
#3858 or land as a follow-up, whichever the maintainers prefer.

## Related

#3650 and #3662 cover the block-wise FP8 variant; this is the per-tensor
sibling neither addresses. #3844 is about FP8 matmul rather than about how a
checkpoint's FP8 layout is classified.

astorise:
## Summary

Fixes #3882.

`detect_backing` only recognized two quantized layouts: NVFP4 (`weight_scale_2`) and block-wise FP8 (`weight_scale_inv`). NVIDIA ModelOpt's third layout — per-tensor FP8, a scalar `weight_scale` on an `F8E4M3` weight, used by ModelOpt's attention projections in mixed-precision NVFP4 checkpoints such as Qwen 3.5 MoE — fell through to `Backing::Dense`. `auto_linear` then read the `F8E4M3` weight at the `VarBuilder`'s dtype and silently dropped `weight_scale`: the layer loaded and ran, off by a constant factor (typically `amax / 448`) on every element, with no signal to the caller.

- Add `Backing::Fp8Tensor`, detected on `weight_scale` alone (after `weight_scale_2` and `weight_scale_inv`, which NVFP4/block-FP8 keep priority over), and `fp8_tensor_linear` to build it — an on-device affine rather than `fp8_block_linear`'s host round-trip, since a scalar scale needs no per-block lookup. `input_scale` is intentionally still unread: this only dequantizes the weight.
- Harden `detect_backing` against the next unrecognized layout: `weight_zero_point` (GPTQ/AWQ-style) now fails to load, naming the tensor, instead of silently falling back to `Dense` (or being misread as `Fp8Tensor`, since those layouts carry both `weight_scale` and `weight_zero_point`). `detect_backing` now returns `Result<Backing>`; it had exactly one caller (`auto_linear`) plus tests.

## Context

This builds on the `candle-transformers::quantized_nvfp4` module from #3858 (NVFP4 support with a packed CPU fallback) — this PR is branched from upstream `main` and carries that module's commits along so it applies and builds standalone; the last two commits are the actual fix for #3882.

## Test plan

- [x] `cargo test -p candle-transformers quantized_nvfp4` — 9 passed, including `fp8_tensor_linear_applies_weight_scale` and `detect_backing_reads_tensor_names` (weight_zero_point rejection, per-tensor FP8 detection/dispatch, and a value-correctness test asserting the scale actually multiplies the output).
- [x] `cargo fmt -p candle-transformers -p candle-nvfp4-kernels -- --check`
- [x] `cargo clippy -p candle-transformers -p candle-nvfp4-kernels --lib --tests -- -D warnings`

### DIFF
--- a/.github/workflows/trufflehog.yml
+++ b/.github/workflows/trufflehog.yml
@@ -13,3 +13,10 @@ jobs:
           fetch-depth: 0
       - name: Secret Scanning
         uses: trufflesecurity/trufflehog@6bd2d14f7a4bc1e569fa3550efa7ec632a4fa67b  # main
+        with:
+          # New-branch pushes have no base commit, so trufflehog falls back to
+          # scanning full history. That resurfaces two known false positives on
+          # candle-pyo3/tests/native/test_tensor.py (Python `test_...` function
+          # names shaped like Lob API keys, e.g. test_tensor_can_be_constructed_from_list)
+          # inherited from upstream history. Lob is unrelated to this project.
+          extra_args: --exclude-detectors=Lob
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
     "candle-core",
     "candle-datasets",
     "candle-examples",
+    "candle-nvfp4-kernels",
     "candle-nn",
     "candle-pyo3",
     "candle-transformers",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ candle-flash-attn-v3 = { path = "./candle-flash-attn-v3", version = "0.11.0" }
 candle-kernels = { path = "./candle-kernels", version = "0.11.0" }
 candle-metal-kernels = { path = "./candle-metal-kernels", version = "0.11.0" }
 candle-nn = { path = "./candle-nn", version = "0.11.0" }
+candle-nvfp4-kernels = { path = "./candle-nvfp4-kernels", version = "0.11.0", default-features = false }
 candle-onnx = { path = "./candle-onnx", version = "0.11.0" }
 candle-transformers = { path = "./candle-transformers", version = "0.11.0" }
 candle-ug = { path = "./candle-ug", version = "0.11.0" }
--- a/candle-nvfp4-kernels/Cargo.toml
+++ b/candle-nvfp4-kernels/Cargo.toml
@@ -11,7 +11,10 @@ license = "MIT OR Apache-2.0"
 
 [features]
 default = []
-cuda = []
+cuda = ["candle-core/cuda"]
+
+[dependencies]
+candle-core = { path = "../candle-core", version = "0.11.0", optional = false }
 
 [build-dependencies]
 anyhow = "1"
--- a/candle-nvfp4-kernels/Cargo.toml
+++ b/candle-nvfp4-kernels/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "candle-nvfp4-kernels"
+version = "0.11.0"
+edition = "2021"
+
+description = "CUDA kernels for NVIDIA ModelOpt NVFP4 quantized linear layers in Candle."
+repository = "https://github.com/huggingface/candle"
+keywords = ["blas", "tensor", "machine-learning", "cuda", "quantization"]
+categories = ["science"]
+license = "MIT OR Apache-2.0"
+
+[features]
+default = []
+cuda = []
+
+[build-dependencies]
+anyhow = "1"
+cudaforge = "0.1.2"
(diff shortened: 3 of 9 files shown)


astorise:

PR #3885, opened 2026-08-14
fix(cuda): name the missing kernel and module in symbol-lookup errors

astorise:
### What

`CudaDevice::get_or_load_func` surfaces a bare cudarc `"named symbol not
found"` when a kernel lookup fails, with no indication of *which* kernel or
module was being looked up.

This is a real diagnostic gap: dtypes like `F8E4M3` have kernels with
non-uniform architecture guards, so on some devices (e.g. `sm_86`) certain
ops work while others fail lookup — a supported, expected configuration
rather than an exceptional one. Without the kernel name, "symbol not found"
gives no signal for diagnosing which op/dtype combination is missing.

### Changes

- `candle-kernels::Module` now carries a human-readable `name` (e.g.
  `"affine"`, `"binary"`, …) alongside its PTX source, exposed via
  `Module::name()`.
- `CudaError::MissingKernel` now carries `kernel_name`, `module_name`, and
  keeps the underlying `DriverError` in the source chain via `#[source]`,
  instead of just `module_name`. New message:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ candle-transformers = { path = "./candle-transformers", version = "0.11.0" }
 candle-ug = { path = "./candle-ug", version = "0.11.0" }
 clap = { version = "4.2.4", features = ["derive"] }
 criterion = { version = "0.8", default-features = false }
-cudarc = { version = "0.19.8", features = [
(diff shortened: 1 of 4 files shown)


astorise:

Issue #3893, opened 2026-08-15
`conv2d_grad_noncontiguous_kernel_gpu` fails on CUDA: PR #3736 fixed the CPU offset bug but not the GPU path


astorise:
## Title
`conv2d_grad_noncontiguous_kernel_gpu` fails on CUDA: PR #3736 fixed the CPU offset bug but not the GPU path

## Summary

#3736 (fixing #3735, *"CPU conv2d backward produces incorrect gradients for non-contiguous kernels"*) added both a CPU and a **GPU** test case in the same commit, but only patched `candle-core/src/cpu_backend/conv2d.rs`. The GPU variant of the test it introduced, `conv2d_grad_noncontiguous_kernel_gpu`, currently fails on CUDA with a real numeric mismatch — the equivalent GPU-side bug (or a related one) was never fixed.

## Reproduction

Run on a CUDA build at or after `21cca0b1` (main), e.g. on commit `bd4cdad3` of `main`:

```
cargo test --features cuda -p candle-core --test conv_tests conv2d_grad_noncontiguous_kernel_gpu
```

Failure output:

```
---- conv2d_grad_noncontiguous_kernel_gpu stdout ----

thread 'conv2d_grad_noncontiguous_kernel_gpu' panicked at candle-core/tests/conv_tests.rs:963:5:
assertion `left == right` failed
  left: 25.926
 right: 57.6732
```

The `_cpu` and `_metal` variants of the same test pass; only `_gpu` (CUDA) fails.

## Test in question

`candle-core/tests/conv_tests.rs`, `conv2d_grad_noncontiguous_kernel` (lines 937–973), registered for all backends via `test_device!` at lines 1017–1020:

```rust
test_device!(
    conv2d_grad_noncontiguous_kernel,
    conv2d_grad_noncontiguous_kernel_cpu,
    conv2d_grad_noncontiguous_kernel_gpu,
    conv2d_grad_noncontiguous_kernel_metal
);
```

The test builds a non-contiguous kernel (a channel slice of a larger weight, `weight.i((.., 1..3, .., ..))`, i.e. non-zero start offset) and compares `conv2d(...).sqr().sum_all().backward()` against the same computation using an explicitly `.contiguous()` copy of the kernel. Loss, `grad_input`, and `grad_weight` should match; on CUDA they don't.

## Root cause reference

[...]

PR #3894, opened 2026-08-15
Fix conv2d CUDA im2col offset bug for non-contiguous kernels

astorise:
## Summary

- Fixes `conv2d_grad_noncontiguous_kernel_gpu` failing on CUDA (#3893): the CUDA im2col path in `conv2d` allocated and filled a contiguous `kernel_c` buffer for non-contiguous kernels, but then discarded it and still matmul'd against the original non-contiguous kernel storage at its original (non-zero) start offset — reading the wrong bytes.
- Mirrors the fix already applied to the CPU backend in #3736: after `copy_strided_src` materializes the kernel into `kernel_c` starting at offset 0, the matmul must use `kernel_c` with offset 0, not the original kernel's start offset.

## Test plan

- [x] Verified the change compiles (`cargo check -p candle-core`, no `cuda` feature available in this sandbox)
- [x] Ran `cargo test --features cuda` on a real GPU runner against this branch — full suite green, including `conv2d_grad_noncontiguous_kernel_gpu`
- [ ] CI (`CI / cuda`) on this PR

### DIFF
--- a/candle-core/src/cuda_backend/mod.rs
+++ b/candle-core/src/cuda_backend/mod.rs
@@ -2022,15 +2022,16 @@ impl BackendStorage for CudaStorage {
                 Layout::contiguous_with_offset((n, k), kernel_l.start_offset()).transpose(0, 1)?;
             col.matmul(kernel, (1, b * m, n, k), &col_l, &kernel_l)?
         } else {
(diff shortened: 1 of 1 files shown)


ivarflakstad (approved):
Thank you ☺️

PR #3895, opened 2026-08-15
Fix F8E4M3 CUDA kernel naming + related CUDA correctness/diagnostics issues

### LINKED ISSUE #3884: `get_or_load_func` drops the kernel name from CUDA symbol-lookup errors
## Summary

When a CUDA kernel symbol is missing from a loaded module, the resulting error
does not say which kernel. `CudaDevice::get_or_load_func` takes the name as an
argument and discards it on the failure path:

```rust
// candle-core/src/cuda_backend/device.rs
pub fn get_or_load_func(&self, fn_name: &str, mdl: &kernels::Module) -> Result<CudaFunc> {
    let ms = self.modules.read().unwrap();
    if let Some(mdl) = ms.mdls[mdl.index()].as_ref() {
        let func = mdl.load_function(fn_name).w()?;   // <- fn_name dropped
        ...
    }
    drop(ms);
    let mut ms = self.modules.write().unwrap();
    let cuda_module = self.context.load_module(mdl.ptx().into()).w()?;
    ms.mdls[mdl.index()] = Some(cuda_module.clone());
    let func = cuda_module.load_function(fn_name).w()?; // <- and here
    ...
}
```

`.w()` wraps cudarc's `DriverError` unchanged, so the caller gets:

```
DriverError(CUDA_ERROR_NOT_FOUND, "named symbol not found")
```

The name that would make this actionable was in scope at the point of failure
and was thrown away.

## Why this is worse than it looks

Two things conspire to make it hard to recover the information afterwards.

**A backtrace does not identify the kernel.** The error propagates upward as a
value through `?`; the eventual panic happens at whatever `unwrap`/`expect`
consumes it, often several layers up in user code. `RUST_BACKTRACE=1` names
that consumer, not the `get_or_load_func` call site. The natural first
debugging step yields nothing.

**A missing symbol is a legitimate, reachable state, not a corrupt build.**
`candle-kernels` guards kernels by target architecture, and the guards are not
uniform across a dtype. For `F8E4M3` specifically, the casts sit under one
threshold and the elementwise and indexing ops under a higher one:

| kernels | guard |
|---|---|
| `cast_f8_e4m3_f32`, `cast_f8_e4m3_bf16`, … (`cast.cu`) | `__CUDA_ARCH__ >= 800` |
| `affine_f8_e4m3` (`affine.cu`) | `__CUDA_ARCH__ >= 890` |
| `is_*_f8_e4m3`, `gather_*_f8_e4m3`, `scatter_add_f8` (`indexing.cu`) | `__CUDA_ARCH__ >= 890` |

So a build for, say, `sm_86` has a genuinely partial `F8E4M3` kernel set: some
operations on such a tensor work and others fail at lookup time. That is a
reasonable design — but it means "symbol not found" is a *routine* outcome of
running a supported dtype on a supported device, and the one thing the user
needs in order to understand it is the name that is currently discarded.

Concretely: we hit `CUDA_ERROR_NOT_FOUND` while building a model on an
`sm_86` device and, after reading the guards above and tracing the load path by
hand, still cannot say which kernel is missing. That is the entire content of
this report — not that the lookup failed, but that a failed lookup is
undiagnosable from its own error.

## Why not fix this in cudarc

`DriverError` is a newtype over the raw driver result:

```rust
pub struct DriverError(pub sys::CUresult);
```

It has a public field, and its `std::error::Error` impl is behind
`#[cfg(feature = "std")]`. Carrying a symbol name would require an owned
`String` — allocation in a `no_std`-friendly FFI binding, a change to a
pattern-matchable public shape, and a breaking change across every API that
returns it. cudarc already surfaces the driver's own description; the caller's
identifier is the caller's to attach.

## Shape of the fix

Attach the name at both `load_function` sites, and the module with it, since
"which module was searched" is the natural follow-up question:

```
cuda: kernel `affine_f8_e4m3` not found in module `affine`
  (this kernel may not be compiled for the target architecture)
```

The parenthetical matters as much as the name: for a user on a pre-`sm_89`
device, "not found" reads as a broken installation, when it is in fact the
architecture guard doing its job.

## Acceptance

- A failed lookup names the kernel and the module it was searched in.
- The underlying `DriverError` remains in the source chain, so
  `CUDA_ERROR_NOT_FOUND` is still distinguishable from other driver failures.
- Both call sites are covered — the cached-module path and the
  load-then-lookup path fail identically today and must keep doing so.

### LINKED ISSUE #3886: F8E4M3 CUDA kernels are unreachable: `DType::as_str()` yields `f8e4m3`, the kernels are named `f8_e4m3`
## Summary

Every CUDA kernel for `F8E4M3` is declared with an underscore between `f8` and
`e4m3`:

```cuda
// candle-kernels/src/cast.cu:159
CAST_OP_FP8(__nv_fp8_e4m3, float,    cast_f8_e4m3_f32)
```

but the Rust side builds kernel names from `DType::as_str()`, which returns the
token *without* that underscore:

```rust
// candle-core/src/dtype.rs:87
Self::F8E4M3 => "f8e4m3",
```

So the lookup asks for `cast_f8e4m3_f32`, the module contains
`cast_f8_e4m3_f32`, and the call fails at runtime with
`CUDA_ERROR_NOT_FOUND` / "named symbol not found".

## Reproduction

Any CUDA device:

```rust
let t = Tensor::zeros((2, 2), DType::F8E4M3, &dev)?;
let _ = t.to_dtype(DType::F32)?;   // CUDA_ERROR_NOT_FOUND: named symbol not found
```

## Where the names diverge

| site | name requested | name declared | |
|---|---|---|---|
| `cuda_backend/mod.rs:1578` — `format!("cast_{}_{}", self.dtype().as_str(), dtype.as_str())` | `cast_f8e4m3_f32` | `cast_f8_e4m3_f32` | ✗ |
| `cuda_backend/mod.rs:130` — `kernel_name::<T>(root)` → `{root}_{dtype}` | `affine_f8e4m3`, `ucopy_f8e4m3`, … | `affine_f8_e4m3`, … | ✗ |
| `cuda_backend/mod.rs:2498` — hardcoded `"ucopy_f8e4m3"` | `ucopy_f8e4m3` | `ucopy_f8_e4m3` | ✗ |
| `cuda_backend/mod.rs:1521` — hardcoded `"const_set_f8_e4m3"` | `const_set_f8_e4m3` | `const_set_f8_e4m3` | ✓ |

The only call site that works is the one that spells the name out by hand.
Everything routed through `as_str()` misses.

## Scope

`candle-kernels` declares `F8E4M3` kernels across `affine.cu`, `binary.cu`,
`cast.cu`, `fill.cu`, `indexing.cu`, `unary.cu` and friends — `affine_f8_e4m3`,
`badd/bdiv/bmul/bsub/bmaximum/bminimum_f8_e4m3`, `eq/ge/gt_f8_e4m3`,
`copy2d_f8_e4m3`, `fill_f8_e4m3`, the `gather_*`/`is_*`/`ia_*` family, every
`cast_*` pair. All of them are addressed through the generated name, so none of
them is reachable.

In practice an `F8E4M3` tensor on the CUDA backend cannot be cast, copied,
scaled, compared or indexed. It can be allocated and filled — that path uses
the hardcoded name.

## This is not an architecture problem

Worth stating explicitly, because the failure mode invites the wrong
conclusion. `candle-kernels` does gate `F8E4M3` kernels by architecture, and
the thresholds are not uniform — casts sit under `__CUDA_ARCH__ >= 800` while
`affine_f8_e4m3` and the indexing family sit under `>= 890`. That makes
"symbol not found" look like a compute-capability story, and we spent
considerable time on that reading before checking the name itself.

It is not. `cast_f8_e4m3_f32` is under the `>= 800` guard and is present in the
PTX for `sm_86`, where we hit this. The lookup fails on every architecture,
Blackwell included, because the requested string does not exist in any build.

(The architecture guards remain a real and separate constraint: fixing the
naming makes the `>= 800` kernels reachable everywhere they are compiled, and
leaves the `>= 890` ones legitimately absent below `sm_89`.)

## Diagnosability

Today this surfaces as a bare `DriverError(CUDA_ERROR_NOT_FOUND, "named symbol
not found")` with no indication of which kernel was missing. The name is known
at the failing call site — `get_or_load_func(fn_name, mdl)` — and discarded.
A backtrace does not recover it either, since the error travels up as a value
through `?` and the eventual panic lands wherever the caller unwraps it.

That gap is the subject of a separate report; mentioning it here because
without it this bug is very hard to reach. Naming the kernel in the error turns
this from a multi-day investigation into a one-line diagnosis.

## Fix

Three ways, in increasing order of preference:

1. **Rename the CUDA kernels** `f8_e4m3` → `f8e4m3`. Touches every declaration
   across the `.cu` files and requires updating the one hardcoded Rust name
   that is currently correct.
2. **Change `as_str()`** to return `"f8_e4m3"`. Note it has other consumers:
   `display.rs` prints it as the user-visible dtype of a tensor, and
   `metal_backend/mod.rs:1929` builds Metal kernel names from it (Metal
   declares no `F8E4M3` kernels today, so only the display output would move).
3. **Separate the two concerns** — give `DType` a dedicated kernel-name token
   distinct from its display string, and build CUDA kernel names from that.

Option 3 addresses the actual defect: one string is doing duty both as a
human-facing dtype label and as an ABI-level symbol fragment, and nothing keeps
the two in agreement. Whichever is chosen, the fix is worth pairing with a test
that resolves each declared `F8E4M3` kernel by its generated name, so a future
divergence fails in CI rather than on a user's device.

## Acceptance

- `Tensor::to_dtype` between `F8E4M3` and `F32`/`F16`/`BF16` succeeds on CUDA.
- The generated name for every `F8E4M3` kernel matches a declared symbol, and
  something in CI checks that rather than leaving it to runtime.
- `const_set_f8_e4m3` keeps working — whichever direction the rename goes, that
  hardcoded name has to move with it.

### LINKED ISSUE #3893: `conv2d_grad_noncontiguous_kernel_gpu` fails on CUDA: PR #3736 fixed the CPU offset bug but not the GPU path

astorise:
## Summary

This branch bundles four small, independent CUDA-side fixes discovered while chasing down #3886. They can be reviewed/merged together or split into separate PRs if preferred — happy to split on request.

- **#3886** — F8E4M3 CUDA kernels were unreachable: `DType::as_str()` returns `"f8e4m3"` (no underscore), but the kernels in `affine.cu`, `binary.cu`, `cast.cu`, `fill.cu`, `indexing.cu`, and `unary.cu`'s `ucopy` op were declared as `..._f8_e4m3` (with underscore). Every generated lookup missed. Renamed the declared kernel symbols to drop the underscore, matching `as_str()`, and updated the one hardcoded caller (`const_set_f8_e4m3` → `const_set_f8e4m3`). Restores F8E4M3 affine/binary/cast/fill/copy/indexing ops on CUDA.
- **#3884** — `get_or_load_func` / `get_or_load_custom_func` surfaced cudarc's bare "named symbol not found" on a missing kernel, with no indication of which kernel or module was being looked up — a real diagnostic gap for dtypes like F8E4M3 whose kernels have non-uniform architecture guards, where a lookup miss is an expected, not exceptional, path. `kernels::Module` now carries a human-readable module name; `CudaError::MissingKernel` carries the kernel name, module name, and keeps the underlying `DriverError` in the source chain. This is what surfaced #3886 with a useful error message instead of a bare cudarc error.
- **#3893** — `conv2d`'s CUDA im2col path had the same non-contiguous-kernel offset bug that #3736 fixed on CPU: when a non-contiguous kernel (e.g. a channel slice of a larger weight) is materialized into a contiguous scratch buffer, the matmul must read that buffer at offset 0. The CUDA path allocated and filled a contiguous `kernel_c` buffer but then discarded it, still matmul'ing against the original non-contiguous kernel storage at its original (non-zero) start offset — reading the wrong bytes. Fixed to matmul against the materialized buffer at offset 0, mirroring the CPU fix. This also unblocks `conv2d_grad_noncontiguous_kernel_gpu` in `conv_tests.rs`, which #3736 added alongside the CPU test but which has been failing on CUDA ever since (the CPU fix never got its CUDA counterpart).
- **Dependency bump**: `cudarc` `0.19.8` → `0.19.9`, a plain upstream release bump with no other changes.

## Changes

- `candle-kernels/src/{affine,binary,cast,fill,indexing,unary}.cu`: renamed `..._f8_e4m3` kernel symbols to `..._f8e4m3`.
- `candle-core/src/cuda_backend/mod.rs`: updated the one hardcoded `const_set_f8_e4m3` caller; fixed the conv2d im2col non-contiguous-kernel offset/buffer bug.
- `candle-core/src/cuda_backend/device.rs`, `candle-core/src/cuda_backend/error.rs`, `candle-kernels/src/lib.rs`: `CudaError::MissingKernel` now names the missing kernel and module and preserves the underlying `DriverError`.
- `Cargo.toml`: `cudarc` `0.19.8` → `0.19.9`.

## Testing

Ran on a CUDA 13.0.2 / compute-capability-8.6 GPU runner:

```
cargo test --features cuda
```

- `candle-core` unit tests, `bilinear_tests`, `conv_tests` (including `conv2d_grad_noncontiguous_kernel_gpu`), `custom_op_tests`, and `tensor_tests` (F8E4M3 cast/affine/binary/fill/indexing coverage) all pass.
- No regressions observed in the rest of the CUDA suite.

## Related issues

- Fixes #3886
- Fixes #3884
- Fixes #3893


astorise:
Closing — this bundles five unrelated changes (cudarc bump, error messages, f8e4m3 rename, conv1d and conv2d fixes). I'm sending them separately; #3894 is the first.

PR #3897, opened 2026-08-15
Add fused Metal kernels for gated-DeltaNet decode and preprocessing

alytaphoenix:
## Summary

Gated-DeltaNet (gated linear attention) decode, as used by Qwen3.5 / Qwen3-Next style hybrid architectures, normally executes as a chain of ~9 serialized, dependency-ordered elementwise/reduction candle ops per layer per decode token (decay multiply, key-weighted read, delta computation, state write, query-weighted read). At batch=1, seq_len=1 the tensors involved are small, so this is cheap in FLOPs but expensive in fixed per-dispatch overhead — every op depends on the previous op's output, so each pays a full pipeline/bind/encode cycle and a hazard-tracking barrier under `HazardTrackingModeUntracked`.

This PR adds `kernel_gdn_decode_step_f32`, fusing the whole step into one dispatch: one thread per `(batch, head, output-column)`, computing the decay/read/delta/write/read chain entirely in registers, no cross-thread communication or threadgroup memory.

- **State is written functionally** — a fresh output buffer; the input state buffer is read-only and untouched — rather than mutated in place, so callers holding other references to the input state tensor (e.g. an externally-taken checkpoint) are unaffected by a decode step running.
- **Every read input takes a `BufferOffset`, not a bare `Buffer`.** Callers commonly derive `q`/`k`/`v` from `narrow()`'d slices of a shared QKV-projection buffer, which can carry a real nonzero byte offset even when the resulting tensor is itself contiguous. Binding at a fixed offset 0 regardless of the tensor's actual layout silently reads the wrong region of the buffer — caught during integration by an end-to-end differential against a real model reference, not by this crate's own unit tests (which used independent, always-offset-0 buffer allocations and so never exercised this). The included regression test packs six tensors into one shared buffer at real nonzero offsets to reproduce that shape directly.

Measured on a real Qwen3.5-family model's decode path (single-token, production shape): replacing the equivalent candle op sequence with this kernel took full end-to-end decode throughput from ~20 to ~43 tok/s (2.1x) on Apple M4 Pro, with the DeltaNet/SSM recurrence's own share of decode time going from the dominant term to roughly on par with the model's MoE FFN.

## Test plan

- [x] `cargo test -p candle-metal-kernels --lib gdn` — pipeline-load, numeric parity against a scalar Rust reference at small/odd shapes and a real production shape (`b=1, h=32, hk=128, hv=128`, max diff 9e-8), and the nonzero-offset regression test
- [x] `cargo test -p candle-metal-kernels --lib` — full crate suite, 65/65
- [x] `cargo fmt --check` and `cargo clippy --tests -- -D warnings` clean on every file this PR touches (some pre-existing clippy warnings remain elsewhere in the crate, unrelated to this change and not touched here)
- [x] Integrated and validated end-to-end against a real cached GGUF model's decode path with a bit-exact/near-bit-exact differential against an independent HF reference implementation (external project, not part of this repo)

https://claude.ai/code/session_01XwN6xpWjh1aFcNdWmd1mAM

### DIFF
--- a/candle-metal-kernels/src/kernel.rs
+++ b/candle-metal-kernels/src/kernel.rs
@@ -1,5 +1,5 @@
 use crate::source::{
-    AFFINE, BINARY, CAST, CONV, FILL, GEMV, INDEXING, MLX_GEMM, MLX_SORT, QUANTIZED, RANDOM,
+    AFFINE, BINARY, CAST, CONV, FILL, GDN, GEMV, INDEXING, MLX_GEMM, MLX_SORT, QUANTIZED, RANDOM,
     REDUCE, SDPA, SORT, TERNARY, UNARY,
 };
 use crate::utils::get_env_bool;
@@ -90,6 +90,7 @@ impl Kernels {
             Source::Cast => CAST,
             Source::Conv => CONV,
             Source::Fill => FILL,
+            Source::Gdn => GDN,
             Source::Gemm => MLX_GEMM,
             Source::Gemv => GEMV,
             Source::Indexing => INDEXING,
--- a/candle-metal-kernels/src/kernels/gdn.rs
+++ b/candle-metal-kernels/src/kernels/gdn.rs
@@ -103,3 +103,286 @@ pub fn call_gdn_decode_step_f32(
     encoder.dispatch_threads(grid_dims, group_dims);
     Ok(())
 }
+
+#[repr(C)]
+struct GdnConv1dArgs {
+    seq_len: u32,
+    hist_len: u32,
+    channels: u32,
+    kernel_size: u32,
+}
+
+impl EncoderParam for GdnConv1dArgs {
+    fn set_param(encoder: &ComputeCommandEncoder, position: usize, data: Self) {
+        encoder.set_bytes(position, &data);
+    }
+}
+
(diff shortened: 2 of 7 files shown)


alytaphoenix:
Pushed an update extending the scope beyond the single-token decode step: three more fused kernels covering the preprocessing pipeline that runs ahead of the recurrent scan.

- `kernel_gdn_causal_conv1d_output_f32` / `_state_f32`: fused causal depthwise conv1d + silu. A naive implementation dispatches `O(seq_len * kernel_size)` separate ops (one narrow/multiply/add per tap per timestep) — the dominant per-op-dispatch cost at small batch sizes. Replaces that with one dispatch for the output and one small dispatch for the next conv state, both written functionally.
- `kernel_gdn_l2_normalize_scale_f32`: fused L2-normalize + scale for q/k.
- `kernel_gdn_decay_beta_gate_f32`: fuses the softplus decay-gate chain and sigmoid beta gate, and indexes its per-head inputs directly instead of requiring a pre-broadcast.

Verified against independent scalar Rust references (11 new tests: multiple shapes including non-threadgroup-aligned sizes, a zero-conv-history edge case, and nonzero-buffer-offset regression tests mirroring the same offset bug class the existing decode-step tests guard against). Full crate test suite clean.

PR #3904, opened 2026-08-17
Fix SIMD128 half-precision CPU kernels

### LINKED ISSUE #3835: candle-core fails to build for wasm32 with +simd128 (the repo's own default): CurrentCpuF16 undefined on the simd128 path
`candle-core` fails to build for `wasm32-unknown-unknown` with `-C target-feature=+simd128`: `CurrentCpuF16` is not defined on the simd128 path.

### Summary

`candle-core/src/cpu/mod.rs` exports the full CPU trait set for AVX2 and NEON, but only `CurrentCpu` for simd128:

```rust
#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
#[cfg(target_feature = "avx2")]
pub use avx::{CurrentCpu, CurrentCpuBF16, CurrentCpuF16};

#[cfg(target_arch = "wasm32")]
#[cfg(target_feature = "simd128")]
pub use simd128::CurrentCpu;                    // <-- no CurrentCpuF16 / CurrentCpuBF16

#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
#[cfg(target_feature = "neon")]
pub use neon::{CurrentCpu, CurrentCpuBF16, CurrentCpuF16};
```

The f16/bf16 helpers below are then gated on `any(neon, avx2, simd128)` and use `CurrentCpuF16`, which the simd128 path never provides:

```rust
#[cfg(any(
    target_feature = "neon",
    target_feature = "avx2",
    target_feature = "simd128"
))]
pub(crate) unsafe fn vec_dot_f16(...) {
    let mut sum = CurrentCpuF16::zero_array();
    ...
}
```

The net effect is that **wasm builds can have SIMD or they can compile, but not both** — `+simd128` is the only way to reach `src/cpu/simd128.rs`, and enabling it breaks the build.

This is not an exotic flag. candle's own root `.cargo/config.toml` sets it as the default for that target:

```toml
[target.wasm32-unknown-unknown]
rustflags = ["-C", "target-feature=+simd128", "--cfg", 'getrandom_backend="wasm_js"']
```

so the failing configuration is the one the repository itself prescribes.

### Reproduction

```bash
cargo new --lib repro && cd repro
cargo add candle-core@0.11.0
# getrandom arrives transitively and cannot pick its own wasm backend;
# without this you hit its compile_error! before reaching candle.
cargo add getrandom@0.3 --features wasm_js
rustup target add wasm32-unknown-unknown

# fails:
RUSTFLAGS='--cfg getrandom_backend="wasm_js" -C target-feature=+simd128'   cargo check --target wasm32-unknown-unknown

# control — succeeds, differing only in +simd128:
RUSTFLAGS='--cfg getrandom_backend="wasm_js"'   cargo check --target wasm32-unknown-unknown
```

### Actual

```
error[E0433]: cannot find type `CurrentCpuF16` in this scope
   --> candle-core-0.11.0/src/cpu/mod.rs:259:15
    |
259 |     while i + CurrentCpuF16::STEP <= k {
    |               ^^^^^^^^^^^^^ use of undeclared type `CurrentCpuF16`
    |
note: found an item that was configured out
```

repeated at `mod.rs:259`, `:260`, `:261` and each further use of
`CurrentCpuF16` in `vec_dot_f16` / `vec_add_f16`.

The control command above, identical but for `+simd128`, prints `Finished`.

### Expected

The crate compiles, and `src/cpu/simd128.rs` is used for f32 kernels.

### Environment

- Reproduces on **`candle-core` 0.11.0** (current release) and on **`main`** (`6e823a4`), at the same three lines.
- rustc 1.95.0, host `x86_64-pc-windows-msvc`, target `wasm32-unknown-unknown`
- Without `+simd128` the crate compiles cleanly on wasm32; the failure is specific to enabling SIMD.

### Suggested fix

Either would do, and the first is smaller:

1. **Narrow the `cfg` on the f16/bf16 helpers** to `any(target_feature = "neon", target_feature = "avx2")`, so the simd128 path keeps the scalar f16 fallback that already exists directly above them, and gets vectorised f32 via `CurrentCpu`.
2. **Implement `CurrentCpuF16` (and `CurrentCpuBF16`) in `simd128.rs`**, matching what `avx.rs` and `neon.rs` provide.

Happy to open a PR for either — a maintainer's steer on which is preferred would be welcome, since (1) changes what wasm gets and (2) is more work but keeps the ISAs symmetric.

### Why it matters

`candle-wasm-examples` ships ten browser demos including `segment-anything`,
and `candle-wasm-tests` exists — so wasm is clearly a supported target rather
than an afterthought, and this is the only thing standing between it and
vectorised inference there. For context on the size of it: on a pure-Rust YOLO26 detector running on candle, disabling AVX2 natively — the closest available proxy for losing SIMD — costs **1.25×** (32.5 ms → 40.6 ms per image, min-of-80). Without threads as well, which is the default situation on `wasm32-unknown-unknown`, it is **2.9×** (93.9 ms). Recovering the SIMD half needs only that the crate build.

damageboy:
## Summary

- add `CurrentCpuF16` and `CurrentCpuBF16` implementations to the WebAssembly SIMD128 backend
- widen half-precision lanes to `f32x4` for arithmetic, matching the AVX and NEON fallback approach
- add browser tests for F16/BF16 vector and scalar addition across both a SIMD block and a scalar tail
- update the WASM test crate to select the `getrandom` 0.3 `wasm_js` backend

## Motivation

Fixes #3835.

`cpu/mod.rs` selects the SIMD128 vector and scalar-add kernels for F16 and BF16, but the SIMD128 backend previously exported only `CurrentCpu`. Consequently, the repository's configured `wasm32-unknown-unknown +simd128` target failed to compile because `CurrentCpuF16` and `CurrentCpuBF16` did not exist.

This implements the full SIMD128 option proposed in the issue instead of falling back to scalar half operations. WebAssembly SIMD has no native half-precision arithmetic, so values are widened to four F32 lanes. BF16 loads use bit widening; stores use `half` conversions to preserve round-to-nearest-even behavior.

## Validation

- `cargo check --locked -p candle-wasm-tests --target wasm32-unknown-unknown --tests`
- `cargo test --locked -p candle-wasm-tests --target wasm32-unknown-unknown --test cpu_tests` using `wasm-bindgen-test-runner` and headless Chrome: 4 passed

### DIFF
--- a/candle-core/src/cpu/simd128.rs
+++ b/candle-core/src/cpu/simd128.rs
@@ -1,5 +1,6 @@
-use super::Cpu;
+use super::{Cpu, CpuBF16, CpuF16};
 use core::arch::wasm32::*;
+use half::{bf16, f16};
 
 pub struct CurrentCpu {}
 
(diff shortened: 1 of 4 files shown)


candle-wasm-tests/tests/cpu_tests.rs
damageboy, line 1: @ivarflakstad is this acceptable?

ivarflakstad: 👍

candle-core/src/cpu/simd128.rs
```diff
@@ -59,3 +60,108 @@ impl Cpu for CurrentCpu {
...
+    }
+
+    unsafe fn load(mem_addr: *const f16) -> Self::Unit {
+        let values = [
+            (*mem_addr).to_f32(),
+            (*mem_addr.add(1)).to_f32(),
+            (*mem_addr.add(2)).to_f32(),
+            (*mem_addr.add(3)).to_f32(),
+        ];
+        v128_load(values.as_ptr().cast())
```
ivarflakstad, line 95: Here you are using v128 to hold 4xf32. I'm guessing it would be more efficient to hold 8xf16. Then as you are about to perform an operation you convert to 8xf32, run your operation, and convert back.

However I don't think an optimization like that should block a PR that fixes a currently broken target though, so we can test the theory at some later point :)

ivarflakstad (approved):
lgtm!

Just waiting on CI :)

candle-core/src/cpu/simd128.rs
damageboy: Let me see if I can check that, still new to the wasm "world" in that sense, so that tooling is still new to me

PR #3908, opened 2026-08-20
fix(onnx): honour allowzero in Reshape and fix -1 inference past a copied zero

### LINKED ISSUE #3907: Reshape ignores the allowzero attribute (candle-onnx)
`candle-onnx`'s `Reshape` does not read the `allowzero` attribute. A `0` in the target shape is
always treated as "copy the corresponding input dimension", which is the `allowzero=0` behaviour;
with `allowzero=1` it should mean a literal zero-length dimension.

## Where it happens

[`candle-onnx/src/eval.rs`](https://github.com/huggingface/candle/blob/0.11.0/candle-onnx/src/eval.rs#L396),
the `"Reshape"` arm:

```rust
"Reshape" => {
    let input0 = get(&node.input[0])?;
    let input1 = get(&node.input[1])?.to_vec1::<i64>()?;
    // TODO: Check that there is at most a single -1 or 0, handle other neg values.
    let mut other_than_minus1 = 1usize;
    for &v in input1.iter() {
        if v != -1 && v != 0 {
            other_than_minus1 *= v as usize
        }
    }
    let input1 = input1
        .iter()
        .enumerate()
        .map(|(idx, &v)| match v {
            -1 => Ok(input0.elem_count() / other_than_minus1),
            0 => input0.dim(idx),          // <-- always the allowzero=0 reading
            _ => Ok(v as usize),
        })
        .collect::<Result<Vec<usize>>>()?;
    let output = input0.reshape(input1)?;
    values.insert(node.output[0].clone(), output);
}
```

The arrow comment is mine; everything else in that block is verbatim.

`node`'s attributes are never consulted here: there is no `get_attr_opt::<i64>(node, "allowzero")`
in this arm, and the string `allowzero` does not appear anywhere in `candle-onnx` at all, whose
only evaluator is `eval.rs`. Other operators in the same file do read their attributes this way
(`LogSoftmax` and `Softmax` immediately below both read `axis`), so the mechanism exists and is
simply not used for this one.

## What ONNX specifies

From the [`Reshape` reference](https://onnx.ai/onnx/operators/onnx__Reshape.html):

> "If 'allowzero' is set, and the new shape includes 0, the dimension will be set explicitly to
> zero (i.e. not taken from input tensor)."

`allowzero` was added in opset 14.

## Reproduction

A single-node model where the two readings give different answers:

```
Reshape, opset 22, attribute allowzero = 1
  input  a : float32[3, 0]   (empty, zero elements)
  input  b : int64[1] = [0]  (initializer, the target shape)
```

```
onnx.reference (onnx 1.22.0)  ->  ok, float32[0], empty
onnxruntime 1.29.0            ->  ok, float32[0], empty
candle-onnx 0.11.0            ->  Error: shape mismatch in reshape, lhs: [3, 0], rhs: [3]
```

The error matches the code exactly. The target is `[0]`, so `idx` is `0` and `input0.dim(0)` is
`3`, hence `rhs: [3]`. With `allowzero=1` the target should stay `[0]`: zero elements requested
against zero elements held, which is the shape `onnx.reference` and ONNX Runtime both produce.

### Steps to reproduce

Builds the model, checks it, saves it, and prints what the specification's own implementation
returns. Needs only `pip install onnx` (tested with onnx 1.22.0). Verified to run as shown.

```python
import numpy as np, onnx
from onnx import TensorProto as T, helper as h, numpy_helper as nh
from onnx.reference import ReferenceEvaluator

shape = nh.from_array(np.array([0], np.int64), "shape")
g = h.make_graph([h.make_node("Reshape", ["a", "shape"], ["y"], allowzero=1)], "reshape",
                 [h.make_tensor_value_info("a", T.FLOAT, [3, 0])],
                 [h.make_tensor_value_info("y", T.FLOAT, [0])], [shape])
m = h.make_model(g, opset_imports=[h.make_opsetid("", 22)]); m.ir_version = 10
onnx.checker.check_model(m); onnx.save(m, "repro.onnx")

r = ReferenceEvaluator(m).run(None, {"a": np.zeros((3, 0), np.float32)})[0]
print(r.shape, r.size)   # (0,) 0  -- candle-onnx errors with rhs: [3]
```

`repro.onnx` is a complete single-node model; loading it with `candle_onnx::simple_eval` is the whole reproduction.

## A second bug in the same function

While writing the fix I found that `other_than_minus1` skips `0` entirely:

```rust
for &v in input1.iter() {
    if v != -1 && v != 0 { other_than_minus1 *= v as usize }
}
```

With `allowzero=0` a `0` means "copy the input dimension", so that copied dimension **is** part of
the volume and must be in the product. Skipping it makes the `-1` inference too large.

ONNX's own test data covers this shape of case, as `zero_and_negative_dim` in the `Reshape`
examples. The run below uses ONNX's own input and target for that case:

```
input  float32[2, 3, 4]   (24 elements)     <- ONNX's original_shape
shape  int64[4] = [2, 0, 1, -1]             <- ONNX's zero_and_negative_dim
                                               (allowzero not set)

onnx.reference (onnx 1.22.0)  ->  [2, 3, 1, 4]
onnxruntime 1.29.0            ->  [2, 3, 1, 4]
tract 0.23.4                  ->  [2, 3, 1, 4]
candle-onnx 0.11.0            ->  Error: shape mismatch in reshape, lhs: [2, 3, 4], rhs: [2, 3, 1, 12]
```

`12` is `24 / (2 x 1)`, the product with the copied `3` left out. This one does not need
`allowzero` at all, so it affects the default path.

## Suggested fix

Resolving every non-`-1` dimension first fixes both, because the `-1` inference then divides by
the true product:

```rust
"Reshape" => {
    let input0 = get(&node.input[0])?;
    let input1 = get(&node.input[1])?.to_vec1::<i64>()?;
    // ONNX and its reference evaluator compare this with 1 rather than testing non-zero.
    let allowzero = get_attr_opt::<i64>(node, "allowzero")?.copied().unwrap_or(0) == 1;

    // "At most one dimension of the new shape can be -1."
    if input1.iter().filter(|&&v| v == -1).count() > 1 {
        bail!("Reshape: at most one dimension of the target shape can be -1")
    }

    let mut resolved: Vec<Option<usize>> = Vec::with_capacity(input1.len());
    for (idx, &v) in input1.iter().enumerate() {
        resolved.push(match v {
            -1 => None,
            0 if allowzero => Some(0),
            0 => Some(input0.dim(idx)?),
            v if v > 0 => Some(v as usize),
            v => bail!("Reshape: invalid dimension {v} in target shape"),
        });
    }

    let known: usize = resolved.iter().flatten().product();
    let dims = resolved
        .into_iter()
        .map(|d| match d {
            Some(d) => Ok(d),
            None if known == 0 => bail!(
                "Reshape: -1 cannot be inferred when another dimension is zero"
            ),
            None => Ok(input0.elem_count() / known),
        })
        .collect::<Result<Vec<usize>>>()?;

    let output = input0.reshape(dims)?;
    values.insert(node.output[0].clone(), output);
}
```

The `known == 0` guard follows the specification when `allowzero` is set:

> "If the attribute 'allowzero' is set, it is invalid for the specified shape to contain both a
> zero value and -1, as the value of the dimension corresponding to -1 cannot be determined
> uniquely."

When `allowzero` is not set, a zero can still reach `known` by being copied from the input, and
that is the unsettled territory of #6104. I chose to reject there as well rather than divide by
zero, but that is a choice rather than a rule, and I am happy to do whatever you prefer.

Two cases from ONNX's `Reshape` test data make good regression tests, alongside the empty case
above. **All three fail on candle-onnx 0.11.0 today** — run, not inferred:

| case | input | target | candle-onnx 0.11.0 |
|---|---|---|---|
| `allowzero_reordered`, `allowzero=1` | `[0,3,4]` | `[3,4,0]` | `lhs: [0, 3, 4], rhs: [3, 4, 4]` |
| `zero_and_negative_dim` | `[2,3,4]` | `[2,0,1,-1]` | `lhs: [2, 3, 4], rhs: [2, 3, 1, 12]` |
| the empty case above, `allowzero=1` | `[3,0]` | `[0]` | `lhs: [3, 0], rhs: [3]` |

`onnx.reference` and ONNX Runtime 1.29.0 return `[3,4,0]`, `[2,3,1,4]` and `[0]`. tract returns
`[2,3,1,4]` for the middle one and rejects the two zero-volume cases for an unrelated reason of
its own, so it is listed here only as a third data point.

One case this does not settle: a negative other than `-1`, such as a target of `[-2]`. ONNX's
text specifies inference for `-1` only and says nothing about other negatives, while
`onnx.reference` accepts them and infers, because it delegates to `np.reshape` and numpy treats
any negative as "infer this dimension". The snippet above rejects them explicitly rather than
casting to `usize` as the current code does, so the observable behaviour is unchanged, only the
error message improves. If you would rather follow numpy and treat any negative as inference, that
is a one-line change to the last arm.

## Version

candle-onnx 0.11.0, macOS aarch64, opset 22. The quoted source is from tag `0.11.0`; `main` is
identical in this arm at the time of writing.

Found by differential testing of single-node models against `onnx.reference`, ONNX Runtime and
tract. Ranks 2, 3 and 4 behave the same way, at every element type tried.

keevin-xu:
Fixes #3907.

Two problems in the `Reshape` arm of `candle-onnx/src/eval.rs`, both about a
`0` in the target shape.

`allowzero` was never read, so a `0` was always treated as "copy the
corresponding input dimension". With `allowzero=1` it should mean a literal
zero-length dimension.

Separately, the product that `-1` is inferred against skipped zeros. Under the
default `allowzero=0` a `0` copies an input dimension, so it is part of the
volume and belongs in that product; leaving it out made the inferred `-1` too
large. This one needs no attribute, and ONNX's own `zero_and_negative_dim` test
shape reaches it.

Resolving every non-`-1` dimension first fixes both, because the inference then
divides by the true product. While there I also rejected more than one `-1`,
which the previous shape of the code could silently accept.

Three regression tests using ONNX's own `Reshape` test shapes, plus one for
multiple `-1`s. All fail without the fix.

The `known == 0` guard rejects `-1` inference whenever the rest of the volume is
zero, which is broader than the spec sentence about `allowzero=1`. It also covers
a copied zero under `allowzero=0`, which onnx/onnx#6104 left unsettled. Rejecting
seemed better than dividing by zero, but I am happy to narrow it.

A case still left open: a negative other than `-1`, such as `[-2]`. ONNX specifies
inference for `-1` only; `onnx.reference` accepts any negative because it
delegates to `np.reshape`. This patch rejects them explicitly instead of casting
to `usize`, so behaviour is unchanged and only the message improves.

### DIFF
--- a/candle-onnx/src/eval.rs
+++ b/candle-onnx/src/eval.rs
@@ -396,20 +396,37 @@ fn simple_eval_(
             "Reshape" => {
                 let input0 = get(&node.input[0])?;
                 let input1 = get(&node.input[1])?.to_vec1::<i64>()?;
-                // TODO: Check that there is at most a single -1 or 0, handle other neg values.
-                let mut other_than_minus1 = 1usize;
-                for &v in input1.iter() {
-                    if v != -1 && v != 0 {
-                        other_than_minus1 *= v as usize
-                    }
+                // A 0 in the target shape copies the corresponding input dimension, unless
+                // allowzero=1, where it means a literal zero-length dimension.
+                let allowzero = get_attr_opt::<i64>(node, "allowzero")?
+                    .copied()
+                    .unwrap_or(0)
+                    == 1;
+                if input1.iter().filter(|&&v| v == -1).count() > 1 {
+                    bail!("Reshape: at most one dimension of the target shape can be -1")
                 }
-                let input1 = input1
-                    .iter()
-                    .enumerate()
-                    .map(|(idx, &v)| match v {
-                        -1 => Ok(input0.elem_count() / other_than_minus1),
-                        0 => input0.dim(idx),
-                        _ => Ok(v as usize),
+                // Resolve everything but -1 first: a copied 0 is part of the volume, so it
+                // has to be in the product that -1 is inferred against.
+                let mut resolved: Vec<Option<usize>> = Vec::with_capacity(input1.len());
+                for (idx, &v) in input1.iter().enumerate() {
+                    resolved.push(match v {
+                        -1 => None,
+                        0 if allowzero => Some(0),
+                        0 => Some(input0.dim(idx)?),
+                        v if v > 0 => Some(v as usize),
+                        v => bail!("Reshape: invalid dimension {v} in target shape"),
+                    });
+                }
+                let known: usize = resolved.iter().flatten().product();
+                let input1 = resolved
+                    .into_iter()
+                    .map(|d| match d {
+                        Some(d) => Ok(d),
+                        // A -1 has no unique value when the rest of the volume is zero.
+                        None if known == 0 => {
+                            bail!("Reshape: -1 cannot be inferred when another dimension is zero")
+                        }
+                        None => Ok(input0.elem_count() / known),
                     })
                     .collect::<Result<Vec<usize>>>()?;
                 let output = input0.reshape(input1)?;
--- a/candle-onnx/tests/ops.rs
+++ b/candle-onnx/tests/ops.rs
@@ -451,6 +451,124 @@ fn test_reshape_operation() -> Result<()> {
     Ok(())
 }
 
+// Build a single-node Reshape graph, optionally with an `allowzero` attribute.
+fn reshape_graph(allowzero: Option<i64>) -> ModelProto {
+    let attribute = allowzero
+        .map(|i| AttributeProto {
+            name: "allowzero".to_string(),
+            ref_attr_name: "allowzero".to_string(),
+            i,
+            doc_string: "allowzero".to_string(),
+            r#type: 2,
+            f: 0.0,
+            s: vec![],
+            t: None,
+            g: None,
+            sparse_tensor: None,
+            tp: None,
+            floats: vec![],
+            ints: vec![],
+            strings: vec![],
+            tensors: vec![],
+            graphs: vec![],
+            sparse_tensors: vec![],
+            type_protos: vec![],
+        })
+        .into_iter()
+        .collect();
+
+    create_model_proto_with_graph(Some(GraphProto {
+        node: vec![NodeProto {
+            op_type: "Reshape".to_string(),
+            domain: "".to_string(),
+            attribute,
+            input: vec![INPUT_X.to_string(), INPUT_Y.to_string()],
+            output: vec![OUTPUT_Z.to_string()],
+            name: "".to_string(),
+            doc_string: "".to_string(),
+        }],
+        name: "".to_string(),
+        initializer: vec![],
+        input: vec![
+            ValueInfoProto {
+                name: INPUT_X.to_string(),
+                doc_string: "".to_string(),
+                r#type: None,
+            },
+            ValueInfoProto {
+                name: INPUT_Y.to_string(),
+                doc_string: "".to_string(),
+                r#type: None,
+            },
+        ],
+        output: vec![ValueInfoProto {
+            name: OUTPUT_Z.to_string(),
+            doc_string: "".to_string(),
+            r#type: None,
+        }],
+        value_info: vec![],
+        doc_string: "".to_string(),
+        sparse_initializer: vec![],
+        quantization_annotation: vec![],
+    }))
+}
+
+fn run_reshape(allowzero: Option<i64>, x: Tensor, shape: Vec<i64>) -> Result<Tensor> {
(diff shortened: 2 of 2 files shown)


ivarflakstad (approved):
👍

ivarflakstad:
Clippy error is unrelated

PR #3919, opened 2026-08-28
Improved in-place functionality. Avoids deadlocking.

ivarflakstad:
In-place ops now check if input tensors share underlying allocation, and apply correct locking mechanisms to avoid deadlocks. Unsupported access patterns are rejected.

We use `parking_lot::RwLock::read_recursive` when reading tensor storage, allowing shared read access without acquiring several read locks.

All in-place ops now call through the new `Tensor::inplace_op` using the new `InplaceOpN<const N: usize>` trait - including ops defined through the old `InplaceOp1/2/3` trait.

`Tensor::inplace_op` checks the validity of the in-place op access pattern wrt `N` tensor operands.
If valid then all (distinct) tensors are locked in order of `Tensor::storage_key` to access the underlying storage. This avoids deadlock by cycle.

Then `Storage::inplace_op` is called, which either delegates to the backend specific `*_fwd` or `*_fwd_aliased`. `*_fwd_aliased` is for when one or more of the input tensors shares storage with the destination tensor.

### DIFF
--- a/candle-core/src/custom_op.rs
+++ b/candle-core/src/custom_op.rs
@@ -388,13 +388,13 @@ pub enum AccessPattern {
 
 impl AccessPattern {
     fn supports(self, rel: LayoutRelation) -> bool {
-        match (self, rel) {
-            (AccessPattern::Elementwise, LayoutRelation::Identical | LayoutRelation::Disjoint) => {
-                true
-            }
-            (AccessPattern::Arbitrary, LayoutRelation::Disjoint) => true,
-            _ => false,
-        }
+        matches!(
+            (self, rel),
+            (
+                AccessPattern::Elementwise,
+                LayoutRelation::Identical | LayoutRelation::Disjoint
+            ) | (AccessPattern::Arbitrary, LayoutRelation::Disjoint)
+        )
     }
 }
 
@@ -692,8 +692,9 @@ impl Tensor {
             let s = match (&guards[i], rels[i]) {
                 (Some(g), None) => Src::Distinct(&**g),
                 (None, Some(rel)) => Src::Aliased(rel),
-                // `guards[i]` is `Some` exactly when `rels[i]` is `None`.
-                _ => unreachable!("guard presence tracks classification"),
+                _ => unreachable!(
+                    "Source is either distinct or aliased. Other match patterns should be impossible"
+                ),
             };
             (s, srcs[i].layout())
         });
--- a/candle-core/src/custom_op.rs
+++ b/candle-core/src/custom_op.rs
@@ -1,6 +1,7 @@
+use crate::layout::LayoutRelation;
 use crate::op::{BackpropOp, Op};
 use crate::tensor::from_storage;
-use crate::{CpuStorage, CudaStorage, Layout, MetalStorage, Result, Shape, Tensor};
+use crate::{bail, CpuStorage, CudaStorage, Layout, MetalStorage, Result, Shape, Storage, Tensor};
 use std::sync::Arc;
 
 /// Unary ops that can be defined in user-land.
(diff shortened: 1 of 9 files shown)


candle-core/src/custom_op.rs
```diff
@@ -352,27 +629,91 @@ pub trait InplaceOp3 {
...
+            Some(g) => g,
+            None => self.storage_mut(),
+        };
+
+        let operands: [(Src<'_, Storage>, &Layout); N] = std::array::from_fn(|i| {
+            let s = match (&guards[i], rels[i]) {
+                (Some(g), None) => Src::Distinct(&**g),
+                (None, Some(rel)) => Src::Aliased(rel),
+                // `guards[i]` is `Some` exactly when `rels[i]` is `None`.
+                _ => unreachable!("guard presence tracks classification"),
```
ivarflakstad: Cryptic. I'll update.

The point is that a source is either distinct or alias - other match patterns are impossible.

candle-core/src/custom_op.rs
```diff
@@ -272,6 +273,200 @@ pub trait InplaceOp1 {
...
+            let [(s, sl)] = srcs;
+            match s {
+                Src::Distinct(s) => InplaceOp2::$fwd(self, dst, dl, s, sl),
+                Src::Aliased(ref rel) => InplaceOp2::$fwd_aliased(self, dst, dl, sl, rel),
+            }
+        }
+    };
+}
+
+impl<C: InplaceOp2> InplaceOpN<1> for C {
```
EricLBuehler, line 462: `src_access_pattern()` remains `None` here, so aliased inputs will bail before `*_fwd_aliased` can run. As a result these new methods are currently unreachable for this generic impl, perhaps we should forward an opt-in from InplaceOp2?

candle-core/src/custom_op.rs
```diff
@@ -352,27 +629,92 @@ pub trait InplaceOp3 {
...
 impl Tensor {
+    /// Applies a custom op in-place for the `self` tensor.
+    ///
+    /// Tensors sharing underlying storage with `self` are classified and passed as [`Src::Aliased`].
+    /// Separate tensors are locked and passed as [`Src::Distinct`].
+    fn inplace_op<const N: usize, C: InplaceOpN<N>>(&self, srcs: [&Self; N], c: &C) -> Result<()> {
+        let name = c.name();
+
+        // Ensure writes cannot collide with themselves
+        if self.layout().has_internal_overlap() {
```
EricLBuehler, line 609: This only detects zero-stride overlap, but some ops like unfold can create overlapping nonzero strides. For example, unfold(0, 3, 1) can produce shape [2, 3] with strides [1, 1], where multiple logical indices address the same cell. Maybe we could conservatively reject layouts that aren’t provably injective (therefore rejecting unfold as a tradeoff but being more correct)?

candle-core/src/tensor.rs
```diff
@@ -2755,33 +2757,41 @@ impl Tensor {
...
+    pub fn storage_and_layout(&self) -> (StorageRef<'_>, &Layout) {
+        let storage = self.storage.read();
         (storage, &self.layout)
     }
 
+    /// Unique key for this tensor's storage. Equal keys mean the tensors share the same allocation.
+    #[inline]
+    pub(crate) fn storage_key(&self) -> usize {
+        let lock: &RwLock<Storage> = self.storage.as_ref();
+        std::ptr::from_ref(lock).addr()
```
EricLBuehler, line 2788: This identifies our Rust-side container but, it seems, not necessarily the underlying allocation for the device. On Metal for instance cloning `Arc<Buffer>` behind a new lock seems like a edge case. Maybe we can make this backend-specific?

candle-core/src/custom_op.rs
ivarflakstad: True. I was thinking we would deprecate `InplaceOp1/2/3` and use `InPlaceOpN` instead.
Having `InplaceOp2` unable to reach `*_fwd_aliased` means it has the same amount of functionality as before - except now it errors instead of deadlocking, so it's mostly just an improvement.

Imo I should either remove `InplaceOp1/2/3::*_fwd_aliased` (as you can't reach them anyway) or we'd have to add `src_access_pattern` to those traits as well.

I'm thinking remove the unreachable fns

candle-core/src/tensor.rs
ivarflakstad: Hmm yeah. I kinda extracted this from `same_storage`, so this has been the case for a while. You're still right though.

Maybe add something like `fn buffer_key` to `Storage`?

candle-core/src/custom_op.rs
ivarflakstad: Think I managed to fix this with the new impl

candle-core/src/custom_op.rs
ivarflakstad: Updated 🫡

EricLBuehler (approved):
The remaining backend allocation comment is still true but I agree that this can be handled as a follow-up.

Looks good!

Issue #3915, opened 2026-08-24
Unable to build whisper example


inferrna:
rustc 1.98.0
Ubuntu 24.04
Freshly cloned candle repo with no modifications

Build with symphonia as it is required
`cargo run --example whisper --release --features symphonia`
```Rust
   //Few lines skipped..
   Compiling candle-transformers v0.11.0 (/home/inferno/.dev/Alien/candle/candle-transformers)
warning: ignoring -C extra-filename flag due to -o flag

error[E0432]: unresolved imports `symphonia::core::audio::AudioBufferRef`, `symphonia::core::audio::Signal`
  --> candle-examples/src/audio.rs:33:34
   |
33 |     use symphonia::core::audio::{AudioBufferRef, Signal};
   |                                  ^^^^^^^^^^^^^^  ^^^^^^ no `Signal` in `audio`
   |                                  |
   |                                  no `AudioBufferRef` in `audio`
   |
help: a similar name exists in the module
   |
33 -     use symphonia::core::audio::{AudioBufferRef, Signal};
33 +     use symphonia::core::audio::{AudioBuffer, Signal};
   |

error[E0432]: unresolved imports `symphonia::core::codecs::DecoderOptions`, `symphonia::core::codecs::CODEC_TYPE_NULL`
  --> candle-examples/src/audio.rs:34:35
   |
34 |     use symphonia::core::codecs::{DecoderOptions, CODEC_TYPE_NULL};
   |                                   ^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^ no `CODEC_TYPE_NULL` in `codecs`
   |                                   |
   |                                   no `DecoderOptions` in `codecs`

error[E0432]: unresolved import `symphonia::core::conv`
  --> candle-examples/src/audio.rs:35:26
   |
35 |     use symphonia::core::conv::FromSample;
   |                          ^^^^ could not find `conv` in `core`

error[E0433]: cannot find `sample` in `core`
  --> candle-examples/src/audio.rs:41:29
   |
41 |         T: symphonia::core::sample::Sample,
   |                             ^^^^^^ could not find `sample` in `core`

error[E0433]: cannot find `conv` in `core`
  --> candle-examples/src/audio.rs:42:31
   |
42 |         f32: symphonia::core::conv::FromSample<T>,
   |                               ^^^^ could not find `conv` in `core`

error[E0433]: cannot find `probe` in `core`
  --> candle-examples/src/audio.rs:54:33
   |
[...]

PR #3931, opened 2026-08-31
Fix the symphonia-gated examples build against the 0.6 audio API

HuzaifaAbdulRehman:
`cargo run --example whisper --release --features symphonia` has not compiled since 28 May 2026, when
#3546 bumped symphonia from 0.5.3 to 0.6.0 and moved most of the modules `pcm_decode` uses. No CI job
enables the `symphonia` feature, so `cargo check --workspace` stayed green the whole time. Reported in
#3915.

The first failure is `candle-examples/src/audio.rs:33`, importing `AudioBufferRef` and `Signal`,
neither of which exists in 0.6. Five more imports moved with it: `core::conv`, `core::sample`,
`core::probe`, `codecs::DecoderOptions` and `codecs::CODEC_TYPE_NULL`.

This ports `pcm_decode` in `candle-examples/src/audio.rs` and in the three copies under
`examples/{encodec,mimi,snac}/audio_io.rs` to `GenericAudioBufferRef`, the `Audio` trait's `plane(0)`,
`first_track_known_codec(TrackType::Audio)`, `make_audio_decoder`, and a `next_packet` that now yields
an `Option`. No public API changes.

Two behaviour differences, both narrowing. Track selection is restricted to audio tracks, where the old
scan took the first track of any type with a known codec. And `plane(0)` returns an `Option`, which I
handle by skipping that buffer.

I have left `resample` alone. It is separately broken against rubato 2, since #3544 bumped rubato 1 to
2 and `rubato::FftFixedInOut` is gone. That still blocks `encodec`, `mimi`, `snac` and `voxtral`. It
is a much bigger migration to the new `Fft`/`FixedSync` buffer model, so it belongs in its own issue
rather than buried here.

## Test Plan

WSL Ubuntu, x86_64, rustc 1.97.1, at 79462df.

```
before: cargo check -p candle-examples --features symphonia
        error: could not compile `candle-examples` (lib) due to 6 previous errors
after:  cargo check -p candle-examples --example whisper --features symphonia   ok
        cargo clippy -p candle-examples --lib --features symphonia -- -D warnings   ok
        cargo fmt --all -- --check   ok
```

The three `audio_io.rs` files pull in `cpal` and this box has no ALSA headers, so rather than assume
that code compiles I lifted the patched `pcm_decode` out of `encodec/audio_io.rs` verbatim into a
scratch crate on symphonia 0.6.1 and decoded two generated 16-bit WAVs:

```
mono   sample_rate=16000 n=1000  max abs error vs expected = 0e0
stereo sample_rate=16000 n=1000  max abs error vs left channel = 0e0
```

Stereo returning 1000 samples rather than 2000 interleaved is the check that matters. It shows
`plane(0)` keeps the single-channel behaviour `chan(0)` had.


ivarflakstad (approved):
Thank you!

PR #3934, opened 2026-08-31
Initial integration of cutile and fused MoE kernel

EricLBuehler:
This PR adds [cutile-rs](https://github.com/NVlabs/cutile-rs) integration to Candle, allowing users to write JIT-compiled CUDA kernels in Rust and launch them from Candle custom ops!

This effectively enables Triton-like custom kernel development for Candle while remaining fully opt-in through a new `cutile` feature.

It provides:

- Interop with Candle’s existing CUDA context, stream, tensors, and synchronization.
- Explicit kernel warmup/precompilation and persistent JIT caching.
- An optimized BF16 routed grouped-matmul primitive for MoE, with reusable GPU routing metadata and optional router weights.


candle-core/src/cuda_backend/mod.rs
```diff
@@ -15,6 +15,8 @@ use std::sync::{Arc, Mutex, OnceLock};
 
 #[cfg(feature = "cudnn")]
 pub mod cudnn;
+#[cfg(feature = "cutile")]
+pub mod cutile;
```
ivarflakstad, line 19: Do we actually need to hide cutile behind a feature flag?
It makes sense in terms of adding support incrementally / having it be opt-in, but since it's JIT we could detect if the machine has `>=sm_80` etc at runtime and just launch cutile kernels if true?

.github/workflows/ci_cuda.yaml
ivarflakstad, line 1: The cuda (not cutile) job errored out with:
> Canceling since a higher priority waiting request for CI / cuda--candle_cutile_integration exists

Let's ensure that either both run or that `test-cutile` also covers all of `test-cuda` 👍

ivarflakstad (commented):
Scrumptious 🧑‍🍳 🔥 

Just need to fix the CI issue and I'll approve.

We don't have an older machine `< sm_80` laying around we could test this on, do we? To ensure backwards compatibility isn't an issue.

candle-core/src/cuda_backend/mod.rs
EricLBuehler: Right now cutile-rs has build-time requirements (CUDA 13.2+ toolkit headers for bindgen, clang/libclang, Rust 1.89) and at runtime (tileiras, sm_80+), so enabling it unconditionally would break some uses of `--features cuda`. That's why I made it opt-in for now. Longer term I would like runtime enabling, but it needs more than an sm_80 check, as the system has to be set up correctly, and the supported arch set depends on the toolkit (sm_90 needs 13.3). It's a lot of logic, and I think we can do it in the future if we think it'd make things easier to use.

EricLBuehler:
cutile-rs doesn't support < sm_80 and the feature is opt-in, so existing users are unaffected.

.github/workflows/ci_cuda.yaml
EricLBuehler: Fixed!

ivarflakstad:
> cutile-rs doesn't support < sm_80 and the feature is opt-in, so existing users are unaffected.

That is the behaviour I expected to see, but I've been surprised before. I'm mostly wondering how confused a user would be if they had `< sm_80` and tried to use `--features=cutile`.

ivarflakstad (approved):
🎉

PR #3406, opened 2026-03-16
Add candle-mi library to useful libraries list

PCfVW:
Hi,

This adds `candle-mi` to *Useful External Resources*, following [discussion #3368](https://github.com/huggingface/candle/discussions/3368):

> "I think that `candle-mi` would be a great name. Additionally, you can open a PR to add this crate into the `Useful External Resources` of the readme." — @EricLBuehler

**What it is.** Mechanistic interpretability for language models in Rust — as far as I can tell still the only Rust crate for it; the Python ecosystem has TransformerLens, nnsight and pyvene. TransformerLens-style hook points are built into the forward pass (capture plus typed interventions, zero overhead when inactive), with attention knockout and steering, logit lens, cross-layer transcoders and SAEs, over a config-driven generic transformer (LLaMA, Qwen, Gemma, Phi-3, StarCoder2), RWKV-6/7, and masked-diffusion LMs. Everything runs on consumer hardware (developed on an RTX 5060 Ti 16GB).

[crates.io](https://crates.io/crates/candle-mi) (v0.1.21) · [docs.rs](https://docs.rs/candle-mi)

**Since I opened this PR**, using candle-mi in anger has fed back into candle itself:

- [#3819](https://github.com/huggingface/candle/pull/3819) — `AdamW::step_t`, `set_step_t` and `moments`, so training runs can be checkpointed
- [#3822](https://github.com/huggingface/candle/pull/3822) — `GradStore`: store the first gradient instead of adding it into fresh zeros
- [#3823](https://github.com/huggingface/candle/pull/3823) — an analytic backward for the fused `softmax_last_dim` and `layer_norm`, addressing [issue #3011](https://github.com/huggingface/candle/issues/3011) and deferring to the open [#3613](https://github.com/huggingface/candle/pull/3613) and [#3724](https://github.com/huggingface/candle/pull/3724)

The diff is one line. Thanks for candle! It's been a pleasure to work with for this kind of research. :)

-- Éric Jacopin.

PR #3944, opened 2026-09-02
Fix CUDA copy_strided_src over-copying from an offset source view

### LINKED ISSUE #3940: CUDA `copy_strided_src` copies past the end of an offset source view, silently corrupting `slice_scatter` (CPU disagrees)
## Summary

On the CUDA backend, `copy_strided_src` sizes its copy from the **whole storage lengths** rather than from the source view. When the source is an offset view whose storage continues past the view (anything produced by `narrow`, which is stride-contiguous but starts partway into its parent), and the destination has room, it copies too many elements.

`Tensor::slice_scatter` is the visible casualty: scattering an offset source into position `p` also overwrites the elements **after** `p` with whatever follows the source in its own storage. No error is raised, and the CPU backend gives a different (correct) answer for the same call.

Observed on 0.11.0 and on current `main` (`d5fee525`); the code below is unchanged between them.

## The discrepancy

```rust
// candle-core/src/cpu_backend/mod.rs:902 -- copies exactly the view's length
StridedBlocks::SingleBlock { start_offset, len } =>
    dst[dst_offset..dst_offset + len]
        .copy_from_slice(&src[start_offset..start_offset + len])
```

```rust
// candle-core/src/cuda_backend/mod.rs:1208 -- derives the length from storage
fn slice_src_and_dst<'a, T>(
    src: &'a CudaSlice<T>,
    src_l: &Layout,
    dst: &'a mut CudaSlice<T>,
    dst_offset: usize,
) -> (cudarc::driver::CudaView<'a, T>, cudarc::driver::CudaViewMut<'a, T>) {
    let src_offset = src_l.start_offset();
    let to_copy = dst
        .len()
        .saturating_sub(dst_offset)
        .min(src.len().saturating_sub(src_offset));
    let src = src.slice(src_offset..src_offset + to_copy);
    let dst = dst.slice_mut(dst_offset..dst_offset + to_copy);
    (src, dst)
}
```

`to_copy` should be `src_l.shape().elem_count()`. It only equals that by accident, when the destination is no larger than the source view. That happens to hold for `Tensor::copy` / `try_clone`, which is presumably why this has gone unnoticed. It does not hold for `slice_scatter0`, where `dst` is the whole output and `dst_offset` points into it.

## Reproduction

```rust

    for device in [Device::Cpu, Device::new_cuda(0)?] {
        // Recipient [1, 4, 3].
        let base = Tensor::new(
            &[[[0f32, 1., 2.], [3., 4., 5.], [6., 7., 8.], [9., 10., 11.]]],
            &device,
        )?;
        // Donor [4, 3]; take row 2 as a view, so the view starts at offset 6
        // and its storage still holds rows 3 onwards.
        let donor = Tensor::new(
            &[
                [90f32, 91., 92.],
                [93., 94., 95.],
                [96., 97., 98.],
                [99., 100., 101.],
            ],
            &device,
        )?;
        let row = donor.narrow(0, 2, 1)?.unsqueeze(0)?; // [1, 1, 3]

        let out = base.slice_scatter(&row, 1, 2)?;
        println!("{device:?}: {:?}", out.flatten_all()?.to_vec1::<f32>()?);
    }
    Ok(())
}
```

Only position 2 was asked for, so both devices should print

```
[0, 1, 2, 3, 4, 5, 96, 97, 98, 9, 10, 11]
```

Actual:

```
Cpu                       : [0, 1, 2, 3, 4, 5, 96, 97, 98,  9,  10,  11]   correct
Cuda(CudaDevice(DeviceId(1))): [0, 1, 2, 3, 4, 5, 96, 97, 98, 99, 100, 101]   position 3 clobbered
```

Position 3 has been overwritten with the donor's row 3, which the caller never referenced. Making the source own its storage (`Tensor::new`, or any op that allocates by `elem_count`) hides the bug; `.contiguous()` does **not**, because a `narrow` view is already contiguous and keeps its parent's storage.

`Tensor::cat` over the same narrowed pieces was tested alongside and is correct on both backends, so this appears reachable through `slice_scatter`'s use of the helper rather than through every caller of it.

## Suggested fix

```rust
let to_copy = src_l.shape().elem_count();
```

keeping the clamp against `dst.len() - dst_offset` if a defensive bound is still wanted. `slice_src_and_dst` is only reached from the `src_l.is_contiguous()` branch of `copy_strided_src`, where `elem_count` is exactly the number of elements the caller means to move.

## How it surfaced

Downstream, an activation-patching intervention overwrites one sequence position of a `[batch, seq_len, hidden]` residual stream with a row taken from another forward pass. That donor row is naturally a `narrow` view into a captured activation. On CUDA the patch also overwrote every position after the patch site.

The symptom was not a crash but a plausible number: a causal-tracing table reporting 100% recovery at every layer for every token position but the last, including the final layer, where patching a non-final position cannot affect the logits at all. Unit tests passed on CPU **and** CUDA, because they built the source with `Tensor::new`; only a donor row taken as a view reproduces it.

## Related

This is the mirror image of #3874, where the CPU backend read a view as though it were the whole storage and the GPU backends were correct. Here CUDA does it and the CPU is correct. It sits with the other recent view/offset fixes (#3735 / #3736, #3893 / #3894, #3853, #3836), which is the argument for fixing the shared helper rather than each call site.

PCfVW:
Fixes #3940.

## The bug

`copy_strided_src` is expected to copy the elements described by `src_l`, that
is `src_l.shape().elem_count()` of them. The CPU backend does. The CUDA backend
derives the count from the **whole storage lengths** instead:

```rust
// candle-core/src/cuda_backend/mod.rs, slice_src_and_dst
let to_copy = dst
    .len()
    .saturating_sub(dst_offset)
    .min(src.len().saturating_sub(src_offset));
```

That equals `elem_count` only by accident, when the destination happens to be no
larger than the source view. It holds for `Tensor::copy` / `try_clone`, which is
presumably why this has gone unnoticed. It does not hold for `slice_scatter0`,
where `dst` is the whole output and `dst_offset` points into it.

So when the source is a view whose storage continues past it (anything produced
by `narrow`, which is stride-contiguous but starts partway in), the copy runs
past the end of the view and overwrites the elements after the scatter point.
Silently, and only on CUDA.

CUDA is the only backend that does this. The other two already take the length
from the layout, which is what this PR changes CUDA to do:

```rust
// candle-core/src/cpu_backend/mod.rs, copy_strided_src_
StridedBlocks::SingleBlock { start_offset, len } =>
    dst[dst_offset..dst_offset + len]
        .copy_from_slice(&src[start_offset..start_offset + len])
```

```rust
// candle-core/src/metal_backend/mod.rs, copy_strided_src, contiguous path
let src_offset = src_l.start_offset() * self.dtype.size_in_bytes();
let length = src_l.shape().elem_count() * self.dtype.size_in_bytes();
```

## Reproduction

```rust

    for device in [Device::Cpu, Device::new_cuda(0)?] {
        let base = Tensor::new(
            &[[[0f32, 1., 2.], [3., 4., 5.], [6., 7., 8.], [9., 10., 11.]]],
            &device,
        )?;
        // Donor rows 0..4; take row 2 as a view, so the view starts at offset 6
        // and its storage still holds row 3.
        let donor = Tensor::new(
            &[[90f32, 91., 92.], [93., 94., 95.], [96., 97., 98.], [99., 100., 101.]],
            &device,
        )?;
        let row = donor.narrow(0, 2, 1)?.unsqueeze(0)?; // [1, 1, 3]

        let out = base.slice_scatter(&row, 1, 2)?;
        println!("{device:?}: {:?}", out.flatten_all()?.to_vec1::<f32>()?);
    }
    Ok(())
}
```

Only position 2 was asked for, so both devices should print

```
[0, 1, 2, 3, 4, 5, 96, 97, 98, 9, 10, 11]
```

Before this change:

```
Cpu : [0, 1, 2, 3, 4, 5, 96, 97, 98,  9,  10,  11]   correct
Cuda: [0, 1, 2, 3, 4, 5, 96, 97, 98, 99, 100, 101]   position 3 clobbered
```

Position 3 has been overwritten with the donor's row 3, which the caller never
referenced.

## The change

Size the copy from the source layout, keeping the two storage bounds as a
defensive clamp so the value can only ever shrink relative to today:

```rust
let to_copy = src_l
    .shape()
    .elem_count()
    .min(dst.len().saturating_sub(dst_offset))
    .min(src.len().saturating_sub(src_offset));
```

`slice_src_and_dst` is only reached from the `src_l.is_contiguous()` branch of
`copy_strided_src`, where `elem_count` is exactly the number of elements the
caller means to move.

## Test

Extended `slice_scatter` in `candle-core/tests/tensor_tests.rs`, so it runs on
all three backends through the existing `test_device!` registration. Two cases,
one with a view starting at offset 0 and one starting partway into its storage,
since only the latter exercises `start_offset` as well as the length.

The new assertions fail on CUDA before this change and pass after; CPU and Metal
are unaffected either way.

## Notes

`Tensor::cat` over equivalent narrowed pieces was checked alongside and is
correct on both backends, so this appears reachable through `slice_scatter`'s
use of the helper rather than through every caller of it.

Found downstream, where an activation-patching intervention overwrites one
sequence position of a `[batch, seq_len, hidden]` residual stream with a row
taken from another forward pass. That donor row is naturally a `narrow` view
into a captured activation. The symptom was not a crash but a plausible number:
a causal-tracing table reporting 100% recovery at every layer for every token
position but the last, including the final layer, where patching a non-final
position cannot affect the logits at all.


PCfVW:
@ivarflakstad, a note about CI rather than about the patch itself.

This is a fork PR, so its runs sit at `action_required` until a maintainer approves them. Four earlier ones are in the same state: #3406 has been pending since 3 August, and #3819, #3822 and #3823 since 31 August.

After thirty days GitHub expires an unapproved run and stamps it `conclusion: failure` with zero jobs executed. That is what happened to the August batch, so the resulting notifications read as test failures when nothing was ever compiled. The three from 31 August come due on 30 September.

No urgency on review. All four are rebased on current `main` and green locally. I am flagging it once here rather than four times.

EricLBuehler (approved):
Thanks @PCfVW

PR #3949, opened 2026-09-03
Fix rotary embedding convention in MiMi transformer

### LINKED ISSUE #3948: mimi uses the interleaved RoPE convention but the checkpoints are stored in the transformers layout
The mimi transformer applies the interleaved (GPT-J style) rotary embedding, but the
`kyutai/mimi` checkpoint that `Model::new` loads is stored in the `transformers` layout, which
rotates the two halves of the head dimension (NeoX style). Attention scores are therefore
computed against the wrong frequency pairing at every position but the first, and both the
encoder and the decoder drift away from the reference implementation as the sequence grows.

https://github.com/huggingface/candle/blob/c30b26cd3bb59e5b39927eebff00884b0c0729d5/candle-transformers/src/models/mimi/transformer.rs#L82

Transformers uses `rotate_half` for this model, [models/mimi/modeling_mimi.py](https://github.com/huggingface/transformers/blob/v4.57.3/src/transformers/models/mimi/modeling_mimi.py#L542),
which is what `candle_nn::rotary_emb::rope` implements.

## Impact

Everything that goes through `models::mimi`, which is the `mimi` and `csm` examples. Encoded
codes are wrong and decoded audio is measurably degraded, but it is a quiet failure. Audio still
sounds like speech, and nothing errors out.

## Reproduction

A 5.04 s mono clip at 24 kHz, 63 frames, f32 on CPU, candle at d5fee525, where this file is
identical to the permalink above.

Codes from `transformers` 4.57.3:

```python
import torch, soundfile as sf
from transformers import MimiModel

wav, _ = sf.read("audio.wav", dtype="float32")            # mono, 24 kHz
m = MimiModel.from_pretrained("kyutai/mimi", dtype=torch.float32).eval()
with torch.no_grad():
    codes = m.encode(torch.from_numpy(wav)[None, None, :], num_quantizers=16).audio_codes
```

Codes from candle, same checkpoint, same audio:

```bash
cargo run --example mimi --features mimi -r -- --cpu audio-to-code audio.wav codes.safetensors
```

**359 of the 1008 codes match, and not one of the 63 frames matches in full.** Agreement is
highest in the first frame, 14 of its 16 codes, and falls off immediately after, which is the
signature of a position-dependent difference rather than a numerical one.

Feeding the `transformers` codes back through `Model::decode` and comparing samples with
`MimiModel.decode`:

| quantizers | max abs sample difference | correlation |
|---|---|---|
| 16 | 1.2e-01 | 0.9959 |
| 32, as used by `csm` | 5.1e-02 | 0.9978 |

Comparing codes is the cleaner check of the model itself, since the `mimi` example applies
`normalize_loudness` before writing its wav.

jorge-menjivar:
fixes #3948

`RotaryEmbedding::apply_rotary_emb` in the mimi transformer applies the interleaved (GPT-J
style) rotary embedding, but the checkpoints are stored in the `transformers` layout, which
rotates the two halves of the head dimension (NeoX style). Attention is computed against the
wrong frequency pairing at every position but the first, so the encoder and the decoder both
drift away from the reference implementation as the sequence grows.

This switches that one call to `candle_nn::rotary_emb::rope`, which is what `rotate_half` in
`transformers` implements.

## Verification

Against `MimiModel` from `transformers` 4.57.3 on `kyutai/mimi`, a 5.04 s mono clip at 24 kHz,
63 frames, f32 on CPU. Encoding the same audio through `cargo run --example mimi`:

| | codes matching `transformers` | frames matching in full |
|---|---|---|
| before | 359 of 1008 | 0 of 63 |
| after | 1008 of 1008 | 63 of 63 |

Decoding the `transformers` codes with `Model::decode` and comparing samples with
`MimiModel.decode`:

| quantizers | max abs sample difference | correlation |
|---|---|---|
| 16, before | 1.2e-01 | 0.9959 |
| 16, after | 7.9e-07 | 1.000000 |
| 32 as used by `csm`, before | 5.1e-02 | 0.9978 |
| 32 as used by `csm`, after | 1.2e-06 | 1.000000 |

Both consumers of the module still build and run. The `mimi` example round trips audio to codes
and back, including streaming with frame sized chunks. The `csm` example builds; its weights are
gated on the hub, so instead of an end to end run I exercised the 32 quantizer decode path it
uses, which is the last two rows above.

### DIFF
--- a/candle-transformers/src/models/mimi/transformer.rs
+++ b/candle-transformers/src/models/mimi/transformer.rs
@@ -79,7 +79,9 @@ impl RotaryEmbedding {
         let qk_dtype = qk.dtype();
         let c = self.cos.narrow(0, seqlen_offset, seqlen)?;
         let s = self.sin.narrow(0, seqlen_offset, seqlen)?;
-        candle_nn::rotary_emb::rope_i(&qk.to_dtype(DType::F32)?, &c, &s)?.to_dtype(qk_dtype)
+        // The checkpoints are stored in the transformers layout, which rotates the two halves
+        // of the head dimension rather than adjacent pairs.
+        candle_nn::rotary_emb::rope(&qk.to_dtype(DType::F32)?, &c, &s)?.to_dtype(qk_dtype)
     }
 }
 


EricLBuehler (approved):
Thank you @jorge-menjivar!

Issue #3253, opened 2025-12-20
broadcast_matmul to handle stride-0 broadcast dimensions


DrJesseGlass:
While implementing CPU Grouped-Query Attention, I encountered a limitation with `broadcast_matmul` and stride-0 dimensions. I'm taking a different approach (fused SIMD kernels) for my use case, but wanted to log this for others who may benefit from the fix.

Currently `broadcast_as` creates stride-0 view and `matmul` rejects stride-0 as "non-contiguous". So, I needed to use the workaround: `.broadcast_as(...).contiguous()` but this physically expands memory.

```rust
// GQA: Q has 16 heads, K/V have 8 heads (2 groups)
let q = ...;  // [1, 8, 2, 2, 128]  
let k = ...;  // [1, 8, 1, 128, 2] ← size-1 dim should broadcast

let scores = q.broadcast_matmul(&k)?;  // Error: "non-contiguous lhs"
```

I expected it to broadcast dim 2 of rhs (1 → 2) during matmul  
But instead I got an error after internal broadcast creates stride-0

PR #3960, opened 2026-09-06
Fix silently incorrect CPU matmul for a stride-zero broadcast batch

### LINKED ISSUE #3744: CPU matmul silently returns incorrect values for stride-zero broadcast batch
### Bug

On Candle 0.11.0, the baseline CPU backend silently returns incorrect values when batched `matmul` receives a stride-zero batch dimension created by `broadcast_as`. The Accelerate CPU provider returns the correct values for the same tensors.

This is related to #3253, but that report covers `broadcast_matmul` rejecting a stride-zero layout. This report concerns direct `broadcast_as(...).matmul(...)` succeeding with incorrect output.

### Minimal reproducer

```rust
use candle_core::{DType, Device, Result, Tensor};

fn main() -> Result<()> {
    let device = Device::Cpu;
    let left = Tensor::ones((1, 32, 32), DType::F32, &device)?;
    let right = Tensor::ones((32, 32, 32), DType::F32, &device)?;

    let stride_zero = left.broadcast_as((32, 32, 32))?.matmul(&right)?;
    let eager = left.repeat((32, 1, 1))?.matmul(&right)?;

    assert_eq!(
        stride_zero.flatten_all()?.to_vec1::<f32>()?,
        eager.flatten_all()?.to_vec1::<f32>()?,
    );
    Ok(())
}
```

With baseline `candle-core = "0.11.0"`, the assertion fails. With `candle-core = { version = "0.11.0", features = ["accelerate"] }` on the same Apple Silicon host, it passes.

### Expected behavior

`matmul` should either honor the stride-zero batch layout and match the eagerly expanded input, or reject the unsupported layout. It should not succeed with incorrect values.

### Environment

- Candle 0.11.0
- Rust 1.94.1, edition 2024
- macOS arm64 / Apple M4 Max
- Baseline CPU provider: incorrect values
- Accelerate CPU provider: correct values

I can prepare a PR once the desired baseline CPU behavior and regression-test location are confirmed.

akashjainn:
Fixes #3744.

`MatMul::f` folds a batched matmul into a single gemm when one side is
batch-invariant. The lhs-broadcast arm is not valid:

```rust
} else if a_skip == 0 && b_skip == n * k {
    (1, m, b * n, k)
}
```

Merging the batch into `n` needs the batches of rhs to be adjacent columns of a
`(k, b * n)` matrix, but rhs is batch-major. And `dst_rs` stays `n` while each
merged row is `b * n` wide, so the gemm writes over itself: only
`(m - 1) * n + b * n` elements are written and the rest keep the zero they were
initialised with. For the reporter's `(32, 32, 32)` case that is 2016 of 32768
correct, the other 30752 silently zero.

The mirrored arm above it is fine, which is what makes this look plausible. A
batch-invariant rhs stacks the batches of lhs into the *rows* of a `(b * m, k)`
matrix, and the destination already matches row for row.

This drops the fold. No replacement is needed: `ab_skip` already yields
`a_skip == 0` for a stride-zero batch, and the per-batch loop then feeds the same
lhs to every step, which is the correct broadcast semantics. I could not find a
valid fold to put back — the transposed formulation would need the destination
laid out `(b, n, m)` and it is `(b, m, n)` — so a broadcast lhs now issues `b`
gemm calls, the same as any other batched matmul. I left a comment saying why,
since the symmetry is inviting.

Only the baseline CPU path was affected. Metal and Accelerate already gave the
right answer, and the CUDA path sets `stride_b = 0` in its `StridedBatchedConfig`,
which is how cuBLAS broadcasts an operand.

`broadcast_matmul` never hit this, because it calls `.contiguous()` on a
broadcast lhs (`tensor.rs:1585`) — only a direct `broadcast_as(...).matmul(...)`
reaches it. Worth noting for the `// TODO: Avoid concretising the broadcasted
matrixes via contiguous.` on line 1563: that fast path needs the backend to
honour a stride-zero batch, which it now does at rank 3.

Related: #3253 is the same `ab_skip` match at higher rank, where three or more
leading strides fall to the catch-all and error instead. Left alone here.

## Test Plan

`broadcast_matmul_stride_zero_lhs` runs `broadcast_as(...).matmul(...)` against
the per-batch products it stands for, over `(32,32,32,32)` plus `m=1`, `k=1` and
`n=1` shapes. Registered for cpu/gpu/metal like the neighbouring tests. It fails
on main and passes with this change.

macOS arm64, rustc 1.98.1, at ddf1b879:

```
cargo test --workspace --lib --bins --tests           68 binaries, 0 failures
cargo test --workspace --doc                          ok
cargo test -p candle-core --features metal            broadcast_matmul_stride_zero_lhs_metal ok
cargo test -p candle-core --features accelerate       ok
cargo clippy --workspace --tests --examples --benches -- -D warnings   ok
cargo fmt --all -- --check                            ok
cargo check --workspace --all-targets                 ok
cargo check --workspace --all-targets --features metal ok
```

The reporter's original program now prints instead of panicking.

CUDA checked separately on an RTX 3060 (CUDA 12.9, `CUDA_COMPUTE_CAP=86`), since
`ci_cuda.yaml` skips forks:

```
cargo test -p candle-core --features cuda --test matmul_tests
...
test broadcast_matmul_stride_zero_lhs_gpu ... ok
test result: ok. 22 passed; 0 failed
```

That confirms the reasoning for leaving `cuda_backend` alone: `gemm_config` builds
a `StridedBatchedConfig` with `stride_b = 0`, which is how cuBLAS broadcasts an
operand, so the batch-invariant lhs was already correct there.


EricLBuehler (approved):
Thank you @akashjainn! Looks great.

PR #3961, opened 2026-09-07
fix(pickle): avoid unbounded allocation in BinUnicode string read (CWE-770)

li-jin-quan:
I found a small DoS in the pickle reader.

`OpCode::BinUnicode` reads a 4-byte length from the file, then allocates that much **before** reading any of the string:

```rust
let len = r.read_u32::<LittleEndian>()?;
let mut data = vec![0u8; len as usize];   // <-- size comes from the file
r.read_exact(&mut data)?;
```

A `u32` can say up to 4 GiB. So a **12-byte** pickled string can ask for 1 GiB.

I measured it with a counting allocator: **1 GiB allocated from 12 bytes** (8 bytes of real payload). On a small machine this aborts the process.

This path runs on every `.pth` / `.ckpt` load (`PthTensors`, `read_all`, `VarBuilder::from_pth`), and those files are usually downloaded from a model hub.

Fix: read with `take(len)` instead of pre-allocating, then check the length.

| input | before | after |
|---|---|---|
| 12 bytes, 1 GiB string | 1,073,741,824 bytes | 8 KB + a clean "truncated" error |
| well-formed checkpoint | unchanged | unchanged |

`take` caps how much is consumed, so buffered bytes stay available for the next opcode — well-formed files behave exactly as before.

Added two tests: string roundtrip, and the truncated case.

Does this look right to you? Happy to adjust anything.


li-jin-quan:
Hi — gentle ping, this one's been open about a week.

Quick summary: BinUnicode trusts the length prefix in the pickle stream, so a crafted file can trigger a huge allocation before any data is actually read. This checks the remaining bytes first.

Happy to change the approach if you'd prefer something else. Worth taking?

PR #3963, opened 2026-09-07
quantized: validate the raw GGML byte slice before reinterpreting it as blocks

### LINKED ISSUE #3815: quantized/ggml_file.rs: from_raw_parts on a 1-byte-aligned pointer without an alignment check
`candle-core/src/quantized/ggml_file.rs:128` builds a typed slice from a byte pointer with no alignment check:

```rust
std::slice::from_raw_parts(raw_data_ptr as *const T, n_blocks)
```

`raw_data_ptr` is a `*const u8` derived from the file buffer, so it carries 1-byte alignment. Creating a `&[T]` where `T` has stricter alignment is undefined behaviour regardless of whether the access happens to work.

Miri flags it directly:

```
error: Undefined Behavior: constructing invalid value: encountered an unaligned reference
(required 4 byte alignment but found 1)
```

This is unconditional on that code path — it is not triggered by any malformed input, it happens on ordinary GGUF loads. In practice it is benign on the usual allocators (malloc returns ≥16-byte-aligned blocks) and on x86/ARM64 where unaligned loads work, which is presumably why it hasn't caused visible problems.

The reason I think it's worth fixing rather than ignoring: **candle's own sibling helper already does the check.** `quantized/mod.rs::as_t_slice` asserts both the length-multiple and the alignment. `from_raw_data` doesn't. That inconsistency is the actual defect — one of the two raw-slice constructors validates and the other doesn't.

Suggested fix: mirror the `as_t_slice` assertions in `from_raw_data`, or route it through the same helper.

Reported as a soundness/UB issue, not a security vulnerability — I found no way to turn it into an out-of-bounds access.

apollo-2006:
Fixes #3815.

`from_raw_data` in `candle-core/src/quantized/ggml_file.rs` turned a `&[u8]` into a `&[T]` with no checks:

```rust
let raw_data_ptr = raw_data.as_ptr();
let n_blocks = size_in_bytes / std::mem::size_of::<T>();
let data = unsafe { std::slice::from_raw_parts(raw_data_ptr as *const T, n_blocks) };
```

Two separate preconditions were missing. `qtensor_from_ggml` is public and safe, so both are reachable from safe code.

### 1. Length is never checked (not in the issue)

`size_in_bytes` is computed from `dims` and never compared against `raw_data.len()`. A buffer shorter than the dims imply produces a slice running past the end of the allocation. Miri:

```
error: Undefined Behavior: constructing invalid value of type &[BlockQ4_0]:
encountered a dangling reference (going beyond the bounds of its allocation)
 --> candle-core/src/quantized/ggml_file.rs:128:24
    candle_core::quantized::ggml_file::from_raw_data::<BlockQ4_0>
    candle_core::quantized::ggml_file::qtensor_from_ggml
```

This one is a plain out-of-bounds read, so it is now rejected with an error.

### 2. Alignment, as reported

A `&[u8]` carries 1-byte alignment while `&[T]` requires `align_of::<T>()`. Worth adding to the issue: this does not just fail under Miri. A debug build aborts on the standard library's own precondition check, so an unaligned caller crashes outright:

```
unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be
aligned and non-null, and the total size of the slice not to exceed `isize::MAX`
signal: 6, SIGABRT
```

A zero-element tensor hits the same abort today, because an empty `Vec<u8>` has a dangling pointer of `0x1`, which is not aligned for the block type.

### Why this does not assert the alignment

The issue suggests mirroring the assertions in `quantized/mod.rs::as_t_slice`. I started there, and it is not safe to do: the alignment a `Vec<u8>` ends up with is decided by the allocator, not the caller, so rejecting an unaligned buffer rejects valid data.

Miri makes this concrete, since it gives a `Vec<u8>` only the 1-byte alignment it actually guarantees. With an assertion in place, an ordinary well-formed tensor is rejected:

```
called `Result::unwrap()` on an `Err` value: ggml tensor data is not aligned to 4 bytes
```

and in the same run a deliberately offset buffer came out aligned and was accepted. The alignment of a byte buffer simply is not something to branch correctness on.

So the slice is borrowed when the buffer happens to be aligned and copied into an aligned `Vec<T>` when it is not. That is always correct, costs nothing on the usual path where the allocator over-aligns, and removes the zero-element abort as a side effect.

Copying is sound for these types for the same reason `GgmlType::zeros` is: the ggml block types are plain data.

Worth flagging separately: `as_t_slice` has the same latent fragility, since its `assert_eq!` on alignment would panic on valid data under such an allocator. I left it alone to keep this focused, happy to follow up if you want it changed.

### Tests

`candle-core/tests/ggml_file_tests.rs`, asserting on loaded contents rather than on which branch is taken, so they hold under either allocator. Against the unfixed code:

| test | before | after |
| --- | --- | --- |
| `unaligned_data_loads_correctly` | SIGABRT | pass |
| `accepts_empty_tensor` | SIGABRT | pass |
| `rejects_truncated_data` | silently accepted | pass |
| `aligned_data_loads_correctly` (control) | pass | pass |

Verified: `cargo test -p candle-core` 232 passed 0 failed, `cargo clippy -p candle-core --all-targets -- -D warnings` clean, `cargo fmt --check` clean, and `cargo +nightly miri test -p candle-core --test ggml_file_tests` reports no undefined behaviour.


apollo-2006:
@ivarflakstad flagging one thing up front, since it is easy to read this PR as not doing what #3815 asked for.

The issue proposes mirroring the assertions in `quantized/mod.rs::as_t_slice`, and treats the inconsistency between the two raw-slice constructors as the defect. This PR deliberately does not do that. Under Miri, where a `Vec<u8>` gets only the 1-byte alignment it actually guarantees, that assertion rejects an ordinary well-formed tensor: "ggml tensor data is not aligned to 4 bytes". The alignment a byte buffer ends up with is the allocator's choice, not the caller's, so asserting on it rejects valid data. Hence borrow when aligned, copy when not.

Which leaves the part I cannot decide on my own. `as_t_slice` still carries that assertion, and it sits on the Metal load path for every GgmlDType at mod.rs:99-113, so it has the same latent panic on an allocator that does not over-align. I left it alone to keep this reviewable, and I see tomsanbear already fixed a UAF in that same function in #3493.

Do you want as_t_slice softened in this PR while the reasoning is in front of you, or kept as a follow-up so this one stays scoped to ggml_file.rs?

Issue #3679, opened 2026-06-28
test_gelu_operation fails on Apple Silicon (1 ULP f32 precision mismatch)


jaweed3:
**Describe the bug**

`cargo test test_gelu_operation` fails on Apple Silicon (tested on M4 Mac Mini) with a 1 ULP difference in `gelu_erf`:

```
assertion left == right failed
  left: [[0.0, 0.8413447], [1.9544997, 2.9959502]]
 right: [[0.0, 0.8413448], [1.9544997, 2.9959502]]
```

The expected value `0.8413448` is the x86 rounding of `GELU(1.0) = x · 0.5 · (1 + erf(x / √2))` = `0.841344745...`. Apple Silicon rounds the same f32 value to `0.8413447`. Both are valid representations of the true value.

**Reproduction**

```
cargo test -p candle-onnx -- test_gelu_operation
```

**Expected behavior**

Tests should pass on all architectures.

**Suggested fix**

Use `to_vec2_round(z, 4)` instead of exact `to_vec2::<f32>` comparison, matching the pattern already used by `test_sin_operation` and `test_cos_operation`.

PR #3975, opened 2026-09-14
fix(onnx): LogSoftmax without an axis attribute returned softmax

yuefdev:
When a `LogSoftmax` node has no `axis` attribute, the `None` branch in `candle-onnx/src/eval.rs` called `softmax_last_dim`, so the op returned softmax instead of log-softmax. With an explicit `axis` it was already correct.

The default axis is -1 since opset 13, so the fix calls `log_softmax` on the last dimension, which matches what the `Softmax` arm does for its default.

`test_logsoftmax_operation` covers exactly this case, but its expected values were the softmax ones (`[0.2689, 0.7311]` for `[1, 2]`), so it passed with the bug. I changed them to the log-softmax values (`[-1.3133, -0.3133]`). The test fails without the fix. I also compared no axis, `axis=1` and `axis=-1` against onnxruntime on a 3D input and all three match now.

Unrelated to this change: `test_gelu_operation` also fails on main for me on x86_64 Windows, looks like the same thing as #3679.


EricLBuehler (commented):
Thank you @yuefdev!

PR #3993, opened 2026-09-21
Only look for a layer norm bias tensor when the norm has one

jordiboehme:
Since #1888, `layer_norm` looks up the bias tensor name before it checks `config.affine`. The lookup now runs for a norm that was asked for without a bias. `layer_norm_no_bias` therefore fails on any checkpoint that ships no bias tensor, which is every ModernBERT checkpoint. The in-tree modernbert example has not been able to load `answerdotai/ModernBERT-base` since then: `Error: Failed to find weight tensor`.

The lookup moves inside the affine branch, where its result is used. The gamma and beta handling from #1888 is unchanged for an affine norm. The error text for a missing bias said "weight tensor" and now says "bias tensor".

The new test covers three cases: a no-bias norm loads from a map that holds only a weight, an affine norm still reads gamma and beta, and an affine norm still fails when the bias is missing. On main the first case fails, on this branch all three pass. The modernbert example predicts Paris again with no other change.

Checks: `cargo fmt --all -- --check`, `cargo clippy --workspace --tests --examples --benches -- -D warnings`, `cargo test -p candle-nn`. All clean.


EricLBuehler:
Hey @jordiboehme! Thanks for the PR. This seems like a duplicate of #3973, closing as such.

